PR #14, opened 2016-02-18
Include devDependencies in default 'test'

adam-moss:
By default npm will install dev dependencies unless the `--production` flag is provided or the NODE_ENV environment variable is set to `production`.

This change adds the `--dev` flag to the default test script generated by `snyk wizard` to ensure dev dependencies, if present, are checked for vulnerabilities.

### DIFF
--- a/cli/commands/protect/wizard.js
+++ b/cli/commands/protect/wizard.js
@@ -300,7 +300,7 @@ function processAnswers(answers, policy, options) {
     }
 
     var test = pkg.scripts.test;
-    var cmd = 'snyk test';
+    var cmd = 'snyk test --dev';
     if (test && test !== 'echo "Error: no test specified" && exit 1') {
       // only add the test if it's not already in the test
       if (test.indexOf(cmd) === -1) {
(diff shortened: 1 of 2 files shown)


remy:
@adam-moss Thanks for this PR. I can see the tests are failing, but I also suspect you weren't able to run any tests yourself (since it makes use of private modules). Our external contrib isn't quite setup (to my liking) yet, so I've made a note to try to fix that up.

I _think_ this PR makes sense, but I need to loop a few others in to make sure.

Thanks again.

adam-moss:
Yep, couldn't run a proper test suite I'm afraid :(

I'd reviewed the travis errors and decided it was still a work in progress at your end :)

remy:
I've pushed an update to our repo that removes the internal dependencies
now, so contributing should be easier!

adam-moss:
Righto, I've rebased based on your changes but Travis still doesn't want to play ball I'm afraid :(

remy:
Yeah, I suspected there was one last thing that might cause a trip up. I'm
off today until Tuesday but I'll get this raised internally.

adam-moss:
Any further thoughts on this one?

PR #18, opened 2016-03-11
saving snyk as a devdependency

antialias:
wizard saves `snyk` in `devDependencies` instead of `dependencies`

### DIFF
--- a/cli/commands/protect/wizard.js
+++ b/cli/commands/protect/wizard.js
@@ -356,7 +356,7 @@ function processAnswers(answers, policy, options) {
           undefsafe(pkg, 'optionalDependencies.snyk')) {
         installPromise = Promise.resolve(); // do nothing
       } else {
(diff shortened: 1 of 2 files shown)


johannakoll:
Thanks for the PR @antialias! @remy will take a look once he is back at the office on Wednesday next week.

remy:
Snyk actually needs to go into dependencies (not dev) if it's being used to apply a protect. 

Say for example you're using the request package that live patches a vuln in hawk. If request puts snyk in the devDeps, then when request is installed as your production dep, request's own devDeps won't be installed, and thus the patch won't be applied to hawk (as intended by the authors of the request package).

It's intentionally that snyk [isn't searched for in the devDeps](https://github.com/Snyk/snyk/blob/master/cli/commands/protect/wizard.js#L354-L356) because of this particular logic.

I hope that makes sense.

antialias:
Thanks for the explanation. That makes sense.

PR #26, opened 2016-05-07
feat: Include devDependencies in default test from Wizard

adam-moss:
- [X] Ready for review
- [X] Follows CONTRIBUTING rules
- [ ] Reviewed by @remy (Snyk internal team)
#### What does this PR do?

(This was previously raised as PR [#14](https://github.com/Snyk/snyk/pull/14) but I... errr... broke it 😊)

Adds --dev to the default `snyk test` script generated by the Wizard.
- Create a new package (e.g. `npm init`)
- Run `snyk wizard` and allow it to update the test script
- Review `package.json`

By default `npm install` will include devDependencies unless `--production` or the relevant environment variable is set.

Therefore it would be sensible to reflect this in the implementation of `snyk` to ensure consistency.
#### Additional questions
- Test `wizard detects existing snyk in scripts.test` now fails because of the way it has been implemented.  I haven't changed this as I'm unsure if you'd prefer to mock the entire thing.

### DIFF
--- a/cli/commands/protect/wizard.js
+++ b/cli/commands/protect/wizard.js
@@ -270,7 +270,7 @@ function processAnswers(answers, policy, options) {
     }
 
     var test = pkg.scripts.test;
-    var cmd = 'snyk test';
+    var cmd = 'snyk test --dev';
     if (test && test !== 'echo "Error: no test specified" && exit 1') {
       // only add the test if it's not already in the test
       if (test.indexOf(cmd) === -1) {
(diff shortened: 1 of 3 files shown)


adam-moss:
Fixed broken test 👍

Niponkong:
เมื่อ 13 ก.ย. 2017 08:16 "Niponkong" <notifications@github.com> เขียนว่า

maban:
Hi @adam-moss! We've been thinking about this use case a lot, but we don't want to test dev dependencies by default as they are not usually exploitable in production. 

However, enabling the testing of dev dependencies when the user opts-in to do so (eg when running `snyk wizard --dev`) would be cool. Would you consider changing your code to make testing dev dependencies possible using a flag?

adam-moss:
Sure, I can rebase and enable `--dev` on `wizard`.

PR #128, opened 2018-01-24
feat: check that there are actually modules in node_modules

jehy:
- [x] Ready for review

#### What does this PR do?

This checks that there are actually some modules in node_modules directory.
I also added a ToDo because it is better to check that all deps from package.json are installed.

From the beginning, of cause.

There can be three cases:
1) node_modules exist and there is something - should not change (test it)
2) node_modules exist but is empty - should return new error message
3) node_modules does not exist - should not change (old error message)

This will break testing projects which have zero dependencies and have an empty `node_modules` directory but I don't think that this testing makes any sense - it is clearly a false positive.

https://github.com/snyk/snyk/issues/126

### DIFF
--- a/lib/snyk-test/npm/index.js
+++ b/lib/snyk-test/npm/index.js
@@ -21,14 +21,23 @@ function test(root, options) {
   var promise = Promise.all([
     fs.exists(root),
     fs.exists(path.join(root, 'node_modules')),
+    fs.readdir(path.join(root, 'node_modules')).catch(function () {return [];}),
   ]).then(function (res) {
     var exists = res[0];
     var nodeModulesExist = res[1];
+    var nodeModulesExistSome = res[2];
     var hasDevDependencies = false;
 
     if (exists && !nodeModulesExist) {
       // throw a custom error
       throw new Error('Missing node_modules folder: we can\'t test ' +
+        'without dependencies.\nPlease run `npm install` first.');
+    }
+    // ToDo: check if all actually used in package.json modules are installed
+    if (nodeModulesExistSome.length === 0) {
+      // throw a custom error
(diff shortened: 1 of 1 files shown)


jehy:
Travis CI seems broken - it fails because snyk is not authorized.

adrukh:
@jehy thanks for this contribution! Overall a very clean implementation!

I'm slightly concerned with this flow:
> 2. node_modules exist but is empty - should return new error message

When using `npm@3` on a project with no dependencies, `npm install` creates an empty `node_modules` folder. I'd like `snyk test` to successfully test 0 dependencies on this project. Your suggestion will make this case result in a failure, and be a breaking change for our users.

Can you suggest a way to overcome this? I'd attempt to run `npm -v` and check the version, acting accordingly to the result of this run.

It would also be great to see tests added for this scenario - see https://github.com/snyk/snyk/blob/master/test/missing-node-modules.test.js#L11 for a relevant test.

Thanks for your effort, it is much welcome! I'd like to help seeing this through :)

jehy:
> 'd like snyk test to successfully test 0 dependencies on this project. Your suggestion will make this case result in a failure, and be a breaking change for our users.

I don't think that it's right to test project with zero dependencies, but I understand that breaking changes are pretty bad.

In https://github.com/snyk/snyk/issues/126 I described why relying on npm version is unreliable. But I added another commit to this pull request - it checks whether all dependencies from `package.json` exist in `node_modules`. That should be fine and projects will zero dependencies should pass. Also it will throw error if not all modules are currently installed.

lili2311:
Closing this as the issue is resolved as per confirmation on the issue: https://github.com/snyk/snyk/issues/126

PR #164, opened 2018-06-22
Fix html report generation and write monitor json result

julienduchesne:
- [x] Ready for review

#### What does this PR do?
- Fixes the HTML report feature. Currently, the output file is text not json but jq is run on it. It returns `parse error: Invalid numeric literal at line 2, column 8
Failed to run the process ...`
- Adds a `snyk-monitor-result.json`


adrukh:
@julienduchesne this is indeed an important fix, but it highlights a shortcoming in our docker image entrypoint. We aimed at allowing `monitor` to run after `test` via the `MONITOR` env var, and broke the flow where a user just wants to run `snyk monitor` via the docker image.

Your suggestion addresses this issue, but I wonder whether it breaks the ability to run `snyk test` via the docker image, expecting human-readable and not JSON output...

julienduchesne:
We would have to run snyk test twice to get both the json result and the human readable output. Would that be acceptable?

PR #176, opened 2018-07-16
Add python3 docker

adrukh:
Grabbing @julienduchesne 's addition to support python-based docker images of the CLI. Will go through our internal release process, stay tuned.

### DIFF
--- a/docker/Dockerfile.python-2
+++ b/docker/Dockerfile.python-2
@@ -1,0 +1,31 @@
+FROM node:8-slim
+
+MAINTAINER Snyk Ltd
+
+# Install python
+RUN apt-get update && \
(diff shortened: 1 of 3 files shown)


adrukh:
@julienduchesne THANK YOU for this contribution!

We merged the commit with the Dockerfiles to serve as a reference in our repo (release 1.89.0), but have yet linked them to our dockerhub. We'll consider doing so in the future, as we apply some testing to the docker image generation process.

tommilligan:
@adrukh any estimate of when this will be released to docker hub?

adrukh:
@tommilligan there's no specific timeline I can share right now. Will update once I do!

PR #188, opened 2018-08-09
Feat/lockfile parser lib

miiila:
**Moved from https://github.com/snyk/snyk-internal/pull/417**

- [x] Ready for review

#### What does this PR do?
- Adds in a fork for `opt-in` lockfile testing
- Includes the new lockfile parser lib
- Splits up some functionality for re-use

Assuming the https://github.com/snyk/nodejs-lockfile-parser/pull/5/files returns a tree the way we expect, i hope this is enough code to be able to get back a tree in the same format and carry on the test in the same way as before.

- We are missing tests for this, need to get 1 in at least!

Look in `snyk-test/npm/index.js`

`snyk-dev test -d` => will test using the old flow by looking at your node_modules
`snyk-dev test -d --file=package-lock.json` will go down the new flow and use the new lib

SC-5873


lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,122 @@ var isCI = require('../../is-ci');
...
+}
+
+function generateDependenciesFromLockfile(root, options) {
+  debug('Lockfile detected, generating dependency tree from lockfile');
+  var targetFileFullPath = path.resolve(root, 'package.json');
+  var lockFileFullPath = path.resolve(root, options.file);
+  var shrinkwrapFullPath = path.resolve(root, 'npm-shrinkwrap.json');
+
+  if (!fileSystem.existsSync(targetFileFullPath)) {
+    throw new Error('LockFile package.json not found at location: ' +
```
michael-go: a typo - `package.json` is not a lock

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,122 @@ var isCI = require('../../is-ci');
...
 
+  if (options.file.endsWith('package-lock.json')) {
+    return generateDependenciesFromLockfile(root, options);
+  }
+  return getDependenciesFromNodeModules(root, options);
+}
+
+function generateDependenciesFromLockfile(root, options) {
+  debug('Lockfile detected, generating dependency tree from lockfile');
+  var targetFileFullPath = path.resolve(root, 'package.json');
```
michael-go: let's think about the case of monorepo and make sure it behaves as we expect, in example what about a case where user runs:
```bash
snyk test --file=node-subproject/package-lock.json
```
in this case the `pacakge.json` should be picked from the same location of the lock, but maybe it's ok as the user can also run:
```bash
snyk test node-project --file=package-lock.json
```

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,122 @@ var isCI = require('../../is-ci');
...
+
+  if (options.file.endsWith('package-lock.json') && fileSystem.existsSync(shrinkwrapFullPath)) {
+    throw new Error('`npm-shrinkwrap.json` was found while using package-lock.json.\n'
+    + 'Please use `--file=npm-shrinkwrap.json` flag instead.');
+  }
+
+  var targetFile = fileSystem.readFileSync(targetFileFullPath);
+  var lockFile = fileSystem.readFileSync(lockFileFullPath);
+
+  analytics.add('local', true);
```
michael-go: maybe also add analytics metadata for the fact a lockfile is used

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,122 @@ var isCI = require('../../is-ci');
...
+
+  var resolveModuleSpinnerLabel = 'Analyzing npm dependencies for ' +
+  lockFileFullPath;
```
michael-go: an indentation is needed here

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,122 @@ var isCI = require('../../is-ci');
...
+  debug(resolveModuleSpinnerLabel);
+  var payload = {
```
michael-go: is it necessary to clone the payload creation (and this is not only the payload, but parameters for the http request) and policy collection into `generateDependenciesFromLockfile()` ? Would expect this function to just generate a tree and continue with the flow that already exists

package.json
```diff
@@ -48,6 +48,7 @@
     "snyk-gradle-plugin": "1.3.0",
     "snyk-module": "1.8.2",
     "snyk-mvn-plugin": "1.2.0",
+    "snyk-nodejs-lockfile-parser": "^1.2.2",
```
michael-go: note we always pin-point our own dependencies to an exact version - maybe we shouldn't, but easier to stick to what we already do 😅

test/acceptance/cli.acceptance.test.js
```diff
@@ -487,6 +487,32 @@ test('`test npm-package` sends pkg info', function (t) {
...
 
+test('`test npm-package --file=package-lock.json ` sends pkg info', function (t) {
+  chdirWorkspaces();
+  return cli.test('npm-package', {file: 'package-lock.json'})
+    .then(function () {
+      var req = server.popRequest();
+      var pkg = req.body;
+      t.equal(req.method, 'POST', 'makes POST request');
+      t.match(req.url, '/vuln/npm', 'posts to correct url');
+      t.ok(pkg.dependencies['to-array'], 'dependency');
```
michael-go: would check a transitive dependency too

test/acceptance/cli.acceptance.test.js
```diff
@@ -487,6 +487,32 @@ test('`test npm-package` sends pkg info', function (t) {
...
+test.only('`test npm-package-shrinkwrap --file=package-lock.json ` with npm-shrinkwrap errors', function (t) {
+  t.plan(1);
+  chdirWorkspaces();
+  return cli.test('npm-package-shrinkwrap', {file: 'package-lock.json'})
+    .catch((e) => {
+      t.includes(e.message, '--file=npm-shrinkwrap.json', 'Contains enough info about error');
+    });
+});
+
 test('`test` on a yarn package does work and displays appropriate text',
```
michael-go: maybe also add a test that if lockfile exists but not given via `--file` it's not used?

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,122 @@ var isCI = require('../../is-ci');
...
+  }
+
+  if (!targetFileFullPath && lockFileFullPath) {
+    throw new Error('Detected a lockfile at location: '
+    + lockFileFullPath + '\n However the package.json is missing!');
+  }
+
+  if (options.file.endsWith('package-lock.json') && fileSystem.existsSync(shrinkwrapFullPath)) {
+    throw new Error('`npm-shrinkwrap.json` was found while using package-lock.json.\n'
+    + 'Please use `--file=npm-shrinkwrap.json` flag instead.');
```
michael-go: and we support that?

package.json
miiila: Hmm, I am not sure I am happy with this approach, however I can see the reasons... Will do it for now.

test/acceptance/cli.acceptance.test.js
miiila: There's no transitive dependecny right now :-(

package.json
gjvis: @miiila It means that the version of the main CLI library allows us to tell exactly the state of the Snyk modules on a user's machine. Very useful for debugging.

lib/snyk-test/npm/index.js
miiila: We don't and we will not => changing the message.

lib/snyk-test/npm/index.js
gjvis: defo need to make sure that the correct lockfile is used (according to NPM rules)

lib/snyk-test/npm/index.js
miiila: Thank you - I had it in mind, but then I forgot :-)

package.json
miiila: @gjvis And for those situations, where you break the flow by fixing the bug 😇

test/acceptance/cli.acceptance.test.js
michael-go: so lets use a better fixture ...

lib/snyk-test/npm/index.js
michael-go: I can try and help refactor this, don't think we should merge a PR with a code that is not in good shape - unless there's good reason for it to be

lib/snyk-test/npm/index.js
lili2311: `LockFile` => `Lockfile`

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,124 @@ var isCI = require('../../is-ci');
...
+  };
+  var modules;
+  options.hasDevDependencies = false;
+  var policyLocations = [options['policy-path'] || root];
+  const p = spinner(resolveModuleSpinnerLabel)
+    .then(function () {
+      return lockFileParser.buildDepTree(manifestFile, lockFile, options);
+    })
+    .then(function (pkg) {
+      modules = pkg;
```
michael-go: my comment here became outdated with some recent changes - but still, this needs refactoring. Why duplicate the http requests params creation & policy file collection code?

michael-go: and also the name `getDependenciesFrom...` is misleading as the function also query for vulns.

lib/snyk-test/npm/index.js
```diff
@@ -11,20 +11,124 @@ var isCI = require('../../is-ci');
 var _ = require('lodash');
 var analytics = require('../../analytics');
 var common = require('../common');
+var fileSystem = require('fs');
```
michael-go: any reason you're not calling the var `fs` like we always do?

lib/snyk-test/npm/index.js
```diff
@@ -11,67 +11,66 @@ var isCI = require('../../is-ci');
...
+    headers: {
+      'x-is-ci': isCI,
+      authorization: 'token ' + snyk.api,
+    },
+  };
+  options.hasDevDependencies = false;
+  options.root = root;
+
+  // if the file exists, let's read the package file and post
+  // that up to the server.
```
michael-go: the comment is misleading, as we'll upload the fully resolved dependency tree, not the package file

lib/snyk-test/npm/index.js
```diff
@@ -90,125 +89,179 @@ function test(root, options) {
...
-        analytics.add('packageName', data.package.name);
-        analytics.add('packageVersion', data.package.version);
-        analytics.add('package', data.package.name + '@' + data.package.version);
-
-        return new Promise(function (resolve, reject) {
-          request(data.payload, function (error, res, body) {
-            if (error) {
-              return reject(error);
-            }
+  return queryForVulns(p, modules, options);
```
michael-go: maybe it will be nicer to do:
```javascript
return p.then(function(payload) {
  return queryForVulns(payload, modules, options);
})
```
or even add it to he promise chain above,
instead of passing a promise as a parameter to `queryVulns`

lib/snyk-test/npm/index.js
```diff
@@ -90,125 +89,179 @@ function test(root, options) {
...
+        path.relative('.', path.join(root, options.file));
+      return spinner(resolveModuleSpinnerLabel)
+        .then(function () {
+          return snyk.modules(
+            root, Object.assign({}, options, {noFromArrays: true}));
+        }).then(spinner.clear(resolveModuleSpinnerLabel));
+    });
+}
+
+function queryForVulns(p, modules, options) {
```
michael-go: note that in the lockfile case - `modules` is just the pkgTree, but in the node_modules case it's an object with methods and such .. and in this function `.pluck()` is called on it and `.numDependencies` is checked

lib/snyk-test/npm/index.js
```diff
@@ -11,67 +11,66 @@ var isCI = require('../../is-ci');
...
+    // options.vulnEndpoint is only used for file system tests
+    url: config.API + (options.vulnEndpoint || '/vuln/npm'),
+    json: true,
+    headers: {
+      'x-is-ci': isCI,
+      authorization: 'token ' + snyk.api,
+    },
+  };
+  options.hasDevDependencies = false;
+  options.root = root;
```
michael-go: this doesn't look right - if we do this to pass data down to `queryForVulns()`, lets instead just use extra parameters for it, and not "abuse" the `options` object

lib/snyk-test/npm/index.js
miiila: I agree (and will address), however, this exactly the thin ice, when you are refactoring and suddenly realise, you are changing behavouir :-D

lib/snyk-test/npm/index.js
miiila: Well, not sure if I agree - what is the `options` object for then? I am not a big fan of methods like `queryForVulns(data, modules, hasDevDependencies, root, options)`.

PR #197, opened 2018-08-19
fix: adjust spacing on 'test' output

adrukh:
- [x] Ready for review

#### What does this PR do?

A cosmetics issue, aligning spaces in output like this:
```
Testing /Users/anton/code/snyk/snyk...

Organisation:      snyk
Package manager:   npm
Target file:       package.json
Open source:       no
Project path:      /Users/anton/code/snyk/snyk
Local Snyk policy: found
Licenses:          enabled

✓ Tested 246 dependencies for known issues, no vulnerable paths found.
```

### DIFF
--- a/cli/commands/test.js
+++ b/cli/commands/test.js
@@ -431,31 +431,42 @@ function createTruncatedVulnsPathsText(vulnList) {
   }
 }
 
+function rightPadWithSpaces(s, desiredLength) {
+  var padLength = desiredLength - s.length;
+  if (padLength <= 0) {
+    return s;
+  }
+
+  var padding = Array.from({length: padLength + 1}).join(' ');
+  return s + padding;
+}
+
 function metaForDisplay(res, options) {
+  var padToLength = 19; // chars to align
   var packageManager = options.packageManager || res.packageManager;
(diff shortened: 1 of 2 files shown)


cli/commands/test.js
```diff
@@ -431,31 +431,42 @@ function createTruncatedVulnsPathsText(vulnList) {
   }
 }
 
+function rightPadWithSpaces(s, desiredLength) {
+  var padLength = desiredLength - s.length;
+  if (padLength <= 0) {
+    return s;
+  }
+
+  var padding = Array.from({length: padLength + 1}).join(' ');
```
miiila: `return s.padEnd(desiredLength, ' ')` should be all you need.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd

darscan: Not sure it would work on Node 4. Alternatively, here's another thing that probably wouldn't work: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat

adrukh: Grrr was too fast with the merge button!

Will amend this

PR #216, opened 2018-09-17
Add typescript

miiila:
Introducing TypeScript will happen in ~3~ 2 phases (flow verified in `registry`):
- [x] [make sure `lib` and `cli` mapping works before/after compilation](https://github.com/snyk/snyk/pull/217)
- [ ] **add typescript support for a sourcode and tests** *

This PR introduces typescript support for source files and tests. One source file and test are converted in order to verify everything works as intended.

As a side effect, `package-lock.json` was added and `npm prune` was removed from Travis setup.

\* Sourcode and tests had to be merged together, because conversion of one file led to different requires in tests


adrukh:
Let's not make this all `chore`s? I understand that there are no functional changes, but this will not create a new release. Meaning that this will be pushed to our users together with another functional change, sometime later... Prefer to make sure it works as expected sooner, and without any additional changes. What do you folks think?

.snyk
```diff
@@ -1,5 +1,12 @@
...
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
-ignore: {}
+ignore:
+  'npm:mem:20180117':
+    - nyc > yargs > os-locale > mem:
+        reason: DoS vulnerability is not valid for CLI tool
+        expires: '2018-10-19T10:35:25.346Z'
+    - tap > nyc > yargs > os-locale > mem:
+        reason: DoS vulnerability is not valid for CLI tool
+        expires: '2018-10-19T10:35:25.346Z'
```
adrukh: `tap` and `nyc` are dev deps - how come we're ignoring them in the policy? How about simply disregarding them?

miiila: I believe this is because of the way, how we are resolving `node_modules`. In the dependency tree, everything from `nyc` down is not a `dev` dependency, only `tap` is. I am not entirely sure, how do we evaulate whole path then.

Not sure what do you mean by disregard, but test suite contains `snyk test` and it wouldn't pass without ignore.

michael-go:
think now we should update `.nvmrc` to `8`?

package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "snyk",
   "description": "snyk library and cli utility",
-  "main": "src/lib/index.js",
+  "main": "dist/lib/index.js",
   "files": [
-    "src",
     "help",
+    "dist",
```
michael-go: maybe include also `src` ? not sure ... 🤔

package.json
```diff
@@ -67,21 +72,26 @@
...
+    "@types/lodash": "^4.14.116",
+    "@types/node": "^10.10.0",
     "eslint": "^4.11.0",
     "nock": "^9.2.5",
     "proxyquire": "^1.7.4",
     "restify": "^4.1.1",
     "sinon": "^1.17.2",
     "strip-ansi": "^4.0.0",
-    "tap": "^12.0.0",
-    "tap-only": "0.0.5",
```
michael-go: maybe if you remove `tap-only`, can also remove the `check-tests` scripts above? as it's purpose it to check for `only()` accidently commited

miiila: @michael-go `tap` supports `.only()` as well (https://www.node-tap.org/only/), that's why I have left it.

package.json
```diff
@@ -67,21 +72,26 @@
...
-    "tape": "^4.0.0"
+    "tap": "github:snyk/node-tap#alternative-runtimes",
+    "tape": "^4.0.0",
```
michael-go: is `tape` really needed? maybe it can easily be replaced with `tap`?

miiila:
@adrukh In my opinion - it's not a functional change and we should be confident we haven't broken anything. 

I also believe that adoption of CLI tool is not as wide as SaaS, so we would get a feedback after a while. If we would publish a new version ( with functional change) in the meantime and there would be a problem - what's a safe version for a rollback?

package.json
miiila: @michael-go Based on https://docs.npmjs.com/files/package.json#files, `src` shouldn't be there. You shouldn't need non-compiled source, when you install this as a dependency.

src/cli/index.ts
```diff
@@ -1,32 +1,34 @@
...
+import * as runtime from './runtime';
 if (!runtime.isSupported(process.versions.node)) {
   console.error(process.versions.node +
     ' is an unsupported nodejs runtime! Supported runtime range is \'' +
     runtime.supportedRange + '\'');
   console.error('Please upgrade your nodejs runtime version ' +
     'and try again.');
   process.exit(1);
-};
+}
```
michael-go: with real ES6 modules (which is not the case here, but might be in the future with node 10), all the imports happen before the the script starts, so I think it's better to put all imports on top before any code to be more clear about it.

src/cli/index.ts
```diff
@@ -1,32 +1,34 @@
...
-var analytics = require('../lib/analytics');
-var args = require('./args')(process.argv);
-var copy = require('./copy');
-var alerts = require('../lib/alerts');
-var sln = require('../lib/sln');
-var exitcode = 0;
+import * as analytics from '../lib/analytics';
+import * as alerts from '../lib/alerts';
+import * as sln from '../lib/sln';
+import argsImport = require('./args');
```
michael-go: maybe rename to `argsLib`?

src/cli/index.ts
```diff
@@ -1,32 +1,34 @@
...
+import copy = require('./copy');
+const args = argsImport(process.argv);
```
michael-go: would add a blank line above this to seprate code from imports

src/cli/index.ts
```diff
@@ -37,10 +39,10 @@ var cli = args.method.apply(null, args.options._).then(function (result) {
     }
   }
   return res;
-}).catch(function (error) {
-  var spinner = require('../lib/spinner');
+}).catch((error) => {
+  const spinner = require('../lib/spinner');
```
michael-go: maybe `import spinner = require('...')` above instead?

src/cli/index.ts
```diff
@@ -63,26 +65,26 @@ var cli = args.method.apply(null, args.options._).then(function (result) {
...
+  const res = analytics({
     args: args.options._,
+    command,
   });
 
   if (args.options.debug) {
     console.log(error.stack);
   } else {
-    var errors = require('../lib/error');
+    const errors = require('../lib/error');
```
michael-go: use `import` at file top instead?

src/cli/index.ts
```diff
@@ -63,26 +65,26 @@ var cli = args.method.apply(null, args.options._).then(function (result) {
...
 
-      var result = errors.message(error);
+      const result = errors.message(error);
       if (args.options.copy) {
         copy(result);
         console.log('Result copied to clipboard');
       } else {
         if ((error.code + '').indexOf('AUTH_') === 0) {
-          var ansiEscapes = require('ansi-escapes');
+          const ansiEscapes = require('ansi-escapes');
```
michael-go: `import` at file top instead?

package.json
michael-go: yes it's not necessary, it may make debugging easier for someone who imports `snyk`

test/acceptance/cli.acceptance.test.js
```diff
@@ -59,7 +59,7 @@ test('test cli with multiple params: good and bad', function (t) {
   .then(function () {
     t.fail('expect to error');
   }).catch(function (error) {
-    errObj = JSON.parse(error.message);
+    var errObj = JSON.parse(error.message);
```
michael-go: `let` if you're at it ? 😅

test/alerts.test.ts
```diff
@@ -1,55 +1,54 @@
-var test = require('tap-only');
-var testUtils = require('./utils');
-var alerts = require('../src/lib/alerts');
+import tap = require('tap');
+const test = tap.test;
```
michael-go: `import {test} from 'tap'` instead ?

test/analytics.test.js
```diff
@@ -1,5 +1,5 @@
 var tap = require('tap');
-var test = require('tap-only');
+var test = require('tap').test;
```
michael-go: guess you don't need the `tap` line above?

test/fixtures/demo-private/node_modules/autocache/test/cache.test.js
```diff
@@ -2,7 +2,7 @@
 /*global describe:true, it: true */
 var test = require('tape');
 
-var cache = require('../')();
```
michael-go: is that correct? this is `cache` not the `snyk` lib

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+        "target": "es5",
+        "lib": [
+          "es2017"
+        ],
+        "module": "commonjs",
+        "allowJs": true,
+        "sourceMap": true,
+        // "declaration": true,
+        "strict": true,
+        "noImplicitAny": false // Needed to compile tap
```
michael-go: instead - can have a separate `tsconfig.json` in the `test/` folder that extends this one - we have such thing in our node-express teamplate e.g.

.snyk
adrukh: Gotcha!

package.json
miiila: Sorry, I disagree here, because it would require to have installed `typescript` etc... And basically, the message we send by that would be: in order to use this, you need to have `src` as well as `dist`.

michael-go: no problem - as I said I wasn't sure myself - wanted to hear you opinion

test/analytics.test.js
miiila: I do - because of https://github.com/snyk/snyk/pull/216/files/bdf2554850913021eed8800070ecb34cf08c3bf4#diff-7dab017e750fb8d7a8cf01c7ef0807efR9

michael-go: the link didn't point me to something specific 🙈

adrukh:
Re `chores` I’d argue that adoption is very quick - a lot of times the cli is a dependency with `^`, and updates immediately after a release.

I’d make a dummy release to get faster feedback that it works (silence on our support channel is a strong signal), so that if there’s a problem we can fix it while the iron is hot :)

If you consider the risk low enough, we can stay with `chore`s.

test/analytics.test.js
miiila: It should 😂 It's needed because of `tap.beforeEach(function (done)`.

package.json
miiila: TBH, I don't know, but don't want to do this decision now :-)

test/analytics.test.js
michael-go: ok cool 👍

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+        "outDir": "./dist",
+        "pretty": true,
+        "target": "es5",
+        "lib": [
+          "es2017"
+        ],
+        "module": "commonjs",
+        "allowJs": true,
+        "sourceMap": true,
+        // "declaration": true,
```
michael-go: any reason leaving this commented out?

miiila: 🙉 🙈 🙊

PR #217, opened 2018-09-18
Move lib and cli to src

miiila:
Introducing TypeScript will happen in 3 phases (flow verified in `registry`):
- [ ] **make sure `lib` and `cli` mapping works before/after compilation**
- [ ] add typescript support for a sourcode
- [ ] add typescript support for tests

This is a first PR, which groups `lib` and `cli` folders in `src` in order to be ready to have them in `dist` folder in the future.


lili2311:
tested locally, looking good!

miiila:
Stand alone binaries works fine on 🍎 and 🐧.

PR #228, opened 2018-10-02
Lockfile autodetection for test, wizard, protect and monitor

lili2311:
- [x] Ready for review

#### What does this PR do?
- Introduce yarn & package-lock.json autodetection to use lockfile only (without node_modules traversal)
- Leave `node_module` traversal as fall back for package.json & shrinkwrap projects

The state of NPM lockfile support in this pr:
- projects with `package.json` &  `package-lock.json` will be tested via the new lockfile parser lib for `snyk protect`, `test`, `monitor`

The state of Yarn lockfile support in this pr:
- projects with `package.json` &  `yarn.lock` will be tested via the new lockfile parser lib for `snyk 
 test`, `monitor`
- projects with `package.json` &  `yarn.lock` will be tested via the old `node_modules` traversal lib (`resolve-deps`) for `snyk protect` and `wizard`
- projects with `package.json` &  `yarn.lock` on node@4 will be tested via the old `node_modules` traversal lib (`resolve-deps`) for all `snyk` commands. Yarn parser lib we use is not compatible with node@4

- most changes are in the npm & yarn plugin to test if a project is lockfile based before choosing which way to generate the tree
- run test, monitor, protect and wizard locally on a lockfile based project

https://snyksec.atlassian.net/browse/SC-6240
https://snyksec.atlassian.net/browse/SC-6277


lili2311:
**WORK IN PROGRESS**
Need to add some changes to wizard to better support this

src/cli/commands/protect/index.js
```diff
@@ -16,16 +16,21 @@ var unsupportedPackageManagers = {
...
   options.loose = true; // replace missing policies with empty ones
   options.vulnEndpoint = '/vuln/npm/patches';
+  // TODO: fix this by providing better patch support for yarn
+  // yarn hoists packages up a tree so we can't assume their location
+  // on disk without traversing node_modules
+  // currently the npm@2 nd npm@3 plugin resolve-deps can do this
+  // but not the latest node-lockfile-parser
+  // HACK: if yarn set traverseNodeModules option to
+  // bypass lockfile test for wizard
+  options.traverseNodeModules = true;
```
lili2311: Fallback for yarn projects where we currently do not support traversing node_modules with the new lockfile-parser

src/cli/commands/protect/wizard.js
```diff
@@ -291,9 +270,23 @@ function processAnswers(answers, policy, options) {
...
+  const isLockFileBased = targetFile.endsWith('package-lock.json') || targetFile.endsWith('yarn.lock');
+
+  // TODO: fix this by providing better patch support for yarn
+  // yarn hoists packages up a tree so we can't assume their location
+  // on disk without traversing node_modules
+  // currently the npm@2 nd npm@3 plugin resolve-deps can do this
+  // but not the latest node-lockfile-parser
+  // HACK: if yarn set traverseNodeModules option to
+  // bypass lockfile test for wizard
+  if (targetFile.endsWith('yarn.lock')) {
```
lili2311: Fallback for yarn projects where we currently do not support traversing node_modules with the new lockfile-parser

miiila:
Great job, waiting for tests to pass before final approvement.

src/cli/commands/protect/wizard.js
```diff
@@ -43,63 +43,41 @@ var unsupportedPackageManagers = {
...
 
-function wizard(options) {
-  if (!options) {
-    options = {};
-  }
-  if (config.org) {
-    options.org = config.org;
-  }
+function wizard(options = {}) {
+  options.org = config.org || null;
```
adrukh: I think this overrides `options.org` if the config is set, which is different from the original code.
Maybe `options.org = options.org || config.org || null;`?

miiila: Good catch! But original code overrides it as well.

```
  if (config.org) {
    options.org = config.org;
  }
```

So config always wins.
`options.org = config.org || options.org || null;` should be the same then.

adrukh: You are right, my suggestion was wrong. Use yours!

Not sure re API entitlements, I think it's unrelated.

miiila: It's unrelated in terms of `wizard`, but could this behaviour be replicated through the codebase?

src/lib/snyk-test/npm/index.js
```diff
@@ -181,6 +181,10 @@ function getDependenciesFromNodeModules(root, options, targetFile) {
         path.dirname(path.resolve(root, targetFile));
       return spinner(resolveModuleSpinnerLabel)
         .then(() => {
+          // yarn projects fall back to node_module traversal if node < 6
```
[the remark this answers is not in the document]
lili2311: Here we bypass the usual `snyk.test` call and directly call `snyk.modules` which simply requires in `snyk-resolve-deps`. We have inconsistent way of conducting a test which we should tackle in one of the future improvements.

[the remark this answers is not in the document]
miiila: Got it, had to read a bit more for a context, it looks 👌

PR #232, opened 2018-10-05
Fix/alert if path with filename

Kirill89:
- [x] Ready for review

#### What does this PR do?

Check if user specify package file name as part of path and throw error if so.

src/cli/index.ts

`snyk test rubygems/Gemfile.lock`

#### Jira?

[SC-6271](https://snyksec.atlassian.net/browse/SC-6271)

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -12,6 +12,7 @@ import {copy} from './copy';
 import spinner = require('../lib/spinner');
 import errors = require('../lib/error');
 import ansiEscapes = require('ansi-escapes');
+import {isPathToPackageFile} from '../lib/detect';
 
 async function runCommand(args) {
   const result = await args.method(...args.options._);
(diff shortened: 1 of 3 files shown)


src/lib/detect.ts
```diff
@@ -27,8 +25,9 @@ var DETECTABLE_FILES = [
 ];
 
 // when file is specified with --file, we look it up here
-var DETECTABLE_PACKAGE_MANAGERS = {
-  Gemfile: 'rubygems',
+/*tslint:disable object-literal-sort-keys*/
```
lili2311: can we just disable this rule in the config please? It is too restrictive I think.

Kirill89: Done

src/cli/index.ts
```diff
@@ -95,6 +96,17 @@ function checkRuntime() {
   }
 }
 
+// Check if user specify package file name as part of path
+// and throw error if so.
+function checkPaths(args) {
+  for (const path of args.options._) {
+    if (typeof path === 'string' && isPathToPackageFile(path)) {
+      throw new Error(`Path "${path}" contains package file name, ` +
```
lili2311: I am a little worried `package` is not clear enough, maybe we can do something like:
`Not a recognised option did you mean --file=${path}`

PR #239, opened 2018-10-09
fix: do not traverse node_modules on wizard monitor for yarn

lili2311:
- [x] Ready for review

#### What does this PR do?
Fixing a bug where if a yarn project is fixed with wizard we traversed the node modules via a specially set flag but then did not set this back just before monitoring.

While traversing node_modules for yarn is a must for now for patching (old style node_module traversal populates where each package lives on disk which is used during patch time) this fall back is not needed for yarn projects node > 6 when we run a test just before monitoring the project & sending the snapshot to snyk web.

To replicate:
- clone https://github.com/lili2311/npm-lockfiles
- run `snyk wizard` in the `yarn-lockfiles` folder and see the error appear: not a node project (this is an error encountered because we are sending yarn.lock as a filename to the `resolve-deps` lib instead pf `package.json`)

### DIFF
--- a/src/cli/commands/protect/wizard.js
+++ b/src/cli/commands/protect/wizard.js
@@ -282,8 +282,9 @@ function processAnswers(answers, policy, options) {
   // on disk without traversing node_modules
   // currently the npm@2 nd npm@3 plugin resolve-deps can do this
   // but not the latest node-lockfile-parser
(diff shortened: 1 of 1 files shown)


lili2311:
Thanks I made it myself 🙄

PR #243, opened 2018-10-12
fix: bump nodejs-lockfile-parser to allow 0 deps

michael-go:
nodejs-lockfile-parser PR: https://github.com/snyk/nodejs-lockfile-parser/pull/23


michael-go:
also verified manually, it works

PR #252, opened 2018-10-21
fix: don't print vulnerable paths count for docker tests

karniwl:
- [x] Ready for review

#### What does this PR do?
Removes display of number of vulnerable paths for docker, since currently the count is inaccurate for several reasons:
1. the trees are trimmed quite aggressively in snyk-docker-plugin to reduce size.
2. for rpm & apk based images, we don't differentiate between packages installed manually vs. transitives that were installed automatically.

[Jira ticket SC-6459](https://snyksec.atlassian.net/browse/SC-6459)

### DIFF
--- a/src/cli/commands/test.js
+++ b/src/cli/commands/test.js
@@ -242,7 +242,8 @@ function displayResult(res, options) {
   var vulnCountText = 'found ' + res.uniqueCount + ' '
     + (res.uniqueCount === 1 ? singleVulnText : multipleVulnsText);
 
-  if (options.showVulnPaths) {
+  // Docker is currently not supported as num of paths is inaccurate due to trimming of paths to reduce size.
+  if (options.showVulnPaths && !options.docker) {
     vulnCountText += ', ' + vulnCount + ' vulnerable ';
 
     if (vulnCount === 1) {
--- a/test/acceptance/cli.acceptance.test.js
+++ b/test/acceptance/cli.acceptance.test.js
@@ -1162,6 +1162,39 @@ function (t) {
   });
 });
 
+test('`test foo:latest --docker vulnerable paths`',
+function (t) {
+  var plugin = {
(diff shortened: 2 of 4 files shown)


michael-go:
this better be a `fix: ` commit, rather than `feat: `.

and please note that the commit message appears later in the Release Notes (https://github.com/snyk/snyk/releases) so maybe reword with:
`fix: don't print vulnerable paths count for docker tests`

and we better give a reason in the PR body to why we do it. Maybe can just say it's because right now the count is not accurate for several reasons:
1. the trees are trimmed quite aggressively in `snyk-docker-plugin` to reduce size
2. for rpm & apk based images, we don't differentiate between packages installed manually vs. transitives there were installed automatically

test/acceptance/cli.acceptance.test.js
```diff
@@ -1162,6 +1162,39 @@ function (t) {
...
+    },
+  };
+  sinon.spy(plugin, 'inspect');
+
+  sinon.stub(plugins, 'loadPlugin')
+    .withArgs(sinon.match.any, sinon.match({docker: true}))
+    .returns(plugin);
+  t.teardown(plugins.loadPlugin.restore);
+
+  var vulns = require('./fixtures/npm-package-policy/vulns.json');
```
darmalovan: I would maybe add a new fixtures file for docker

src/cli/commands/test.js
```diff
@@ -242,7 +242,7 @@ function displayResult(res, options) {
   var vulnCountText = 'found ' + res.uniqueCount + ' '
     + (res.uniqueCount === 1 ? singleVulnText : multipleVulnsText);
 
-  if (options.showVulnPaths) {
+  if (options.showVulnPaths && !options.docker) {
```
darmalovan: maybe add a small comment here saying why docker is not supported?

PR #250, opened 2018-10-19
chore: upgrade sinon

DanielRuf:
- [ ] Ready for review
- [ ] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules
- [ ] Reviewed by Snyk internal team

#### What does this PR do?
Upgrade sinon

#### Any background context you want to provide?
Resolve the deprecated warnings.

PR #255, opened 2018-10-27
Test/upgrade sinon

adrukh:
- [x] Ready for review

#### What does this PR do?

Merges #250 via an internal PR. This is needed as PRs from forks are failing our travis builds due to sensitive env vars and travis' security policy.

cc @DanielRuf

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "nock": "^9.2.5",
     "proxyquire": "^1.7.4",
     "restify": "^4.1.1",
-    "sinon": "^1.17.2",
+    "sinon": "^4.0.0",
     "strip-ansi": "^4.0.0",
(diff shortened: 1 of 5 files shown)


adrukh:
@DanielRuf can you advise please - tests fail as the commit doesn't bump `sinon` in `package.json` to a newer version. See https://travis-ci.org/snyk/snyk/jobs/447102624

Considering we still want to be able to run tests on node 4, what would be the right version of `sinon` to upgrade to? I'm guessing you picked something newer than `^1` when you were working on this, and I'm also guessing `^7` won't fly due to the need to run tests on node 4...

adrukh:
I picked `sinon@^4` seeing it's the last one to support nodejs 4, seems like we're good.

PR #257, opened 2018-10-30
fix: fix for assumed patch paths

lili2311:
Also amend how we re-lock the lockfile

- [x] Ready for review

#### What does this PR do?
- Fixes a bug where for package-lock.json projects we are relying on the assumed patch paths for patching (since the lockfile is the exact shape of the `node_modules`)
The bug was around us using `cwd` which is mostly used for tests and this was returning path paths like: `/npm/node_modules/path-to-module/` instead of  `/node_modules/path-to-module/`

- Also fixing how we regenerate the lockfile if user selected to add snyk to dependencies. Update is relying on having the vuln from paths present, so skipping straight past all this and directly calling `install` && `installDev` since adding `snyk` is not part of the vuln testing flow 

https://github.com/snyk/snyk/compare/fix/file-path-fix?expand=1#diff-7d011e4c9a3d66fcafd27dede7916a27R17
Project `goof` can be used after adding a lockfile as an example

### DIFF
--- a/src/cli/commands/protect/wizard.js
+++ b/src/cli/commands/protect/wizard.js
@@ -13,7 +13,6 @@ const chalk = require('chalk');
 const url = require('url');
 const _ = require('lodash');
 const exec = require('child_process').exec;
-const undefsafe = require('undefsafe');
 const auth = require('../auth');
 const getVersion = require('../version');
 const allPrompts = require('./prompts');
@@ -276,7 +275,6 @@ function processAnswers(answers, policy, options) {
   var packageManager = detect.detectPackageManager(cwd, options);
   var targetFile = options.file || detect.detectPackageFile(cwd);
   const isLockFileBased = targetFile.endsWith('package-lock.json') || targetFile.endsWith('yarn.lock');
-
   // TODO: fix this by providing better patch support for yarn
   // yarn hoists packages up a tree so we can't assume their location
   // on disk without traversing node_modules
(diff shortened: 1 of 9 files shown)


src/lib/protect/get-vuln-source.js
```diff
@@ -12,7 +12,9 @@ function getVulnSource(vuln, cwd, live) {
   });
 
   var viaPath = path.resolve(
-    cwd || process.cwd(),
+    // cwd is optional and mostly used for testing
+    // according to an old note in the code in patch
+    process.cwd() || cwd,
```
miiila: This will be always resolved to `process.cwd()`, there's no way how it could be falsy. Not sure, if it's the right change, but if it is, `|| cwd` part can be removed.

src/cli/commands/protect/wizard.js
```diff
@@ -435,7 +437,10 @@ function processAnswers(answers, policy, options) {
         // and how do we reconcile if the global install is older
         // than the local version?
         } else {
-          if (answers['misc-add-protect']) {
+          const addSnykToProdDeps = answers['misc-add-protect'];
+          const addSnykToDevDeps = undefsafe(pkg, 'devDependencies.snyk');
```
miiila:
```diff
- undefsafe(pkg, 'devDependencies.snyk');
+ _.get(pkg, 'devDependencies.snyk');
```

PR #267, opened 2018-11-08
chore: async/await and TS in lib/snyk-test/run-test

michael-go:
because it's about time


src/lib/snyk-test/run-test.js
```diff
@@ -17,167 +17,153 @@ var snyk = require('../');
...
+      if (!seen[curr.id]) {
+        seen[curr.id] = true;
+        acc++;
+      }
+      return acc;
+    }, 0);
+
+    return res;
+  } finally {
+    spinner.clear(spinnerLbl)();
```
Shesekino: `await`?

PR #292, opened 2018-12-05
feat: add a CLI message when newer version is available

yuliabaron:
- [x] Ready for review

#### What does this PR do?
Adding a message to CLI when a newer version is available

https://snyksec.atlassian.net/browse/SC-6399

#### Screenshots
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "tempfile": "^2.0.0",
     "then-fs": "^2.0.0",
     "undefsafe": "^2.0.0",
+    "update-notifier": "^2.5.0",
     "uuid": "^3.2.1"
   },
   "devDependencies": {
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -13,6 +13,32 @@ import spinner = require('../lib/spinner');
 import errors = require('../lib/error');
 import ansiEscapes = require('ansi-escapes');
 import {isPathToPackageFile} from '../lib/detect';
+import * as updateNotifier from 'update-notifier';
+import * as fs from 'fs';
+import * as p from 'path';
+
+function updateCheck() {
+    const root = p.resolve(__dirname, p.normalize('../..'));
+    const pkgPath = p.resolve(root, 'package.json');
+    const isPkgFilePresent = fs.existsSync(pkgPath);
+
+    if (!isPkgFilePresent) {
+      return;
+    }
+
+    const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+
+    // if there's no version (f.e. during tests) - do not proceed
+    if (!pkg.version) {
+      return;
+    }
+
+    // Checks for available update and returns an instance
+    // Default updateCheckInterval is once a day
+    const notifier = updateNotifier({pkg});
+    notifier.notify();
+    return;
+}
 
 async function runCommand(args) {
   const result = await args.method(...args.options._);
@@ -108,6 +134,7 @@ function checkPaths(args) {
 }
 
 async function main() {
+  updateCheck();
   checkRuntime();
 
   const args = argsLib(process.argv);


package.json
```diff
@@ -110,5 +111,6 @@
       "help/**/*.txt"
     ]
   },
-  "snyk": true
+  "snyk": true,
+  "version": "1.109.0"
```
darscan: Howdy! We should not commit the version number into our package.json, semantic release will take care of that during the publish step.

src/cli/index.ts
```diff
@@ -14,6 +14,19 @@ import errors = require('../lib/error');
...
 import {isPathToPackageFile} from '../lib/detect';
 
+import * as updateNotifier from 'update-notifier';
+var pkg = require('../../package.json');
+
+async function updateCheck() {
+    // Checks for available update and returns an instance
+    // Default updateCheckInterval is once a day
+    var notifier = updateNotifier({
+        pkg: pkg,
```
miiila: You can write this as `updateNotifier({pkg}); when object key is same as a variable it points to, you can use this syntactic sugar :-)

src/cli/index.ts
```diff
@@ -14,6 +14,19 @@ import errors = require('../lib/error');
...
+    });
+    // Notify using the built-in convenience method
```
miiila: I'm not sure I understand this comment. Do you think we need it?

src/cli/index.ts
```diff
@@ -14,6 +14,19 @@ import errors = require('../lib/error');
 import ansiEscapes = require('ansi-escapes');
 import {isPathToPackageFile} from '../lib/detect';
 
+import * as updateNotifier from 'update-notifier';
+var pkg = require('../../package.json');
+
+async function updateCheck() {
```
miiila: As notify is not asycnhronous (i.e. we don't do `await notifier.notify()`, `async` keyword before the function name is misleading.

src/cli/index.ts
```diff
@@ -108,6 +121,7 @@ function checkPaths(args) {
 }
 
 async function main() {
+  await updateCheck();
```
miiila: Related to the comment above, `updateCheck()` can be fully synchronous.

src/cli/index.ts
```diff
@@ -14,6 +14,19 @@ import errors = require('../lib/error');
 import ansiEscapes = require('ansi-escapes');
 import {isPathToPackageFile} from '../lib/detect';
 
+import * as updateNotifier from 'update-notifier';
+var pkg = require('../../package.json');
```
miiila:
```diff
- var
+ const
```

src/cli/index.ts
miiila:
```diff
- var
+ const
```

src/cli/index.ts
miiila: Personal sidenote - I don't like using `require` to load .json files, I would suggest `fs.readFile`, but I fear we can saw this pattern in this app more often, so it's probably good to go.

src/cli/index.ts
```diff
@@ -14,6 +14,19 @@ import errors = require('../lib/error');
...
+
+async function updateCheck() {
+    // Checks for available update and returns an instance
```
miiila: I'm not sure about this comment either - do you think it brings in something important?

src/cli/index.ts
```diff
@@ -13,6 +13,34 @@ import spinner = require('../lib/spinner');
...
+    const pkgFile = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+
+    // Checks for available update and returns an instance
+    // Default updateCheckInterval is once a day
+    // set the try/catch to catch the errors if for some reason package.json doesn't have the required info
+    let notifier;
+    try {
+      notifier = updateNotifier({pkgFile});
+    } catch (err) {
+      console.warn('Failed to start updater, ERR=', err);
```
yuliabaron: Not sure whether to keep this warning. The code works from different directories, errors were observed during the tests on travis only

darscan: I imagine that we get here when `version` is not present in the package.json as would be the case on Travis etc.

yuliabaron: exactly

src/cli/index.ts
```diff
@@ -13,6 +13,34 @@ import spinner = require('../lib/spinner');
...
+      return;
+    }
+
+    notifier.notify();
```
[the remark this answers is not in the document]
yuliabaron: i return from catch, but you're right, will move

PR #293, opened 2018-12-05
feat: sha256 checksums for binary releases

michael-go:
With this, sha256 checksums will be computed for the binary packages of the CLI and will be uploaded together with the binaries as release assets to GitHub, and will be named as `snyk-<platform>.sha256`.


.releaserc
```diff
@@ -3,7 +3,7 @@
     "@semantic-release/npm",
     {
       "path": "@semantic-release/exec",
-      "cmd": "npm i -g pkg && pkg ."
+      "cmd": "npm i -g pkg && pkg . && shasum -a 256 snyk-linux > snyk-linux.sha256 && shasum -a 256 snyk-macos > snyk-macos.sha256 && shasum -a 256 snyk-win.exe > snyk-win.exe.sha25"
```
adrukh: Missing `6` at the end

michael-go: right! I remember saying to myself "add a 6 after you paste" 😅  thanks ! 🙏

aarlaud: sha25 is probably secure too !

michael-go:
I'll also verify now that the `shasum` command exists and works the same on a travis shell

michael-go:
cool - `shasum` exists on travis and has same params

PR #301, opened 2018-12-07
fix: add tests for update-notifier and adjust code

yuliabaron:
- [x] Ready for review

#### What does this PR do?
Adds tests for update-notifier API

### DIFF
--- a/src/lib/updater.ts
+++ b/src/lib/updater.ts
@@ -1,0 +1,25 @@
+import * as updateNotifier from 'update-notifier';
+import * as fs from 'fs';
+import * as p from 'path';
+
+export function updateCheck() {
+  const pkgPath = p.join(__dirname, '../..', 'package.json');
+  const isPkgFilePresent = fs.existsSync(pkgPath);
+
+  if (!isPkgFilePresent) {
+    return false;
+  }
+
+  const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+
+  // if there's no version (f.e. during tests) - do not proceed
+  if (!pkg.version) {
+    return false;
+  }
+
+  // Checks for available update and returns an instance
+  // Default updateCheckInterval is once a day
+  const notifier = updateNotifier({pkg});
+  notifier.notify();
+  return true;
+}
--- a/test/updater.test.js
+++ b/test/updater.test.js
@@ -1,0 +1,33 @@
+const tap = require('tap');
+const test = tap.test;
+const updateCheck = require('../src/lib/updater').updateCheck;
+const path = require('path');
+const sinon = require('sinon').createSandbox();
+const updateNotifier = require('update-notifier');
+
+// Fake location of the package.json file and verify the code behaves well
+test('missing package.json', (t) => {
+  t.plan(1);
+  let resolveStub = sinon.stub(path, 'resolve');
+  resolveStub.onFirstCall().returns('falseDir');
+  resolveStub.onSecondCall().returns('falseFile');
+
+  t.tearDown(() => {
+    resolveStub.restore();
+  });
+
+  t.equal(updateCheck(), false, 'Notifier was not started on missing package.json');
+  t.end();
+});
+
+// Run updateNotifier API for the basic package. THe target is to verify API still stands
+test('verify updater', (t) => {
+   var pkg = {
+    name: 'snyk',
+    version: '1.0.0'
+  };
+  const notifier = updateNotifier({pkg});
+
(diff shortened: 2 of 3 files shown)


test/updater.test.js
```diff
@@ -0,0 +1,31 @@
...
+  t.equal(updateCheck(), false, 'Notifier was not started on missing package.json');
+  resolveStub.restore();
+  t.end();
+});
+
+// Run updateNotifier API for hte basic package. THe target is to verify API still stands
+test('verify updater', function (t) {
+  t.plan(1);
+  var pkg = {
+    "name": "snyk",
```
Kirill89: `"` -> `'`

Kirill89: `"name"` -> `name`

test/updater.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Fake location of the package.json file and verify the code behaves well
+test('missing package.json', function (t) {
+  t.plan(1);
+  var resolveStub = sinon.stub(path, 'resolve');
+  resolveStub.onFirstCall().returns('falseDir');
+  resolveStub.onSecondCall().returns('falseFile');
+
+  t.equal(updateCheck(), false, 'Notifier was not started on missing package.json');
+  resolveStub.restore();
```
Kirill89: All the `.restore()` stuff should be in `t.teardown` callback. Because if test fails unrestored stub can affect other tests.

test/updater.test.js
```diff
@@ -0,0 +1,31 @@
+const tap = require('tap');
+const test = tap.test;
+const updateCheck = require('../src/lib/updater').updateCheck;
+const path = require('path');
+const sinon = require('sinon').createSandbox();
+const updateNotifier = require('update-notifier');
+
+// Fake location of the package.json file and verify the code behaves well
+test('missing package.json', function (t) {
```
Kirill89: `function (t)` -> `(t) =>`

src/lib/updater.ts
```diff
@@ -0,0 +1,26 @@
+import * as updateNotifier from 'update-notifier';
+import * as fs from 'fs';
+import * as p from 'path';
+
+export function updateCheck() {
+  const root = p.resolve(__dirname, p.normalize('../..'));
```
Kirill89: You don't need `p.normalize` here - it is not user input (already normalised).

Kirill89: I believe we can do something like `const pkgPath = p.join(root, '../../package.json');` instead of this.

test/updater.test.js
Kirill89: Can we add test which actually checks that notifier works (I mean produces some output)?

test/updater.test.js
```diff
@@ -0,0 +1,31 @@
...
+  t.plan(1);
+  var resolveStub = sinon.stub(path, 'resolve');
+  resolveStub.onFirstCall().returns('falseDir');
+  resolveStub.onSecondCall().returns('falseFile');
+
+  t.equal(updateCheck(), false, 'Notifier was not started on missing package.json');
+  resolveStub.restore();
+  t.end();
```
Kirill89: I'm not sure you need `t.plan(x);` and `t.end();` in this tests.

src/lib/updater.ts
yuliabaron: tests also run on Windows env, so I do need p.normalize. And this file is not a test, it runs in CLI.

test/updater.test.js
yuliabaron: no. I can't run notifier during the test as it spawns a child process and probably during testing this is not possible. Plus, I do not want to test external library. All I want is to make sure I'm able to init it successfully with the API they claimed. This way, if API changes I'll catch that during tests.

test/updater.test.js
yuliabaron: tests fail without that.

src/lib/updater.ts
Kirill89: Just check it `path.win32.join('bla1', 'bla2', 'bla3', 'Bla4', '../package.json');` I believe `join` already doing `normalize` inside.

test/updater.test.js
Kirill89: But in the test "verify updater" you're exactly testing external library? At least maybe better to mock `fs.readFileSync`?

Kirill89: But in the test "verify updater" you're exactly testing external library? At least maybe better to mock fs.readFileSync?

test/updater.test.js
Kirill89: Let's remove `t.plan(x);` at least.

PR #309, opened 2018-12-18
Feat/support docker binaries

orkamara:
- [x] Ready for review

#### What does this PR do?
Add the new binaries issues into the output of the CLI

The change support both regular output + `--json` flag
We had to change some logic around `introducedThrough` and `fromPaths` to make it work.

https://snyksec.atlassian.net/secure/RapidBoard.jspa?projectKey=DC&rapidView=62&selectedIssue=DC-7

### DIFF
--- a/src/cli/commands/test.js
+++ b/src/cli/commands/test.js
@@ -26,7 +26,7 @@ function test() {
   }
 
   // populate with default path (cwd) if no path given
-  if (args.length ===  0) {
+  if (args.length === 0) {
     args.unshift(process.cwd());
   }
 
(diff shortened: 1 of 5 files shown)


src/cli/commands/test.js
```diff
@@ -557,7 +565,7 @@ function groupVulnerabilities(vulns) {
       map[curr.id].severity = curr.severity;
       map[curr.id].isNew = isNewVuln(curr);
     }
-    if (curr.upgradePath) {
+    if (curr.upgradePath && curr.upgradePath[0]) {
```
adrukh: What is the purpose of this change?

orkamara: It's basically a big fix - we didn't check if the list is empty - only if its declared.
The result is that we might give the user remediation advice even if `upgradePath.length == 0`.
The reason why we didn't discover the bug before is because the check `curr.upgradePath[1] === curr.from[1];` - but in our case those 2 lists are empty

adrukh: Hmmmm seems really important! Is this on a separate commit? If it doesn’t break the new flow, maybe it deserves its own test, PR and release?

orkamara: changed it - need to invest more and figure out how to test it

PR #313, opened 2018-12-19
feat: Split dockerfile / base image vulns

robcresswell:
This patch adds an extra line to vulnerabilities found in a Dockerfile, indicating whether they came from the base Docker image (FROM: ) or your Dockerfile. In the latter case, the instruction that installed the vulnerable package is also highlighted.

### DIFF
--- a/src/cli/commands/test.js
+++ b/src/cli/commands/test.js
@@ -358,7 +358,11 @@ function formatIssues(vuln, options) {
       ? createRemediationText(vuln, packageManager)
       : '',
     fixedIn: options.docker ? createFixedInText(vulnerableRange, version) : '',
+    dockerfilePackage: options.docker && vuln.dockerfileInstruction
+      ? `\n  Introduced in your Dockerfile by '${ vuln.dockerfileInstruction }'`
+      : `\n  Introduced by your base image (${ vuln.dockerBaseImage })`,
   };
+
   return (
     vulnOutput.issueHeading + '\n' +
     vulnOutput.description + '\n' +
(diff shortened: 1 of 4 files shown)


darmalovan:
Please put description & resolve conflicts when you can :)

src/cli/commands/test.js
```diff
@@ -358,7 +358,11 @@ function formatIssues(vuln, options) {
       ? createRemediationText(vuln, packageManager)
       : '',
     fixedIn: options.docker ? createFixedInText(vulnerableRange, version) : '',
+    dockerfilePackage: options.docker && vuln.dockerfileInstruction
+      ? `\n  Introduced in your Dockerfile by '${ vuln.dockerfileInstruction }'`
```
orkamara:
```suggestion
      ? `\n  Dockerfile layer: '${ vuln.dockerfileInstruction }'`
      : `\n  Dockerfile layer: FROM (${ vuln.dockerBaseImage })`,
```

src/lib/snyk-test/run-test.ts
```diff
@@ -61,6 +67,17 @@ async function runTest(packageManager: string, root: string , options): Promise<
       return acc;
     }, 0);
 
+    if (res.docker && dockerfilePackages) {
+      res.vulnerabilities = res.vulnerabilities.map((vuln) => {
+        const dockerfilePackage = dockerfilePackages[vuln.name.split('/')[0]];
```
orkamara: consider extracting vuln.name.split('/')[0] into a different const (maybe `pkgSource`)?

PR #317, opened 2018-12-30
feat: fixed in key in binaries output

karniwl:
- [x] Ready for review

#### What does this PR do?
Adds a `Fixed in` version key in binaries/output, using npm semver rules.
Refactors formatting issues logic.

Run `snyk test --docker node:5.10.1` for example to see the fixed versions of found vulnerabilities.

#### Screenshots

image

### DIFF
--- a/src/cli/commands/test.js
+++ b/src/cli/commands/test.js
@@ -1,6 +1,7 @@
 module.exports = test;
 
 var _ = require('lodash');
+var semver = require('semver');
 var chalk = require('chalk');
 var debug = require('debug')('snyk');
 var snyk = require('../../lib/');
@@ -195,7 +196,6 @@ function summariseErrorResults(errorResults) {
 function displayResult(res, options) {
   var meta = metaForDisplay(res, options) + '\n\n';
   var dockerAdvice = dockerRemediationForDisplay(res);
-  var binariesIssues = binariesIssuesForDisplay(res);
   var packageManager = options.packageManager;
   var prefix = chalk.bold.white('\nTesting ' + options.path + '...\n\n');
 
@@ -236,7 +236,6 @@ function displayResult(res, options) {
     return (
       prefix + meta + summaryOKText + (
         isCI ? '' :
-          binariesIssues +
           dockerAdvice +
           nextStepsText +
           dockerSuggestion)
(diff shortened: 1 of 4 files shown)


src/cli/commands/test.js
```diff
@@ -333,13 +339,24 @@ function displayResult(res, options) {
...
 
-function createFixedInText(groupedVuln) {
-  var vulnerableRange = groupedVuln.list[0].semver.vulnerable[0];
-  if (/^<\S+$/.test(vulnerableRange)) {
-    // removing the first char from the version. For example: <7.50.1-1
-    return chalk.bold('\n  Fixed in: ' + vulnerableRange.substr(1));
+function createFixedInText(versionRangeList, pkgVersion) {
+  let fixedVersion = '';
+  if (!pkgVersion && versionRangeList &&
+      versionRangeList.length && /^<\S+$/.test(versionRangeList[0])) {
```
leongold: This `if` clause is a bit difficult to read, how do you feel about extracting it to a function?

src/cli/commands/test.js
```diff
@@ -314,7 +320,7 @@ function displayResult(res, options) {
       remediationInfo: vuln.metadata.type !== 'license'
         ? createRemediationText(vuln, packageManager)
         : '',
-      fixedIn: options.docker ? createFixedInText(vuln) : '',
+      fixedIn: options.docker ? createFixedInText(vuln.list[0].semver.vulnerable, version) : '',
```
leongold: I understand there's special meaning to the first index of the vuln.list, perhaps we could store it in a variable? (same for versionRangeList[0] and versionList[1])

src/cli/commands/test.js
```diff
@@ -333,13 +340,28 @@ function displayResult(res, options) {
...
-  var vulnerableRange = groupedVuln.list[0].semver.vulnerable[0];
-  if (/^<\S+$/.test(vulnerableRange)) {
-    // removing the first char from the version. For example: <7.50.1-1
-    return chalk.bold('\n  Fixed in: ' + vulnerableRange.substr(1));
+function createFixedInText(versionRangeList, pkgVersion) {
+  let fixedVersion = '';
+  const lesserThan = /^<\S+$/;
+  if (!pkgVersion) {
+    if (versionRangeList &&
+        versionRangeList.length &&
```
orkamara: versionRangeList condition is unnecessary

src/cli/commands/test.js
```diff
@@ -333,13 +340,28 @@ function displayResult(res, options) {
...
 
-function createFixedInText(groupedVuln) {
-  var vulnerableRange = groupedVuln.list[0].semver.vulnerable[0];
-  if (/^<\S+$/.test(vulnerableRange)) {
-    // removing the first char from the version. For example: <7.50.1-1
-    return chalk.bold('\n  Fixed in: ' + vulnerableRange.substr(1));
+function createFixedInText(versionRangeList, pkgVersion) {
+  let fixedVersion = '';
+  const lesserThan = /^<\S+$/;
+  if (!pkgVersion) {
```
orkamara: maybe add an explanation for `pkgVersion` reason being undefined?

src/cli/commands/test.js
karniwl: we can assume it'll never be undefined?

src/cli/commands/test.js
```diff
@@ -333,13 +340,28 @@ function displayResult(res, options) {
...
+    if (versionRangeList &&
+        versionRangeList.length &&
+        (lesserThan.test(versionRangeList[0]))) {
```
leongold: Thanks for the changes! Can we also store `versionRangeList[0]` in a variable? I don't understand the significance of the first index (I guess it's ordered and we need the first item?) :)

karniwl: there isn't a special meaning, it's not ordered, we actually have only one item in the list in that case. I added a comment explaining it

src/cli/commands/test.js
```diff
@@ -299,6 +300,12 @@ function displayResult(res, options) {
       }))
       .join(', ');
 
+    var version = undefined;
+    var vulnerableRange = vuln.list[0].semver.vulnerable;
+    if (vuln.metadata.packageManager === 'upstream') {
```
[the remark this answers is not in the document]
orkamara: yes

src/cli/commands/test.js
karniwl: and which did you mean? l 347 or l 348?

orkamara: ignore me :)

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+    if (versionRangeList &&
+        versionRangeList.length &&
+        (lesserThan.test(versionRangeList[0]))) {
+      // trim relational operator `<` from first version in list
+      // as non binaries are simple case of a single version range
+      fixedVersion = versionRangeList[0].substr(1);
+    }
+  } else {
+    for (const versionRange of versionRangeList) {
+      if (semver.valid(pkgVersion) && semver.satisfies(pkgVersion, versionRange)) {
```
orkamara:
```suggestion
       if (!semver.valid(pkgVersion) || !semver.satisfies(pkgVersion, versionRange)) {
       continue;
       }
```

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+        const fixedVersionCandidate = versionRange.split(' ')[1];
```
orkamara: maybe add a comment with the example of the array of ranges

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
-      vulnOutput.fromPaths +
-      // Optional - not always there
-      vulnOutput.remediationInfo +
-      vulnOutput.fixedIn +
-      vulnOutput.extraInfo
-    );
+  var nonBinariesSortedGroupedVulns = sortedGroupedVulns.filter(function (vuln) {
+    return (vuln.metadata.packageManager !== 'upstream');
+  });
+  var binariesSortedGroupedVulns = sortedGroupedVulns.filter(function (vuln) {
```
orkamara:
```suggestion
   var dockerBinariesSortedGroupedVulns = sortedGroupedVulns.filter(function (vuln) {
```

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
-      vulnOutput.description + '\n' +
-      vulnOutput.info + '\n' +
-      vulnOutput.introducedThrough + '\n' +
-      vulnOutput.fromPaths +
-      // Optional - not always there
-      vulnOutput.remediationInfo +
-      vulnOutput.fixedIn +
-      vulnOutput.extraInfo
-    );
+  var nonBinariesSortedGroupedVulns = sortedGroupedVulns.filter(function (vuln) {
```
orkamara:
```suggestion
   var sortedGroupedVulns = sortedGroupedVulns.filter(function (vuln) {
```

orkamara: Let's not change the all code to be docker related :)

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+    return (vuln.metadata.packageManager !== 'upstream');
+  });
+  var binariesSortedGroupedVulns = sortedGroupedVulns.filter(function (vuln) {
+    return (vuln.metadata.packageManager === 'upstream');
   });
+  var groupedVulnInfoOutput = nonBinariesSortedGroupedVulns.map(vuln => formatIssues(vuln, options));
+  var groupedBinariesVulnInfoOutput = res.docker && res.docker.binariesVulns ?
```
orkamara:
```suggestion
   var groupedDockerBinariesVulnInfoOutput = (res.docker && res.docker.binariesVulns) ?
```

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+      if (i.from[1]) {
+        return i.from && i.from[1];
+      }
+      return i.from;
+    }))
+    .join(', ');
+
+  var version = undefined;
+  var vulnerableRange = vuln.list[0].semver.vulnerable;
+  if (vuln.metadata.packageManager === 'upstream') {
```
orkamara: I think it's better to add an optional `version` param to `formatIssues()`.
Then, pass it `vuln.metadata.version`  when calling it from `formatBinariesIssues()`

It will remove another reference to `upstream`

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+    groupedBinariesVulnInfoOutput.join('\n\n') + '\n\n' + meta + summary;
   return prefix + body + dockerAdvice + dockerSuggestion;
 };
 
-function createFixedInText(groupedVuln) {
-  var vulnerableRange = groupedVuln.list[0].semver.vulnerable[0];
-  if (/^<\S+$/.test(vulnerableRange)) {
-    // removing the first char from the version. For example: <7.50.1-1
-    return chalk.bold('\n  Fixed in: ' + vulnerableRange.substr(1));
+function formatBinariesIssues(binariesSortedGroupedVulns, binariesVulns, options) {
```
orkamara:
```suggestion
 function formatDockerBinariesIssues(dockerBinariesSortedGroupedVulns, binariesVulns, options) {
```

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+    vulnOutput.remediationInfo +
+    vulnOutput.fixedIn +
+    vulnOutput.extraInfo
+  );
+}
+
+function createFixedInText(versionRangeList, pkgVersion) {
+  let fixedVersion = '';
+  const lesserThan = /^<\S+$/;
+  // pkgVersion is undefined for non binaries vulns
```
orkamara:
```suggestion
   // pkgVersion is undefined for OS pkgs vulns
```

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
+  if (!pkgVersion) {
+    if (versionRangeList &&
+        versionRangeList.length &&
+        (lesserThan.test(versionRangeList[0]))) {
```
orkamara: agree about taking `versionRangeList[0]` into a different var with explanation about its meaning

test/acceptance/cli.acceptance.test.ts
```diff
@@ -1664,6 +1664,7 @@ test('`test foo:latest --docker with binaries vulnerabilities`', async (t) => {
       'docker should not includes number of vulnerable paths');
     t.match(msg, 'Detected 1 vulnerabilities for node@5.10.1');
     t.match(msg, 'High severity vulnerability found in node');
+    t.match(msg, 'Fixed in: 5.13.1');
```
orkamara: maybe add another vuln with `fixed in`? to make sure that we are running on array of indices?

src/cli/commands/test.js
```diff
@@ -288,58 +287,108 @@ function displayResult(res, options) {
...
 
-function createFixedInText(groupedVuln) {
-  var vulnerableRange = groupedVuln.list[0].semver.vulnerable[0];
-  if (/^<\S+$/.test(vulnerableRange)) {
-    // removing the first char from the version. For example: <7.50.1-1
-    return chalk.bold('\n  Fixed in: ' + vulnerableRange.substr(1));
+function formatBinariesIssues(binariesSortedGroupedVulns, binariesVulns, options) {
+  const binariesIssuesOutput = [];
+  for (const pkgInfo of _.values(binariesVulns.affectedPkgs)) {
+    binariesIssuesOutput.push(chalk.bold.white(
```
orkamara: split the push() call and the param into 2 different lines. ... it's a bit long :)

src/cli/commands/test.js
karniwl: that's the name I first gave and it caused a bug that wasted too much time because that's a variable name we already declared and changing it deletes all binaries vulns from it and then when filtering by `upstream` package manager we get an empty result...
I can change the order of the lines but that's too error-prone imo, I'll think of another solution but you're welcome to offer one as well.

src/cli/commands/test.js
karniwl: I think this is much less readable than the current condition, we want to find the version range that matches our package version, imo it makes more sense to show what we are looking for, not what we're NOT looking for

orkamara: I guess it's a style opinion - you're choice :)

src/cli/commands/test.js
```diff
@@ -288,58 +287,118 @@ function displayResult(res, options) {
...
+      if (i.from[1]) {
+        return i.from && i.from[1];
+      }
+      return i.from;
+    }))
+    .join(', ');
+
+  var version = undefined;
+  var vulnerableRange = vuln.list[0].semver.vulnerable;
+  if (vuln.metadata.packageManager.toLowerCase() === 'upstream') {
```
karniwl: @orkamara

src/cli/commands/test.js
orkamara: The problem I'm thinking about is a developer that has no idea about what's binary is, so `nonBinariesSortedGroupedVulns` might be a bit confusing for him

PR #322, opened 2019-01-04
chore: remove chalk

robcresswell:
Small commit, replaces chalk with ansi-colors, which is faster and has
fewer dependencies. Hooray for small wins!


orkamara:
Hooray!

PR #325, opened 2019-01-08
chore: Run CI against Node 10 (current LTS)

robcresswell:


adrukh:
Consider a similar change in `appveyor.yml`?

PR #326, opened 2019-01-09
fix: highlight the use of machine name for orgs help

lirantal:
- [x] Ready for review

#### What does this PR do?
Current help advises to use specify the org name but it isn't clear about the fact that you need to specify the "machine name" representation of the org name.

If I have an org called "Liran demo", it has a machine name of "liran-demo-123" associated with it and the latter is the one that should be provided to the CLI.


robcresswell:
I think filling out the text with an example should be enough. More bothered about clarity than pretty help output :)

lirantal:
@robcresswell how does it read now?

PR #329, opened 2019-01-10
fix: Use unreleased 'pkg' to fix release process

robcresswell:
The CLI build process is currently broken due to an interaction between pkg's "globby" dependency and "dir-glob".

See more at https://github.com/zeit/pkg/issues/603 and https://github.com/kevva/dir-glob/pull/8


robcresswell:
Will do

robcresswell:
No longer needed following the release of `pkg@4.3.6`

PR #258, opened 2018-10-31
fix: ensure patch utility is gnu version

weyusi:
- [x] Ready for review
- [x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules
- [ ] Reviewed by Snyk internal team

#### What does this PR do?
This pull adds a check to test whether `patch` binary is GNU version and outputs an error message if it's not.

#### How should this be manually tested?
1. Use alpine as CI build environment
2. Install a patch-able node package (e.g. [renovate](https://www.npmjs.com/package/renovate)).
3. `snyk protect`
4. There should be [error message](https://support.snyk.io/snyk-cli/snyk-protect-requires-the-patch-binary) `Snyk couldn't patch the specified vulnerabilities because GNU's patch is not available. Please install 'patch' and try again.`.
5. Repeat 1-4 again, but this time with GNU patch installed `apk add patch`.

#### Current behavior
- Alpine 3.8: https://gitlab.com/curben/snyk-test/-/jobs/114648947

#### With this pull
- Alpine 3.8:
  - [Without GNU patch](https://gitlab.com/curben/snyk-test/-/jobs/114669522)
  - [With GNU patch](https://gitlab.com/curben/snyk-test/-/jobs/114670161)

- TrueOS 18.03 (FreeBSD 12):
  - Without GNU patch

  - With GNU patch

#### Any background context you want to provide?
Alpine and *BSD ships with BusyBox and BSD versions of `patch` respectively, which don't support the arguments (i.e. `--backup` in BusyBox and `--verbose` in BSD) used by Synk.

This pull helps those OS users to locate the issue, even without `--debug`.

#### What are the relevant tickets?
Closes #108 @welwood08 

#### Additional questions
1. This [comment](https://github.com/snyk/snyk/issues/48#issuecomment-270350881) mention Snyk ships with `patch` binary. Please clarify.
2. `hasbin()` can be replaced by `if(error)` in `exec()`. Shall I remove it?
3. Is test case required? not sure what the following line does currently,
https://github.com/snyk/snyk/blob/c98b8b39779410f70027eab31a97dd910994f332/test/patch-ensure-patch.js#L33

PR #335, opened 2019-01-18
feat: migrate from GNU patch binary to jsdiff

### LINKED ISSUE #48: Snyk Windows Support - Snyk protect requires unix patch utility
- `node -v`: node v6.7.0
- `npm -v`: npm v3.10.3
- `snyk -v`: 1.19.1
- Command run: snyk protect
### Expected behaviour

`snyk protect` requires the Unix patch utility, which is not available on windows. 

A logical workaround is to install git command line tools prior to snyk, or use a bash like emulation (cygwin etc.)

A more real life scenario can be followed on the https://github.com/FreeCodeCamp/FreeCodeCamp/issues/11323
### Actual behaviour

Requesting better Windows support.. 😞 !
### Steps to reproduce

You can install the Free Code Camp Development stack on a Windows machine. Suggesting this because its having a lot of dependencies that need `snyk` patches and is a good candidate for debug.

---

If applicable, please append the `--debug` flag on your command and include the output here **ensuring to remove any sensitive/personal details or tokens.

### LINKED ISSUE #170: Windows incompatibility due to `patch`
Trying to install Snyk on windows will crash with the notice "Snyk couldn't patch the specified vulnerabilities because GNU's patch is not available. Please install 'patch' and try again." which is not a straight forward thing to do on a windows machine, given that unlike linux distros there is no way to get the right thing.

Ideally, Snyk would be able to use a cross-platform patching tool (and even ideally-er, use a node based tool so that there is no need to even download anything beyond what npm install already takes care of). However, barring that, could this error be change to inform people what the authoritative location for `patch` is?

(And not something along the lines of "install msys" or something, which is total overkill. We've reached a point where most tools are finally reasonably cross-platform in the node/npm ecosystem, it would be fantastic if Snyk also "just worked" regardless of OS)

saschanaz:
- [x] Ready for review

Closes #170, closes #48, closes #258


saschanaz:
What should I do to pass authentication?

adrukh:
@saschanaz thanks, this is much appreciated!

Some context - we explored this avenue a while back, and found no reliable way to *apply* patches in JS. Not saying there wasn't any, just that we searched, tried a few alternatives, and decided to postpone this task.

You already know this via https://github.com/snyk/snyk/issues/48#issuecomment-363628283, I'm just stating it here for reference :)

Your PR looks promising, let's see what our tests say about it. They are failing on this PR as we require a valid API token to run the test suite, and keep it as a sensitive env var on travis. These vars don't get exposed to forked PRs to avoid leaking them to malicious PRs...

You can run the test suite locally while making sure that your API token exists in the `SNYK_TOKEN` env var. Just `npm test`, focus on the tests that have `patch` and `protect` in their names.

saschanaz:
Hi all, this now passes all the tests on my side 🎉

adrukh:
@saschanaz I'm merging this PR into a local branch so we can push this forward. Cannot say exactly when this will happen, but it's on our radar!

PR #355, opened 2019-02-01
feat: use jsdiff instead of 'patch' to reduce dependency on OS binaries

adrukh:
- [x] Ready for review

#### What does this PR do?

Applying patches on `snyk protect` requires the `patch` binary to be available. This is problematic especially on Windows envs where GNU binaries are not installed by default.

With the help of @saschanaz in #335 we're giving `jsdiff` another go, with the intent to remove the dependency on the `patch` binary, and apply patches with pure JS.


adrukh:
I tested this manually by applying the `patch` and the `jsdiff` versions against https://github.com/snyk/goof, comparing the resulting `node_modules` folder. The only difference has been the content of the 'patch flag' files, which is expected (they contain the timestamp of patching).

We will huddle internally on how to best move forward in the least risky way. Really excited about this opportunity, thanks again @saschanaz !

miiila:
Hi @saschanaz, I just wanna let you know we haven't forgotten your awesome contribution and we've proceed a bit more with internal testing and we hope to merge this PR soon.

Thank you again for your help!

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
-    }, function (error, stdout) { // stderr is ignored
-      var out = stdout.trim();
-      if (error || out.indexOf('FAILED') !== -1) {
-        debug('patch command failed', relative, error, out);
-        return patchError(error, out, relative, vuln).catch(reject);
-      }
-      if (!live) {
-        return resolve();
-      }
+    jsDiff(patchContent, relative, live).then(function () {
```
michael-go: maybe `async/away` it ?

src/lib/protect/apply-patch.js
```diff
@@ -1,6 +1,7 @@
 module.exports = applyPatch;
 
 var debug = require('debug')('snyk');
+var diff = require('diff');
```
michael-go: we added the @types pkg for it, so why not `.ts` it? 😉

package.json
```diff
@@ -51,11 +51,13 @@
   "dependencies": {
     "@snyk/dep-graph": "1.1.2",
     "@snyk/gemfile": "1.1.0",
+    "@types/diff": "^3.5.2",
```
michael-go: why this, and the other types are actually here and not in `devDependencies` ? 🤔

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
-      exec(cmd, {
-        cwd: cwd,
-        env: process.env,
-        maxBuffer: STDOUT_BUFFER_SIZE,
-      }, function (error, stdout) {
-        var out = stdout.trim();
-        if (error || out.indexOf('FAILED') !== -1) {
-          debug('patch command failed', relative, error, out);
-          return patchError(error, out, relative, vuln).catch(reject);
+function jsDiff(patchContent, relative, live) {
```
michael-go: what is `live` ? I see we had such a var before, but maybe it's a chance to improve the naming?

saschanaz: `live` is a boolean that decides whether the patch result should be written to the file system or not.

michael-go: Cool, thanks. Would suggest renaming it to `isDryRun` (and reversing the logic checks 😅).

package.json
michael-go: yep - `lodash` & `node` types are in `devDependencies` so all these are here by mistake

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
+        } catch (err) {
+          reject(err);
+        }
+      },
     });
   });
 }
 
-function patchError(error, stdout, dir, vuln) {
+function stripFirstSlash(fileName) {
```
michael-go: would rename this to `stripBeforeFirstSlash` or `trimBeforeFirstSlash` - only understood this when debugging. Also would suggest add comment with a sample path(s) at the caller site, to illustrate why we do this.

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
-        if (error || out.indexOf('FAILED') !== -1) {
-          debug('patch command failed', relative, error, out);
-          return patchError(error, out, relative, vuln).catch(reject);
+function jsDiff(patchContent, relative, live) {
+  const files = {};
+  return new Promise(function (resolve, reject) {
+    diff.applyPatches(patchContent, {
+      loadFile: function (index, callback) {
+        try {
+          var fileName = stripFirstSlash(index.oldFileName);
```
michael-go: would suggest add comment with a sample path(s), to illustrate why we do this.

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
-        cwd: cwd,
-        env: process.env,
-        maxBuffer: STDOUT_BUFFER_SIZE,
-      }, function (error, stdout) {
-        var out = stdout.trim();
-        if (error || out.indexOf('FAILED') !== -1) {
-          debug('patch command failed', relative, error, out);
-          return patchError(error, out, relative, vuln).catch(reject);
+function jsDiff(patchContent, relative, live) {
+  const files = {};
```
michael-go: maybe rename to `patchedFiles`?

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
         }
-
-        debug('patch succeed', out);
-
-        resolve();
-      });
+      },
+      patched: function (index, content, callback) {
+        try {
+          if (content === false) {
```
michael-go: is it documented that upon failure `content === false` ? didn't find it.
and also, maybe we need a test to cover this check?

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,84 @@ function applyPatch(patch, vuln, live) {
...
+            files[oldFileName] = null;
+          }
+          files[newFileName] = content;
+          callback();
+        } catch (err) {
+          callback(err);
+        }
+      },
+      compareLine: function (_, line, operation, patchContent) {
+        if (operation === ' ') {
```
michael-go: a comment will help here. what is operation `' '` ?

src/lib/protect/apply-patch.js
saschanaz: I filed an issue for this behavior: https://github.com/kpdecker/jsdiff/issues/247

src/lib/protect/apply-patch.js
saschanaz: `' '` means there is no operation. I needed this because jsdiff by default compare all lines and fails when an operation-less diff line mismatches, which differs from GNU patch behavior which only checks for `+`/`-`/`!`.

src/lib/protect/apply-patch.js
michael-go: thanks! Maybe it's safer to make a less strict check, like?
```js
if (!content) {
...
```
or can a valid patch result be empty? (I guess not)

saschanaz: We probably can do that. I just wanted to be as strict as possible here, as it's a weird behavior anyway 😁

src/lib/protect/apply-patch.js
adrukh: This goes way up into the flow (https://github.com/snyk/snyk/blob/master/src/cli/commands/protect/index.js#L78), used in `wizard` flow too. Suggest to change this on a separate `chore` PR.

package.json
adrukh: I'll move this.

src/lib/protect/apply-patch.js
adrukh: Doing this.

adrukh:
Force-pushed after rebasing over `master` and attending to the majority of the comments.
@saschanaz please see last two comments re `diff` operations and how we can document them.

src/lib/protect/apply-patch.js
saschanaz: The comment would be:

```suggestion
        if (operation === ' ') {
          // Ignore when no patch operators as GNU patch does
```

src/lib/protect/apply-patch.js
saschanaz: Maybe:

```suggestion
          if (content === false) {
            // `false` means the patch does not match the original content.
```

adrukh:
@michael-go can you please ✅ so I can squash the commits and merge this? ✨

adrukh:
@saschanaz 👋 
We've been running with this change since Sunday, monitoring our success vs failure rates. We are seeing some rare cases where the patching logic is not behaving properly. We had quite limited visibility into patching failures until this change, so it's hard to tell whether this is a new case, or simply deeper visibility into an older issue.

I was wondering if you could shed some light on the situation as we see it. Sometimes, we end up processing an empty hunk. This code runs - https://github.com/snyk/snyk/blob/49dc318db8a5e46bb324c10909239d954cdd3217/src/lib/protect/apply-patch.js#L46-L59 - with the `index` var being just `{ "hunks": [] }`. We bail out as the `index.oldFilename` is undefined.

This is happening for our users here and there, but we are not able to reproduce the issue locally. Any thoughts about this would be greatly appreciated 🙏

saschanaz:
https://github.com/kpdecker/jsdiff/blob/2c1a29c0309fef22219bcf2cbb475f20508a955b/src/patch/parse.js#L55-L69

Looks like it may happen when a patch file has no header. Is there a headerless patch on snyk?

adrukh:
Hmmm

Not only do we not have headerless patches, but this is also not persistent at all in terms of reproduction.

Could this be a race condition when we have yet created the patch file locally, and are already trying to read it? Will check!

src/lib/protect/apply-patch.js
```diff
@@ -29,48 +26,88 @@ function applyPatch(patch, vuln, live) {
...
+        }
+        if (!live) {
+          return resolve();
+        }
+        try {
+          for (var fileName of patchedFiles) {
+            if (patchedFiles[fileName] === null) {
+              fs.unlinkSync(path.resolve(relative, fileName));
+            }
+            fs.writeFileSync(path.resolve(relative, fileName), patchedFiles[fileName]);
```
saschanaz: This line looks like a potential bug generator where `null` would trigger removal and then rewrite.

adrukh:
@saschanaz thanks for your assistance here! I think we found the culprit - indeed a race condition where we `resolve` fetching the patch once the data has been downloaded, but not once it has been written to disk. See https://github.com/snyk/snyk/blob/master/src/lib/protect/fetch-patch.js#L33-L35

We're going to fix this now, have been carrying this problem with us for a very long time! 😌

src/lib/protect/apply-patch.js
adrukh: This area writes the patched dependency source back to the disk, seems ok :)

saschanaz: The value `null` was intended to delete the file but here it instead overwrites with the text `null`.

adrukh: Aha! I see your point. Will make this safer.

adrukh: See https://github.com/snyk/snyk/pull/379/commits/2acfd242642fec537b86dbc3f56dfa86e577a8c4

PR #372, opened 2019-02-17
chore: remove get-uri and hasbin

Kirill89:
- [x] Ready for review

#### What does this PR do?

Remove dependencies:

- `get-uri`
- `hasbin`

Because nobody use it in the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,8 +57,6 @@
     "configstore": "^3.1.2",
     "debug": "^3.1.0",
     "diff": "^4.0.1",
-    "get-uri": "2.0.2",
-    "hasbin": "^1.2.3",
     "inquirer": "^3.0.0",
     "lodash": "^4.17.5",
     "needle": "^2.2.4",


michael-go:
@Kirill89 if you're at it - did you try running https://www.npmjs.com/package/depcheck ?

Kirill89:
@michael-go.

```
> depcheck
Unused dependencies
* get-uri
Unused devDependencies
* @types/diff
* @types/lodash
* @types/node
```

michael-go:
interesting it didn't flag `hasbin`

Kirill89:
I bet it's because you can find it in [assets](https://github.com/snyk/snyk/blob/68d5c6d0ed8efe1c9124ce9834973e57fe79ec69/test/fixtures/protect-via-snyk/yarn.lock#L339). But it's not related to code.

package.json
```diff
@@ -57,8 +57,6 @@
     "configstore": "^3.1.2",
     "debug": "^3.1.0",
     "diff": "^4.0.1",
-    "get-uri": "2.0.2",
-    "hasbin": "^1.2.3",
```
adrukh: Maybe worth relocking `test/fixtures/protect-via-snyk` to get rid of `hasbin` there as well?

Kirill89: We don't have lock file here.

Kirill89: Does it make sense? It's only test fixtures.

adrukh: I'm 90% sure that without this change, running `npm t` locally will generate uncommitted changes on the repo. OTOH, we'd need to re-lock the test project only after this change is released... Just a nag, not a must :)

PR #379, opened 2019-02-19
fix: robust patch fetching to avoid a race condition

adrukh:
- [x] Ready for review

#### What does this PR do?

Stream-based implementation of patch fetching used to have a race condition where the promise would `resolve` as soon as the patch content was downloaded, while letting the file write stream hanging to complete whenever it felt like. This resulted in sporadic patch failures where patch files were read before data was fully written, resulting in an attempt to apply an empty patch.

This is now fixed by making patch fetching logic more robust, summoning the power of TypeScript!


src/lib/protect/fetch-patch.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as fs from 'fs';
+import * as analytics from '../analytics';
+import * as debugModule from 'debug';
+import * as proxyFromEnv from 'proxy-from-env';
+import * as ProxyAgent from 'proxy-agent';
+
+const debug = debugModule('snyk:fetch-patch');
+const getProxyForUrl = proxyFromEnv.getProxyForUrl;
+
+async function getPatchFile(url, filename) {
```
michael-go: can add type annotations

src/lib/protect/fetch-patch.ts
```diff
@@ -0,0 +1,49 @@
...
+  let options;
+  const proxyUri = getProxyForUrl(url);
+  if (proxyUri) {
+    debug('Using proxy: ', proxyUri);
+    options = { agent: new ProxyAgent(proxyUri) };
+  }
+
+  let res;
```
michael-go: if you install `@types/needle` you can then do 
```
let res: NeedleResponse;
```

src/lib/protect/fetch-patch.ts
```diff
@@ -0,0 +1,49 @@
...
+  let patchData;
```
michael-go:
```typescript
let patchData: string;
```

src/lib/protect/fetch-patch.ts
michael-go: would rename `filename` to `outFilename` or somethin'

and b.t.w. now that we use jsDiff - do we need to actually write this to file, or this function can just return the contents instead?

src/lib/protect/fetch-patch.ts
```diff
@@ -0,0 +1,49 @@
...
+  try {
+    res = await needle('get', url, options || {});
```
michael-go: will `needle` throw if it's not 200?

src/lib/protect/fetch-patch.ts
adrukh: I'd rather not make a functional change right now... The serial patching process is managed by a reducing promises chain, needs a refactor of its own :)

src/lib/protect/fetch-patch.ts
adrukh: OMG it doesn't

src/lib/protect/fetch-patch.ts
adrukh: Couldn't make it work for some reason...

michael-go: sorry, this:
```typescript
let res: needle.NeedleResponse;
```

adrukh: Done!

PR #393, opened 2019-03-16
Feat/yarn project type

lili2311:
- [x] Ready for review

We can support Yarn projects in the main app, allow any `yarn` CLI projects to be created as the `yarn` type from now on instead of `npm`.

Previously added projects are unaffected until they are deleted in the UI.
Any new project with a yarn.lock will be created as `yarn` type.

#### Screenshots

Screen Shot 2019-03-16 at 15 54 38


orsagie:
Do we need the node_modules inside the yarn-package-with-subfolders?

help/file.txt
```diff
@@ -14,13 +14,14 @@ looking for files in following order:
...
-  requirements.txt
   build.gradle
   build.gradle.kts
   build.sbt
+  Pipfile
+  requirements.txt
   Gopkg.lock
   vendor/vendor.json
-  obj/project.assets.json
+  project.assets.json
```
orsagie: why is this changing?

lili2311: We look in both locations https://github.com/snyk/snyk/blob/f5511061847d3ec21c2e49b7afc8bea15393aff6/src/lib/detect.ts#L23-L24

Might as well only show the name of the file like for all others was my thinking.

lili2311:
> yarn-package-with-subfolders

Indeed, yarn lockfiles parsing is not supported on node 4 so we have to fall back to traversing node_modules

PR #411, opened 2019-03-28
feat: plugin returns multiple trees

orsagie:
- [ ] Ready for review

#### What does this PR do?
This is done in preparation for getting multiple trees from gradle plugin.
Currently gradle plugin will return multiple test results iff the `--all-subprojects` flag is passed.

We don't support multiple trees for monitor as of yet, because it breaks our upstream handling of projects. so we fail if the argument `--scan-all-subprojects` is passed.

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -54,6 +54,8 @@ Options:
   --gradle-sub-project=<string>
                        For Gradle "multi project" configurations,
                        test a specific sub-project.
+  --all-sub-projects   For "multi project" configurations, test all
+                       sub-projects.
   -q, --quiet ........ Silence all output.
   -h, --help ......... This help information.
   -v, --version ...... The CLI version.
--- a/src/cli/commands/monitor.js
+++ b/src/cli/commands/monitor.js
@@ -35,6 +35,13 @@ function monitor() {
     snyk.id = options.id;
   }
 
+
+  // This is a temporary check for gradual rollout of subprojects scanning
+  // TODO: delete once supported for monitor
+  if (options['all-sub-projects']) {
+    throw new Error('`--all-sub-projects` is currently not supported for `snyk monitor`');
+  }
+
   return apiTokenExists('snyk monitor')
     .then(function () {
       return args.reduce(function (acc, path) {
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -80,14 +80,20 @@ async function test(...args) {
     }
 
     // Not all test results are arrays in order to be backwards compatible
-    // with sctripts that use a callback with test. Coerce results/errors to be arrays
+    // with scripts that use a callback with test. Coerce results/errors to be arrays
     // and add the result options to each to be displayed
-    if (!Array.isArray(res)) {
-      res = [res];
-    }
-    for (const r of res) {
-      results.push(_.assign(r, {path}));
-      resultOptions.push(testOpts);
+    const resArray: any[] = Array.isArray(res) ? res : [res];
+
+    for (let i = 0; i < resArray.length; i++) {
+      results.push(_.assign(resArray[i], {path}));
+      // currently testOpts are identical for each test result returned even if it's for multiple projects.
+      // we want to return the project names, so will need to be crafty in a way that makes sense.
+      if (!testOpts.subProjectNames) {
+        resultOptions.push(testOpts);
+      } else {
+        resultOptions.push(_.assign(_.cloneDeep(testOpts),
+          {subProjectName: testOpts.subProjectNames[i]}));
+      }
     }
   }
 
@@ -519,6 +525,9 @@ function metaForDisplay(res, options) {
   ];
   if (options.file) {
     meta.push(chalk.bold(rightPadWithSpaces('Target file: ', padToLength)) + options.file);
+  }
+  if (options.subProjectName) {
+    meta.push(chalk.bold(rightPadWithSpaces('Sub project: ', padToLength)) + options.subProjectName);
   }
   if (options.docker) {
     meta.push(chalk.bold(rightPadWithSpaces('Docker image: ', padToLength)) + options.path);
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -50,7 +50,6 @@ async function test(...args) {
     // these options later.
     const testOpts = _.cloneDeep(options);
     testOpts.path = path;
-    resultOptions.push(testOpts);
 
     let res;
 
@@ -81,7 +80,13 @@ async function test(...args) {
     }
 
     // add the tested path to the result of the test (or error)
-    results.push(_.assign(res, {path}));
+    if (!Array.isArray(res)) {
+      res = [res];
+    }
+    for (const r of res) {
+      results.push(_.assign(r, {path}));
+      resultOptions.push(testOpts); // doesnt this put testOps in there too many times? line:53
+    }
   }
 
   // resultOptions is now an array of 1 or more options used for
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -79,13 +79,15 @@ async function test(...args) {
       }
     }
 
-    // add the tested path to the result of the test (or error)
+    // Not all test results are arrays in order to be backwards compatible
+    // with sctripts that use a callback with test. Coerce results/errors to be arrays
+    // and add the result options to each to be displayed
     if (!Array.isArray(res)) {
       res = [res];
     }
     for (const r of res) {
       results.push(_.assign(r, {path}));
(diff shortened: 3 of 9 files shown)


kyegupov:
typo in a commit msg: "add help texst"

src/cli/commands/test.ts
```diff
@@ -81,7 +80,13 @@ async function test(...args) {
     }
 
     // add the tested path to the result of the test (or error)
-    results.push(_.assign(res, {path}));
+    if (!Array.isArray(res)) {
```
kyegupov: can you just pin down the return type of snyk.test() to always be an array? 

(also see below, and also wrap the error above into an array)

src/lib/snyk-test/index.js
```diff
@@ -27,11 +27,18 @@ function executeTest(root, options) {
...
-          res.packageManager = packageManager;
+      .then(function (results) {
+        for (const res of results) {
+          if (!res.packageManager) {
+            res.packageManager = packageManager;
+          }
         }
-        return res;
+        if (results.length === 1) {
+          // Return only one result if only one found as this is the default usecase
```
kyegupov: can you instead return an array always (see above)? So that we know that test() and executeTest() always return multiple results.

Or does it break too many tests?

help/help.txt
```diff
@@ -54,6 +54,8 @@ Options:
   --gradle-sub-project=<string>
                        For Gradle "multi project" configurations,
                        test a specific sub-project.
+  --all-subprojects    For gradle "multi project" configurations,
```
lili2311: the previous flag is `--gradle-sub-project` I think we should follow the same spelling `--all-sub-projects`

src/lib/snyk-test/run-test.ts
```diff
@@ -151,87 +188,109 @@ function assemblePayload(root: string, options, policyLocations: string[]): Prom
...
+
     await spinner(spinnerLbl);
-    const inspectRes = await moduleInfo.inspect(root, options.file, options);
+    const deps = await getDepsFromPlugin(root, options);
 
-    const pkg = inspectRes.package;
-    if (_.get(inspectRes, 'plugin.packageManager')) {
-      options.packageManager = inspectRes.plugin.packageManager;
-    }
+    for (const depRoot of deps.depRoots) {
```
kyegupov: here we are implying that deps always has `depRoots` and not `package`, because we have patched it in line 206. It'd be nicer for getDepsFromPlugin to return a type of `PackageMultiRoots {plugin..., deproots...}` so that we are clear what the contract is. Not urgent though, we can add this refactoring later.

test/acceptance/cli.acceptance.test.ts
```diff
@@ -2139,6 +2207,56 @@ test('`monitor pip-app --file=requirements.txt`', async (t) => {
...
+  const req = server.popRequest();
+  t.equal(req.method, 'PUT', 'makes PUT request');
+  t.match(req.url, '/monitor/gradle', 'puts at correct url');
+  t.same(spyPlugin.getCall(0).args,
+    ['gradle-app', 'build.gradle', {
+      args: null,
+    }], 'calls gradle plugin');
+});
+
+test('`monitor gradle-app --scan-all-subprojects`', async (t) => {
```
kyegupov: should be `all-subprojects`?

help/help.txt
kyegupov: Also, do we maybe want "gradle" prefix and capitalized Gradle?

orsagie: Changing to `--all-sub-projects` 
As for adding a `gradle` prefix, I don't think we want this flag will also be useful for yarn workspaces if we hoose to still stay behind a flag. (and we stick to [kebab-casing](http://wiki.c2.com/?KebabCase) for our flags, so no capitalizations)

src/cli/commands/monitor.js
```diff
@@ -35,6 +35,13 @@ function monitor() {
     snyk.id = options.id;
   }
 
+
+  // This is a temporary check for gradual rollout of subprojects scanning
+  // TODO: delete once supported for monitor
```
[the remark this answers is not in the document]
orsagie: 😱

src/lib/snyk-test/run-test.ts
orsagie: yeah, I was not sure what the pretty way to do this is, any way I wrote it seemed like it was a hack. But I will pin down the type to be more clear

help/help.txt
orsagie: Or you meant `Gradle` in the description, changing

src/cli/commands/test.ts
orsagie: `test` can be run with a callback, So I don't want to break functionality for those that are using us via script and expecting a single object back

src/lib/snyk-test/index.js
orsagie: see above: it's not about breaking tests per-se, it's about potentially breaking customers using `test` with callback

src/lib/snyk-test/run-test.ts
```diff
@@ -93,7 +40,101 @@ interface Payload {
...
+  targetFile?: string;
+}
+
+interface Package {
+  plugin: PluginMetadata;
+  depRoots?: DepRoot[]; // currently only returned by gradle
+  package?: DepTree;
+}
+
+async function runTest(packageManager: string, root: string, options): Promise<object[]> {
```
kyegupov: It was above "interface payload", so it's harder to read the diff now. What is the logic change here?

test/acceptance/cli.acceptance.test.ts
```diff
@@ -550,6 +551,95 @@ test('`test gradle-app` returns correct meta', async (t) => {
     'local policy not displayed');
 });
 
+test('`test gradle-app --all-sub-projects` returns sends `multiDepRoots` argument to plugin', async (t) => {
```
kyegupov: extra word: "returns sends"

test/acceptance/cli.acceptance.test.ts
```diff
@@ -550,6 +551,95 @@ test('`test gradle-app` returns correct meta', async (t) => {
...
+    async inspect() {
+      return {plugin: {}};
+    },
+  };
+  sinon.spy(plugin, 'inspect');
+  const loadPlugin = sinon.stub(plugins, 'loadPlugin');
+  t.teardown(loadPlugin.restore);
+  loadPlugin.withArgs('gradle').returns(plugin);
+
+  try {
```
kyegupov: note: a shorter way to do it is to use t.rejects(); although the test won't be async then and you would have to use t.plan()

test/acceptance/cli.acceptance.test.ts
```diff
@@ -633,9 +633,10 @@ test('`test gradle-app --all-sub-projects` returns correct multi tree meta', asy
     t.match(meta[1], /Package manager:\s+gradle/,
       'package manager displayed');
     t.match(meta[2], /Target file:\s+build.gradle/, 'target file displayed');
-    t.match(meta[3], /Open source:\s+no/, 'open source displayed');
-    t.match(meta[4], /Project path:\s+gradle-app/, 'path displayed');
-    t.notMatch(meta[5], /Local Snyk policy:\s+found/,
+    t.match(meta[3], /Sub project:\s+tree(1|2)/, 'subproject displayed');
```
orsagie: Here is the visual change (only an additional line in the resulting meta): 
`Sub project: tree1` will be displayed for the first result, and `Sub project: tree2` for the second

src/lib/snyk-test/run-test.ts
orsagie: I like to have interfaces next to where they are used. Can move back if you have a reasoning for it being there.

test/acceptance/cli.acceptance.test.ts
orsagie: edited

test/acceptance/cli.acceptance.test.ts
orsagie: leaving as is  🙃

src/lib/snyk-test/run-test.ts
```diff
@@ -210,6 +209,13 @@ async function getDepsFromPlugin(root, options): Promise<MultiRootsPackage> {
       inspectRes.package.targetFile = inspectRes.plugin.targetFile;
     }
     inspectRes.depRoots = [{depTree: inspectRes.package}];
+  } else {
+    // if we have multiple dep roots, we are dealing with a multi project solution
+    // add the name of the projects to the options?
+    // how do we know to add the right one??? we are returning all the result at one time, but the display
+    // does some weird shit with the testOpts
```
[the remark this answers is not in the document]
orsagie: I accidentally commited this comment. This is just me rambling until I had an idea of what to do

PR #414, opened 2019-03-29
fix: Raise get-uri version

robcresswell:
This prevents a vulnerable version of extend (<3.0.2) being pulled in during
deduping.

See https://snyk.io/vuln/npm:extend:20180424


adrukh:
So maybe wait till #415 is released, and then bump the `proxy-agent` deps and get rid of `get-uri` that was added as a hack to pin its version?

adrukh:
Obsolete with #418

PR #415, opened 2019-03-29
feat: Update unsupported runtimes

robcresswell:
This PR raises our minimum support Node version to 6. See https://snyk.io/blog/snyk-cli-drops-support-for-node-js-4-argon/ for more detail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
   - "10"
   - "8"
   - "6"
-  - "4"
 cache:
   directories:
     - node_modules
--- a/src/cli/runtime.ts
+++ b/src/cli/runtime.ts
@@ -1,8 +1,8 @@
-import {gte} from 'semver';
+import { gte } from 'semver';
 
-const MIN_RUNTIME = '4.0.0';
+const MIN_RUNTIME = '6.5.0';
 
-export const supportedRange = '>= 4';
+export const supportedRange = `>= ${MIN_RUNTIME}`;
 
 export function isSupported(runtimeVersion) {
   return gte(runtimeVersion, MIN_RUNTIME);
--- a/test/runtime.test.js
+++ b/test/runtime.test.js
@@ -4,9 +4,11 @@ var runtime = require('../src/cli/runtime');
 test('nodejs runtime versions support ', function (t) {
   t.ok(runtime.isSupported(process.versions.node),
     'Current runtime is supported');
-  t.ok(runtime.isSupported('4.0.0'), '4.0.0 is supported');
+  t.ok(runtime.isSupported('6.16.0'), '6.16.0 is supported');
   t.ok(runtime.isSupported('11.0.0-pre'), 'pre-release is supported');
   t.notOk(runtime.isSupported('0.10.48'), '0.10 is not supported');
   t.notOk(runtime.isSupported('0.12.18'), '0.12 is not supported');
+  t.notOk(runtime.isSupported('4.0.0'), '4.0.0 is not supported');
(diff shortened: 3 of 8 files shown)


src/cli/runtime.ts
```diff
@@ -1,9 +1,9 @@
...
 
-const MIN_RUNTIME = '4.0.0';
+const MIN_RUNTIME = '6';
 
-export const supportedRange = '>= 4';
+export const supportedRange = `>= ${MIN_RUNTIME}`;
 
 export function isSupported(runtimeVersion) {
-  return gte(runtimeVersion, MIN_RUNTIME);
+  return satisfies(runtimeVersion, supportedRange);
```
darscan: satisfies does not do what you think it does, please revert :trollface:

.travis.yml
```diff
@@ -20,8 +19,6 @@ install:
   - npm install
 script:
   - npm run snyk-auth
-  # tests are failing on npm@5, downgrade to npm@3 if >3 is installed
-  - test "$(npm -v | cut -d'.' -f1)" -le 3 || npm i -g npm@^3.0.0
```
adrukh: This is actually more needed then ever, since nodejs6 and up bundle npm@5 and up (iirc)

src/cli/runtime.ts
```diff
@@ -1,8 +1,8 @@
-import {gte} from 'semver';
+import { gte } from 'semver';
 
-const MIN_RUNTIME = '4.0.0';
+const MIN_RUNTIME = '6.5.0';
```
adrukh: I remembered `6.0.0` was not good due to some good reasons, can we elaborate what they are?

test/runtime.test.js
```diff
@@ -4,9 +4,10 @@ var runtime = require('../src/cli/runtime');
 test('nodejs runtime versions support ', function (t) {
   t.ok(runtime.isSupported(process.versions.node),
     'Current runtime is supported');
-  t.ok(runtime.isSupported('4.0.0'), '4.0.0 is supported');
+  t.ok(runtime.isSupported('6.16.0'), '6.16.0 is supported');
   t.ok(runtime.isSupported('11.0.0-pre'), 'pre-release is supported');
   t.notOk(runtime.isSupported('0.10.48'), '0.10 is not supported');
   t.notOk(runtime.isSupported('0.12.18'), '0.12 is not supported');
+  t.notOk(runtime.isSupported('4.0.0'), '4.0.0 is not supported');
```
adrukh: Maybe also test that `6.4.0` is not supported, just to assert the actual value of `6.5.0` that is being used?

.travis.yml
robcresswell: Do we still need to make sure tests run on npm@3 though? I assumed this was some workaround...

test/runtime.test.js
robcresswell: Sure will do

src/cli/runtime.ts
robcresswell: Related to JS compatibility. `6.0.0` doesn't have full ES2015 support IIRC, which we wanted.

.travis.yml
adrukh: That's exactly the thing... Our tests fixtures assume npm@3 logic when it comes to deduping deps etc. Newer versions of npm break some of the older tests.

adrukh:
Let's also update https://github.com/snyk/snyk/blob/master/package.json#L22 please

.travis.yml
robcresswell: Ahh, I see, so we'd need to update fixture. I'll revert tomorrow

PR #417, opened 2019-03-31
fix: create copies of original files prior to patching

adrukh:
- [x] Ready for review

#### What does this PR do?

Fixes a bug where the patched files object was iterated over as an array `for ... of`, instead of an object `for ... in`.

This revealed another issue where the `.orig` copies of patched files were not being created. This is needed as our logic undoes a patch using the `.orig` pre-patched file when the same patch is re-applied twice. This is needed in cases where a protected project is being protected again, for the chance the used patch has been updated in the meantime.

😌

### DIFF
--- a/src/lib/protect/apply-patch.js
+++ b/src/lib/protect/apply-patch.js
@@ -8,7 +8,7 @@ var fs = require('fs');
 var uuid = require('uuid/v4');
 var errorAnalytics = require('../analytics').single;
 
-function applyPatch(patch, vuln, live, patchUrl) {
+function applyPatch(patchFileName, vuln, live, patchUrl) {
   var cwd = vuln.source;
 
   return new Promise(function (resolve, reject) {
@@ -27,7 +27,7 @@ function applyPatch(patch, vuln, live, patchUrl) {
       debug('Failed loading package.json of package about to be patched', err);
     }
 
-    var patchContent = fs.readFileSync(path.resolve(relative, patch), 'utf8');
+    var patchContent = fs.readFileSync(path.resolve(relative, patchFileName), 'utf8');
 
     jsDiff(patchContent, relative, live).then(function () {
       debug('patch succeed');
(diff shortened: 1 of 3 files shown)


src/lib/protect/apply-patch.js
```diff
@@ -49,6 +49,12 @@ function jsDiff(patchContent, relative, live) {
           if (patchedFiles[fileName]) {
             return callback(null, patchedFiles[fileName]);
           }
+
+          // create an `.orig` copy of the file prior to patching it
+          // used in case we need to revert a patch
+          fs.copyFileSync(path.resolve(relative, fileName),
+            path.resolve(relative, fileName) + '.orig');
```
adrukh: This is the actual fix. When we load a file prior to patching it, we create an `'.orig'` copy of it.

src/lib/protect/apply-patch.js
```diff
@@ -92,7 +98,7 @@ function jsDiff(patchContent, relative, live) {
         }
         try {
           // write patched files back to disk, unlink files completely removed by patching
-          for (var fileName of patchedFiles) {
+          for (var fileName in patchedFiles) {
```
adrukh: This is yet another fix that was very much due.

src/lib/protect/apply-patch.js
```diff
@@ -8,7 +8,7 @@ var fs = require('fs');
 var uuid = require('uuid/v4');
 var errorAnalytics = require('../analytics').single;
 
-function applyPatch(patch, vuln, live, patchUrl) {
+function applyPatch(patchFile, vuln, live, patchUrl) {
```
michael-go: `patchFileName`?

src/lib/protect/apply-patch.js
michael-go: so ... we didn't create these at all? or is this a special flow where we forgot to?

src/lib/protect/apply-patch.js
adrukh: Will do

src/lib/protect/apply-patch.js
adrukh: We didn't! I looked back, prior to the `jsdiff` change, and could NOT find a commit where `.orig` files were being created.

adrukh: GRRRR `fs.copyFileSync` added only in nodejs 8.5.0. Will replace.

PR #418, opened 2019-03-31
feat: bump deps

adrukh:
- [x] Ready for review

#### What does this PR do?

Now that we support nodejs 6.5.0 and upwards, a generous dep bump to catch up with time.
Notable changes:
* `inquirer` bump from `3` to `6` will resolve Windows 10 `snyk wizard` hanging issues
* `proxy-agent` bump from `2` to `3` will remove a vulnerable transitive dependency


package.json
```diff
@@ -83,17 +82,17 @@
...
     "update-notifier": "^2.5.0",
     "uuid": "^3.3.2"
   },
   "devDependencies": {
     "@types/diff": "^3.5.2",
-    "@types/lodash": "^4.14.121",
-    "@types/node": "^10.10.0",
+    "@types/lodash": "^4.14.123",
     "@types/needle": "^2.0.4",
+    "@types/node": "^10.14.4",
```
robcresswell: This should be in line with `.nvmrc`: so `8.x`

robcresswell: Although thinking this, shouldn't we make `.nvmrc` the same as the minimum version we support? (6?)

adrukh: I'll align with `8`, but we shouldn't use `6` - I think the idea is 'what version should we develop in', not 'we still support an EOL-ed version'.

adrukh: Nope, gotta have `10`, using `8` fails `tsc` with:
```
node_modules/type-fest/index.d.ts:56:10 - error TS2339: Property 'observable' does not exist on type 'SymbolConstructor'.

56  [Symbol.observable](): ObservableLike;
            ~~~~~~~~~~
```

michael-go: I think `.nvmrc` should be 6 - so that code that isn't supported in 6 will break locally too. This will save time waiting for the tests to fail on Travisl

adrukh: Updated!

PR #424, opened 2019-04-02
fix: do not unpatch unwanted vulns when patching other vulns

adrukh:
- [x] Ready for review

#### What does this PR do?

When patching vulnerabilities, we would restore previously patched dependencies to re-apply patches. The 'restore' process involves renaming backup files that were created during the previous patching process over their patched counterparts. We used to use `recursive-readdir` for this, which reads all the files in the patched dependency folder. Improving the performance of this flow by using `glob` instead, asking it to fetch relevant files only.

Taking the opportunity to fix a bug where we would unintentionally unpatch a patched transitive dependency while patching its parent dependency by eagerly traversing `node_modules` folders.


src/lib/protect/patch.js
```diff
@@ -67,18 +67,18 @@ function patch(vulns, live) {
               ' | Restoring file back to original to apply the patch again');
               // else revert the patch
               return new Promise(function (resolve, reject) {
-                recursive(vuln.source, function (error, files) {
+                // find all backup files that do not belong to transitive deps
+                glob('**/*.orig', {cwd: vuln.source, ignore: '**/node_modules/**'}, function (error, files) {
```
lili2311: ❤️

src/lib/protect/patch.js
```diff
@@ -67,18 +67,18 @@ function patch(vulns, live) {
...
-                  var origFiles = files.filter(function (file) {
-                    return file.slice(-5) === '.orig';
-                  });
-
-                  for (var file of origFiles) {
-                    fs.renameSync(file, path.dirname(file) + '/' +
-                      path.basename(file).slice(0, -5));
+                  // copy '.orig' backups over the patched files
+                  for (var file of files) {
+                    var backupFile = path.resolve(vuln.source, file);
```
lili2311: maybe use `const` instead?

test/protect-apply-same-patch-again.test.js
```diff
@@ -31,10 +31,27 @@ const patch = proxyquire('../src/lib/protect/patch', {
...
+const transitiveDepBackupFilePath = path.resolve(alreadyPatchedTransitiveDepFolder, 'somecode.js.orig');
+const transitiveDepBackupContent = 'console.log("some original code here")';
+const transitiveDepPatchFlagPath = path.resolve(alreadyPatchedTransitiveDepFolder, '.some-vuln-id.flag');
+const transitiveDepPatchFlagContent = new Date().toISOString();
+
 test('setup', (t) => {
-  var fixture = fs.readFileSync(fixturesModuleFolder + '/node-fixture.js');
-  fs.writeFileSync(fixturesModuleFolder + '/node.js', fixture);
-  t.pass();
+  var fixture = fs.readFileSync(path.resolve(fixturesModuleFolder, 'node-fixture.js'));
```
lili2311: also `const` perhaps?

test/protect-apply-same-patch-again.test.js
```diff
@@ -89,15 +115,25 @@ test('Same patch is applied multiple times without issue', (t) => {
       // that need patching, both successfully patched
 
       const vulnsToPatch = res.patch[packagesToPatch[0]];
-      var pacthedVulns = 0;
+      var patchedVulns = 0;
```
lili2311: `let`?

PR #456, opened 2019-04-25
feat: Release improved project naming for csproj behind a flag

lili2311:
- [x] Ready for review

#### What does this PR do?
Bump the nuget plugin to release improved naming behind the flag `--assets-project-name`
```snyk monitor --file=pathtosln --org=snykorg --assets-project-name```
https://github.com/snyk/snyk-nuget-plugin/pull/50

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -39,26 +39,35 @@ Options:
                        to true). Applicable to `snyk test`.
   --project-name=<string>
                        Specify a custom Snyk project name (`snyk monitor` only).
+  --policy-path ...... Manually pass a path to a policy file.
+  --insecure ......... Ignore unknown certificate authorities.
+  --json ............. Return results in JSON format.
+  --dry-run .......... Don't apply updates or patches during protect.
+  --severity-threshold=<low|medium|high>
+                       Only report vulnerabilities of provided level or higher.
+  -q, --quiet ........ Silence all output.
+  -h, --help ......... This help information.
+  -v, --version ...... The CLI version.
+
+Gradle options:
+  --gradle-sub-project=<string>
+                       For Gradle "multi project" configurations,
+                       test a specific sub-project.
+  --all-sub-projects   For "multi project" configurations, test all
+                       sub-projects.
+
+.Net (Nuget) options:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by adrukh** [comment]
@lili2311 do we want to refer to this shiny new flag in the help text?

File: help/help.txt [hunk 1]
```diff
@@ -56,6 +56,9 @@ Options:
                        test a specific sub-project.
   --all-sub-projects   For "multi project" configurations, test all
                        sub-projects.
+  --assets-project-name
+                       For Nuget projects, forces us to try using the project name
+                       from the project.assets.json file when possible.
```
adrukh:
```
For Nuget project, forces us to try using the project name from the project.assets.json file when possible.
```
While very accurate and true, I'm still left a bit puzzled - what happens when I don't use this option? Also, the use of `forces us` is a bit off to me :)

Suggestion:
```
When monitoring a .NET project, use the project.assets.json name, if found.
```

nrodrigues1: We might want to specify .NET project using NuGet PackageReference project style. 
If you scan a .NET framework that use packages.config insteand of PackageReference, the argument will have no effects.

Also, we could specify that we check for the projectname in project.assets.json or something like that. "Use the project.assets.json project name" might be more accurate

orsagie: Can we actually move this documentation to the plugin? 

We have many options that are plugin specific, and that is where they should live until we decide otherwise.

help/help.txt
```diff
@@ -39,26 +39,35 @@ Options:
...
                        Only report vulnerabilities of provided level or higher.
+  -q, --quiet ........ Silence all output.
+  -h, --help ......... This help information.
+  -v, --version ...... The CLI version.
+
+Gradle options:
   --gradle-sub-project=<string>
                        For Gradle "multi project" configurations,
                        test a specific sub-project.
   --all-sub-projects   For "multi project" configurations, test all
```
orsagie: this one is CLI-wide. but only gradle uses it at the moment. It's fine here now, but we should remember to move it when we support other PM's like yarn workspaces etc.

PR #462, opened 2019-04-30
feat: add git metadata to monitor payload

odinn1984:
- [ ] Ready for review

#### What does this PR do?
We are adding git metadata to monitor when sending the payload to snyk.io for grouping CLI projects

### DIFF
--- a/src/lib/monitor.ts
+++ b/src/lib/monitor.ts
@@ -7,6 +7,8 @@ import * as _ from 'lodash';
 import * as isCI from './is-ci';
 import * as analytics from './analytics';
 import { SingleDepRootResult, MonitorError } from './types';
+import * as projectMetadata from './project-metadata';
+import * as path from 'path';
 
 // TODO(kyegupov): clean up the type, move to snyk-cli-interface repository
 
@@ -14,6 +16,8 @@ interface MonitorBody {
   meta: Meta;
   policy: string;
   package: {}; // TODO(kyegupov): DepTree
+  target: {};
+  targetFileRelativePath: string;
   targetFile: string;
 }
 
@@ -35,7 +39,7 @@ interface Meta {
   projectName: string;
 }
 
-export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
+export function monitor(root, meta, info: SingleDepRootResult, targetFile): Promise<any> {
   const pkg = info.package;
   const pluginMeta = info.plugin;
   let policyPath = meta['policy-path'];
@@ -52,8 +56,12 @@ export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
         return snyk.policy.create();
       }
       return snyk.policy.load(policyLocations, opts);
-    }).then((policy) => {
+    }).then(async (policy) => {
       analytics.add('packageManager', packageManager);
+
+      const target = await projectMetadata.getInfo(pkg);
+      const targetFileRelativePath = targetFile ? path.relative(root, targetFile) : '';
+
       // TODO(kyegupov): async/await
       return new Promise((resolve, reject) => {
         request({
@@ -79,7 +87,9 @@ export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
             package: pkg,
             // we take the targetFile from the plugin,
             // because we want to send it only for specific package-managers
+            target,
             targetFile: pluginMeta.targetFile,
+            targetFileRelativePath,
           } as MonitorBody,
           gzip: true,
           method: 'PUT',
--- a/src/lib/project-metadata/index.ts
+++ b/src/lib/project-metadata/index.ts
@@ -1,0 +1,18 @@
+import * as gitTargetBuilder from './target-builders/git';
+import { GitTarget } from './types';
+
+const TARGET_BUILDERS = [
+  gitTargetBuilder,
+];
+
+export async function getInfo(packageInfo): Promise<GitTarget|null> {
+  for (const builder of TARGET_BUILDERS) {
+    const target = await builder.getInfo(packageInfo);
+
+    if (target) {
+      return target;
+    }
+  }
+
+  return null;
+}
--- a/src/lib/project-metadata/target-builders/git.ts
+++ b/src/lib/project-metadata/target-builders/git.ts
@@ -1,0 +1,33 @@
+import fs = require('fs');
+import GitUrlParse = require('git-url-parse');
+
+import subProcess = require('../../sub-process');
+import { GitTarget } from '../types';
+
+export async function getInfo(packageInfo): Promise<GitTarget|null> {
+  let origin: string|null|undefined;
+
+  if (packageInfo.docker) {
+    return null;
+  }
+
+  try {
+    origin = (await subProcess.execute('git', ['remote', 'get-url', 'origin'])).trim();
+
+    if (!origin) {
+      return null;
+    }
+
+    const parsedOrigin = GitUrlParse(origin);
+    const branch = (await subProcess.execute('git', ['rev-parse', '--abbrev-ref', 'HEAD'])).trim();
+
+    return {
+      remoteUrl: parsedOrigin.toString('http'),
+      branch,
+    };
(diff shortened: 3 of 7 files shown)


src/lib/project-metadata/git.ts
```diff
@@ -0,0 +1,43 @@
...
+  return {
+    repo: hostedGitMetadata.project,
+    owner: hostedGitMetadata.user,
+    branch,
+    commitSha,
+    targetFile: gitRelativeTargetFile,
+  };
+}
+
+function findGitRoot(): string {
```
darscan: please consider `git rev-parse --show-toplevel`

src/lib/monitor.ts
```diff
@@ -52,8 +54,11 @@ export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
         return snyk.policy.create();
       }
       return snyk.policy.load(policyLocations, opts);
-    }).then((policy) => {
+    }).then(async (policy) => {
       analytics.add('packageManager', packageManager);
+
+      const gitTarget = await projectMetadata.getInfo('git', root, targetFile) as GitInfo;
```
lili2311: Can this targetfile ever be different to the https://github.com/snyk/snyk/pull/462/files#diff-e2e0f6cf5732ba336dc3ba0de17258faR88?

Will it matter? Not all plugins send targetFile and some choose what is best to process in the end

src/lib/project-metadata/git.ts
odinn1984: 👍

src/lib/monitor.ts
odinn1984: @lili2311 the difference here would be that this is the relative path to the target file (relative to the git project root) and we pass it to monitor later in the chain. i think it would make more sense to present the file in this way in snyk.io. wdyt? basically meaning that if u have a monorepo for example we will save the target file as the relative path from the top level dir to the file even if ur running it inside of a sub dir

gjvis: see other comment, but commenting here too: how does this know already that this is `git`?

src/lib/project-metadata/index.ts
```diff
@@ -0,0 +1,13 @@
+import * as gitMetadata from './git';
+import { GitInfo } from './types';
+
+export async function getInfo(metadataSource: string, root: string, targetFile: string): Promise<GitInfo|null> {
```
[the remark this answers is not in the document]
odinn1984: 1. yes the monitor is responsible for adding the monitor data, I wanted this factory to return only target related data
2 and 3. so would it make sense to do something like... does a git command work then it's git if not try to see if we have a docker argument etc?
4. if the root dir changes that's fine no since we're just using it to get the relative path to the target file?

src/lib/project-metadata/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface GitInfo {
+  repo: string;
```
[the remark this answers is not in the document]
odinn1984: if we do a remote url (which was origin URL in michael's original PR) it means that we're leaving the parsing part to registry to build the target... or we can just use the URL somehow to represent the project at registry?

src/lib/project-metadata/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface GitTarget {
+  user: string;
+  project: string;
+  branch: string;
+}
```
gjvis: this is still too hosted-git centric. A good test case (direct from https://git-scm.com/) is `user@server:path/to/repo.git` -> that should be easy to unambiguously represent in this `GitTarget` type

noticed that the `hosted-git-info` only supports things cloned from GitHub, GitLab, and Bitbucket so its not really appropriate here. Perhaps https://www.npmjs.com/package/git-url-parse is better? Or something else (I only searched for 2 minutes... :joy:)

Maybe an `https` representation is the most appropriate? there's not right answer here :'(

src/lib/project-metadata/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { GitTarget } from './types';
+
+export async function getInfo(packageInfo): Promise<GitTarget|null> {
+  // This is meant to be a factory once we introduce more flavours like Docker for example
+  // currently only handles git. The docker is just an example on how to expand it
+  if (packageInfo.docker) {
+    return null;
+  } else if ((await subProcess.execute('git', ['remote', 'get-url', 'origin'])).trim()) {
+    return gitMetadata.getInfo();
+  }
```
gjvis: This is crossing into the specific-code territory, but while i'm here:

```typescript
const TARGET_BUILDER = [gitMetadata, dockerImageMetadata];

// ...
for (const targetBuilder in TARGET_BUILDERS) {
  const target = targetBuilder.getInfo(packageInfo);
  if (target) {
    return target;
  }
}
```

an implementation like that lets you add new target builders (or some other term, not precious about the name)  easily.

They return a target, or undefined if they can't, and you get to keep all target-specific knowledge out of this file, its responsibility becomes simply the registration and order-of-preference for each target builder.

src/lib/project-metadata/types.ts
gjvis: regarding your other comment about registry needing to do processing - I don't follow? As long as the targets are both standardised and hold the git upstream repo details then we can both group by targets and support the future desire to match up CLI with hosted source code projects in some way.

what am I missing?

src/lib/monitor.ts
lili2311: What happens when a user runs a snyk test with a file `../../somefile`? After checking analytics this happens today!

lili2311:
@gjvis @odinn1984 

The plugin will always return a file that was used in the end to generate the Tree, the plugin has better understanding to do this.

However not all plugins return it relevant to the git root of course.

I think if we for now send what you propose we can then check the data for discrepancies. For example some users run test via `snyk test ../../somefile` What will this translate to in the git target?

Do we have a test for this?

odinn1984:
@lili2311 this would be translated to the relative path from the root if we chose to send that, the question here is if it's useful to send that as well.

We're not in the tests stage yet, I will add tests once all the questions are answered :)

src/lib/project-metadata/types.ts
odinn1984: yeah, that's a good point... I think I was overcomplicating things here because my head was too hosted solution oriented...

src/lib/project-metadata/target-builders/git.ts
```diff
@@ -0,0 +1,33 @@
+import fs = require('fs');
+import GitUrlParse = require('git-url-parse');
+
+import subProcess = require('../../sub-process');
+import { GitTarget } from '../types';
+
+/* tslint:disable:no-unused-variable */
```
ArturSnyk: is this necessary?

src/lib/project-metadata/target-builders/git.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const parsedOrigin = GitUrlParse(origin);
+    const branch = (await subProcess.execute('git', ['rev-parse', '--abbrev-ref', 'HEAD'])).trim();
+
+    return {
+      name: parsedOrigin.toString('http'),
+      branch,
+    };
+  } catch (err) {
+    return null;
```
ArturSnyk: can we handle it differently? maybe log reason for erroring

src/lib/monitor.ts
```diff
@@ -52,8 +56,12 @@ export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
         return snyk.policy.create();
       }
       return snyk.policy.load(policyLocations, opts);
-    }).then((policy) => {
+    }).then(async (policy) => {
```
jonathan-fielding: why add async

jonathan-fielding: this is returning a promise not using async await

odinn1984: @jonathan-fielding if you look a few lines below I'm using await there, an async function is a promise behind the scenes so this still holds true and works fine.

PR #466, opened 2019-04-30
fix: do not offer remediation advice when scanning a non-local package

kyegupov:
- [x] Ready for review

#### What does this PR do?
Currently we display remediation advice even when a non-local package is scanned.
We should not, since the user does not own the package and cannot fix it.

https://snyksec.atlassian.net/browse/BST-591

image

### DIFF
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -9,16 +9,21 @@ import {exists as apiTokenExists} from '../../lib/api-token';
 import {SEVERITIES, WIZARD_SUPPORTED_PMS} from '../../lib/snyk-test/common';
 import * as docker from '../../lib/docker-promotion';
 import * as Debug from 'debug';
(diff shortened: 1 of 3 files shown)


src/cli/commands/test.ts
```diff
@@ -9,15 +9,25 @@ import {exists as apiTokenExists} from '../../lib/api-token';
 import {SEVERITIES, WIZARD_SUPPORTED_PMS} from '../../lib/snyk-test/common';
 import * as docker from '../../lib/docker-promotion';
 import * as Debug from 'debug';
+import {isLocalFolder} from '../../lib/detect';
 
 const debug = Debug('snyk');
 const SEPARATOR = '\n-------------------------------------------------------\n';
 
+interface OptionsAtDisplayStage {
```
lili2311: ❤️

PR #487, opened 2019-05-07
fix: display monitor error code in the error message

kyegupov:
- [x] Ready for review

#### What does this PR do?

Improves the error message to make debugging easier.

The existing error just produces `unexpected error: undefined` in the stack trace, which is not helpful.

### DIFF
--- a/src/lib/monitor.ts
+++ b/src/lib/monitor.ts
@@ -97,7 +97,8 @@ export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
           if (res.statusCode === 200 || res.statusCode === 201) {
             resolve(body);
           } else {
-            const e = new MonitorError('unexpected error: ' + body.message);
+            const e = new MonitorError('Server returned unexpected error for the monitor request. ' +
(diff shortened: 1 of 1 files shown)


src/lib/monitor.ts
```diff
@@ -97,7 +97,8 @@ export function monitor(root, meta, info: SingleDepRootResult): Promise<any> {
           if (res.statusCode === 200 || res.statusCode === 201) {
             resolve(body);
           } else {
-            const e = new MonitorError('unexpected error: ' + body.message);
+            const e = new MonitorError('Server returned unexpected error for the monitor request. ' +
+              `Status code: ${res.statusCode}, response: ${res.body}`);
```
lili2311: this doesn't seem safe to display the whole body, please keep it to only `body.message` as before as we could be showing too much info.  
If `userMessage` is present we prefer it.
What if there is no `res.statuscode`? We don't always have it and it is not always `statusCode`

kyegupov: Ok, switched to userMessage/message. `statusCode` will always be present, because it's generated by the `request` library.

PR #528, opened 2019-05-21
feat: vulns exit code 1, errors 2

lili2311:
- [x] Ready for review

#### What does this PR do?
- Add error code 2 for all failure states apart from *vulns found*, give's user a chance to act on this and retry if needed
- Stop treating *vulns* as an error that is treated as a `bad-command`, stop further processing and display vuln results.

Start in command/test

Run snyk locally and try a few failed commands vs normal test runs

We send all the vuln found commands back and pollute our analytics, it is hard to differentiate between actual failures in this way

MultipleTestsExistCodes2

### DIFF
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -153,7 +153,8 @@ async function test(...args): Promise<string> {
       summariseErrorResults(errorResults) + '\n';
   }
 
-  const notSuccess = vulnerableResults.length > 0 || errorResults.length > 0;
+  const notSuccess = errorResults.length > 0;
+  const foundVulnerabilities = vulnerableResults.length > 0;
 
   if (notSuccess) {
     response += chalk.bold.red(summaryMessage);
@@ -162,9 +163,15 @@ async function test(...args): Promise<string> {
     // translation
     // HACK as there can be different errors, and we pass only the
     // first one
-    error.code = (vulnerableResults[0] || errorResults[0]).code;
-    error.userMessage = (vulnerableResults[0] || errorResults[0]).userMessage;
+    error.code = errorResults[0].code;
+    error.userMessage = errorResults[0].userMessage;
     throw error;
+  }
+
+  if (foundVulnerabilities) {
+    response += chalk.bold.red(summaryMessage);
+    console.error(response);
+    process.exit(1);
   }
 
   response += chalk.bold.green(summaryMessage);
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -38,12 +38,9 @@ async function runCommand(args) {
 
 async function handleError(args, error) {
   spinner.clearAll();
-  let command = 'bad-command';
+  const command = 'bad-command';
 
-  if (error.code === 'VULNS') {
-    // this isn't a bad command, so we won't record it as such
-    command = args.command;
-  } else if (!error.stack) { // log errors that are not error objects
+  if (!error.stack) { // log errors that are not error objects
     analytics.add('error', JSON.stringify(error));
     analytics.add('command', args.command);
   } else {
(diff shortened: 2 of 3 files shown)


src/lib/errors/legacy-errors.js
```diff
@@ -71,10 +69,6 @@ module.exports = function error(command) {
 module.exports.message = function(error) {
   let message = error; // defaults to a string (which is super unlikely)
   if (error instanceof Error) {
-    if (error.code === 'VULNS') {
```
lili2311: this is not set anywhere

src/lib/errors/legacy-errors.js
```diff
@@ -46,7 +45,6 @@ const codes = {
   404: errors.notfound,
   411: errors.endpoint, // try to post to a weird endpoint
   403: errors.endpoint,
-  401: errors.auth,
```
lili2311: this error was refactored into a new custom error, removing from legacy errors

src/cli/commands/test.ts
```diff
@@ -162,11 +163,17 @@ async function test(...args): Promise<string> {
...
-    error.userMessage = (vulnerableResults[0] || errorResults[0]).userMessage;
+    error.code = errorResults[0].code;
+    error.userMessage = errorResults[0].userMessage;
     throw error;
   }
 
+  if (foundVulnerabilities) {
+    response += chalk.bold.red(summaryMessage);
+    console.error(response);
+    process.exit(1);
```
darscan: I don't think we can exit like this.. due to things like `sln` and the multi test thing (`snyk test folder1 folder2 folder3`)

kyegupov: I think both are handled ok: `for (const path of args) {` is done by the line 104 and sln.updateArgs already was applied to args.

src/cli/index.ts
```diff
@@ -38,12 +38,9 @@ async function runCommand(args) {
...
-  let command = 'bad-command';
+  const command = 'bad-command';
 
-  if (error.code === 'VULNS') {
-    // this isn't a bad command, so we won't record it as such
-    command = args.command;
-  } else if (!error.stack) { // log errors that are not error objects
+  if (!error.stack) { // log errors that are not error objects
     analytics.add('error', JSON.stringify(error));
     analytics.add('command', args.command);
```
kyegupov: is "bad-command" actually of any use as the analytics data?

src/cli/index.ts
```diff
@@ -55,6 +52,10 @@ async function handleError(args, error) {
     if (error.message && error.message.vulnerabilities) {
       delete error.message.vulnerabilities;
     }
+    if (error.vulnerabilities) {
+      delete error.message.vulnerabilities;
```
kyegupov: delete error.vulnerabilities

src/cli/index.ts
```diff
@@ -132,15 +133,15 @@ async function main() {
   }
 
   if (!process.env.TAP && failed) {
-    process.exit(1);
+    process.exit(2);
```
kyegupov: 1. Maybe extract EXIT_CODE_VULNS and EXIT_CODE_ERRORS?

2. These need to be documented in the main README.md of the repo

src/cli/index.ts
lili2311: yes it is every actual error + vulns and we have dashboardsrelying on this

src/cli/commands/test.ts
lili2311: Will verify, also need to test `sln` which relies on multiple results back too

lili2311: Multiple test runs works :D Will document it all and add screenshots / gifs. Need to check SLN

lili2311: That also seems to work, will look into the tests

PR #542, opened 2019-05-29
Make human-readable error message if --file value wasn't passed correctly

dkontorovskyy:
- [X] Ready for review

#### What does this PR do?
Gives nice, human-readable error message, if --file value was passed with space, but now as expected (--file=[filename])

From root folder of this project run command:
`node ./dist/cli/index.js test --file package.json`

[BST-606](https://snyksec.atlassian.net/browse/BST-606)

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -15,7 +15,7 @@ import errors = require('../lib/errors/legacy-errors');
 import ansiEscapes = require('ansi-escapes');
 import {isPathToPackageFile} from '../lib/detect';
 import {updateCheck} from '../lib/updater';
-import { MissingTargetFileError } from '../lib/errors/missing-targetfile-error';
+import { MissingTargetFileError, FileFlagBadInputError } from '../lib/errors';
 
 const debug = Debug('snyk');
 const EXIT_CODES = {
@@ -134,9 +134,12 @@ async function main() {
   let failed = false;
   let exitCode = EXIT_CODES.ERROR;
   try {
-    if (args.options.file && (args.options.file as string).match(/\.sln$/)) {
+    if (args.options.file && typeof args.options.file === 'string' && (args.options.file as string).match(/\.sln$/)) {
       sln.updateArgs(args);
+    } else if (typeof args.options.file === 'boolean') {
+      throw new FileFlagBadInputError();
     }
+
     checkPaths(args);
     res = await runCommand(args);
   } catch (error) {
--- a/src/lib/errors/file-flag-bad-input.ts
+++ b/src/lib/errors/file-flag-bad-input.ts
@@ -1,0 +1,12 @@
+import {CustomError} from './custom-error';
+
+export class FileFlagBadInputError extends CustomError {
+    private static ERROR_CODE: number = 422;
+    private static ERROR_MESSAGE: string = 'Empty --file argument. Did you mean --file=path/to/file ?';
+
+    constructor() {
+        super(FileFlagBadInputError.ERROR_MESSAGE);
+        this.code = FileFlagBadInputError.ERROR_CODE;
+        this.userMessage = FileFlagBadInputError.ERROR_MESSAGE;
+    }
+}
--- a/src/lib/errors/index.ts
+++ b/src/lib/errors/index.ts
@@ -1,2 +1,3 @@
-export {NoSupportedManifestsFoundError} from './no-supported-manifests-found';
+export {FileFlagBadInputError} from './file-flag-bad-input';
 export {MissingTargetFileError} from './missing-targetfile-error';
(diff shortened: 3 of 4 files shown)


src/cli/index.ts
```diff
@@ -134,9 +135,12 @@ async function main() {
   let failed = false;
   let exitCode = EXIT_CODES.ERROR;
   try {
-    if (args.options.file && args.options.file.match(/\.sln$/)) {
+    if (args.options.file && !isBoolean(args.options.file) && args.options.file.match(/\.sln$/)) {
       sln.updateArgs(args);
+    } else if (isBoolean(args.options.file)) {
+      throw new Error('--file flag and it\'s value should be separated by equals sign (e.g. --file=package.json)');
```
lili2311: how about something that speaks less about the fact we have `=` instead of ` ` (we want to fix this soon) and instead something like:
`Did you mean `--file=path/to/file`?`

lili2311:
Please add a test for this scenario? We want to assert a nice error message is shown to the user

src/cli/index.ts
lili2311: Also please create this as a custom error, see example: https://github.com/snyk/snyk/tree/master/src/lib/errors we are moving to having all errors defined this way + error code etc

dkontorovskyy: Sounds legit 👍

src/cli/index.ts
```diff
@@ -134,9 +135,12 @@ async function main() {
   let failed = false;
   let exitCode = EXIT_CODES.ERROR;
   try {
-    if (args.options.file && args.options.file.match(/\.sln$/)) {
+    if (args.options.file && !isBoolean(args.options.file) && args.options.file.match(/\.sln$/)) {
```
kyegupov: first, isBoolean has been derpecated: https://nodejs.org/api/util.html#util_util_isboolean_object

second, whitelisting (typeof === string) is generally safer (imagine someone refactoring the args parser so that it allows array arguments)

src/cli/index.ts
kyegupov: I think tslint will suggest you to use `"it's"`

src/cli/index.ts
```diff
@@ -134,9 +134,12 @@ async function main() {
   let failed = false;
   let exitCode = EXIT_CODES.ERROR;
   try {
-    if (args.options.file && args.options.file.match(/\.sln$/)) {
+    if (args.options.file && typeof args.options.file !== 'boolean' && args.options.file.match(/\.sln$/)) {
```
kyegupov: I'd still prefer `typeof === string`

test/acceptance/cli-args.test.ts
```diff
@@ -0,0 +1,10 @@
+import {test} from 'tap';
+import {exec} from 'child_process';
+
+test('should file, cos --file and value are not passed correctly', (t) => {
```
kyegupov: 'snyk test command should fail when --file is not specified correctly'

test/acceptance/cli-args.test.ts
```diff
@@ -0,0 +1,10 @@
...
+    t.plan(1);
+
+    exec('node ./dist/cli/index.js test --file package-lock.json', (_, stdout) => {
+        t.match(stdout, /Did you mean --file=path\/to\/file ?/, 'correct error output');
```
kyegupov: match against a string would be a better idea (you won't have to escape slashes and "?")

test/acceptance/cli-args.test.ts
```diff
@@ -0,0 +1,10 @@
+import {test} from 'tap';
+import {exec} from 'child_process';
+
+test('should file, cos --file and value are not passed correctly', (t) => {
+    t.plan(1);
+
+    exec('node ./dist/cli/index.js test --file package-lock.json', (_, stdout) => {
```
kyegupov: the _ here is the error code. You probably should verify that too.

src/lib/errors/file-flag-bad-input.ts
```diff
@@ -0,0 +1,12 @@
+import {CustomError} from './custom-error';
+
+export class FileFlagBadInput extends CustomError {
+    private static ERROR_CODE: number = 422;
+    private static ERROR_MESSAGE: string = 'Did you mean --file=path/to/file ?';
```
kyegupov: maybe "Emtpy --file argument. Did you mean --file=path/to/file ?"

src/lib/errors/index.ts
```diff
@@ -1,2 +1,3 @@
-export {NoSupportedManifestsFoundError} from './no-supported-manifests-found';
+export {FileFlagBadInput} from './file-flag-bad-input';
```
lili2311: `FileFlagBadInputError` to be consistent? But not a blocker

PR #543, opened 2019-05-29
feat: add branch and pkg name to monitor analytics

odinn1984:
- [ ] Ready for review

#### What does this PR do?

We want to add branch and project name to the data that we send to analytics.

### DIFF
--- a/src/lib/monitor.ts
+++ b/src/lib/monitor.ts
@@ -59,6 +59,10 @@ export async function monitor(root, meta, info: SingleDepRootResult, targetFile)
   const target = await projectMetadata.getInfo(pkg);
   const targetFileRelativePath = targetFile ? path.relative(root, targetFile) : '';
 
+  if (target && target.branch) {
+    analytics.add('targetBranch', target.branch);
+  }
+
   // TODO(kyegupov): async/await
   return new Promise((resolve, reject) => {
     request({


src/lib/monitor.ts
```diff
@@ -62,6 +62,12 @@ export function monitor(root, meta, info: SingleDepRootResult, targetFile): Prom
       const target = await projectMetadata.getInfo(pkg);
       const targetFileRelativePath = targetFile ? path.relative(root, targetFile) : '';
 
+      if (target && target.branch) {
+        analytics.add('branch', target.branch);
+      }
+
+      analytics.add('projectName', pkg.name);
```
lili2311: project name will be calculated in Registry in the end and will unlikely be this name. Another important property is `--project-name` option as it is the project name override that is also used to calculate project name on the other end

odinn1984: thanks @lili2311 that's good to know. so that's why we're not sending it from CLI i guess... ok I'll talk to the BI team about it since it was their request and see if pkg name is enough for them and why they need that. pkg.name will always b there though no?

odinn1984: ok @lili2311 looks like we can send the ID as the project name from my few tests... wdyt?

lili2311: which ID?  Cli never knows if there is an existing project or not until we monitor. So when `monitor` returns an id thats the only time we have something solid. Will that be enough?

odinn1984: the meta.id one... looks like it's similar to the project name? or is it just in npm and maven?

PR #548, opened 2019-06-02
chore: refactor auth to async await

lili2311:
- [x] Ready for review

#### What does this PR do?
- Refactor auth handling to use `async/await` instead of promises
- Slightly refactor to simplify + for readability

`auth/index.ts`

- Build with `npm run build` & then test out using `snyk auth`

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -12,24 +12,32 @@ import { MisconfiguredAuthInCI } from '../../../lib/errors/misconfigured-auth-in
 import { AuthFailedError } from '../../../lib/errors/authentication-failed-error';
 import {verifyAPI} from './is-authed';
 
-module.exports = auth;
-
 const apiUrl = url.parse(config.API);
 const authUrl = apiUrl.protocol + '//' + apiUrl.host;
 const debug = Debug('snyk-auth');
 let attemptsLeft = 0;
+
+module.exports = auth;
 
 function resetAttempts() {
   attemptsLeft = 30;
 }
 
 type AuthCliCommands = 'wizard' | 'ignore';
+interface WebAuthResponse {
+  res: object;
+  body: object;
+}
 
-async function webAuth(via: AuthCliCommands) {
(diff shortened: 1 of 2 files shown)


src/cli/commands/auth/index.ts
```diff
@@ -12,61 +12,56 @@ import { MisconfiguredAuthInCI } from '../../../lib/errors/misconfigured-auth-in
...
-    console.log(msg);
-  } else {
-    return Promise.reject(MisconfiguredAuthInCI());
-  }
-
-  const lbl = 'Waiting...';
+    urlStr + '\n');
 
-  return spinner(lbl).then(() => {
+  try {
```
kyegupov: you are dropping spinner here? Why?

src/cli/commands/auth/index.ts
```diff
@@ -12,61 +12,56 @@ import { MisconfiguredAuthInCI } from '../../../lib/errors/misconfigured-auth-in
...
-
 const apiUrl = url.parse(config.API);
 const authUrl = apiUrl.protocol + '//' + apiUrl.host;
 const debug = Debug('snyk-auth');
 let attemptsLeft = 0;
 
+module.exports = auth;
+
 function resetAttempts() {
   attemptsLeft = 30;
```
kyegupov: 30 is a lot. Do you know how was this number chosen?

lili2311: I don't perhaps worth asking in slack, it does seem like a lot

src/cli/commands/auth/index.ts
lili2311: Two reasons:
- it prints `Waiting...` and then is clears it a few seconds later. Seems not very helpful
- the way previous code was written means it cleared the spinner after getting the value back, i a promise chain couldn't see a nice way to add this into the refactored code as we return the value.

kyegupov: you could return "testAuthComplete().then(spinner.clear)"

But if you feel that the spinner was useless, LET IT DIE.

PR #550, opened 2019-06-03
fix: protect failing due to missing auth

orsagie:
- [x] Ready for review

#### What does this PR do?
Fixes issue where `snyk protect` fails due to missing api token.

The change was reverted here: https://github.com/snyk/snyk/pull/549

This change fixes the underlying issue, and adds a test to make sure protect auth behaviour is as expected.


lili2311:
Please include the original commit in this PR as well?

src/cli/index.ts
```diff
@@ -66,6 +66,7 @@ async function handleError(args, error) {
         copy(result);
         console.log('Result copied to clipboard');
       } else {
+        console.log('error', 'error');
```
lili2311: please remove

src/lib/errors/missing-api-token.ts
```diff
@@ -0,0 +1,13 @@
+import {CustomError} from './custom-error';
+
+export class MissingApiTokenError extends CustomError {
+    private static ERROR_CODE: number = 401;
+    private static ERROR_MESSAGE: string =
+        '`snyk` requires an authenticated account. Please run `snyk auth` and try again.';
+
+    constructor() {
+        super(MissingApiTokenError.ERROR_MESSAGE);
+        this.code = MissingApiTokenError.ERROR_CODE;
```
lili2311: missing `strCode` we need to preserve the error codes as they were please for our dashboard etc example: https://github.com/snyk/snyk/blob/de5346c9a31655f6d84c619fa12db81751624d26/src/lib/errors/token-expired-error.ts#L9

lili2311: `NO_API_TOKEN` was the original code

test/acceptance/protect.test.ts
```diff
@@ -0,0 +1,12 @@
+import {test} from 'tap';
+import {exec} from 'child_process';
+import * as userConfig from '../../src/lib/user-config';
+
+test('`protect` for unsupported package managers', (t) => {
```
lili2311: the test name seems to be completely different?

test/acceptance/protect.test.ts
```diff
@@ -0,0 +1,12 @@
+import {test} from 'tap';
+import {exec} from 'child_process';
+import * as userConfig from '../../src/lib/user-config';
+
+test('`protect` should not fail for unauthorized users', (t) => {
+  t.plan(1);
+  userConfig.delete('api');
+
+  exec('node ./dist/cli/index.js protect', (_, stdout) => {
```
lili2311: do we normally run tests this way? Is it dafe to delete the api prop in testing?

src/lib/index.js
```diff
@@ -10,14 +10,14 @@ module.exports = snyk;
...
-var apiToken = require('./api-token');
+const apiToken = require('./api-token');
 
 // make snyk.api *always* get the latest api token from the config store
 Object.defineProperty(snyk, 'api', {
   enumerable: true,
   configurable: true,
   get: function () {
-    return apiToken();
+    return apiToken.api();
```
kyegupov: that's the key change, right?

test/acceptance/protect.test.ts
dkontorovskyy: We should delete this to test for unauthenticated users. (cos if you have `api` that means you had been authenticated)
We can set it back after this test

kyegupov: I think what we normally do is also in this review, see "test without authentication" below.
We probably should do the similar thing for `protect`?

lili2311: I guess I am wondering if we the order of tests changes, would this break other tests. Would indeed be safer to re-set after the test

dkontorovskyy: @kyegupov @lili2311 make sense. Applying changes

src/lib/index.js
dkontorovskyy: Yeap. This actually caused all failures.

dkontorovskyy:
@kyegupov @lili2311 changes applied

test/cli.test.js
```diff
@@ -249,18 +249,19 @@ test('snyk ignore - not authorized', function (t) {
     .catch((err) => t.pass('no policy file saved'));
 });
 
-test('test without authentication', function (t) {
-  t.plan(1);
-  return cli.config('unset', 'api').then(function () {
-    return cli.test('semver@2');
-  }).then(function (res) {
+test('test without authentication', async (t) => {
+  t.plan(2);
```
kyegupov: you shouldn't need t.plan() for async test functions

lili2311:
tested locally:
- protect ok with no auth
- test fails without auth, then after auth OK

PR #553, opened 2019-06-04
feat: select Gradle confugiration via name/attributes

kyegupov:
- [x] Ready for review

#### What does this PR do?
Gradle plugin now supports selecting configurations via regexes/attributes (especially useful for Android builds).

Drive-by: simplify error analytics.


lili2311:
Please update the help doc for gradle to add the new cli option? later we can link to live docs

kyegupov:
@lili2311 help added. Now, the interesting question: do we rely anywhere on analytics `error-message`? The message is duplicated in the `error` field, so we can retain only that, but do we have any dashdoards etc that rely on `error-message` specifically?

PR #561, opened 2019-06-06
Go Modules support in `snyk test`

kyegupov:
- [x] Ready for review

#### What does this PR do?
Enables go.mod files support for test command via `--file=go.mod` flag. Sending organisation has to be enabled on the backend.

Monitor command will go in soon.

As usual, with the tests.

Running test command on this branch with `--file=go.mod` using entitled organisation.

This change means that go.mod files is scanned and tree is created always, but backend than denies the request if sending organisation is not entitled.

This is not ideal, but I believe it pays off given the temporary nature of the beta release. Otherwise, we would need to create whole new communication flow for feature flag pre-check.

### DIFF
--- a/src/lib/detect.ts
+++ b/src/lib/detect.ts
@@ -18,7 +18,6 @@ const DETECTABLE_FILES: string[] = [
   'build.gradle.kts',
   'build.sbt',
   'Pipfile',
-  'requirements.txt',
   'Gopkg.lock',
   'vendor/vendor.json',
   'obj/project.assets.json',
@@ -26,6 +25,9 @@ const DETECTABLE_FILES: string[] = [
   'packages.config',
   'paket.dependencies',
   'composer.lock',
+  // Python has the lowest priority. Python scripts are sometimes included in projects written in
+  // more "heavyweight" languages, so we should prefer those.
+  'requirements.txt',
 ];
 
 // when file is specified with --file, we look it up here
@@ -45,6 +47,7 @@ const DETECTABLE_PACKAGE_MANAGERS: {
   'Pipfile': 'pip',
   'requirements.txt': 'pip',
   'Gopkg.lock': 'golangdep',
+  'go.mod': 'gomod',
   'vendor.json': 'govendor',
   'project.assets.json': 'nuget',
   'packages.config': 'nuget',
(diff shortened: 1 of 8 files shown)


src/lib/detect.ts
```diff
@@ -18,14 +18,16 @@ const DETECTABLE_FILES: string[] = [
...
   'Gopkg.lock',
   'vendor/vendor.json',
   'obj/project.assets.json',
   'project.assets.json',
   'packages.config',
   'paket.dependencies',
   'composer.lock',
+  // Python has the lowest priority. Python scripts are sometimes included in projects written in
+  // more "heavyweight" languages, so we should prefer those.
+  'requirements.txt',
```
adrukh: This is quite a breaking change for our users - we preferred `requirements.txt` much earlier before, and will now risk picking up another file for the same `snyk test` command. Let's avoid this please?

kyegupov: At any rate, I'd suggest adding go.mod above it.

src/lib/monitor.ts
```diff
@@ -54,6 +54,9 @@ export async function monitor(root, meta, info: SingleDepRootResult, targetFile)
   policy = await snyk.policy.load(policyLocations, {loose: true});
 
   const packageManager = meta.packageManager;
+  if (packageManager === 'gomod') {
+    throw new Error('Monitoring go.mod files is not supported yet.');
```
kyegupov: even with the flag enabled?

test/acceptance/cli.acceptance.test.ts
```diff
@@ -1381,6 +1381,59 @@ test('`test paket-app --file=paket.dependencies`', async (t) => {
...
+      args: null,
+      file: 'go.mod',
+      org: null,
+      packageManager: 'gomod',
+      path: 'golang-mod',
+      showVulnPaths: true,
+    }], 'calls golang plugin');
+});
+
+test('`test golang-app` does not auto-detects golang-mod', async (t) => {
```
kyegupov: detects -> detect

src/lib/monitor.ts
miiila: There'll be a second set of PRs for monitor

PR #570, opened 2019-06-10
fix: remove auto authentication on all commands other than auth

orsagie:
- [ ] Ready for review

#### What does this PR do?
`Wizard` and `Ignore` commands have been doing automated authentication in case of no authentication, this behaviour is different to what test or monitor do - which is to bail out if there is no auth. This PR aims to regulate auth behaviour across all the commands. 

In addition, wording for the CI auth error is changed to 

run 
`snyk test`
`snyk protect`
`snyk monitor`
`snyk ignore`
`snyk wizard`
See that we require authentication.

### DIFF
--- a/src/lib/errors/misconfigured-auth-in-ci-error.ts
+++ b/src/lib/errors/misconfigured-auth-in-ci-error.ts
@@ -1,8 +1,8 @@
 import {CustomError} from './custom-error';
 
 export function MisconfiguredAuthInCI() {
-  const errorMsg = 'Web auth cannot be used whilst inside CI. You must include ' +
-  'your API token as an environment value: `API=12345678`';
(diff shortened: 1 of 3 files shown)


src/lib/errors/misconfigured-auth-in-ci-error.ts
```diff
@@ -1,7 +1,7 @@
 import {CustomError} from './custom-error';
 
 export function MisconfiguredAuthInCI() {
-  const errorMsg = 'Web auth cannot be used whilst inside CI. You must include ' +
+  const errorMsg = 'Snyk is missing auth token in order to run inside CI. You must include ' +
   'your API token as an environment value: `API=12345678`';
```
adrukh: Please adjust the second part of the message, from `API=12345678` to `SNYK_TOKEN=12345678`.

PR #574, opened 2019-06-11
fix: show stacktrace only when in debug mode

orsagie:
- [x] Ready for review

#### What does this PR do?
Most stacktraces are only visible in debug mode, but there were a few that fell through the cracks. This PR fixes this.

`https://github.com/snyk/snyk/blob/master/src/cli/index.ts#L45` will handle most of the errors. But some places were still using the legacy error handling to show errors. Removed the stack from the message, and only logging it in debug mode.

### DIFF
--- a/src/lib/errors/legacy-errors.js
+++ b/src/lib/errors/legacy-errors.js
@@ -2,6 +2,7 @@
 const config = require('../config');
 const chalk = require('chalk');
 const {SEVERITIES} = require('../snyk-test/common');
+const analytics = require('../analytics');
 
 const errors = {
   connect: 'Check your network connection, failed to connect to Snyk API',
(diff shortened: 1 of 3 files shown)


src/lib/errors/legacy-errors.js
```diff
@@ -86,7 +86,7 @@ module.exports.message = function(error) {
       message = chalk.bold.red(message);
     } else if (error.code) { // means it's a code error
       message = 'An unknown error occurred. Please include the trace below ' +
```
kyegupov: maybe "please include the output below" ?

orsagie: hmm, or rather "run with `-d` to see full trace"? The stack will not show unless they specify it. 
I will also add to analytics, so we know what these non user friendly error codes are.

PR #579, opened 2019-06-13
fix: remove old auth from ignore

orsagie:
- [x] Ready for review

#### What does this PR do?
Test and Monitor are no longer using `is-authed` and `authorization` . Removing them and only authing with `apiTokenExists`

run ignore without and with auth. 

https://github.com/snyk/snyk/issues/556

### DIFF
--- a/src/cli/commands/ignore.ts
+++ b/src/cli/commands/ignore.ts
@@ -1,70 +1,49 @@
 module.exports = ignore;
 
 import * as policy from 'snyk-policy';
-import chalk from 'chalk';
-import * as authorization from '../../lib/authorization';
-import * as auth from './auth/is-authed';
 import {apiTokenExists} from '../../lib/api-token';
-import {isCI} from '../../lib/is-ci';
 
 import * as Debug from 'debug';
 const debug = Debug('snyk');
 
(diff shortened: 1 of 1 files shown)


src/cli/commands/ignore.ts
```diff
@@ -1,70 +1,49 @@
 module.exports = ignore;
 
 import * as policy from 'snyk-policy';
-import chalk from 'chalk';
-import * as authorization from '../../lib/authorization';
-import * as auth from './auth/is-authed';
```
miiila: Can we remove `./auth/is-authed` then?

src/cli/commands/ignore.ts
```diff
@@ -1,70 +1,49 @@
...
-    if (!authed) {
-      if (isCI()) {
-        throw MisconfiguredAuthInCI();
-      }
-    }
-    apiTokenExists();
-  }).then(() => {
-    return authorization.actionAllowed('cliIgnore', options);
-  }).then((cliIgnoreAuthorization) => {
-    if (!cliIgnoreAuthorization.allowed) {
```
miiila: What exactly was this for? Do we know if somebody is using this?

src/cli/commands/ignore.ts
orsagie: It is still used in wizard. I will open another PR to remove it from there.

PR #606, opened 2019-06-26
fix: preserve all additional deptree fields when pruning

kyegupov:
- [x] Ready for review

#### What does this PR do?
Makes sure we preserve all the additional fields of the tree.

### DIFF
--- a/src/lib/monitor.ts
+++ b/src/lib/monitor.ts
@@ -77,18 +77,10 @@ async function pruneTree(tree: DepTree, packageManagerName: string): Promise<Dep
   const graph = await depGraphLib.legacy.depTreeToGraph(tree, packageManagerName);
   const prunedTree: DepTree = await depGraphLib.legacy
     .graphToDepTree(graph, packageManagerName, {deduplicateWithinTopLevelDeps: true}) as DepTree;
-  // Transplant metadata not preserved by the transformation:
-  if (tree.docker) {
-    prunedTree.docker = tree.docker;
-  }
-  if (tree.packageFormatVersion) {
-    prunedTree.packageFormatVersion = tree.packageFormatVersion;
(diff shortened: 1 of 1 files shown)


src/lib/monitor.ts
```diff
@@ -78,14 +78,10 @@ async function pruneTree(tree: DepTree, packageManagerName: string): Promise<Dep
...
-  if (tree.docker) {
-    prunedTree.docker = tree.docker;
-  }
-  if (tree.packageFormatVersion) {
-    prunedTree.packageFormatVersion = tree.packageFormatVersion;
-  }
-  if (tree.targetFile) {
-    prunedTree.targetFile = tree.targetFile;
+  for (const key of Object.keys(tree)) {
+    if (!prunedTree[key]) {
```
darscan: `if (prunedTree[key] === undefined)`

src/lib/monitor.ts
```diff
@@ -78,14 +78,10 @@ async function pruneTree(tree: DepTree, packageManagerName: string): Promise<Dep
   const prunedTree: DepTree = await depGraphLib.legacy
     .graphToDepTree(graph, packageManagerName, {deduplicateWithinTopLevelDeps: true}) as DepTree;
   // Transplant metadata not preserved by the transformation:
-  if (tree.docker) {
```
lili2311: Are any of these relevant? https://github.com/snyk/snyk/blob/b91330244930a2d018fe6c603fc25feceb021d25/src/lib/types.ts#L5

I feel like we need a test for this to ensure pre-pruned vs pruned has the same data, how do we know this is all the data that is missing?

kyegupov: well, they are outside the deptree, so they are unaffected.

PR #616, opened 2019-07-01
feat: use cli-interface package to define the plugin interface

kyegupov:
- [x] Ready for review

#### What does this PR do?
Enables using the new `cli-interface` repo for plugin type definitions.

The only plugin which is now 100% supported is Gradle (plugin v 3.0.0+), the rest have only partially compatible types.

Also changes names of some options (e.g. multiDepRoot -> allSubProjects).
Option transformation for allSubProjects has moved to args.ts

Be familiar with https://github.com/snyk/snyk-cli-interface repo.

Note that DepTree has a lot of fields of poorly known origin or purpose. This might affect switching the plugins to DepGraph interface.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   "author": "snyk.io",
   "license": "Apache-2.0",
   "dependencies": {
+    "@snyk/cli-interface": "^1.5.0",
     "@snyk/dep-graph": "1.10.0",
     "@snyk/gemfile": "1.2.0",
     "@types/agent-base": "^4.2.0",
@@ -72,7 +73,7 @@
     "snyk-config": "^2.2.1",
     "snyk-docker-plugin": "1.25.1",
     "snyk-go-plugin": "1.11.0",
-    "snyk-gradle-plugin": "2.12.5",
+    "snyk-gradle-plugin": "^3.0.1",
     "snyk-module": "1.9.1",
     "snyk-mvn-plugin": "2.3.1",
     "snyk-nodejs-lockfile-parser": "1.15.0",
--- a/src/cli/args.ts
+++ b/src/cli/args.ts
@@ -157,8 +157,9 @@ export function args(rawArgv: string[]): Args {
   // should be added here
   for (const dashedArg of [
     'package-manager',
-    'packages-folder',
+    'packages-folder', // TODO(kyegupov): find out where it is used
     'severity-threshold',
+    'all-sub-projects',
   ]) {
     if (argv[dashedArg]) {
       const camelCased = dashToCamelCase(dashedArg);
--- a/src/cli/commands/monitor.ts
+++ b/src/cli/commands/monitor.ts
@@ -16,9 +16,6 @@ import * as detect from '../../lib/detect';
 import * as plugins from '../../lib/plugins';
 import {ModuleInfo} from '../../lib/module-info'; // TODO(kyegupov): fix import
 import {
-  SingleDepRootResult,
-  MultiDepRootsResult,
-  isMultiResult,
   MonitorOptions,
   MonitorMeta,
   MonitorResult,
@@ -30,6 +27,9 @@ import {
   MonitorError,
   UnsupportedFeatureFlagError,
 } from '../../lib/errors';
+import {
+  InspectResult, SinglePackageResult, isMultiResult, MultiProjectResult, InspectOptions,
+} from '@snyk/cli-interface/dist/legacy/plugin';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 
@@ -61,10 +61,10 @@ async function promiseOrCleanup<T>(p: Promise<T>, cleanup: () => void): Promise<
 // or an error message.
 async function monitor(...args0: MethodArgs): Promise<any> {
   let args = [...args0];
-  let options: MonitorOptions = {};
+  let options = {} as ArgsOptions & MonitorOptions;
   const results: Array<GoodResult | BadResult> = [];
   if (typeof args[args.length - 1] === 'object') {
-    options = args.pop() as ArgsOptions as MonitorOptions;
+    options = args.pop() as ArgsOptions & MonitorOptions;
   }
 
   args = args.filter(Boolean);
@@ -78,7 +78,7 @@ async function monitor(...args0: MethodArgs): Promise<any> {
     snyk.id = options.id;
   }
 
-  if (options['all-sub-projects'] && options['project-name']) {
+  if (options.allSubProjects && options['project-name']) {
     throw new Error('`--all-sub-projects` is currently not compatible with `--project-name`');
   }
 
@@ -126,13 +126,11 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
       // Scan the project dependencies via a plugin
 
-      const pluginOptions = plugins.getPluginOptions(packageManager, options);
       analytics.add('packageManager', packageManager);
-      analytics.add('pluginOptions', pluginOptions);
 
-      // TODO: the type should depend on multiDepRoots flag
-      const inspectResult: SingleDepRootResult|MultiDepRootsResult = await promiseOrCleanup(
-          moduleInfo.inspect(path, targetFile, { ...options, ...pluginOptions }),
+      // TODO: the type should depend on allSubProjects flag
+      const inspectResult: InspectResult = await promiseOrCleanup(
+          moduleInfo.inspect(path, targetFile!, { ...options } as ArgsOptions & InspectOptions),
           spinner.clear(analyzingDepsSpinnerLabel));
 
       analytics.add('pluginName', inspectResult.plugin.name);
@@ -146,7 +144,7 @@ async function monitor(...args0: MethodArgs): Promise<any> {
       const meta: MonitorMeta = {
         'method': 'cli',
         'packageManager': packageManager,
-        'policy-path': options['policy-path'],
+        'policy-path': options['policy-path'] as string,
         'project-name': options['project-name'] || config.PROJECT_NAME,
         'isDocker': !!options.docker,
         'prune': !!options['prune-repeated-subdependencies'],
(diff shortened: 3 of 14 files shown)


src/cli/commands/monitor.ts
```diff
@@ -26,6 +23,9 @@ import { MethodArgs, ArgsOptions } from '../args';
 import { maybePrintDeps } from '../../lib/print-deps';
 import * as analytics from '../../lib/analytics';
 import {MonitorError} from '../../lib/errors';
+import {
+  InspectResult, SinglePackageResult, isMultiResult, MultiProjectResult,
+} from '@snyk/cli-interface/dist/legacy/plugin';
```
miiila: I'm confused by `dist` part here, should it be there?

src/lib/module-info/index.ts
```diff
@@ -1,21 +1,31 @@
...
 
-export function ModuleInfo(plugin, policy) {
+interface DepTreeWithPolicy extends DepTree {
+  policy: string;
+}
+
+export function ModuleInfo(plugin: Plugin | SingleSubprojectPlugin, policy) {
   return {
-    async inspect(root, targetFile, options) {
+    async inspect(root: string, targetFile: string | undefined, options: InspectOptions) {
```
miiila: What about making `targetfile` optional?

miiila: `options: InspectOptions & ArgsOptions`? We treat it that way on line 19.

src/lib/module-info/index.ts
```diff
@@ -1,21 +1,31 @@
...
       }, options);
 
       debug('calling plugin inspect()', {root, targetFile, pluginOptions});
       const info = await plugin.inspect(root, targetFile, pluginOptions);
       debug('plugin inspect() done');
 
       // attach policy if not provided by plugin
-      if (policy && !info.package.policy) {
+
+      if (policy && !isMultiResult(info) && !(info.package.policy)) {
```
miiila: Why not for `isMultiResult`? Shouldn't it be applied to all `scannedProjects` then?

miiila: I'm not sure how do I feel about function (`isMultiResult`) in something called interface (`@snyk/cli-interface`). Just saying it loud, no opinions.

src/lib/package-managers.ts
```diff
@@ -18,6 +18,12 @@ export const SUPPORTED_PACKAGE_MANAGER_NAME: {
   nuget: 'NuGet',
   paket: 'Paket',
   composer: 'Composer',
+
+  // Docker
+  rpm: 'rpm (Linux)',
```
miiila: Not sure if this has any functional implication, would be good to double check.

src/lib/module-info/index.ts
orsagie: Think of these functions as `typeof` which are necessary and I think belong next to interface.

src/lib/package-managers.ts
orsagie: @orkamara Can you confirm these namings for docker?

src/lib/module-info/index.ts
orsagie: Because the options are not optional as well

src/lib/plugins/rubygems/index.ts
```diff
@@ -1,17 +1,26 @@
...
+  root: string,
+  targetFile: string,
+  options: MultiSubprojectInspectOptions):
+  Promise<MultiProjectResult>;
 
-export async function inspect(root: string, targetFile: string): Promise<RubyGemsInspectResult> {
-  if (!targetFile ) {
+export async function inspect(root: string, targetFile: string, options?: InspectOptions): Promise<InspectResult> {
+  if (options && isMultiSubProject(options)) {
+    throw new Error('Returning multiple sub-projects is not supported for Node.js');
```
orsagie: not supported for ruby

src/lib/module-info/index.ts
orsagie: because multi result returns `depRoots` and not `package` - we should double check that the policy gets updated properly in multi projects

src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -2,10 +2,28 @@ import * as modulesParser from './npm-modules-parser';
...
+  root: string,
+  targetFile: string,
+  options: MultiSubprojectInspectOptions):
+  Promise<MultiProjectResult>;
 
 export async function inspect(root: string, targetFile: string, options: types.Options = {}):
-Promise<types.InspectResult> {
-  if (!targetFile ) {
+    Promise<InspectResult> {
+  if (isMultiSubProject(options)) {
```
lili2311: Please have a test for every error you have introduced

src/lib/plugins/rubygems/index.ts
lili2311: test for this please

src/lib/module-info/index.ts
kyegupov: added an error for that case

kyegupov: Implementing sum types from dirt and sticks...

src/cli/commands/monitor.ts
kyegupov: yup. That's the only way to do imports from inside the package.

package.json
```diff
@@ -72,7 +73,7 @@
     "snyk-config": "^2.2.1",
     "snyk-docker-plugin": "1.25.1",
     "snyk-go-plugin": "1.10.2",
-    "snyk-gradle-plugin": "2.12.5",
+    "snyk-gradle-plugin": "^3.0.0",
```
kyegupov: I'd like it to be a 3.0.1 - that version would have updated type definitions.

src/cli/args.ts
```diff
@@ -157,8 +157,9 @@ export function args(rawArgv: string[]): Args {
   // should be added here
   for (const dashedArg of [
     'package-manager',
-    'packages-folder',
+    'packages-folder', // TODO(kyegupov): find out where it is used
     'severity-threshold',
+    'all-sub-projects',
```
kyegupov: this replaces pluginOptions (see below)

src/lib/plugins/index.ts
```diff
@@ -10,66 +10,54 @@ import * as phpPlugin from 'snyk-php-plugin';
...
+      // this plugin needs an updrade of the types;
+      return adaptSingleProjectPlugin(phpPlugin as SingleSubprojectPlugin);
     }
     default: {
       throw new Error(`Unsupported package manager: ${packageManager}`);
     }
   }
 }
-
-export function getPluginOptions(packageManager: string, options: types.Options): types.Options {
```
kyegupov: see args.ts

src/lib/plugins/nodejs-plugin/index.ts
kyegupov: added for ruby and node.js

src/lib/plugins/types.ts
```diff
@@ -1,26 +1,16 @@
...
-  };
-  package?: any;
-  depRoots?: any;
-}
+import {InspectOptions } from '@snyk/cli-interface/dist/legacy/plugin';
+
+export type Options = CliInspectOptions & InspectOptions;
 
-export interface Options {
+interface CliInspectOptions {
```
kyegupov: I'd like every one of these to be documented.

src/cli/args.ts
lili2311: `packages-folder` is from nuget

src/cli/commands/monitor.ts
lili2311: no `dist` please, same as in the other PR let's use top level exports and import as normal

src/lib/module-info/index.ts
```diff
@@ -1,24 +1,33 @@
 import * as _ from 'lodash';
 import * as Debug from 'debug';
+import {
+  Plugin, InspectOptions, isMultiResult, SingleSubprojectPlugin,
+} from '@snyk/cli-interface/dist/legacy/plugin';
```
lili2311: same here, please import as usual

src/lib/package-managers.ts
lili2311: Why are these added here? We don't auto detect based on these so how does this affect the code?

src/lib/package-managers.ts
```diff
@@ -18,6 +18,12 @@ export const SUPPORTED_PACKAGE_MANAGER_NAME: {
...
+  deb: 'deb (Linux)',
+  apk: 'apk (Linux)',
+  dockerfile: 'Dockerfile',
```
lili2311: dockerFile is not a package manager and we don't auto detect it?

src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -2,10 +2,23 @@ import * as modulesParser from './npm-modules-parser';
 import * as lockParser from './npm-lock-parser';
 import * as types from '../types';
 import { MissingTargetFileError } from '../../errors/missing-targetfile-error';
+import {
+  isMultiSubProject, SingleSubprojectInspectOptions, SinglePackageResult,
+} from '@snyk/cli-interface/dist/legacy/plugin';
```
lili2311: no dist import please

src/lib/plugins/types.ts
```diff
@@ -1,26 +1,16 @@
-export interface InspectResult {
-  plugin: {
-    name: string;
-    runtime?: string;
-  };
-  package?: any;
-  depRoots?: any;
-}
+import {InspectOptions } from '@snyk/cli-interface/dist/legacy/plugin';
```
lili2311: `dist` here

src/lib/snyk-test/run-test.ts
```diff
@@ -16,14 +16,18 @@ import common = require('./common');
...
 import {
   NoSupportedManifestsFoundError,
   InternalServerError,
   FailedToGetVulnerabilitiesError,
 } from '../errors';
 import { maybePrintDeps } from '../print-deps';
 import { SupportedPackageManagers } from '../package-managers';
+import {
+  InspectResult, isMultiResult, MultiProjectResult, InspectOptions,
+} from '@snyk/cli-interface/dist/legacy/plugin';
```
lili2311: `dist` here

src/lib/types.ts
```diff
@@ -1,4 +1,5 @@
 import { SupportedPackageManagers } from './package-managers';
+import { DepTree } from '@snyk/cli-interface/dist/legacy/common';
```
lili2311: `dist` here as well

kyegupov:
was partially implemented in https://github.com/snyk/snyk/pull/732

PR #620, opened 2019-07-01
feat: refactor snyk test http errors

lili2311:
- [x] Ready for review

#### What does this PR do?
- Add custom errors to test results
- Remove old code around handling package vulns
- Add a tip to try testing with `--dev` not as part of error flow
Screen Shot 2019-07-01 at 17 28 03

### DIFF
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -246,17 +246,23 @@ function displayResult(res, options: Options & TestOptions) {
       'no vulnerable paths found.' :
       'none were found.';
     const summaryOKText = chalk.green(`✓ ${testedInfoText}, ${vulnPathsText}`);
-    const nextStepsText =
+    const nextStepsText = localPackageTest ?
       '\n\nNext steps:' +
       '\n- Run `snyk monitor` to be notified ' +
       'about new related vulnerabilities.' +
       '\n- Run `snyk test` as part of ' +
-      'your CI/test.';
+      'your CI/test.' : '';
+    // user tested a package@version and got 0 vulns back, but there were dev deps
+    // to consider
+    const snykPackageTestTip: string = !(localPackageTest || options.dev) ?
+      '\n\nTip: Snyk only tests production dependencies by default (which ' +
+      'this project had none). Try re-running with the `--dev` flag.' : '';
     return (
       prefix + meta + summaryOKText + multiProjAdvice + (
         isCI() ? '' :
           dockerAdvice +
-          nextStepsText)
+          nextStepsText +
+          snykPackageTestTip)
     );
   }
 
--- a/src/cli/commands/test.ts
+++ b/src/cli/commands/test.ts
@@ -255,8 +255,7 @@ function displayResult(res, options: Options & TestOptions) {
     // user tested a package@version and got 0 vulns back, but there were dev deps
     // to consider
     const snykPackageTestTip: string = !(localPackageTest || options.dev) ?
-      '\n\nTip: Snyk only tests production dependencies by default (which ' +
-      'this project had none). Try re-running with the `--dev` flag.' : '';
+      '\n\nTip: Snyk only tests production dependencies by default. You can try re-running with the `--dev` flag.' : '';
     return (
       prefix + meta + summaryOKText + multiProjAdvice + (
         isCI() ? '' :
@@ -290,7 +289,7 @@ function displayResult(res, options: Options & TestOptions) {
   }
   let summary = testedInfoText + ', ' + chalk.red.bold(vulnCountText);
 
-  if (localPackageTest && WIZARD_SUPPORTED_PMS.indexOf(packageManager) > -1) {
+  if (localPackageTest && WIZARD_SUPPORTED_PMS.includes(packageManager)) {
     summary += chalk.bold.green('\n\nRun `snyk wizard` to address these issues.');
   }
   let dockerSuggestion = '';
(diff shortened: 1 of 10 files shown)


src/cli/commands/test.ts
```diff
@@ -8,16 +8,13 @@ import {isCI} from '../../lib/is-ci';
...
-import {Options} from '../../lib/types';
+import {Options, TestOptions} from '../../lib/types';
 import {isLocalFolder} from '../../lib/detect';
 import { MethodArgs } from '../args';
 
 const debug = Debug('snyk');
 const SEPARATOR = '\n-------------------------------------------------------\n';
 
-interface OptionsAtDisplayStage {
-  canSuggestRemediation: boolean;
```
lili2311: not needed, we can just re-calculate rather than pollute global snyk cli options

src/cli/commands/test.ts
```diff
@@ -215,11 +212,11 @@ function summariseErrorResults(errorResults) {
   return '';
 }
 
-function displayResult(res, options: Options & OptionsAtDisplayStage) {
+function displayResult(res, options: Options & TestOptions) {
   const meta = metaForDisplay(res, options) + '\n\n';
   const dockerAdvice = dockerRemediationForDisplay(res);
   const packageManager = options.packageManager;
-  options.canSuggestRemediation = isLocalFolder(options.path);
```
lili2311: renaming to better reflect that it is a test of a package like `snyk test lodash@1.2.3` instead of snyk test of project on disk

src/lib/snyk-test/run-test.ts
```diff
@@ -148,34 +150,16 @@ async function runTest(packageManager: SupportedPackageManagers,
...
-        const err = new Error(body && body.error ?
-          body.error :
-          res.statusCode);
-
-        (err as any).userMessage = body && body.userMessage;
-        // this is the case where a local module has been tested, but
-        // doesn't have any production deps, but we've noted that they
-        // have dep deps, so we'll error with a more useful message
-        if (res.statusCode === 404 && hasDevDependencies) {
-          (err as any).code = 'NOT_FOUND_HAS_DEV_DEPS';
```
lili2311: no analytics have this code, the response is actually `200` on a test example: `snyk test lodash@4.17.11`, lodash only has devDeps. And `hasDevDependencies` on the body comes back false!

So instead showing the `Tip` when user did not use `--dev` flag and it is a remote package test & we returned `0 ` vulns to nudge them to look at dev deps too.

src/lib/types.ts
```diff
@@ -74,6 +74,11 @@ export interface Options {
   json?: boolean;
   vulnEndpoint?: string;
   projectName?: string;
+  insecure?: boolean;
```
lili2311: adding missing options from cli supported options from `snyk help`, did not add language specific options yet

test/system/remote-package.test.ts
```diff
@@ -112,13 +112,40 @@ test('multiple test arguments', (t) => {
...
+    'tip text as expected');
+  } catch (error) {
+    t.fail('should not throw, instead received error: ' + error);
+  }
+});
+
 test('test for non-existing', (t) => {
   t.plan(1);
 
   cli.test('@123').then((res) => {
```
dkontorovskyy: you can make `async/await` it and get rid of promises and I believe from `t.plan()`s

lili2311: Updated!

src/cli/commands/test.ts
```diff
@@ -246,17 +246,23 @@ function displayResult(res, options: Options & TestOptions) {
...
       '\n\nNext steps:' +
       '\n- Run `snyk monitor` to be notified ' +
       'about new related vulnerabilities.' +
       '\n- Run `snyk test` as part of ' +
-      'your CI/test.';
+      'your CI/test.' : '';
+    // user tested a package@version and got 0 vulns back, but there were dev deps
+    // to consider
+    const snykPackageTestTip: string = !(localPackageTest || options.dev) ?
+      '\n\nTip: Snyk only tests production dependencies by default (which ' +
```
kyegupov: "of which this project had none", to my best knowledge of English. 

But even that is incorrect, since this message is shown when there were 0 vulns, not 0 deps.

lili2311: yeah this is out old message it is the same: https://github.com/snyk/snyk/pull/620/files#diff-94bda9bbb4d28bca7d68ff57a89eb635L26 Can reword though

lili2311: Updated to ditch the message about deps, doesn't look like we know anyway so if the user on a package scan get's 0 vulns, just suggest they might want to also try with --dev

test/system/remote-package.test.ts
```diff
@@ -23,132 +23,152 @@ import * as cli from '../../src/cli/commands';
...
-    t.fails('should fail, instead received ' + res);
-  }).catch((error) => {
-    t.match(error.message, '500', 'expected error ' + error.message);
-  });
+test('test for existing remote package with dev-deps only', async (t) => {
+  try {
+    const res = await cli.test('lodash@4.17.11');
+    const lastLine = res.trim().split('\n').pop();
+    t.deepEqual(lastLine, 'Tip: Snyk only tests production dependencies by default ' +
+    '(which this project had none). Try re-running with the `--dev` flag.',
```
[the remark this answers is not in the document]
lili2311: will check

PR #626, opened 2019-07-02
feat: autodetect go.mod file

lwywoo:
- [x] Ready for review

#### What does this PR do?
Enable autodetection of go.mod file

https://snyksec.atlassian.net/browse/BST-725

#### Screenshots
Screenshot 2019-07-02 at 19 02 52

### DIFF
--- a/src/lib/detect.ts
+++ b/src/lib/detect.ts
@@ -20,6 +20,7 @@ const DETECTABLE_FILES: string[] = [
   'Pipfile',
   'requirements.txt',
   'Gopkg.lock',
+  'go.mod',
   'vendor/vendor.json',
   'obj/project.assets.json',
   'project.assets.json',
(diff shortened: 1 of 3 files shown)


src/lib/detect.ts
```diff
@@ -20,6 +20,7 @@ const DETECTABLE_FILES: string[] = [
   'Pipfile',
   'requirements.txt',
   'Gopkg.lock',
+  'go.mod',
```
kyegupov: I have seen some projects with both go.mod and requirements.txt.
And I'd think our customers would prefer scanning Go over Python (which is likely some build / helper scripts).

Ideally, we should crash if we find more than one matching file and require the `--file` option.

miiila: @kyegupov I agree, especially with multi-matching. There's an effort planned for next Q/H, so we'll address it. For now, I'm OK with leaving it as it is in order to keep projects working; we should not switch from current Python project to Go silently (yes, that means it should be added to the end of the list).

PR #631, opened 2019-07-03
Upgrade Python3 to 3.7 instead of 3.5

julienduchesne:
Some packages require higher versions to be installed

- [x] Ready for review

#### What does this PR do?
Some packages need features from newer python versions so Python should be kept up-to-date. The current Python docker image has Python 3.5.3 which is pretty old.

Here is an example error from incompatible packages:
`ERROR: dacite requires Python '>=3.6' but the running Python is 3.5.3`

The Python 2 image was also migrated to use the same format (`FROM python`) to facilitate changes to both images

### DIFF
--- a/docker/Dockerfile.python-2
+++ b/docker/Dockerfile.python-2
@@ -6,7 +6,7 @@ RUN mkdir /home/node
 WORKDIR /home/node
 
 # Install Python utilities, node, Snyk CLI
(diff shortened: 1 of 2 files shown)


adrukh:
Merging to a side-branch to pull this in properly, will open a separate PR for this change.

PR #652, opened 2019-07-10
fix(protect): only patch when version on disk satisfies vuln

lirantal:
- [x] Ready for review

#### What does this PR do?
Updates the logic for checking whether to apply patch A on dep B to fix vuln C.

Prior to this change, our logic would sometime lead to a misdetection of the following kind:
Patch A is applicable to `pkg@1.0.0`
Dep B is located in node_modules as `pkg@2.0.0`
Vuln C was detected on `pkg@1.0.0`, which was then resolved on the disk as dep B.

This would cause a clash and fail when applying the patch due to version incompatibility.

The fix is a stop-gap measure that validates the version of the dep located in node_modules prior to applying a patch.

### DIFF
--- a/src/lib/protect/apply-patch.js
+++ b/src/lib/protect/apply-patch.js
@@ -6,7 +6,7 @@ const exec = require('child_process').exec;
 const path = require('path');
 const fs = require('fs');
 const uuid = require('uuid/v4');
-const semver = require("semver");
+const semver = require('semver');
 const errorAnalytics = require('../analytics').single;
 
 function applyPatch(patchFileName, vuln, live, patchUrl) {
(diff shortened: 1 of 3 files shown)


src/lib/protect/apply-patch.js
```diff
@@ -19,14 +20,31 @@ function applyPatch(patchFileName, vuln, live, patchUrl) {
...
     debug('DRY RUN: relative: %s', relative);
 
+    let pkg;
     try {
       const packageJson = fs.readFileSync(path.resolve(relative, 'package.json'));
-      const pkg = JSON.parse(packageJson);
+      pkg = JSON.parse(packageJson);
       debug('package at patch target location: %s@%s', pkg.name, pkg.version);
     } catch (err) {
       debug('Failed loading package.json of package about to be patched', err);
```
adrukh: Let's return here too, if the package.json is not ok, maybe we cannot trust the dependency data?

adrukh: Also, if we don't return, `pkg.version` on line 32 will 💥 because `pkg` is undefined.

src/lib/protect/apply-patch.js
```diff
@@ -19,14 +20,31 @@ function applyPatch(patchFileName, vuln, live, patchUrl) {
...
     }
 
+    const versionOfPackageToPatch = pkg.version;
+
+    const vulnerableVersions = vuln.semver.vulnerable;
```
adrukh: What matters here is not the vuln's affected semver range (`vuln.semver.vulnerable`), but the dep version on which the vuln was detected (`vuln.version`) - please change

src/lib/protect/apply-patch.js
```diff
@@ -20,22 +21,38 @@ function applyPatch(patchFileName, vuln, live, patchUrl) {
...
+        const packageJson = fs.readFileSync(packageJsonPath);
+        pkg = JSON.parse(packageJson);
+        debug('package at patch target location: %s@%s', pkg.name, pkg.version);
+      } catch (err) {
+        debug('Failed loading package.json at %s. Skipping patch!', packageJsonPath, err);
+        return resolve();
+      }
+
+      const versionOfPackageToPatch = pkg.version;
+      const patchableVersionsRange = vuln.patches.version;
```
miiila: How does the `patchableVersionRange` look like? I remember some issues with strings, I had to run something like `semver.coalesce` or something like that.

adrukh: Thanks for pointing this out! Indeed we will not assume the version ranges are pure semver notation, and will use `semver.coerce`.

PR #656, opened 2019-07-12
feat: removes npm update prompt

thisislawatts:
- [X] Ready for review

#### What does this PR do?

During the packaging process for the binary adds a `STANDALONE` to the bundle. The `updateCheck` will then bail out early without displaying the upgrade prompt. The aim of this is to prevent users who installed without node being prompted to upgrade `npm`.

### DIFF
--- a/src/lib/updater.ts
+++ b/src/lib/updater.ts
@@ -10,6 +10,13 @@ export function updateCheck() {
     return false;
   }
 
+  const standalonePath = p.join(__dirname, '../', 'STANDALONE');
+  const isStandaloneBuild = fs.existsSync(standalonePath);
+
+  if (isStandaloneBuild) {
+    return false;
+  }
+
   const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
 
   // if there's no version (f.e. during tests) - do not proceed
--- a/test/updater.test.js
+++ b/test/updater.test.js
@@ -1,19 +1,18 @@
 const tap = require('tap');
 const test = tap.test;
 const updateCheck = require('../src/lib/updater').updateCheck;
-const path = require('path');
+const fs = require('fs');
+const p = require('path');
 const sinon = require('sinon').createSandbox();
 const updateNotifier = require('update-notifier');
 
 // Fake location of the package.json file and verify the code behaves well
 test('missing package.json', (t) => {
-  t.plan(1);
-  let resolveStub = sinon.stub(path, 'resolve');
-  resolveStub.onFirstCall().returns('falseDir');
-  resolveStub.onSecondCall().returns('falseFile');
+  const fsStub = sinon.stub(fs, 'existsSync');
+  fsStub.withArgs(p.join(__dirname, '../', 'package.json')).returns(false);
 
   t.tearDown(() => {
-    resolveStub.restore();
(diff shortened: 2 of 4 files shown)


garethr:
@thisislawatts you mentioned running into issues on Windows. This might be *completely irrelevant* but I skimmed the code because I'm annoying. Assuming (large assumption) that the problem you ran into was with `touch` not being available. The following would work cross-platform, given it looks like node is installed.

```
node --eval "const fs = require('fs'); fs.closeSync(fs.openSync('dist/STANDALONE', 'w'));
```

thisislawatts:
Thanks for the pointer @garethr, prompted me to spend some time investigating and appears that this was the solution
https://github.com/snyk/snyk/pull/656/files#diff-60b57a3102dc355d753b2571eda313dcL11-L13

 🤞

src/lib/updater.ts
```diff
@@ -10,6 +10,13 @@ export function updateCheck() {
     return false;
   }
 
+  const standalonePath = p.join(__dirname, '../dist/', 'STANDALONE');
```
darscan: pretty sure this path is incorrect (given that we're currently in `src/lib`.. or rather `dist/lib`)

yep, when debugged this becomes `dist/dist/STANDALONE`

test/updater.test.js
```diff
@@ -1,30 +1,39 @@
...
-    resolveStub.restore();
+    fsStub.restore();
   });
 
   t.equal(updateCheck(), false, 'Notifier was not started on missing package.json');
   t.end();
 });
 
+test('STANDALONE declaration present', (t) => {
+  const fsStub = sinon.stub(fs, 'existsSync').callsFake(() => true);
```
darscan: this is way to broad a stub (hence the bug above not being caught). Can you add `withArgs` so that we only stub calls with the expected arguments

test/updater.test.js
```diff
@@ -1,30 +1,39 @@
...
 const updateNotifier = require('update-notifier');
 
 // Fake location of the package.json file and verify the code behaves well
 test('missing package.json', (t) => {
-  t.plan(1);
-  let resolveStub = sinon.stub(path, 'resolve');
-  resolveStub.onFirstCall().returns('falseDir');
-  resolveStub.onSecondCall().returns('falseFile');
+  const fsStub = sinon.stub(fs, 'existsSync');
+  fsStub.onFirstCall().returns(true);
```
darscan: I would narrow this stub too

src/lib/updater.ts
```diff
@@ -10,6 +10,13 @@ export function updateCheck() {
     return false;
   }
 
+  const standalonePath = p.join(__dirname, '../dist', 'STANDALONE');
```
darscan: hehe.. this still seems wrong.. we're in `src/lb/` (or `dist/lib/`) so `../dist` takes us to `src/dist/` (or `dist/dist/`). I just added `console.log('standalonePath', standalonePath)` and it prints:
```
/Users/xxx...xxx/snyk/dist/dist/STANDALONE
```

thisislawatts: Thanks for taking another look at that @darscan. Apologies it looks like a mis-commit on my behalf, the dist should have been removed from `updater.ts`.

PR #660, opened 2019-07-16
feat: add check for `experimental-dep-graph`  for monitor

dkontorovskyy:
- [X] Ready for review

#### What does this PR do?
Allow to check for `experimentalDepGraph` feature flag in monitor flow.

[BST-771](https://snyksec.atlassian.net/browse/BST-771)

### DIFF
--- a/src/cli/commands/monitor.ts
+++ b/src/cli/commands/monitor.ts
@@ -11,6 +11,7 @@ import chalk from 'chalk';
 import * as pathUtil from 'path';
 import * as spinner from '../../lib/spinner';
 
+import * as request from '../../lib/request';
 import * as detect from '../../lib/detect';
 import * as plugins from '../../lib/plugins';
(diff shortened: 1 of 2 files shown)


src/cli/commands/monitor.ts
```diff
@@ -75,6 +82,14 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
   apiTokenExists();
 
+  if (options['experimental-dep-graph']) {
+    const isSupported = await isFeatureFlagSupportedForOrg(_.camelCase('experimental-dep-graph'));
+
+    if (!isSupported) {
+      throw new Error('`--experimental-dep-graph` is currently not supported');
```
lili2311: usually Registry should send back a response that has a `userMessage` we can show that message directly which is that this feature is not supported for this org.

PR #662, opened 2019-07-16
feat: support project name option in test command

joeholdcroft:
- [x] Ready for review

#### What does this PR do?

- Adds support for the `--project-name` option for the `test` command. If a project name is passed that does not exist, the test will continue to run anyway.
- Fixes an issue where `targetFile` was always sent to the API on `test`, even for languages that do not use it.
- Sends `x-snyk-api-version` header to the API so we can version behaviour at the API side
  - This is needed for the related PR to avoid causing adverse effects to customers who do not upgrade their CLI (i.e. we continue to ignore `targetFile` on the `test` endpoint unless customers are using the latest fixed CLI)

Monitor a project with a custom name and some ignored issues, then run `test` defining the project name as `--project-name`, any ignored issues should not be flagged in the CLI results.

Run the above for a language that requires `targetFile` (go, .net etc).

To test this, you'll need to have the linked PR below set up locally with your CLI pointing to it as the API.

See above.

Related PR for API support here: https://github.com/snyk/registry/pull/8811

N/A

#### Screenshots

N/A

#### Additional questions

N/A


src/cli/commands/test.ts
```diff
@@ -50,6 +50,7 @@ async function test(...args: MethodArgs): Promise<string> {
     // these options later.
     const testOpts = _.cloneDeep(options);
     testOpts.path = path;
+    testOpts.projectName = testOpts['project-name'];
```
lili2311: If the users have been currently sending this anyway on test, it will suddenly create new project. Can we check how many cli analytics events we have with `snyk test` and the `--project-name` flag passed in?

Can help with this tomorrow, so before we merge we can identify how many users may be affected by this change

joeholdcroft: Following yesterday's discussion, this is ready for another review.

joeholdcroft: Also for posterity, noting here that test never creates projects.

src/lib/request/request.ts
```diff
@@ -10,95 +10,100 @@ import * as analytics from '../analytics';
...
-      const json = JSON.stringify(body);
-      if (json.length < 1e4) {
-        debug(JSON.stringify(body, null, 2));
+      if (!payload.headers) {
+        payload.headers = {};
       }
 
-      // always compress going upstream
-      data = zlib.gzipSync(json, {level: 9});
+      payload.headers['x-snyk-api-version'] = versionNumber;
```
joshje: @gjvis I remember previously there was a discussion over "whether to `-x` prefix or not. I can't remember the outcome 😄

PR #674, opened 2019-07-23
Feat/windows compatible scripts

lili2311:
- [x] Ready for review

### BREAKING CHANGE
- Use `&&` instead of `;` to make scripts windows compatible. This will fail the prepare/prepublish script as `&&` only continues when the first command succeeds.

- Convert a wizard text => .ts

## Linked issue
https://github.com/snyk/snyk/issues/646

### DIFF
--- a/test/wizard-package-changes.test.ts
+++ b/test/wizard-package-changes.test.ts
@@ -1,11 +1,12 @@
-const tap = require('tap');
-const test = require('tap').test;
-const sinon = require('sinon').createSandbox();
-const proxyquire = require('proxyquire');
+import * as tap from 'tap';
+import * as _sinon from 'sinon';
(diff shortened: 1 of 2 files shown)


test/wizard-package-changes.test.ts
```diff
@@ -1,11 +1,11 @@
-const tap = require('tap');
-const test = require('tap').test;
-const sinon = require('sinon').createSandbox();
-const proxyquire = require('proxyquire');
+import * as tap from 'tap';
+import {createSandbox as sinon} from 'sinon';
```
miiila: Tests are failing and I believe it's because of this. I believe
```
import * as _sinon from 'sinon';
const sinon = _sinon.createSandbox();
```

could do the trick.

PR #682, opened 2019-07-24
Remove breaking change

miiila:


adrukh:
@miiila this seems to be reverting a bit too much?

PR #711, opened 2019-08-05
refactor: typescriptify request and alerts

kyegupov:
- [x] Ready for review

#### What does this PR do?
A bit of typescriptification.


dkontorovskyy:
@kyegupov you don't want to trigger the release with this one? cos you commit starts with `refactor:`

kyegupov:
@dkontorovskyy yeah fair enough let's make it a fix

src/cli/commands/auth/index.ts
```diff
@@ -3,7 +3,7 @@ import * as open from 'opn';
 import * as snyk from '../../../lib';
 import * as config from '../../../lib/config';
 import {isCI} from '../../../lib/is-ci';
-import * as request from '../../../lib/request';
+import request = require('../../../lib/request');
```
miiila: Why moving from import to require?

kyegupov: because lib/request is now .ts, and TS compiler now can see that it's export = function, and so we have to use the special syntax https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

PR #736, opened 2019-08-18
fix: snyk policy command execution error and missed tests

AlbertKogan:
- [x] Ready for review

#### What does this PR do?
Fix execution error of CLI command `snyk policy`. 
The problem was with named export instead default. In addition tests and TypeScript notations added.

Run in terminal `snyk policy`


src/cli/commands/policy.ts
```diff
@@ -16,3 +16,5 @@ export async function displayPolicy(path) {
       throw error;
     }
 }
+
+export = displayPolicy;
```
AlbertKogan: Changed this because in file `src/cli/commands/index.js` all modules hotloaded as default

PR #746, opened 2019-08-23
feat: switch to open

jamesgeorge007:
- [x] Ready for review
- [x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules
- [ ] Reviewed by Snyk internal team

#### What does this PR do?
Dropped [opn](npmjs.com/package/opn) and made the necessary workarounds to use [open](npmjs.com/package/open) instead.

> [opn](npmjs.com/package/open) was deprecated and renamed to [open](npmjs.com/package/open)

related: [#1679](https://github.com/mozilla/web-ext/pull/1679)

PR #749, opened 2019-08-25
feat: switch to open (opn was deprecated and renamed to open)

orkamara:
- [x] Ready for review

#### What does this PR do?
Switch using `opn` package to `open`. `opn` is deprecated and renamed to `open`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "inquirer": "^6.2.2",
     "lodash": "^4.17.14",
     "needle": "^2.2.4",
-    "opn": "^5.5.0",
+    "open": "^6.4.0",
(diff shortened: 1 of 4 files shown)


dkontorovskyy:
Brought by @jamesgeorge007 in #746

dkontorovskyy:
@jamesgeorge007, unfortunately we need to put this PR on hold, cos of 2 reasons:
1. `open@6+` doesn't support node versions less than 8 (has inside async/await functions). And this is very crucial for us, cos we need to support node 6.
2. Needs to be properly tested, cos could bring bugs in out standalone binaries. (zeit already has this issue https://github.com/zeit/pkg/issues/731)

PR #763, opened 2019-09-06
fix: support subprojects in docker image and update docs

miiila:
- [x] Ready for review

#### What does this PR do?
Allows jq to parse array of results when running in docker.


odinn1984:
@miiila did u make sure that snyk to html is working here on a project with sub projects? from what i tested it's not working and the output html is not correct... it's not showing the vulns

miiila:
@odin1984 Yes, I did :-)

odinn1984:
then it's approved :) woohooo ... i still think we should have gotten rid of it but hopefully soon once we deprecate v1 of the jenkins plugin

PR #772, opened 2019-09-22
Chore/introduce prettier

miiila:
- [x] Ready for review

#### What does this PR do?
Apply standard prettier rules.


package.json
```diff
@@ -25,11 +25,14 @@
     "build": "tsc",
     "build-watch": "tsc -w",
     "find-circular": "npm run build && madge --circular ./dist",
+    "format": "prettier --write '{src,test,scripts}/**/*.{js,ts}'",
```
sfat: it would be nice to have this part of `build` and/or `test` so you don't have to always tell people: "you forgot to format your code".
Even `on save` in your IDE would be cool.

miiila: @sfat Thanks! Idea is that we don't want to change the code without developer knowing. So we prefer to everybody to use their own way. I know some folks prefer git commit hook, while I have my IDE doing this for me.

So we basically require some standards and it's up to you, how you will achieve them :-)

PR #773, opened 2019-09-22
feat: convert nodejs plugin response to use multi format

lili2311:
- [x] Ready for review

#### What does this PR do?
Convert the nodejs plugin code to use the new multi result format in preparation for auto detections and handling of multiple file results.

### DIFF
--- a/src/cli/commands/monitor.ts
+++ b/src/cli/commands/monitor.ts
@@ -28,7 +28,7 @@ interface GoodResult {
   ok: true;
   data: string;
   path: string;
-  subProjectName?: string;
+  projectName?: string;
 }
 
 interface BadResult {
@@ -166,10 +166,10 @@ async function monitor(...args0: MethodArgs): Promise<any> {
       // SinglePackageResult is a legacy format understood by Registry, so we have to convert
       // a MultiProjectResult to an array of these.
 
-      let perSubProjectResults: pluginApi.SinglePackageResult[] = [];
+      let perProjectResult: pluginApi.SinglePackageResult[] = [];
       let advertiseSubprojectsCount: number | null = null;
       if (pluginApi.isMultiResult(inspectResult)) {
-        perSubProjectResults = inspectResult.scannedProjects.map(
+        perProjectResult = inspectResult.scannedProjects.map(
           (scannedProject) => ({
             plugin: inspectResult.plugin,
             package: scannedProject.depTree,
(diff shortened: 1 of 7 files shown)


src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -2,22 +2,25 @@ import * as modulesParser from './npm-modules-parser';
...
 
   const getLockFileDeps = isLockFileBased && !options.traverseNodeModules;
+  const depTree: any = getLockFileDeps ?
+      await lockParser.parse(root, targetFile, options) :
+      await modulesParser.parse(root, targetFile, options);
+
   return {
     plugin: {
       name: 'snyk-nodejs-lockfile-parser',
       runtime: process.version,
```
lili2311: This may be contradicting this change: https://github.com/snyk/snyk-cli-interface/pull/21/files#diff-b2d5ac8d1346936f02713d534de45340R93

src/cli/commands/monitor.ts
```diff
@@ -190,18 +190,18 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         const manageUrl = url.format(endpoint);
 
         endpoint.pathname = leader + '/monitor/' + res.id;
-        const subProjectName = pluginApi.isMultiResult(inspectResult)
+        const projectName = (pluginApi.isMultiResult(inspectResult))
           ? subProjDeps.package.name
```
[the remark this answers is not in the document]
lili2311: Updated @miiila

src/lib/plugins/nodejs-plugin/index.ts
miiila: Don't see an issue there

lili2311: the issue will be that different projects will be scanned soon and they will not share always the exact same version of gradle or maven or sbt for example. Or node in this example.

lili2311: But we can fix that easily I think but adding meta inside each result not on a top level response

lili2311:
Manually tested, works as expected 🎉

src/cli/commands/monitor.ts
```diff
@@ -28,7 +28,7 @@ interface GoodResult {
   ok: true;
   data: string;
   path: string;
-  subProjectName?: string;
+  projectName?: string;
```
orsagie: I like this change of renaming this from `subProjectName` to `projectName`. We don't really have a concept of sub project, as we monitor each sub project as separate project.

PR #774, opened 2019-09-22
feat: convert ruby to send back multi project result

lili2311:
- [x] Ready for review

#### What does this PR do?
- Return a dependency tree instead of files from the ruby handler (dropping legacy code)
- Always return an array of results in preparation of allowing scanning multiple of the same project type in one go.

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -35,7 +35,7 @@ async function webAuth(via: AuthCliCommands) {
 
   // validate that via comes from our code, and not from user & CLI
   if (redirects[via]) {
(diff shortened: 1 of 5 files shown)


src/cli/commands/auth/index.ts
```diff
@@ -35,7 +35,7 @@ async function webAuth(via: AuthCliCommands) {
 
   // validate that via comes from our code, and not from user & CLI
   if (redirects[via]) {
-    urlStr += '&redirectUri=' + new Buffer(redirects[via]).toString('base64');
+    urlStr += '&redirectUri=' + Buffer.from(redirects[via]).toString('base64');
```
miiila: 👍

PR #779, opened 2019-09-27
fix: correctly document behavior of `--strict-out-of-sync` and avoid string typed booleans

mrackwitz:
- [x] Ready for review

#### What does this PR do?

Changes the help text to document the default behavior of the `--strict-out-of-sync` flag and change the options type signature to eliminate the previously present string typed boolean option values `"true" | "false"`.

The current default behavior was manually verified by @darscan and me, by running `snyk test` on a project with a lockfile (npm and yarn). Also I verified that the behavior is still the same.

This will make it easier to get the CocoaPods plugin working. This failed because it had a stricter options definition, not including string typed booleans:

```
src/lib/plugins/index.ts:59:7 - error TS2322: Type 'typeof import("/Users/shaun/dev/work/snyk/repos/snyk/node_modules/@snyk/snyk-cocoapods-plugin/dist/index")' is not assignable to type 'Plugin'.
  Types of property 'inspect' are incompatible.
    Type '(root: string, targetFile?: string | undefined, options?: CocoaPodsInspectOptions | undefined) => Promise<SinglePackageResult>' is not assignable to type '(root: string, targetFile: string, options?: Options | undefined) => Promise<InspectResult>'.
      Types of parameters 'options' and 'options' are incompatible.
        Type 'Options | undefined' is not assignable to type 'CocoaPodsInspectOptions | undefined'.
          Type 'Options' is not assignable to type 'CocoaPodsInspectOptions'.
            Types of property 'strictOutOfSync' are incompatible.
              Type 'boolean | "true" | "false" | undefined' is not assignable to type 'boolean | undefined'.
                Type '"true"' is not assignable to type 'boolean | undefined'.

59       return cocoapodsPlugin;
```

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -89,11 +89,11 @@ Gradle options:
 
 Npm options:
   --strict-out-of-sync=<true|false>
-                       Allow testing out of sync lockfile. Defauls to false.
+                       Allow testing out of sync lockfile. Defauls to true.
 
 Yarn options:
   --strict-out-of-sync=<true|false>
-                       Allow testing out of sync lockfile. Defauls to false.
+                       Allow testing out of sync lockfile. Defauls to true.
 
 Python options:
   --command=<string>   Python interpreter to use. Default: "python", set to
--- a/src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
+++ b/src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
@@ -2,7 +2,6 @@ import * as baseDebug from 'debug';
 const debug = baseDebug('snyk');
 import * as path from 'path';
 import * as spinner from '../../spinner';
-import * as _ from 'lodash';
 import * as analytics from '../../analytics';
 import * as fs from 'fs';
 import * as lockFileParser from 'snyk-nodejs-lockfile-parser';
@@ -46,7 +45,7 @@ export async function parse(root: string, targetFile: string, options: Options):
   debug(resolveModuleSpinnerLabel);
   try {
     await spinner(resolveModuleSpinnerLabel);
-    const strictOutOfSync = _.get(options, 'strictOutOfSync') !== false;
+    const strictOutOfSync = options.strictOutOfSync !== false;
     return lockFileParser
       .buildDepTree(manifestFile, lockFile, options.dev, lockFileType, strictOutOfSync);
   } finally {
--- a/src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
+++ b/src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
@@ -46,7 +46,7 @@ export async function parse(root: string, targetFile: string, options: Options):
   debug(resolveModuleSpinnerLabel);
   try {
     await spinner(resolveModuleSpinnerLabel);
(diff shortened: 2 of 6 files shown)


src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
```diff
@@ -46,7 +46,7 @@ export async function parse(root: string, targetFile: string, options: Options):
   debug(resolveModuleSpinnerLabel);
   try {
     await spinner(resolveModuleSpinnerLabel);
-    const strictOutOfSync = _.get(options, 'strictOutOfSync') !== 'false';
+    const strictOutOfSync = _.get(options, 'strictOutOfSync') !== false;
     return lockFileParser
       .buildDepTree(manifestFile, lockFile, options.dev, lockFileType, strictOutOfSync);
```
darscan: given that this option can now only be `undefined`, `true` or `false` I think we can pass it straight through as-is (i.e drop the lodash stuff)

mrackwitz: Good call, will address this.

mrackwitz: I've kept the default behavior here for now. (`const strictOutOfSync = options. strictOutOfSync !== false`)

darscan: for some reason I just find this really hard to read:
```js
const strictOutOfSync = options.strictOutOfSync !== false;
```
how about:
```js
const strictOutOfSync = !!options.strictOutOfSync;
```
or:
```js
const strictOutOfSync = Boolean(options.strictOutOfSync);
```

mrackwitz: Tbh I find my variant the easiest to read, because it does actually imply what it does, without having to know how Node.js deals with `!undefined` or `Boolean(undefined)`. But if you consider any of these actually considered easier or more aligned with the conventions, you would need to let me know which.

mrackwitz: Ha, wait, the joke is that your variants actually do something else, which wasn't clear to me on first sight.

While the current variant defaults for `undefined` to `true`:
```
> a = undefined; a !== false
true
> a = false; a !== false
false
> a = true; a !== false
true
```

Yours default to `false`:
```
> a = undefined; !!a
false
> a = false; !!a
false
> a = true; !!a
true
```

So what about this instead: `a !== undefined ? a : true`

darscan: I totally forgot about the current behaviour! you initial code is fine

help/help.txt
```diff
@@ -89,11 +89,11 @@ Gradle options:
 
 Npm options:
   --strict-out-of-sync=<true|false>
-                       Allow testing out of sync lockfile. Defauls to false.
+                       Allow testing out of sync lockfile. Defauls to true.
```
lili2311: this is a doc only change the default behaviour is not changed?

mrackwitz: Yes, we realized that the default behavior was different from what was documented.

PR #795, opened 2019-10-04
fix: add packed dependency to prevent download from git

miiila:
- [x] Ready for review

#### What does this PR do?
To prevent problems with installing from git, switch to local packages with bundled dependencies.

Other change is https://github.com/snyk/node-proxy-agent/commit/f05c4640b94e0995ede28983aafadd3cc067f4d4


adrukh:
@miiila if you were to answer `why` and not `what` in the commit message, :anton:, would it not be more reasonable to just add the `git` binary to our docker images instead?

miiila:
@adrukh I don't think so as it fails also for customers not using docker images, but "just" being behind firewall.

adrukh:
Got it, thanks!

PR #810, opened 2019-10-09
feat: refactor auth to use async await

lili2311:
- [x] Ready for review

#### What does this PR do?
Refactor auth to use async / await

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -54,23 +54,17 @@ async function webAuth(via: AuthCliCommands) {
 
   const lbl = 'Waiting...';
 
-  return (
-    spinner(lbl)
-      .then(() => {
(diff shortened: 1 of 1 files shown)


src/cli/commands/auth/index.ts
```diff
@@ -54,23 +54,19 @@ async function webAuth(via: AuthCliCommands) {
...
-  );
+  try {
+    await spinner(lbl);
+    await setTimeout(() => {
+      open(urlStr, { wait: false });
+    }, 2000);
+
+    const res = await testAuthComplete(token);
+    return res;
+  } catch (error) {
```
darscan: This `catch` does nothing but re-throw, can just remove it

lili2311: that makes perfect sense

PR #813, opened 2019-10-09
fix: fetch patch ignoring command line arguments

gitphill:
BST-889
Fetch patch was making http requests by using needle directly
Changed to use lib request which uses command line arguments when invoking needle
Added tests for lib request
Changed fetch patch fail test to typescript
Removed patch fetch proxy test (now covered by lib request test)
Added @types/sinon devDependency to help write test with sinon assertions and matchers

- [X] Ready for review

#### What does this PR do?

Fix issue when snyk protect run as part of snyk wizard and doesn’t keep the --insecure flag
Module fetch-patch was using needle directly
I've changed to use lib/request which handles command line arguments like --insecure

Run `DEBUG=* snyk wizard --insecure` against a project where a patch can be applied. Notice that fetch patch does not use insecure flag and should result in needle using rejectUnauthorized option.

BST-889

##### Screenshots

Screenshot 2019-10-09 at 22 49 55


test/patch-fetch-proxy.test.js
```diff
@@ -1,89 +0,0 @@
-var test = require('tap').test;
```
miiila: I don't see this test being migrated. Is it desired?

gitphill: it's covered by the new request.test.ts

PR #820, opened 2019-10-16
fix: use plugin fix and revert package version lookup

moshikod:
- [x] Ready for review

#### What does this PR do?
Revert the package version lookup patch after the root cause was fixed in the plugin [snyk-docker-plugin/pull/115](https://github.com/snyk/snyk-docker-plugin/pull/115)

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -145,8 +145,7 @@ async function runTest(
 
       if (res.docker && dockerfilePackages) {
         res.vulnerabilities = res.vulnerabilities.map((vuln) => {
-          const pkg = vuln.name.split('/')[0] + '-' + vuln.version;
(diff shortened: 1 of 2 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -145,8 +145,7 @@ async function runTest(
 
       if (res.docker && dockerfilePackages) {
         res.vulnerabilities = res.vulnerabilities.map((vuln) => {
-          const pkg = vuln.name.split('/')[0] + '-' + vuln.version;
-          const dockerfilePackage = dockerfilePackages[pkg];
```
lili2311: is there a test for this already?

moshikod: The plugin includes some tests for this functionality.
I made this change before fixing the root cause in the plugin and now I would like to revert https://github.com/snyk/snyk/pull/791

PR #822, opened 2019-10-18
fix: safely check for vuln info

lili2311:
- [x] Ready for review

#### What does this PR do?

Vulns and remediation get filtered by severity in totally different ways and in different places, we can't expect all vuln data to always be present so check for safety first just like we already do in upgrades.

https://snyk.slack.com/archives/CGHELPDHN/p1571407777053500

To test run `snyk test --severity-threshold=high` in the snyk/goof repo.

### DIFF
--- a/src/cli/commands/test/formatters/remediation-based-format-issues.ts
+++ b/src/cli/commands/test/formatters/remediation-based-format-issues.ts
@@ -189,8 +189,10 @@ function constructPatchesText(
     return [];
   }
   const patchedTextArray = [chalk.bold.green('\nPatchable issues:')];
-
   for (const id of Object.keys(patches)) {
+    if (!basicVulnInfo[id]) {
+      continue;
+    }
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/formatters/remediation-based-format-issues.ts
```diff
@@ -189,8 +189,10 @@ function constructPatchesText(
     return [];
   }
   const patchedTextArray = [chalk.bold.green('\nPatchable issues:')];
-
   for (const id of Object.keys(patches)) {
+    if (!basicVulnInfo[id]) {
```
robcresswell: Seems like you may as well combine this with the condition below rather than two statements

robcresswell:
Merging to prevent errors in production, but we shouldn't really link to a private Slack channel as a public reference; details should be in the commit message

PR #825, opened 2019-10-21
fix: match semver patch

ivanstanev:
A previous commit introduced semver.coerce() on the patch version range, when doing 'snyk protect'.
The problem is that coerce() returns a SemVer type, not a string. The existing code uses semver.satisfies() to check if the given patch applies to the node_module, but since satisfies() expects a string, the patch was not applied since the check never passed.

This resulted in patches being skipped, but we marked them as applied.

The following fix restores the previous functionality of running satisfies() on the patch range but also keeps the existing check of using coerce() while also fixing the output of coerce() to be a string (by reading the version property of the SemVer type). This way we do a check that the patch version is a SemVer range, but we also correctly cover the case where it's not a SemVer range (which was originally the intention).

- [x] Ready for review

[Zendesk Ticket 1968](https://snyk.zendesk.com/agent/tickets/1968)

### DIFF
--- a/test/fixtures/hoek@4.2.0-nopatch.json
+++ b/test/fixtures/hoek@4.2.0-nopatch.json
@@ -1,0 +1,178 @@
+{
+  "vulnerabilities": [
+    {
+      "CVSSv3": "CVSS:3.0/AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:L/A:L",
+      "alternativeIds": [
+        "SNYK-JS-HOEK-12061"
+      ],
+      "creationTime": "2018-02-12T22:28:27.612000Z",
+      "credit": [
+        "Olivier Arteau (HoLyVieR)"
+      ],
+      "cvssScore": 6.3,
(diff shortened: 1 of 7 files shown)


test/protect-semver-patch.test.ts
```diff
@@ -0,0 +1,90 @@
+// tslint:disable-next-line: no-var-requires
+const debug = require('debug')('snyk');
```
gitphill: you can do the following to avoid tslint:disable-next-line

```
import * as debugModule from 'debug';
const debug = debugModule('snyk');
```

Similar on line 6 too

ivanstanev:
@gitphill I've updated the PR!
Please note this change: https://github.com/snyk/snyk/pull/825/files#diff-d97913173ab0f19dd07e005afe851b7aR8-R15
I have added the same version check here because this code path runs before the actual apply-patch. This code path filters out patches that _can_ be applied, so it wouldn't make sense to have different checks at different points in time.

test/protect-semver-patch.test.ts
ivanstanev: I'm requiring a .json file though, I get an error when I try to use TS import 🤔

test/fixtures/hoek@4.2.0-nopatch.json
```diff
@@ -0,0 +1,178 @@
+{
```
orsagie: These files should not be checked in.

orsagie: Once you remove them, tests should pass

gitphill: Not sure that's true @orsagie - this file is used to test when no patch is available, see https://github.com/snyk/snyk/blob/4bb172caa4232565e979f1d19b38adb46e8dd54c/test/protect-semver-patch.test.ts#L68

The failing CI tests are unrelated to this change and I've seen them fail regularly in the past

PR #840, opened 2019-10-29
fix: propagate correct relative path

orsagie:
- [x] Ready for review

#### What does this PR do?
Fix the relative path sent back to registry, to be relative to the path from where the targetFile was found.

### DIFF
--- a/src/lib/monitor.ts
+++ b/src/lib/monitor.ts
@@ -176,7 +176,7 @@ export async function monitor(
 
   const target = await getTarget(pkg, meta);
   const targetFileRelativePath = targetFile
-    ? path.relative(root, targetFile)
+    ? path.join(path.resolve(root), targetFile)
     : '';
 
   if (target && target.branch) {
@@ -285,7 +285,7 @@ export async function monitorGraph(
 
   const target = await getTarget(pkg, meta);
   const targetFileRelativePath = targetFile
-    ? path.relative(root, targetFile)
+    ? path.join(path.resolve(root), targetFile)
     : '';
 
   if (target && target.branch) {
--- a/test/acceptance/cli.acceptance.test.ts
+++ b/test/acceptance/cli.acceptance.test.ts
@@ -3212,11 +3212,21 @@ test('`monitor sbt package --experimental-dep-graph --sbt-graph`', async (t) => 
   t.equal(req.method, 'PUT', 'makes PUT request');
   t.match(req.url, '/monitor/sbt/graph', 'puts at correct url');
   t.ok(req.body.depGraphJSON, 'sends depGraphJSON');
-  t.match(
-    req.body.targetFileRelativePath,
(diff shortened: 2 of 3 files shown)


test/acceptance/cli.acceptance.test.ts
```diff
@@ -3231,6 +3232,7 @@ test('`monitor yarn-package`', async (t) => {
   t.notOk(pkg.dependencies['object-assign'], 'no dev dependency');
   t.notOk(pkg.from, 'no "from" array on root');
   t.notOk(pkg.dependencies.debug.from, 'no "from" array on dep');
+  t.equal(req.body.targetFileRelativePath, 'yarn-package/yarn.lock');
```
lili2311: what was it before?

src/lib/monitor.ts
```diff
@@ -284,9 +282,7 @@ export async function monitorGraph(
   const policy = await snyk.policy.load(policyLocations, { loose: true });
 
   const target = await getTarget(pkg, meta);
-  const targetFileRelativePath = targetFile
-    ? path.relative(root, targetFile)
-    : '';
+  const targetFileRelativePath = targetFile ? path.join(root, targetFile) : '';
```
lili2311: how does this affect existing projects? where is this currently used in the app `targetFileRelativePath`?

orsagie: This does not, as `targetFileRelativePath` was not actually propagated to registry before

test/acceptance/cli.acceptance.test.ts
orsagie: before it was something like `../yarn.lock'

src/lib/monitor.ts
lili2311: Why is this needed? We already have `target` and then this it should be enough? See `monitor-target.test.ts`

test/acceptance/cli.acceptance.test.ts
lili2311: How does running `snyk test --file=A`   vs `snyk test` in the right folder vs `snyk test folder` affect this? Will it correctly pick up the same file path?

orsagie: This doesn't affect test, only monitor.
Once the PR for persisting this data in registry is merged - They will all be the same project, but the relative path will be updated. `snyk monitor --file=A` and `snyk monitor` will result in the full path `/Users/blah/path/to/folder/A`, while `snyk monitor folder` will result in path `folder/A`.

I will change so that 
`snyk monitor --file=A` and `snyk monitor` return `folder/A`
`snyk monitor folder` returns `parentFolder/folder/A`

orsagie: Changed to always return the absolute path (after a talk with @arielorn)

lili2311:
I think the Windows test failure is real, will need to look into it

PR #875, opened 2019-11-20
fix: support show-vulnerable-paths without a value

kyegupov:
- [x] Ready for review

#### What does this PR do?
Allows to specify `--show-vulnerable-paths` flag without a value (that would map to "all")

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -56,8 +56,14 @@ async function test(...args: MethodArgs): Promise<string> {
   options.org = options.org || config.org;
 
   // making `show-vulnerable-paths` 'some' by default.
-  const svpSupplied = (options['show-vulnerable-paths'] || '').toLowerCase();
-  options.showVulnPaths = showVulnPathsMapping[svpSupplied] || 'some';
+  let svpSupplied = options['show-vulnerable-paths'] || '';
+  // Some legacy documentation refers to this as a boolean flag.
+  // We handle the "true" value as "some" (the default).
+  if ((svpSupplied as any) === true) {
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/index.ts
```diff
@@ -56,8 +56,14 @@ async function test(...args: MethodArgs): Promise<string> {
   options.org = options.org || config.org;
 
   // making `show-vulnerable-paths` 'some' by default.
-  const svpSupplied = (options['show-vulnerable-paths'] || '').toLowerCase();
-  options.showVulnPaths = showVulnPathsMapping[svpSupplied] || 'some';
+  let svpSupplied = options['show-vulnerable-paths'] || '';
+  // Some legacy documentation refers to this as a boolean flag.
+  // Let's handle that.
+  if ((svpSupplied as any) === true) {
+    svpSupplied = 'all';
```
adrukh: Hmm this is tricky. Not specifying the flag assumes `'some'` (old code), but specifying the flag without a value assumes `'all'`.

I'm not deep enough into this flag's implication to offer help, but will say this - start from the `help` text and adjust it so that the user understands what happens when the flag is omitted, used as a boolean, used with a valid string value, or used with an invalid string value. Writing some sensible and succinct `help` text will point us to the right code to write.

kyegupov: Yes, you're right, the default is "some".
I think we shouldn't overload the help text, it's pretty long as it is. It already states that the default is "some", so that should apply here as well.

lili2311:
We chatted more on Slack, consensus was this introduces a different expectation so we will not continue with this.

PR #881, opened 2019-11-26
[WIP] fix: check token for validity before making any requests

dkontorovskyy:
- [X] Ready for review

#### What does this PR do?
Fix auth error if `experimental-dep-graph` exists. 

If cli was invoked with `experimental-dep-graph` flag and invalid token, user will get error message
`Feature flag 'experimental-dep-graph' is not currently enabled for your org, to enable please contact snyk support`, cos token wasn't checked for validity and line https://github.com/snyk/snyk/blob/master/src/cli/commands/monitor/index.ts#L91 will return actually `401 Not Authorised`, but because we do `isFFSupported.ok` and it was `undefined`, user will see totally unrelated error message.

### DIFF
--- a/src/lib/api-token.ts
+++ b/src/lib/api-token.ts
@@ -1,17 +1,25 @@
-import { MissingApiTokenError } from '../lib/errors';
+import { AuthFailedError, MissingApiTokenError } from './errors';
 
 import * as config from './config';
 import * as userConfig from './user-config';
+import { isAuthed } from '../cli/commands/auth/is-authed';
 
 export function api() {
   // note: config.TOKEN will potentially come via the environment
   return config.api || config.TOKEN || userConfig.get('api');
 }
 
-export function apiTokenExists() {
+export async function apiTokenExists(): Promise<string> {
   const configured = api();
   if (!configured) {
     throw new MissingApiTokenError();
   }
+
+  const isValid = await isAuthed();
+
+  if (!isValid.ok) {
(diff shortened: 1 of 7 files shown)


src/lib/api-token.ts
```diff
@@ -1,17 +1,38 @@
...
 
+import request = require('./request');
 import * as config from './config';
 import * as userConfig from './user-config';
 
 export function api() {
   // note: config.TOKEN will potentially come via the environment
   return config.api || config.TOKEN || userConfig.get('api');
 }
+async function isTokenValid(token: string) {
```
gitphill: we have very similar function in src/cli/commands/auth/is-authed.ts - could reuse (use this function there) to avoid duplication?

src/lib/api-token.ts
```diff
@@ -1,17 +1,38 @@
...
+    url: `${config.API}/verify/token`,
+    body: {
+      api: token,
+    },
+    gzip: true,
+    json: true,
+  });
 
-export function apiTokenExists() {
+  return (response as any).body;
```
gitphill: type the function return object?

src/lib/api-token.ts
dkontorovskyy: Was looking for this one but haven't found :( Nice catch 👍

dkontorovskyy:
Closing this PR, cos this is not right approach

PR #882, opened 2019-11-26
fix: bug with --experimental-dep-graph and wrong auth token

dkontorovskyy:
- [X] Ready for review

#### What does this PR do?
Fix auth error if experimental-dep-graph exists.

If cli was invoked with experimental-dep-graph flag and invalid token, user will get error message
Feature flag 'experimental-dep-graph' is not currently enabled for your org, to enable please contact snyk support, cos token wasn't checked for validity and line https://github.com/snyk/snyk/blob/master/src/cli/commands/monitor/index.ts#L91 will return actually 401 Not Authorised, but because we do isFFSupported.ok and it was undefined, user will see totally unrelated error message.

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -18,7 +18,11 @@ import { MonitorOptions, MonitorMeta } from '../../../lib/types';
 import { MethodArgs, ArgsOptions } from '../../args';
 import { maybePrintDeps } from '../../../lib/print-deps';
 import * as analytics from '../../../lib/analytics';
-import { MonitorError, UnsupportedFeatureFlagError } from '../../../lib/errors';
+import {
+  AuthFailedError,
+  MonitorError,
+  UnsupportedFeatureFlagError,
+} from '../../../lib/errors';
 import { legacyPlugin as pluginApi } from '@snyk/cli-interface';
 import { isFeatureFlagSupportedForOrg } from '../../../lib/feature-flags';
 import { formatMonitorOutput } from './formatters/format-monitor-response';
@@ -87,6 +91,10 @@ async function monitor(...args0: MethodArgs): Promise<any> {
     const isFFSupported = await isFeatureFlagSupportedForOrg(
       _.camelCase('experimental-dep-graph'),
     );
+
+    if (isFFSupported.code === 401) {
+      throw AuthFailedError(isFFSupported.error, isFFSupported.code);
+    }
 
     if (!isFFSupported.ok) {
       throw new UnsupportedFeatureFlagError(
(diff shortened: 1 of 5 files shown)


src/lib/errors/authentication-failed-error.ts
```diff
@@ -1,13 +1,14 @@
...
-export function AuthFailedError(errorMessage, errorCode) {
-  const message = errorMessage
-    ? errorMessage
-    : 'Authentication failed. Please check the API token on ' + config.ROOT;
-  const error = new CustomError(message);
-  error.code = errorCode || 401;
+export function AuthFailedError(
+  errorMessage: string = 'Authentication failed. Please check the API token on ' +
+    config.ROOT,
+  errorCode = 401,
```
gitphill: nice refactor

src/cli/commands/monitor/index.ts
```diff
@@ -83,6 +88,10 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
   apiTokenExists();
 
+  if (!(await isAuthed())) {
```
gitphill: Other modules perform isAuthed _before_ apiTokenExists, expecting apiTokenExists to throw MissingApiTokenError when auth fails. Seems a bit subtle to me, however we probably have this pattern to support commands that can be run when unauthenticated?

* ignore: https://github.com/snyk/snyk/blob/c965884b69624d7ab293391c0cd9107e3d03c27b/src/cli/commands/ignore.ts#L18-L26
* protect wizard: https://github.com/snyk/snyk/blob/79b792b57407408eb7df71e9cec3efc2605169a8/src/cli/commands/protect/wizard.ts#L99-L108
* test: I couldn't see isAuthed being used in test, but apiTokenExists is called - maybe same bug there?

test/acceptance/cli-monitor.acceptance.test.ts
```diff
@@ -133,11 +133,37 @@ test('monitor for package with no name in lockfile', async (t) => {
   t.pass('succeed');
 });
 
-test('`monitor npm-package with experimental-dep-graph not enabled`', async (t) => {
+test('`monitor npm-package with experimental-dep-graph enabled, but bad auth token`', async (t) => {
```
gitphill: 👍

shaninja:
Just want to make sure - did you test this with FF enabled for the group and not the org? (which was the original issue reported, so it would be nice to validate it)

dkontorovskyy:
Doing a smaller refactor to fix the issue and creating a ticket for later for overall CLI improvement, cos this is much bigger scope that was thought in the beginning.

PR #884, opened 2019-11-27
feat: asynchronously process monitor args

gitphill:
- [X] Ready for review

#### What does this PR do?

Use Promise.all on each path passed to monitor.
Added test for monitor using multiple paths.

Run 'snyk monitor ./path1 ./path2 ./path3 ...'
Use DEBUG=* to see each path is now processed asynchronously.

In preparation for auto detect --all-projects feature.

BST-1079

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -54,12 +54,135 @@ async function promiseOrCleanup<T>(
   });
 }
 
+// detects package manager
+// loads appropriate plugin and inspects for dependencies
+// finally posts dependencies to registry
+async function detectInspectMonitor(
+  path: string,
+  options: MonitorOptions = {},
+): Promise<GoodResult | BadResult> {
+  try {
+    await validateMonitorPath(path, options.docker);
+
+    let packageManager = detect.detectPackageManager(path, options);
+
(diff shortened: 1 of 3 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -54,12 +54,135 @@ async function promiseOrCleanup<T>(
...
+
+    const analysisType = options.docker ? 'docker' : packageManager;
+
+    const analyzingDepsSpinnerLabel =
+      'Analyzing ' + analysisType + ' dependencies for ' + displayPath;
+
+    const postingMonitorSpinnerLabel =
+      'Posting monitor snapshot for ' + displayPath + ' ...';
+
+    await spinner(analyzingDepsSpinnerLabel);
```
gitphill: not sure about these spinners anymore? copied form existing, but since this is now done asynchronously the spinner can flicker as other jobs kick in

orsagie: can we start a spinner around this asyncMonitor? and update it as we start more jobs?

gitphill: I think it's ok for now, does indicate what's currently happening

PR #886, opened 2019-11-28
feat: find files func in prep for auto detection

gitphill:
- [X] Ready for review

#### What does this PR do?
Added find-files module with find function to help search for files.
Added test for find and test fixtures.

Just helper function at the moment, separate PR to integrate with CLI commands and auto detection.

BST-1072

### DIFF
--- a/src/lib/find-files.ts
+++ b/src/lib/find-files.ts
@@ -1,0 +1,109 @@
+import * as fs from 'fs';
+import * as pathLib from 'path';
+// TODO: use util.promisify once we move to node 8
+
+/**
+ * Returns files inside given file path.
+ *
+ * @param path file path.
+ */
+export async function readDirectory(path: string): Promise<string[]> {
+  return await new Promise((resolve, reject) => {
+    fs.readdir(path, (err, files) => {
+      if (err) {
+        reject(err);
+      }
+      resolve(files);
+    });
+  });
+}
+
+/**
+ * Returns file stats object for given file path.
+ *
+ * @param path path to file or directory.
+ */
+export async function getStats(path: string): Promise<fs.Stats> {
+  return await new Promise((resolve, reject) => {
+    fs.stat(path, (err, stats) => {
+      if (err) {
+        reject(err);
+      }
+      resolve(stats);
+    });
+  });
+}
+
+/**
+ * Find all files in given search path. Returns paths to files found.
+ *
+ * @param path file path to search.
+ * @param ignore (optional) files to ignore. Will always ignore node_modules.
+ * @param filter (optional) file names to find. If not provided all files are returned.
+ * @param levelsDeep (optional) how many levels deep to search, defaults to two, this path and one sub directory.
+ */
+export async function find(
+  path: string,
+  ignore: string[] = [],
+  filter: string[] = [],
+  levelsDeep = 2,
+): Promise<string[]> {
+  const found: string[] = [];
+  // ensure we ignore find against node_modules path.
+  if (path.endsWith('node_modules')) {
+    return found;
+  }
+  // ensure node_modules is always ignored
+  if (!ignore.includes('node_modules')) {
+    ignore.push('node_modules');
+  }
+  try {
+    if (levelsDeep < 0) {
+      return found;
+    } else {
+      levelsDeep--;
+    }
+    const fileStats = await getStats(path);
+    if (fileStats.isDirectory()) {
+      const files = await findInDirectory(path, ignore, filter, levelsDeep);
+      found.push(...files);
+    } else if (fileStats.isFile()) {
+      const fileFound = findFile(path, filter);
+      if (fileFound) {
+        found.push(fileFound);
+      }
+    }
+    return found;
+  } catch (err) {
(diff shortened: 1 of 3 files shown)


src/lib/find-files.ts
```diff
@@ -0,0 +1,79 @@
...
+  searchPath: string,
+  ignore: string[] = [],
+  filenames: string[] = [],
+): Promise<string[]> {
+  const found: string[] = [];
+  try {
+    const stats = await getStats(searchPath);
+    if (stats.isDirectory()) {
+      const paths = await readDirectory(searchPath);
+      for (const subPath of paths) {
```
orsagie: Do we want to do this asynchronously for found paths?

src/lib/find-files.ts
```diff
@@ -0,0 +1,79 @@
...
+  });
+}
+
+/**
+ * Returns file stats object for given file path.
+ *
+ * @param filePath path to file or directory.
+ */
+export async function getStats(filePath: string): Promise<fs.Stats> {
+  return await new Promise((resolve, reject) => {
```
lili2311: why is a new Promise needed here?

src/lib/find-files.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Recusively find all files in given search path. Returns paths to files found.
+ *
+ * @param searchPath path to search.
+ * @param ignore (optional) paths to ignore.
+ * @param filenames (optional) filenames to find. If not provided all files are returned.
+ */
+export async function find(
```
lili2311: Why do we need to find all files? We only care about finding supported target files on disk is that not much much faster? 
What if there are thousands of files in a folder and none of them are even a supported one?

src/lib/find-files.ts
gitphill: allow async/await in find method - could be synchronous if you prefer? (would be slower)

src/lib/find-files.ts
gitphill: Yes the two optional filters allow you to focus in on the files of interest - when we come to use in auto detect we'll pass all our detectable files into filenames, and all the directories we want to ignore into ignore, bit like so:

```
export async function detect(searchPath: string): Promise<string[]> {
  const files = await find(searchPath, IGNORE_DIRECTORIES, DETECTABLE_FILES);
  return files;
}
```

Noticed in spike we only go one-level deep from cwd, thought we could support deeply nested project structures

On reflection though, this could be complicated for things like gradle where I think we only want the top-level manifest.?Could add facility to break recursion on these projects though. Do we have a sense from customers what their directory structure is like?

src/lib/find-files.ts
gitphill: Not sure I understand, could you elaborate?

src/lib/find-files.ts
lili2311: We only want 1 level deep on purpose for now this is `v1` we agreed, there are some complications like gradle scanning indeed that we will have to address down the line before properly being able to support scanning. This requires a lot more work and changes to understand that a group of manifests must be processed together and how they relate to each other. 

There is also performance implications of doing this recursively for all folder (imagine users calling snyk test on a root level folder of the machine), so for now we will start 1 level deep and expand further at a later time as we need to build in some other restrictions.

src/lib/find-files.ts
lili2311: ah is all this because we can't yet use `promisify`, got it!

src/lib/find-files.ts
lili2311: I think she is saying you can do `Promise.all()` instead of `for loop` (maybe with a pMap to reduce concurrency)

src/lib/find-files.ts
```diff
@@ -0,0 +1,79 @@
...
+  searchPath: string,
+  ignore: string[] = [],
+  filenames: string[] = [],
+): Promise<string[]> {
+  const found: string[] = [];
+  try {
+    const stats = await getStats(searchPath);
+    if (stats.isDirectory()) {
```
lili2311: this whole function is quite hard to read with so many if else, maybe split it up and return early where possible?

src/lib/find-files.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param filenames (optional) filenames to find. If not provided all files are returned.
+ */
+export async function find(
+  searchPath: string,
+  ignore: string[] = [],
+  filenames: string[] = [],
+): Promise<string[]> {
+  const found: string[] = [];
+  try {
+    const stats = await getStats(searchPath);
```
lili2311: maybe `filePathStats` to be more explicit?

src/lib/find-files.ts
gitphill: yea I'll take a look, it is a bit intricate😃

src/lib/find-files.ts
gitphill: yea promisify is better once we drop node 6

src/lib/find-files.ts
gitphill: ok will re-work

PR #887, opened 2019-11-28
fix: keep inspect synchronous

gitphill:
- [X] Ready for review

#### What does this PR do?
Reverts calls to inspect back to synchronous operations to
avoid problems running plugins concurrently.
Keep calls to registry as asynchronous operations.

After discussing and testing, decided running all monitor operations asynchronously could cause issues (performance, inspects interfering with one another). This change keeps registry calls asynchronous, but changes inspect calls back to synchronous.

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -26,6 +26,7 @@ import {
 import { legacyPlugin as pluginApi } from '@snyk/cli-interface';
 import { isFeatureFlagSupportedForOrg } from '../../../lib/feature-flags';
 import { formatMonitorOutput } from './formatters/format-monitor-response';
+import { SupportedPackageManagers } from '../../../lib/package-managers';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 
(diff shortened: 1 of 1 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -227,12 +196,74 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
+      );
+
+      if (inspectResult.plugin.packageManager) {
+        packageManager = inspectResult.plugin.packageManager;
+      }
+
+      analytics.add('pluginName', inspectResult.plugin.name);
+      await spinner.clear(analyzingDepsSpinnerLabel)(inspectResult);
+
+      snapshots.push({
```
kyegupov: Actually, you don't need to wait for all the inspects to complete to begin sending snapshots. It makes sense to call saveSnapshot here and save the resulting promise. Then, in line 262, you will just need to call Promise.all on that array.

gitphill: updated, thanks

src/cli/commands/monitor/index.ts
```diff
@@ -227,12 +196,77 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
+      );
+    } catch (err) {
+      // push this error, the loop continues
+      results.push({ ok: false, data: err, path });
+    }
+  }
+
+  // Part 2: send snapshot to registry asynchronously
+  const snapshotResults = await Promise.all(snapshots);
+  results = results.concat(...snapshotResults);
```
kyegupov: this .concat means that you don't retain the original order of the arguments. E.g. if you have arguments [a,b,c] and the outcomes are [a=success, b=fail, c=success], the results will be [b-error, a-success, c-success], and they will be printed by the CLI in that order, not mirroring the original.

gitphill: I couldn't get order to come out like that - but I've updated with another step to ensure snapshots are pushed to results in the order they came in

src/cli/commands/monitor/index.ts
```diff
@@ -227,12 +196,86 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
+      );
+    } catch (err) {
+      // push this error, the loop continues
+      snapshots.push(Promise.resolve([{ ok: false, data: err, path }]));
+    }
+  }
+
+  // Part 2: send snapshots to registry asynchronously
+  const snapshotResults = await Promise.all(snapshots);
+  const snapshotsResultsFlat = ([] as Array<GoodResult | BadResult>).concat(
```
kyegupov: is this "concat" necessary? Why can't you just use snapshotResults?

gitphill: snapshotResults is an array of arrays

kyegupov: ah, makes sense now. Yes, one argument can produce multiple results (e.g. gradle subprojects).
But then you will lose those extra results when filtering in line 272!

gitphill: do gradle subprojects have the same path?

gitphill: ah yea good spot, hmm not sure this approach will be possible then.
the original 

```js
const snapshotResults = await Promise.all(snapshots);
results = ([] as Array<GoodResult | BadResult>).concat(...snapshotResults);
```

seemed to respect the added to snapshots array passed to Promise.all - what about concat could reorder array?

i'll add test for gradle projects

kyegupov: "seemed to respect the added to snapshots array passed to Promise.all" - I think this is what we should be striving for

PR #892, opened 2019-12-03
feat: add fail-on arg

gitphill:
- [X] Ready for review

#### What does this PR do?

Added --fail-on argument for snyk test CLI command.
When supplied this argument will check whether any vulns
are upgradable or patchable before failing.
If a vulnerabilities are found be they cannot be upgraded
or patched, a sucessful error code is returned instead.
Test cases added.
Added help text for new arg.

`snyk test --fail-on` with either all, upgradable or patchable

BST-1063


test/acceptance/cli-test/cli-test.fail-on.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+        utils.chdirWorkspaces('fail-on');
+        await params.cli.test('upgradable', {
+          failOn: 'upgradable',
+        });
+        t.fail('expected test to throw exception');
+      } catch (err) {
+        t.equal(err.code, 'VULNS', 'should throw exception');
+      }
+    },
+    'test vulnerable project with pinnable and --fail-on=upgradable': (
```
gitphill: these pinnable tests are failing in CI because python is not installed (I think) - can we install python?

src/cli/commands/test/index.ts
```diff
@@ -230,6 +253,63 @@ async function test(...args: MethodArgs): Promise<string> {
   return response;
 }
 
+function shouldFail(vulnerableResults: any[], failOn: FailOn) {
```
lili2311: this is very nice ❤️

src/cli/commands/test/index.ts
```diff
@@ -66,6 +67,10 @@ async function test(...args: MethodArgs): Promise<string> {
     return Promise.reject(new Error('INVALID_SEVERITY_THRESHOLD'));
   }
 
+  if (options.failOn && !validateFailOn(options.failOn)) {
+    return Promise.reject(new FailOnError());
```
lili2311: this error is not formatted in red like others 🤔 something about this flow in here is different.

lili2311: this fixes it and seems we need the same for `INVALID_SEVERITY_THRESHOLD`
```  if (options.failOn && !validateFailOn(options.failOn)) {
    const error = new FailOnError();
    return Promise.reject(chalk.red.bold(error.message));
  }
```

Not important for this PR though

src/cli/commands/test/index.ts
gitphill: would be better with types - is there a good type to use?

src/cli/commands/test/index.ts
gitphill: yea looks like legacy-errors.js does this, your suggestion works

src/cli/commands/test/index.ts
gitphill: doesn't look possible to type without reworking how we parse results

test/acceptance/cli-fail-on.test.ts
```diff
@@ -0,0 +1,509 @@
+import * as tap from 'tap';
```
gitphill: changed this to a test as apposed to a cli spec - couldn't get sinon stubs to work there

gitphill: had to separate out stub tests into their own file - otherwise they would interfere with other tests

src/cli/commands/test/index.ts
```diff
@@ -81,6 +87,7 @@ async function test(...args: MethodArgs): Promise<string> {
 
     try {
       res = (await snyk.test(path, testOpts)) as LegacyVulnApiResult;
+      // console.log(JSON.stringify(res));
```
lili2311: please remove

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -1424,10 +1425,6 @@ after('teardown', async (t) => {
   }
 });
 
-function chdirWorkspaces(subdir = '') {
```
lili2311: not needed?

test/acceptance/workspace-helper.ts
```diff
@@ -0,0 +1,19 @@
+import * as path from 'path';
+import { readFileSync } from 'fs';
```
lili2311: ❤️

PR #896, opened 2019-12-06
fix: check that provided remote url is a string

AndreDalcher:
- [x] Ready for review

#### What does this PR do?
Check that the value passed to --remote-repo-url is a string (it is possible for it to be a boolean otherwise)

Setting the flag with no argument caused the value `true` to be persisted


gitphill:
Our other error messages get written as bold red text - this seems to output default text color. Think it needs a `chalk.red.bold` wrapper around the error message somewhere.

AndreDalcher:
@joshje changed, and logic centralised

PR #908, opened 2019-12-13
fix: filter each folder to grab 1 file per project type

lili2311:
- [x] Ready for review

#### What does this PR do?

- for every manifest we find we should only return 1 default manifest to process based on project type & which plugin will process it

### DIFF
--- a/src/lib/find-files.ts
+++ b/src/lib/find-files.ts
@@ -1,5 +1,7 @@
 import * as fs from 'fs';
 import * as pathLib from 'path';
+import * as _ from 'lodash';
+import { detectPackageManagerFromFile } from './detect';
 // TODO: use util.promisify once we move to node 8
 
 /**
@@ -73,7 +75,7 @@ export async function find(
         found.push(fileFound);
       }
     }
-    return found;
+    return filterForDefaultManifests(found);
   } catch (err) {
     throw new Error(`Error finding files in path '${path}'.\n${err.message}`);
   }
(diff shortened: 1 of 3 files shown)


src/lib/find-files.ts
```diff
@@ -105,5 +110,95 @@ async function findInDirectory(
...
+    return null;
+  }
+}
+
+function chooseBestManifest(
+  files: Array<{ base: string; path: string }>,
+  projectType: string,
+): string | null {
+  switch (projectType) {
+    case 'node': {
```
lili2311: we should be checking for default files based on `handler` or `plugin`

src/lib/find-files.ts
```diff
@@ -1,5 +1,8 @@
 import * as fs from 'fs';
 import * as pathLib from 'path';
+import _ = require('lodash');
```
dkontorovskyy: Can we don't use `= require` syntax, please?

src/lib/find-files.ts
```diff
@@ -73,7 +76,9 @@ export async function find(
         found.push(fileFound);
       }
     }
-    return found;
+    const filteredFiles = filterForDefaultManifests(found);
```
dkontorovskyy: Can be squashed into 1 line
```suggestion
      return filterForDefaultManifests(found);
```

src/lib/find-files.ts
lili2311: yep! it auto imports like this :(

dkontorovskyy:
Looks good. Is there any tests related to `Gemfiles`? I see only related to `npm`

lili2311:
> Looks good. Is there any tests related to `Gemfiles`? I see only related to `npm`

Thanks for spotting that will fix it

lili2311:
@dkontorovskyy @gitphill added a ruby test

src/lib/find-files.ts
```diff
@@ -105,5 +107,95 @@ async function findInDirectory(
       return find(resolvedPath, ignore, filter, levelsDeep);
     });
   const found = await Promise.all(toFind);
-  return Array.prototype.concat.apply([], found);
+  const foundFiles = Array.prototype.concat.apply([], found);
```
orsagie:
```suggestion
return Array.prototype.concat.apply([], found);
```

src/lib/find-files.ts
```diff
@@ -107,3 +109,92 @@ async function findInDirectory(
...
+      return packageJson.path;
+    }
+    case 'rubygems': {
+      const defaultManifest = files.filter((path) =>
+        ['Gemfile.lock'].includes(path.base),
+      )[0];
+      return defaultManifest.path;
+    }
+    default: {
+      return null;
```
orsagie: If there are projects of other types can we debug saying we don't support those yet?

lili2311: can it be a separate PR?

PR #915, opened 2019-12-18
feat: send targetfile same as for monitor

lili2311:
- [ ] Ready for review

#### What does this PR do?

To match a project on the BE the targetFile must be send only if the plugin exposes it, just like for monitor. This will allow us to always match the same project by name

Original PR that added this: https://github.com/snyk/snyk/pull/662

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -378,7 +378,7 @@ async function assembleLocalPayloads(
       }
 
       let body: PayloadBody = {
-        targetFile: pkg.targetFile,
+        targetFile: deps.plugin.targetFile,
         projectNameOverride: options.projectName,
         policy: policy && policy.toString(),
         docker: pkg.docker,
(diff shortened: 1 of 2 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -378,7 +378,7 @@ async function assembleLocalPayloads(
       }
 
       let body: PayloadBody = {
-        targetFile: pkg.targetFile,
+        targetFile: deps.plugin.targetFile,
```
gitphill: we will need to take this into account on auto detect
https://github.com/snyk/snyk/pull/902/files#diff-8af5d8cc823dc3809a95696f97399198R343-R348
where we only set the plugin name

gitphill: maybe make it a required property on interface?

lili2311: targetFile is not required, it is sometimes there sometimes not (legacy reasons). We will need to validate for auto detect indeed that we still correctly bubble this up. We can add an assertion on a test that we do not send it for ruby & do send it for some others?

lili2311:
More upstream fixes may be needed, investigating

lili2311:
This is a first step towards a fix, this fixes an issue where for `test` ruby right now is sending targetFile but monitor it is not. This is causing a mismatch in name and the policy is not applied.

PR #926, opened 2019-12-26
fix: calculate package.json indentation when running snyk wizard

darmalovan:
- [x] Ready for review

#### What does this PR do?
Calculates `package.json` indentation when running `snyk wizard`

Fixes https://github.com/snyk/snyk/issues/911

We use to have a default value of 2 as the white indentation number.
This change calculates the indentation used in the package.json file
and uses it when updating the file.

### DIFF
--- a/src/cli/commands/protect/wizard.ts
+++ b/src/cli/commands/protect/wizard.ts
@@ -305,6 +305,15 @@ function addProtectScripts(existingScripts, npmVersion, options) {
   return scripts;
 }
 
+function calculatePkgFileIndentation(packageFile: string): number {
+  let pkgIndentation = 2;
+  const whitespaceMatch = packageFile.match(/{\n(\s+)"/);
+  if (whitespaceMatch && whitespaceMatch[1]) {
+    pkgIndentation = whitespaceMatch[1].length;
+  }
+  return pkgIndentation;
+}
+
 interface Pkg {
   scripts: any;
   snyk: boolean;
@@ -335,6 +344,7 @@ function processAnswers(answers, policy, options) {
     targetFile.endsWith('yarn.lock');
 
   let pkg = {} as Pkg;
+  let pkgIndentation = 2;
 
   analytics.add(
     'answers',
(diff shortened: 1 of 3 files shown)


src/cli/commands/protect/wizard.ts
```diff
@@ -419,6 +420,14 @@ function processAnswers(answers, policy, options) {
       // an `npm install` which will change the deps
       return fs
         .readFile(packageFile, 'utf8')
+        .then((packageFileString) => {
```
lili2311: I think this would be nicer as a function instead of adding on to the promise chain, we are trying to refactor all the promise chains here

test/wizard-process-answers.test.js
```diff
@@ -288,3 +289,30 @@ test('wizard maintains whitespace at beginning and end of package.json', functio
       t.end();
     });
 });
+
+test('wizard updates vulns and retains indentation', function(t) {
```
lili2311: ❤️

test/wizard-process-answers.test.js
```diff
@@ -3,6 +3,7 @@ var test = require('tap').test;
 var proxyquire = require('proxyquire');
 var path = require('path');
 var sinon = require('sinon');
+var fs = require('fs');
```
lili2311: how about a quick refactor to use `const`?

test/wizard-process-answers.test.js
```diff
@@ -288,3 +289,30 @@ test('wizard maintains whitespace at beginning and end of package.json', functio
...
+
+test('wizard updates vulns and retains indentation', function(t) {
+  const old = process.cwd();
+  const dir = path.resolve(__dirname, 'fixtures', 'four-spaces');
+  const manifestPath = path.resolve(dir, 'package.json');
+  const original = fs.readFileSync(manifestPath, 'utf-8');
+  writeSpy = sinon.spy(); // create a new spy
+  process.chdir(dir);
+
+  wizard
```
lili2311: please use async / await where possible

PR #928, opened 2019-12-27
chore: refactor shared code out and split out helper functions

lili2311:
- [x] Ready for review

#### What does this PR do?
Split out helper functions into their own files

### DIFF
--- a/src/lib/plugins/get-deps-from-plugin.ts
+++ b/src/lib/plugins/get-deps-from-plugin.ts
@@ -1,15 +1,14 @@
+import * as debugModule from 'debug';
 import { legacyPlugin as pluginApi } from '@snyk/cli-interface';
-import detect = require('../../lib/detect');
-import { TestOptions } from '../types';
-import { Options } from '../types';
-import { NoSupportedManifestsFoundError } from '../errors';
-import { find } from '../find-files';
-import { AUTO_DETECTABLE_FILES } from '../detect';
-import { getSinglePluginResult } from '../plugins/get-single-plugin-result';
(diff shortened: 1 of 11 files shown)


src/lib/plugins/get-deps-from-plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+    const targetFiles = await find(root, [], AUTO_DETECTABLE_FILES, 1);
+    debug(
+      `auto detect manifest files, found ${targetFiles.length}`,
+      targetFiles,
+    );
+    if (targetFiles.length === 0) {
+      throw NoSupportedManifestsFoundError([root]);
+    }
+    inspectRes = await getMultiPluginResult(root, options, targetFiles);
+    return inspectRes;
```
dkontorovskyy: Shouldn't have here return, cos it will never get to next `if..else` block

src/lib/plugins/get-multi-plugin-result.ts
```diff
@@ -0,0 +1,73 @@
+import * as _ from 'lodash';
+import * as cliInterface from '@snyk/cli-interface';
+
+import pathUtil = require('path');
```
dkontorovskyy:
```suggestion
import * as pathUtil from 'path';
```

src/lib/snyk-test/run-test.ts
```diff
@@ -3,12 +3,9 @@ import * as fs from 'fs';
 import pathUtil = require('path');
```
dkontorovskyy: import * as pathUtil from 'path';

PR #931, opened 2019-12-27
chore: refactor manage url into a function

lili2311:
- [ ] Ready for review

#### What does this PR do?

Move constructing manageUrl to a more appropriate place.

### DIFF
--- a/src/cli/commands/monitor/formatters/format-monitor-response.ts
+++ b/src/cli/commands/monitor/formatters/format-monitor-response.ts
@@ -1,15 +1,18 @@
 import * as _ from 'lodash';
 import chalk from 'chalk';
+import * as url from 'url';
+
 import { MonitorResult } from '../../../../lib/types';
+import * as config from '../../../../lib/config';
 
 export function formatMonitorOutput(
   packageManager,
   res: MonitorResult,
-  manageUrl,
   options,
   projectName?: string,
   foundProjectCount?: number,
 ): string {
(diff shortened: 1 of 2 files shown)


src/cli/commands/monitor/formatters/format-monitor-response.ts
```diff
@@ -61,3 +64,17 @@ export function formatMonitorOutput(
...
+function buildManageUrl(resId: string, org?: string): string {
+  const endpoint = url.parse(config.API);
+  let leader = '';
+  if (org) {
+    leader = '/org/' + org;
+  }
+  endpoint.pathname = leader + '/manage';
+  const manageUrl = url.format(endpoint);
+
+  // TODO: what was this meant to do?
```
lili2311: @ivanstanev this looks wrong, what is it meant to do?

dkontorovskyy: Shall we merge with this, even though line#78 doesn't do anything

lili2311: It is the same as it was before, so it should be safe to do so until we can get more answers.

lili2311: WDYT?

dkontorovskyy: Totally agree. But we need to get some clarification what that was supposed to do and clean it after or refactor

PR #952, opened 2020-01-08
fix: increase cli test timeout

gitphill:
- [X] Ready for review

#### What does this PR do?
Increase tap timeout from 300 to 400 due to increased test count we're
regularly hitting this timeout.

npm test:acceptance

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -59,7 +59,7 @@ async function webAuth(via: AuthCliCommands) {
     spinner.start();
     await setTimeout(() => {
       open(urlStr, { wait: false });
-    }, 2000);
+    }, 0);
 
     const res = await testAuthComplete(token);
     return res;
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "prepare": "npm run build",
     "test:common": "npm run check-tests && npm run build && npm run lint && node --require ts-node/register src/cli test --org=snyk",
     "test:acceptance": "tap test/acceptance/**/*.test.* -Rspec --timeout=300 --node-arg=-r --node-arg=ts-node/register",
-    "test:system": "tap test/system/*.test.* -Rspec --timeout=300 --node-arg=-r --node-arg=ts-node/register",
+    "test:system": "tap test/system/*.test.* -Rspec --node-arg=-r --node-arg=ts-node/register",
```
darscan: pretty sure leaving this out means a default timeout of 60 seconds

gitphill: i've tried all sorts (up to 600ms), starting to think timeout is not the issue

gitphill:
OK after lots of experimentation, the real issue was the amount of time we take to open the browser window. Reduced this from 2 seconds to immediate. Allows tests to more consistently pass and not timeout. The timeout was cause because the test teardown brought down the local server before all async tests had returned.

PR #955, opened 2020-01-10
feat: add exclude arg and use to ignore

gitphill:
- [X] Ready for review

#### What does this PR do?
Ignore directories when using --all-projects.
Added tests for test and monitor with --exclude.

`snyk test --all-projects --detection-depth=2 --exclude=example-dir,example-2`

BST-1126

### DIFF
--- a/src/lib/errors/exclude-flag-invalid-input.ts
+++ b/src/lib/errors/exclude-flag-invalid-input.ts
@@ -1,0 +1,13 @@
+import { CustomError } from './custom-error';
+
+export class ExcludeFlagInvalidInputError extends CustomError {
+  private static ERROR_CODE = 422;
+  private static ERROR_MESSAGE =
+    'The --exclude argument must be a comma seperated list of directory names and cannot contain a path.';
+
+  constructor() {
+    super(ExcludeFlagInvalidInputError.ERROR_MESSAGE);
+    this.code = ExcludeFlagInvalidInputError.ERROR_CODE;
+    this.userMessage = ExcludeFlagInvalidInputError.ERROR_MESSAGE;
+  }
+}
--- a/src/lib/errors/option-missing-error.ts
+++ b/src/lib/errors/option-missing-error.ts
@@ -1,0 +1,10 @@
+import { CustomError } from './custom-error';
+
+export class OptionMissingErrorError extends CustomError {
+  constructor(option: string, required: string) {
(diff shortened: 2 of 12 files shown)


src/lib/plugins/get-deps-from-plugin.ts
```diff
@@ -25,7 +25,13 @@ export async function getDepsFromPlugin(
 
   if (options.allProjects) {
     const levelsDeep = options.detectionDepth || 1; // default to 1 level deep
-    const targetFiles = await find(root, [], AUTO_DETECTABLE_FILES, levelsDeep);
+    const ignore = options.exclude ? options.exclude.split(',') : [];
```
lili2311: what is `options.exclude` is badly formatted?

gitphill: lets add a test :)

lili2311: `--exclude folder1,folder2` or even ` --exclude=a/b/c` as long as it doesn't break the current test run.

gitphill: ok added some validation and tests

src/lib/errors/exclude-flag-invalid-input.ts
```diff
@@ -0,0 +1,13 @@
+import { CustomError } from './custom-error';
+
+export class ExcludeFlagInvalidInputError extends CustomError {
```
lili2311: I think we can merge these two errors and provide the custom string into the constructor at call site?

src/lib/errors/option-missing-error.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomError } from './custom-error';
+
+export class OptionMissingErrorError extends CustomError {
```
lili2311: ❤️

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -80,7 +80,6 @@ export const AllProjectsTests: AcceptanceTests = {
         'calls maven plugin twice',
       );
       // maven
-      console.log(result);
```
lili2311: 😅

src/lib/errors/exclude-flag-invalid-input.ts
lili2311: can do it later

PR #965, opened 2020-01-22
feat: enable cocoapods for --all-projects scanning

lili2311:
- [x] Ready for review

#### What does this PR do?

Enable Cocoapods `Podfile` and `Podfile.lock` to be picked up when running `test` or `monitor` with `--all-projects` flag


src/cli/commands/monitor/index.ts
```diff
@@ -85,9 +84,15 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
-      const analysisType =
-        (options.docker ? 'docker' : guessedPackageManager) || 'all';
+      let analysisType = 'all';
+      let packageManager;
+      if (options.allProjects) {
+        analysisType = 'all';
+      } else if (options.docker) {
+        analysisType = 'docker';
+      } else {
+        packageManager = detect.detectPackageManager(path, options);
```
lili2311: only run this check if we are not in `Docker` or `all-projects` mode as it will fail when no supported manifests are found

lili2311:
Includes the PR https://github.com/snyk/snyk/pull/966

src/lib/snyk-test/run-test.ts
```diff
@@ -159,8 +159,10 @@ async function runTest(
       res.filesystemPolicy = !!payloadPolicy;
       if (!options['ignore-policy']) {
         res.policy = res.policy || (payloadPolicy as string);
+        console.log('*****  res.policy', res.policy);
```
dkontorovskyy: It this a left over?

src/lib/snyk-test/run-test.ts
```diff
@@ -159,8 +159,10 @@ async function runTest(
...
         const policy = await snyk.policy.loadFromText(res.policy);
         res = policy.filter(res, root);
+        console.log('***** ', policy);
```
dkontorovskyy: It this a left over?

src/lib/find-files.ts
```diff
@@ -193,6 +193,12 @@ function chooseBestManifest(
       )[0];
       return defaultManifest.path;
     }
+    case 'cocoapods': {
+      const defaultManifest = files.filter((path) =>
+        ['Podfile'].includes(path.base),
```
dkontorovskyy: Not a blocker, just a question why not `'Podfile' === path.base`?

lili2311: no reason copy + 🍝, will update in the next PR

dkontorovskyy: Because it's only relevant for `npm` and `yarn` and we bring it for all of others. We will have refactor after we release everything I think. I bet we could simplify when more ecosystems will be supported with `--all-projects`

lili2311: It is relevant for any ecosystem where more than 1 manifest file can be detected in the same folder and we need to choose. Technically we could only look for Podfile here and remove this entry, but there are other cocopoads manifests we may auto detect later on which are currently supported via `--file`

PR #970, opened 2020-01-24
feat: increase default levels deep to 2 for `--all-projects`

lili2311:
- [x] Ready for review

#### What does this PR do?
Enable the default detection depths to increase from 1 to 2 folders deep.

### DIFF
--- a/src/lib/plugins/get-deps-from-plugin.ts
+++ b/src/lib/plugins/get-deps-from-plugin.ts
@@ -24,7 +24,7 @@ export async function getDepsFromPlugin(
   let inspectRes: pluginApi.InspectResult;
 
   if (options.allProjects) {
-    const levelsDeep = options.detectionDepth || 1; // default to 1 level deep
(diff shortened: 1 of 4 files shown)


src/lib/plugins/get-deps-from-plugin.ts
```diff
@@ -24,7 +24,7 @@ export async function getDepsFromPlugin(
   let inspectRes: pluginApi.InspectResult;
 
   if (options.allProjects) {
-    const levelsDeep = options.detectionDepth || 1; // default to 1 level deep
+    const levelsDeep = options.detectionDepth;
```
lili2311: default defined inside `find` function is `2`

gitphill: I can predict the future!
_goes to buy lottery ticket_

gitphill:
we could update help text with new default depth

PR #975, opened 2020-01-26
Track rubygems projects' targetFilerelativePath

aviadhahami:
- [x] Ready for review

#### What does this PR do?
Added `analytics` logging for rubygems projects using `--file` flag;

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -155,6 +155,12 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         const targetFileRelativePath = tFile
           ? pathUtil.join(pathUtil.resolve(path), tFile)
           : '';
+
+        // target file logging as a port of [Jira: BST-1124]
+        if (packageManager === 'rubygems' && options.file) {
+          analytics.add('targetFileRelativePath', targetFileRelativePath);
+        }
+
         const res: MonitorResult = await promiseOrCleanup(
           snykMonitor(
             path,
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -2,6 +2,7 @@ export = test;
 
 import * as _ from 'lodash';
 import chalk from 'chalk';
+import * as pathUtil from 'path';
 import * as snyk from '../../../lib';
 import * as config from '../../../lib/config';
 import { isCI } from '../../../lib/is-ci';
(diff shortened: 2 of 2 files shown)


src/cli/commands/test/index.ts
```diff
@@ -124,9 +125,17 @@ async function test(...args: MethodArgs): Promise<string> {
     // with scripts that use a callback with test. Coerce results/errors to be arrays
     // and add the result options to each to be displayed
     const resArray: any[] = Array.isArray(res) ? res : [res];
-
     for (let i = 0; i < resArray.length; i++) {
       results.push(_.assign(resArray[i], { path }));
+
+      if (options.file && resArray[i].packageManager === 'rubygems') {
```
orsagie: This might be better done 
 after https://github.com/snyk/snyk/blob/a724b210cd2ad38d7124ff6257ed571ee34d973a/src/lib/snyk-test/run-test.ts#L353 as there might be multiple files that are tested (--all-projects).

src/cli/commands/monitor/index.ts
```diff
@@ -155,6 +155,12 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         const targetFileRelativePath = tFile
           ? pathUtil.join(pathUtil.resolve(path), tFile)
           : '';
+
+        // target file logging as a port of [Jira: BST-1124]
+        if (packageManager === 'rubygems' && options.file) {
```
orsagie: We should log this even if the `options.file` option was not set

lili2311: so we definitely  don't already track this in existing analytics?

orsagie: we don't

lili2311: what about older CLIs? Have you considered doing this in the main app instead?

src/cli/commands/test/index.ts
aviadhahami: Not sure I understand what you mean

src/cli/commands/monitor/index.ts
aviadhahami: @lili2311 you are right, on it

aviadhahami: @lili2311 in `registry` I can only get the file (flag usage and file itself), but not `targetFileRelativeFile`, do we still want to proceed w/ `registry`?

aviadhahami:
ok, IMO this can be dropped in favor of https://github.com/snyk/registry/pull/11380

@lili2311 wdyt?

lili2311:
@aviadatsnyk yep agreed it would be better there

PR #978, opened 2020-01-27
feat: enable paket to be autodetected with --all-projects

orsagie:
- [x] Ready for review

#### What does this PR do?
Enables paket projets to be autodetected with the --all-projects option set.

### DIFF
--- a/src/lib/detect.ts
+++ b/src/lib/detect.ts
@@ -39,6 +39,7 @@ export const AUTO_DETECTABLE_FILES: string[] = [
   'Gemfile.lock',
   'pom.xml',
   'packages.config',
+  'paket.dependencies',
   'project.json',
   'project.assets.json',
   'Podfile',
--- a/test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
+++ b/test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
@@ -26,6 +26,9 @@ export const AllProjectsTests: AcceptanceTests = {
       t.ok(spyPlugin.withArgs('rubygems').calledOnce, 'calls rubygems plugin');
       t.ok(spyPlugin.withArgs('npm').calledOnce, 'calls npm plugin');
       t.ok(spyPlugin.withArgs('maven').calledOnce, 'calls maven plugin');
+      t.ok(spyPlugin.withArgs('nuget').calledOnce, 'calls nuget plugin');
+      t.ok(spyPlugin.withArgs('paket').calledOnce, 'calls nuget plugin');
+
       // npm
       t.match(
         result,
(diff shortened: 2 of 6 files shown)


src/lib/detect.ts
```diff
@@ -39,6 +39,7 @@ export const AUTO_DETECTABLE_FILES: string[] = [
   'Gemfile.lock',
   'pom.xml',
   'packages.config',
+  'paket.dependencies',
```
dkontorovskyy: `paket.lock` needed here also?

orsagie: We only detect `paket.dependencies`, we won't do `paket.lock` by itself (the plugin will detect it)

dkontorovskyy: But why then `Gemfile` we detect both? And for npm we detect `package-lock.json` and for yarn `yarn.lock`?

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -26,6 +26,9 @@ export const AllProjectsTests: AcceptanceTests = {
       t.ok(spyPlugin.withArgs('rubygems').calledOnce, 'calls rubygems plugin');
       t.ok(spyPlugin.withArgs('npm').calledOnce, 'calls npm plugin');
       t.ok(spyPlugin.withArgs('maven').calledOnce, 'calls maven plugin');
+      t.ok(spyPlugin.withArgs('nuget').calledOnce, 'calls nuget plugin');
+      t.ok(spyPlugin.withArgs('paket').calledOnce, 'calls nuget plugin');
```
gitphill: calls paket plugin

test/acceptance/cli-test/cli-test.all-projects.spec.ts
```diff
@@ -21,8 +21,10 @@ export const AllProjectsTests: AcceptanceTests = {
       t.ok(spyPlugin.withArgs('rubygems').calledOnce, 'calls rubygems plugin');
       t.ok(spyPlugin.withArgs('npm').calledOnce, 'calls npm plugin');
       t.ok(spyPlugin.withArgs('maven').calledOnce, 'calls maven plugin');
+      t.ok(spyPlugin.withArgs('nuget').calledOnce, 'calls nuget plugin');
+      t.ok(spyPlugin.withArgs('paket').calledOnce, 'calls nuget plugin');
```
gitphill: calls paket plugin

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
orsagie: The paket plugin is the nuget plugin

src/lib/detect.ts
gitphill: Gemfile is good point - plugin seems to throw error if no Gemfile.lock is found, so should Gemfile be in this list? I think npm plugin can handle either package-lock.json or package.json so that's ok

lili2311: Gemfile is not really needed no, the rest is all determined by the plugin really and what it can find itself and what it needs to be detected by us. Me and Daniel spoke about this in a different PR, we can go back and tidy up some of these afterwards.

lili2311: for ruby we find the `Gemfile.lock` later in the plugin so just Gemfile will be fine, see `gatherSpecs. Let's keep that a separate PR

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -38,18 +41,20 @@ export const AllProjectsTests: AcceptanceTests = {
...
       const requests = params.server
-        .popRequests(4)
+        .popRequests(6)
         .filter((req) => req.url.includes('/monitor/'));
-      t.equal(requests.length, 3, 'Correct amount of monitor requests');
+      t.equal(requests.length, 5, 'Correct amount of monitor requests');
 
       requests.forEach((req) => {
         t.match(req.url, '/monitor/', 'puts at correct url');
-        t.notOk(req.body.targetFile, "doesn't send the targetFile");
```
lili2311: maybe it's time we assert the relevant ones send the targetFile? I am not a fan of removing this because now we have something that will send it rather than change the assertion.

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
gitphill: ah ok 👍

src/lib/detect.ts
dkontorovskyy: Yeah, just wanted to clarify. We should do a tidy up PR afterward

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
orsagie: I agree. I will add plugin specific check.

orsagie: added

PR #999, opened 2020-02-10
chore: convert yarn util to ts

lili2311:
- [x] Ready for review

#### What does this PR do?

- Convert yarn util to ts


src/lib/yarn.ts
```diff
@@ -1,9 +1,18 @@
-module.exports = yarn;
+export = yarn;
```
dkontorovskyy: can we avoid default exports?

lili2311: Updated!

PR #1007, opened 2020-02-17
fix: drop tape

lili2311:
- [ ] Ready for review

#### What does this PR do?

Drop `tape` and use `tap` instead


test/deep-deps.test.ts
```diff
@@ -17,6 +17,6 @@ test('finds all sub-dependencies', async (t) => {
   } catch (e) {
     t.fail(e.message);
     console.log(e.stack);
-    t.bailout();
+    t.end();
```
gitphill: do you need end (async function) ? could also remove console.log or put whole error object in fail if it's useful

lili2311: you are right no need to end, will update

test/no-deps.test.ts
```diff
@@ -0,0 +1,16 @@
...
+const osDir = path.resolve(__dirname, 'fixtures', 'no-deps');
+
+test('works when there are no dependencies', async (t) => {
+  try {
+    const modules = await snyk.modules(osDir);
+    t.ok(true, 'modules did not bail');
+    t.deepEqual(modules.dependencies, {});
+  } catch (e) {
+    console.log(e.stack);
+    t.fail(e.message);
```
gitphill: fail with e object and remove log?

test/policy-populates-prompt.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    t.notEqual(rule.reason, undefined, 'reason should not be undefined');
+    t.equal(
+      prompt.default,
+      rule.reason,
+      'reasons should match in default value',
+    );
+  } catch (e) {
+    console.log(e.stack);
+    t.fail('could not load the policy file');
```
gitphill: fail with e object and remove log?

test/scoped-module.test.js
```diff
@@ -1,4 +1,4 @@
-var test = require('tape');
+var test = require('tap').test;
```
lili2311: did not convert these as the diff was getting huge, will come back to them

test/protect-patch.test.js
```diff
@@ -2,7 +2,7 @@
 var debug = require('debug')('snyk');
 var protect = require('../src/lib/protect');
 var path = require('path');
-var test = require('tape');
+var test = require('tap').test;
```
aviadhahami: `var` ? :owl:

PR #1033, opened 2020-02-28
fix: drop isolate, modules & bus commands

lili2311:
- [x] Ready for review

#### What does this PR do?
- `snyk modules` has been used only 4 times in two weeks and is an undocumented legacy
command to just show deps in `node_modules`, we have a better way to see this now
with `--print-deps` for any project type

- `isolate` command is orphaned, untested and not called by anything else


src/cli/args.ts
```diff
@@ -153,15 +153,9 @@ export function args(rawArgv: string[]): Args {
...
-      'protect',
-      'test',
-      'modules',
-      'monitor',
-      'wizard',
-      'ignore',
-      'woof',
-    ].indexOf(command) !== -1
+    ['protect', 'test', 'monitor', 'wizard', 'ignore', 'woof'].indexOf(
+      command,
```
aviadhahami: maybe [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)?

dkontorovskyy: We still support Node 6. Array.includes is poorly supported there, if not mistaken

lili2311: Let's raise a separate PR to stop testing on 6

lili2311:
https://github.com/snyk/snyk/issues/788

PR #1044, opened 2020-03-11
Feat: support UTM in login URL from environment variables

lirantal:
- [x] Ready for review

#### What does this PR do?
Adds support for reading UTM configuration from environment variables, prefixed with SNYK_UTM_* and supports all 3 UTM options: Source, Medium, Campaign.

When UTM data is available, the authentication (a `snyk auth`) flow will append this data to the query params of the URL that is opened in the browser for authenticating.

Reviewing the auth flow that now gets the UTMs appended to the URL.

To test manually in a cloned repository:

```bash
SNYK_UTM_MEDIUM=test_medium SNYK_UTM_SOURCE=test_source SNYK_UTM_CAMPAIGN=test_campaign node --require ts-node/register src/cli auth
```

and ensure the URL is being opened with the UTMs data as well as shown in the console's stdout from the auth flow.

This allows to better track and monitor sign-ups and registrations with the Snyk CLI.

None

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -34,7 +34,11 @@ async function webAuth(via: AuthCliCommands) {
   };
 
   let urlStr = authUrl + '/login?token=' + token;
-  urlStr += '&' + getUtmsAsString();
+
+  const utmParams = getUtmsAsString();
+  if (utmParams) {
+    urlStr += '&' + utmParams;
+  }
 
   // validate that via comes from our code, and not from user & CLI
   if (redirects[via]) {
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -12,6 +12,7 @@ import { MisconfiguredAuthInCI } from '../../../lib/errors/misconfigured-auth-in
 import { AuthFailedError } from '../../../lib/errors/authentication-failed-error';
 import { verifyAPI } from './is-authed';
 import { CustomError } from '../../../lib/errors';
+import { getUtmsAsString } from '../../../lib/utm';
 
 export = auth;
 
@@ -33,6 +34,7 @@ async function webAuth(via: AuthCliCommands) {
   };
 
   let urlStr = authUrl + '/login?token=' + token;
+  urlStr += '&' + getUtmsAsString();
 
   // validate that via comes from our code, and not from user & CLI
   if (redirects[via]) {
--- a/test/system/cli.test.ts
+++ b/test/system/cli.test.ts
@@ -135,7 +135,7 @@ test('auth with UTMs in environment variables', async (t) => {
 
   // read data from console.log
   let stdoutMessages = '';
-  const stubConsoleLog = (msg) => (stdoutMessages += msg);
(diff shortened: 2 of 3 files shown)


test/system/cli.test.ts
```diff
@@ -116,13 +116,62 @@ test('auth with no args', async (t) => {
...
+    opn: open,
+  });
+  // stub CI check (ensure returns false for system test)
+  const ciStub = sinon.stub(ciChecker, 'isCI').returns(false);
+
+  // read data from console.log
+  let stdoutMessages = '';
+  const stubConsoleLog = (msg: string) => (stdoutMessages += msg);
+  const origConsoleLog = console.log;
+  console.log = stubConsoleLog;
```
lili2311: please drop

src/cli/commands/auth/index.ts
```diff
@@ -33,6 +34,7 @@ async function webAuth(via: AuthCliCommands) {
   };
 
   let urlStr = authUrl + '/login?token=' + token;
+  urlStr += '&' + getUtmsAsString();
```
lili2311: this assumes it will always be non-empty and sent?

test/system/cli.test.ts
lirantal: do we have a utility function that already does this on the cli? if so, I'm happy to use that or move this to a utils module

src/cli/commands/auth/index.ts
lirantal: it can be empty, in which case a login link will look like:
```
https://snyk.io/login?token=273cecc4-c055-4d38-a4c4-46c3d5354f3b&utm_medium=&utm_source=&utm_campaign=
```

perhaps I should update it to just return an empty string then

lirantal: @lili2311 updated in https://github.com/snyk/snyk/pull/1044/commits/db213ad5954df1bda018156992e42169570accdf if that makes more sense

test/system/cli.test.ts
dkontorovskyy: Why not stubbing `opn` package and verify the url sent to it?

src/cli/commands/auth/index.ts
lirantal: I updated the test but wanted to add that now that I think about it and read the code for the tests, the rest of them don’t have the env vars set and they would run the same sanity checks so actually I’m not convinced adding another test would be beneficial

test/system/cli.test.ts
lirantal: We already do that in the test, here: https://github.com/snyk/snyk/pull/1044/files#diff-7bfe4677ecb8c0c16631229833040ff4R153-R158

dkontorovskyy: Just silence logs like here https://github.com/snyk/snyk/blob/master/test/system/cli.test.ts#L104.
You're checking the URL in line 154. No need to double-check the same thing.

lirantal: @dkontorovskyy it's the same text and the same intention but not the same thing. If we have a regression in which we miss out printing something to the console we won't find it unless we have this test that ensures the data is also printed to the console.

PR #1049, opened 2020-03-20
fix: making sure writing to disk has finished before reading

firecode9:
- [X ] Ready for review

#### What does this PR do?
The wizard is trying to read from the disk before we are making sure that writing has finished. This will cause the process to fail later on

### DIFF
--- a/src/cli/commands/protect/wizard.ts
+++ b/src/cli/commands/protect/wizard.ts
@@ -68,16 +68,16 @@ async function processPackageManager(options: Options) {
     );
   }
 
-  return fs.exists(path.join('.', 'node_modules')).then((nodeModulesExist) => {
-    if (!nodeModulesExist) {
-      // throw a custom error
-      throw new Error(
-        "Missing node_modules folder: we can't patch without having installed packages." +
-          `\nPlease run '${packageManager} install' first.`,
-      );
-    }
-    return options;
-  });
+  const nodeModulesExist = await fs.exists(path.join('.', 'node_modules'));
+  if (!nodeModulesExist) {
+    // throw a custom error
+    throw new Error(
+      "Missing node_modules folder: we can't patch without having installed packages." +
+        `\nPlease run '${packageManager} install' first.`,
+    );
+  }
+
+  return Promise.resolve(options);
 }
 
 async function loadOrCreatePolicyFile(options: Options & WizardOptions) {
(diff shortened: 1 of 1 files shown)


src/cli/commands/protect/wizard.ts
```diff
@@ -524,7 +524,7 @@ function processAnswers(answers, policy, options) {
           options.packageTrailing;
         return (
           spinner(lbl)
-            .then(fs.writeFile(packageFile, packageString))
+            .then(fs.writeFileSync(packageFile, packageString))
```
lili2311: the `.then` should be waiting on the `fs.writeFile` to complete. What is the bug? Where is a test to show the problem please?

lili2311: please check the tests they are failing

firecode9:
Related support ticket:
https://snyk.zendesk.com/agent/tickets/4288

src/cli/commands/protect/wizard.ts
firecode9: There is a race condition here:
image

The write to disk promise is done however the content itself is still not written to disk. Later on in: `/src/lib/plugins/nodejs-plugin/npm-lock-parser.ts`

image

We are trying to read before we have the file on the disk

src/cli/commands/protect/wizard.ts
```diff
@@ -68,16 +69,18 @@ async function processPackageManager(options: Options) {
...
-    if (!nodeModulesExist) {
-      // throw a custom error
-      throw new Error(
-        "Missing node_modules folder: we can't patch without having installed packages." +
-          `\nPlease run '${packageManager} install' first.`,
-      );
-    }
-    return options;
-  });
+  return fsPromise
```
dkontorovskyy: Can you make `async/await` this function?

src/cli/commands/protect/wizard.ts
```diff
@@ -524,7 +527,10 @@ function processAnswers(answers, policy, options) {
           options.packageTrailing;
         return (
           spinner(lbl)
-            .then(fs.writeFile(packageFile, packageString))
+            .then(() => {
```
dkontorovskyy: Can you explain how this fixes the issue? Missing a bit.

firecode9: See here:
https://github.com/snyk/snyk/pull/1049#discussion_r395754744

PR #1068, opened 2020-04-16
Feat/add integration name to analytics

maxjeffos:
- [x] Ready for review

#### What does this PR do?
Looks for two env vars - SNYK_INTEGRATION_NAME and SNYK_INTEGRATION_VERSION - and if they are set adds them to the metadata section of the analytics. These are for identifying which (if any) Snyk tool/plugin/extension/pipe/etc might be making the CLI calls. For example, the Snyk Azure Pipelines task, the Bitbucket Pipe, the CircleCI Orb, IDE plugins, etc.

We would like to have production monitoring to have better insights into when things are working or not and how much the various CI/IDE/etc tools are used.

### DIFF
--- a/src/lib/analytics.js
+++ b/src/lib/analytics.js
@@ -22,6 +22,9 @@ function analytics(data) {
     data = {};
   }
 
+  analytics.add('integrationName', process.env.SNYK_INTEGRATION_NAME || '');
+  analytics.add('integrationVersion', process.env.SNYK_INTEGRATION_VERSION || '');
+
   // merge any new data with data we picked up along the way
   if (Array.isArray(data.args)) {
     // this is an overhang from the cli/args.js and we don't want it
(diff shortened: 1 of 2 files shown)


src/lib/analytics.js
```diff
@@ -22,6 +22,9 @@ function analytics(data) {
     data = {};
   }
 
+  analytics.add('integrationName', process.env.INTEGRATION_NAME || '');
```
garethr, line 25: Prefixing these values with `SNYK_` would match the other Snyk specific environment variables we have.

gitphill: agree - could you squash the commits down and focus on 'why' instead of 'what' as they come out in our release notes https://github.com/snyk/snyk/releases, thanks

dkontorovskyy:
Could also make commit messages a bit shorter? We have a specific check for that. 
https://github.com/snyk/snyk/pull/1068#issuecomment-614871690

This one is a bit too long `feat: map INTEGRATION_NAME and INTEGRATION_VERSION env vars to analytics` 🙏

src/lib/analytics.js
maxjeffos: @gitphill I squashed them down to one per your request. Let me know if it's good now.
Also, I added the `SNYK_` prefix.

PR #1071, opened 2020-04-20
feat: support maven reachable vulnerabilities for `snyk test`

darmalovan:
- [x] Ready for review

#### What does this PR do?
Will send a call graph to registry if the --reachable-vulns flag
is being passed.

The function call graph is generated by the `mvn-plugin` (https://github.com/snyk/snyk-mvn-plugin/pull/56)

This is only supported behind the reachableVulns feature flag for
the maven package manager.

Another small commit was pushed to iterate on the output text.

https://snyksec.atlassian.net/browse/FLOW-159

#### Screenshots
image


src/lib/reachable-vulns.ts
```diff
@@ -0,0 +1,59 @@
...
+      reachableVulnsSupportedRes.code,
+    );
+  }
+  if (reachableVulnsSupportedRes.userMessage) {
+    throw new UnsupportedFeatureFlagError(
+      featureFlag,
+      reachableVulnsSupportedRes.userMessage,
+    );
+  }
+  return true;
```
dkontorovskyy, line 58: `reachableVulnsSupportedRes` cannot return `false`? Does it make sense maybe to return result from the body of request rather than hardcoded value?

darmalovan: Do you think it will make sense to have another if clause for `reachableVulnsSupportedRes.ok === false`?

PR #1075, opened 2020-04-22
chore: preps for cloud config support

orkamara:
- [x] Ready for review

#### What does this PR do?
NO logic changes!

2 main parts:
* Break big functions (displayResult and runTest)into smaller ones - to make the next PR much more easy
* Add SupportedProjectTypes for Cloud Config support

### DIFF
--- a/src/cli/commands/test/formatters/remediation-based-format-issues.ts
+++ b/src/cli/commands/test/formatters/remediation-based-format-issues.ts
@@ -418,7 +418,7 @@ function printPath(path: string[]) {
   return path.slice(1).join(' > ');
 }
 
-function formatIssue(
+export function formatIssue(
   id: string,
   title: string,
   severity: SEVERITY,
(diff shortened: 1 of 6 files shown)


src/lib/package-managers.ts
```diff
@@ -14,6 +18,8 @@ export type SupportedPackageManagers =
   | 'composer'
   | 'cocoapods';
 
+export type SupportedCloudConfigs = 'k8sconfig' | 'helmconfig';
```
orsagie, line 21: maybe for now leave this as undefined, as there is no support yet :)

orsagie:
I would like to see the move to incorporating `mode` instead of `projectType`. Just like we have `package-managers.ts` I would like to see a `cloud-project-type.ts` (or whatever the proper name is) and later one for docker, etc. These will be switched on based on the mode the CLI is run with. Might not be for this PR, but, pointing it out

PR #1079, opened 2020-04-22
chore: upgrade restify to V8

aviadhahami:
#### What does this PR do?

Upgrade Restify to v8 due to ReDos in V4

#### Where should the reviewer know?
Restify V8 doesn't accept a domain and protocol but a patch directly 
i.e `http(s)://localhost:123/a/b/c/` will throw error, while `/a/b/c/` will work as expected;

* I've split the protocol prefix from the path string to match the new API
* Changed the following things when instaciating Restify:
  * `mapParams` prop as the new Restify doesn't auto map (https://github.com/restify/node-restify/issues/61)
  * Added `handleUncaughtExceptions: true` as Restify won't throw w/o it and we will encounter TO.
However, this will log the following error message:
  > DEPRECATION WARNING: Due to deprecation of the domain module in node.js, all features in restify that depend on it have been deprecated as well. This includes `handleUncaughtExceptions` and `next.ifError()`. They will continue to work in 5.x, but consider them unsupported and likely to be removed from future versions of restify."

  * Unlike what this message says, apparently this feature is still active https://github.com/restify/node-restify/issues/1653 and for some odd reasons w/o it the tests fail on MS machines? :thinking: 
* 1 file got slipped in by lint (`src/lib/analytics.js`)
* Fixed 1 small bug where no return statement will make the server send HTTP headers twice and fail


test/acceptance/cli-fail-on-docker.test.ts
```diff
@@ -9,13 +9,14 @@ const { test, only } = tap;
...
-process.env.SNYK_API = 'http://localhost:' + port + '/api/v1';
+const BASE_API = '/api/v1';
+process.env.SNYK_API = 'http://localhost:' + port + BASE_API;
 process.env.SNYK_HOST = 'http://localhost:' + port;
 process.env.LOG_LEVEL = '0';
 const apiKey = '123456789';
 let oldkey;
 let oldendpoint;
-const server = fakeServer(process.env.SNYK_API, apiKey);
+const server = fakeServer(BASE_API, apiKey);
```
anthogez, line 19: Why before we were passing `process.env.SNYK_API` and now only `BASE_API`?

test/acceptance/cli-fail-on-pinnable.test.ts
```diff
@@ -9,13 +9,14 @@ const { test, only } = tap;
...
-process.env.SNYK_API = 'http://localhost:' + port + '/api/v1';
+const BASE_API = '/api/v1';
+process.env.SNYK_API = 'http://localhost:' + port + BASE_API;
 process.env.SNYK_HOST = 'http://localhost:' + port;
 process.env.LOG_LEVEL = '0';
 const apiKey = '123456789';
 let oldkey;
 let oldendpoint;
-const server = fakeServer(process.env.SNYK_API, apiKey);
+const server = fakeServer(BASE_API, apiKey);
```
anthogez, line 19: same as I asked on the top

test/acceptance/cli-fail-on-docker.test.ts
aviadhahami: `process.env.SNYK_HOST` includes the protocol and domain  `http` and `localhost`;
This type of "path" was supported in Restify V4 but not in V8;
Restify v8 forces a path specifier (start w/ `/` or `*`)  and will throw for a full path (which makes sense cause the server shouldn't care where it is located)

anthogez: I see, perfect! Thanks :)

anthogez: So why do we still need process.env.SNYK_HOST? since we are not calling it anywhere inside of `cli-fail-on-docker.test.ts`

aviadhahami: further down-the-line functions consume the `SNYK_API` via `process.env.SNYK_API`;
that is why it is being set before every test suite and removed after it;
Removing the full path from the env var will fail the tests (and believe me, I've spent a lot of time there)

src/lib/analytics.js
```diff
@@ -23,7 +23,10 @@ function analytics(data) {
   }
 
   analytics.add('integrationName', process.env.SNYK_INTEGRATION_NAME || '');
-  analytics.add('integrationVersion', process.env.SNYK_INTEGRATION_VERSION || '');
+  analytics.add(
```
dkontorovskyy, line 26: rebase will remove this file

test/acceptance/fake-server.ts
```diff
@@ -14,6 +14,7 @@ export function fakeServer(root, apikey) {
   const server = restify.createServer({
     name: 'snyk-mock-server',
     version: '1.0.0',
+    handleUncaughtExceptions: true,
```
dkontorovskyy, line 17: wouldn't want that. Let if die hard when exceptions are uncaught

test/cli-server.js
```diff
@@ -5,14 +5,16 @@ module.exports = function(root, apikey, notAuthorizedApiKey) {
   var server = restify.createServer({
     name: 'snyk-mock-server',
     version: '1.0.0',
+    handleUncaughtExceptions: true,
```
dkontorovskyy, line 8: Same here though

test/acceptance/fake-server.ts
aviadhahami: issue is it wouldn't die at all on MS machines w/o this flag :thinking:

aviadhahami: @dkontorovskyy 
pull `master`, remove these and test w/ MS machines. Do you have the same?

test/acceptance/cli-fail-on-docker.test.ts
anthogez: This is being followed by https://github.com/snyk/snyk/pull/1088

PR #1084, opened 2020-04-24
fix: bump ruby-semver to use min Node 8 instead of 10

lili2311:
- [ ] Ready for review

#### What does this PR do?
All libraries that are used in CLI must remain using min Node.js version >=8 to maintain support. Reverting the ruby-semver change that upgraded this to 10.

snyk/snyk#1060


darscan:
@gitphill Why was `@snyk/ruby-semver` added here as a direct dep, instead of bumping it where it's actually used: https://github.com/snyk/cocoapods-lockfile-parser/blob/master/package.json#L35 ?

PR #1090, opened 2020-04-28
feat: app+os cli changes. bump docker-plugin. adding target to container/docker projects

firecode9:
- [X] Ready for review

#### What does this PR do?
In order to support app+os projects, we need a way to set the target on the project so they can be properly grouped in the UI

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "proxy-from-env": "^1.0.0",
     "semver": "^6.0.0",
     "snyk-config": "^2.2.1",
-    "snyk-docker-plugin": "2.10.0",
+    "snyk-docker-plugin": "3.0.0",
     "snyk-go-plugin": "1.13.0",
     "snyk-gradle-plugin": "3.2.5",
     "snyk-module": "1.9.1",
--- a/src/lib/project-metadata/target-builders/container.ts
+++ b/src/lib/project-metadata/target-builders/container.ts
@@ -1,0 +1,15 @@
+import { DepTree } from '../../types';
+import { ContainerTarget } from '../types';
+
+export async function getInfo(
+  packageInfo: DepTree,
+): Promise<ContainerTarget | null> {
+  // safety check
+  if (!packageInfo.docker) {
+    return null;
+  }
+
+  return {
+    image: (packageInfo as any).image || packageInfo.name,
+  };
+}
--- a/src/lib/project-metadata/target-builders/git.ts
+++ b/src/lib/project-metadata/target-builders/git.ts
@@ -5,6 +5,7 @@ import { DepTree } from '../../types';
 import { GitTarget } from '../types';
 
 export async function getInfo(packageInfo: DepTree): Promise<GitTarget | null> {
+  // safety check
   if (packageInfo.docker) {
     return null;
   }
--- a/src/lib/project-metadata/types.ts
+++ b/src/lib/project-metadata/types.ts
@@ -2,3 +2,11 @@ export interface GitTarget {
   remoteUrl?: string;
   branch?: string;
 }
+
+export interface ContainerTarget {
+  image?: string;
+}
+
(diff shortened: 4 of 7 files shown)


src/lib/project-metadata/types.ts
```diff
@@ -1,4 +1,10 @@
-export interface GitTarget {
-  remoteUrl?: string;
+export interface Target {
   branch?: string;
 }
+export interface GitTarget extends Target {
```
darscan, line 4: I.. really.. hate inheritance :D

Not sure we need an abstract type here (Target).. also, `branch` does not make sense on `ContainerTarget`

src/lib/project-metadata/target-builders/container.ts
```diff
@@ -0,0 +1,15 @@
...
+  packageInfo: DepTree,
+): Promise<ContainerTarget | null> {
+  // safety check
+  if (!packageInfo.docker) {
+    return null;
+  }
+
+  return {
+    image: (packageInfo as any).image || packageInfo.name,
+  } as ContainerTarget;
```
darscan, line 14: nooooooooo! (no casting, please!)

src/lib/project-metadata/types.ts
firecode9: yes. the problem is here:
image

src/lib/project-metadata/target-builders/container.ts
darscan: (you won't need casting if you get rid of the Target inheritance nonsense ;) )

firecode9: :-) I dont need the `as ContainerTarget` at all. going to remove it

src/lib/project-metadata/types.ts
firecode9: okay, find a way to fix it

firecode9:
@darscan all comments fixed. removed inheritance

src/lib/project-metadata/target-builders/git.ts
```diff
@@ -34,5 +36,9 @@ export async function getInfo(packageInfo: DepTree): Promise<GitTarget | null> {
     // executing git commands.
   }
 
+  if (target && target.branch) {
```
darscan, line 39: Oh lordy, this is much worse 😱 _Screams CQRS_

> every method should either be a command that performs an action, or a query that returns data to the caller, but not both. In other words, Asking a question should not change the answer.

`getInfo` is clearly a query.. it should absolutely not have side effects.

I can see how you ended up here after removing `branch` from the target, but this is not the solution :D

You need a type guard, for example:

```typescript
function isGitTarget(target: any): target is GitTarget {
  return target && target.branch;
}
```
Then just change the original callsites to:
```typescript
if (isGitTarget(target)) {
  // TS now knows that this is a GitTarget, so you can access .branch
  analytics.add('targetBranch', target.branch);	
}
```
Perhaps the typeguard can live in `src/lib/project-metadata/types.ts` in case it is useful elsewhere

firecode9: yeah I was thinking after my commit that getInfo is not a good place to add the analytics call

firecode9: ok fixing

firecode9:
@darscan ready for review (again) :-)

package.json
```diff
@@ -79,7 +79,7 @@
     "proxy-from-env": "^1.0.0",
     "semver": "^6.0.0",
     "snyk-config": "^2.2.1",
-    "snyk-docker-plugin": "2.10.0",
+    "snyk-docker-plugin": "3.0.0",
```
darscan, line 82: I don't see a lockfile update for this?

darscan: my bad.. wrong project!

PR #1093, opened 2020-04-29
feat: bump CLI dependencies to use patched @snyk/lodash

lili2311:
- [x] Ready for review

#### What does this PR do?
Bump CLI dependencies to use non-vulnerable forked version of lodash or remove it entirely.
As part of this fix all transitives that rely on `lodash` have been updated to use non-vulnerable `@snyk/lodash` instead

#### Links

- *Lodash Issue*: https://github.com/lodash/lodash/issues/4744
- *Disclosed vuln*: https://hackerone.com/reports/712065


lili2311:
@anthogez we could go with any but then we lose all types which we have types for that

anthogez:
> @anthogez we could go with any but then we lose all types which we have types for that

I understand your point, I meant as first approach then once we have time we create a side pr wherever we are using `snyk-config` and we apply the properly types

PR #1094, opened 2020-04-30
feat: container manifest file scanning in --experimental flag

firecode9:
- [ X] Ready for review
- [ X] Reviewed by Snyk internal team

#### What does this PR do?
- snyk monitor | container manifest file scanning in --experimental flag
- Bump the snyk-docker-plugin

Notes:
1. All the code changes are under the `--experimental` flag
2. snyk test changes will be done on another PR

This is an example output of an image with a node app inside:

image

### DIFF
--- a/src/lib/container/index.ts
+++ b/src/lib/container/index.ts
@@ -1,0 +1,29 @@
+import { ScannedProject } from '@snyk/cli-interface/legacy/common';
+
+export function isContainer(scannedProject: ScannedProject): boolean {
+  return scannedProject.meta?.imageName !== undefined;
+}
+
+export function getContainerTargetFile(
+  scannedProject: ScannedProject,
+): string | undefined {
+  return scannedProject.targetFile;
+}
+
+export function getContainerName(
+  scannedProject: ScannedProject,
+): string | undefined {
+  if (scannedProject.targetFile) {
+    // for app+os projects the name of project is a mix of the image name
+    // with the target file (if one exists)
+    return scannedProject.meta?.imageName + ':' + scannedProject.targetFile;
+  } else {
+    return scannedProject.meta?.imageName;
+  }
+}
+
+export function getContainerProjectName(
+  scannedProject: ScannedProject,
+): string | undefined {
+  return scannedProject.meta?.imageName;
+}
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -28,6 +28,12 @@ import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
 import { CallGraph, ScannedProject } from '@snyk/cli-interface/legacy/common';
 import { isGitTarget } from '../project-metadata/types';
 import { serializeCallGraphWithMetrics } from '../reachable-vulns';
+import {
+  getNameDepTree,
+  getNameDepGraph,
+  getProjectName,
+  getTargetFile,
+} from './utils';
 
 const debug = Debug('snyk');
 
@@ -133,7 +139,7 @@ export async function monitor(
   }
   const policy = await snyk.policy.load(policyLocations, { loose: true });
 
-  const target = await projectMetadata.getInfo(pkg, meta);
+  const target = await projectMetadata.getInfo(scannedProject, pkg, meta);
 
   if (isGitTarget(target) && target.branch) {
     analytics.add('targetBranch', target.branch);
@@ -167,7 +173,7 @@ export async function monitor(
             pid: process.pid,
             node: process.version,
             master: snyk.config.isMaster,
-            name: pkg.name,
+            name: getNameDepTree(scannedProject, pkg),
             version: pkg.version,
             org: config.org ? decodeURIComponent(config.org) : undefined,
             pluginName: pluginMeta.name,
@@ -178,7 +184,7 @@ export async function monitor(
             dockerfileLayers: pkg.docker
               ? pkg.docker.dockerfileLayers
               : undefined,
-            projectName: meta['project-name'],
+            projectName: getProjectName(scannedProject, meta),
             prePruneDepCount, // undefined unless 'prune' is used,
             monitorGraph: false,
             versionBuildInfo: JSON.stringify(
@@ -192,7 +198,7 @@ export async function monitor(
           // because we want to send it only for specific package-managers
           target,
           // WARNING: be careful changing this as it affects project uniqueness
-          targetFile: pluginMeta.targetFile,
+          targetFile: getTargetFile(scannedProject, pluginMeta),
           targetFileRelativePath,
         } as MonitorBody,
         gzip: true,
@@ -260,7 +266,7 @@ export async function monitorGraph(
   }
   const policy = await snyk.policy.load(policyLocations, { loose: true });
 
-  const target = await projectMetadata.getInfo(pkg, meta);
+  const target = await projectMetadata.getInfo(scannedProject, pkg, meta);
 
   if (isGitTarget(target) && target.branch) {
     analytics.add('targetBranch', target.branch);
@@ -287,7 +293,7 @@ export async function monitorGraph(
             pid: process.pid,
             node: process.version,
             master: snyk.config.isMaster,
-            name: depGraph.rootPkg.name,
+            name: getNameDepGraph(scannedProject, depGraph),
             version: depGraph.rootPkg.version,
             org: config.org ? decodeURIComponent(config.org) : undefined,
             pluginName: pluginMeta.name,
@@ -297,7 +303,7 @@ export async function monitorGraph(
             dockerfileLayers: pkg.docker
               ? pkg.docker.dockerfileLayers
               : undefined,
-            projectName: meta['project-name'],
+            projectName: getProjectName(scannedProject, meta),
             prePruneDepCount, // undefined unless 'prune' is used
             missingDeps: treeMissingDeps,
             monitorGraph: true,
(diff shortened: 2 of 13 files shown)


src/lib/container/index.ts
```diff
@@ -0,0 +1,35 @@
+import { ScannedProject } from '@snyk/cli-interface/legacy/common';
+
+export function isContainer(
+  scannedProject: ScannedProject,
+): string | undefined {
```
shaimendel, line 5: shouldn't this return boolean?

src/lib/container/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  return scannedProject.meta && scannedProject.meta.imageName;
+}
+
+export function getTargetFileIfContainer(
+  scannedProject: ScannedProject,
+): string | undefined {
+  if (isContainer(scannedProject)) {
```
aviadhahami, line 12: maybe 1 line this in order to show returned value?
```js
return isContainer(..) ? scappedProject.targetFile : undefined;

File: src/lib/container/index.ts [hunk 3]
```diff
@@ -0,0 +1,35 @@
+import { ScannedProject } from '@snyk/cli-interface/legacy/common';
+
+export function isContainer(
+  scannedProject: ScannedProject,
+): string | undefined {
+  return scannedProject.meta && scannedProject.meta.imageName;
```
aviadhahami, line 6: can be
```ts
return scannedProject.meta?.imageName

File: src/lib/monitor/index.ts [hunk 4]
```diff
@@ -192,7 +198,8 @@ export async function monitor(
           // because we want to send it only for specific package-managers
           target,
           // WARNING: be careful changing this as it affects project uniqueness
-          targetFile: pluginMeta.targetFile,
+          targetFile:
```
aviadhahami, line 201: I'd suggest creating a general function for each prop `targetFile`, `projectName` and `name`;
This way we won't have long conditions inside a returned object;
Will help future readbility and maintenance :)

src/lib/container/index.ts
firecode9: it should 😄

src/lib/container/index.ts
firecode9: yes. fixed

src/lib/container/index.ts
firecode9: yes. fixed

src/lib/container/index.ts
firecode9: fixed

src/lib/monitor/index.ts
firecode9: Fixed in the 2nd commit

src/lib/monitor/index.ts
```diff
@@ -287,7 +293,7 @@ export async function monitorGraph(
             pid: process.pid,
             node: process.version,
             master: snyk.config.isMaster,
-            name: depGraph.rootPkg.name,
+            name: selectNameDepGraph(scannedProject, depGraph),
```
aviadhahami, line 296: I'd rename these to `getName`, `getProjectName` and `getTargetFile` respectively as they are more of getters than select :)

firecode9: ok. fixed

src/lib/container/index.ts
```diff
@@ -0,0 +1,33 @@
+import { ScannedProject } from '@snyk/cli-interface/legacy/common';
+
+export function isContainer(
+  scannedProject: ScannedProject,
+): boolean | undefined {
```
shaimendel, line 5: this should not return undefined, just boolean :) if the field does not exist then the answer is `false`.
anyway it is weird that we return string as bool, I assume the return statement should be like
`return scannedProject.meta?.imageName !== undefined`

src/lib/container/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  return scannedProject.meta?.imageName;
+}
+
+export function getTargetFileIfContainer(
+  scannedProject: ScannedProject,
+): string | undefined {
```
shaimendel, line 11: I also don't like that it returns undefined.. can we move all of this logic to  src/lib/monitor/utils.ts?
so for example:
```
export function getTargetFile(
  scannedProject: ScannedProject,
  pluginMeta: PluginMetadata,
): string | undefined {
  if (isContainer(scannedProject)) {
    return scannedProject.targetFile;
  }
  return pluginMeta.targetFile;
}
```

src/lib/container/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  return isContainer(scannedProject) ? scannedProject.targetFile : undefined;
+}
+
+export function getNameIfContainer(
+  scannedProject: ScannedProject,
+): string | undefined {
+  if (isContainer(scannedProject)) {
+    if (scannedProject.targetFile) {
+      return scannedProject.meta.imageName + ':' + scannedProject.targetFile;
```
shaimendel, line 20: might worth a comment why we do it

test/container.test.ts
```diff
@@ -0,0 +1,81 @@
+import { test } from 'tap';
```
shaimendel, line 1: that's a great unit-test-like tests, but I think we have to add some kind of system tests to make sure this works. around this area https://github.com/snyk/snyk/blob/master/test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts#L1245

src/lib/monitor/index.ts
```diff
@@ -192,7 +198,7 @@ export async function monitor(
           // because we want to send it only for specific package-managers
           target,
           // WARNING: be careful changing this as it affects project uniqueness
-          targetFile: pluginMeta.targetFile,
+          targetFile: getTargetFile(scannedProject, pluginMeta),
```
gitphill, line 201: Have we checked this does not break projects? Please check all our plugins, projectName and targetFile are sensitive areas of code to change. Check how these properties are processed in backend to gain an understanding of what could break

src/lib/container/index.ts
firecode9: Yes. fixed

src/lib/container/index.ts
firecode9: here I would disagree, because `scannedProject.targetFile` is something that monitor/utils should not know about, but rather hide this logic under `container.ts` file. It also makes unit testing easy as we can just test the file logic

firecode9: as for: `returns undefined`, I dont see another way as target file might undefined

src/lib/container/index.ts
firecode9: ok. ,added

src/lib/monitor/index.ts
firecode9: Hi, 
There is another PR for registry side: https://github.com/snyk/registry/pull/13277

The new code works only if the `scannedProject` has a `target` with `imageName` inside, and that's a new thing, so it all the "old" code would never hit this if case

firecode9: What do you mean check all our plugins? Besides making sure that tests are all passing, what else do you think I should do?

src/lib/container/index.ts
shaimendel: so maybe have a `getContainerTargetFile` so the `getTargetFile` looks like this:
```
export function getTargetFile(
  scannedProject: ScannedProject,
  pluginMeta: PluginMetadata,
): string | undefined {
  if (isContainer(scannedProject)) {
    return getContainerTargetFile(scannedProject);
  }
  return pluginMeta.targetFile;
}
```

src/lib/monitor/index.ts
firecode9: @gitphill I've also added:
https://github.com/snyk/snyk/pull/1094/files#diff-e09af9109bf7e5b6e4fac2827d861fd5
To test the utils and make sure the code works for "old" workflow

src/lib/container/index.ts
firecode9: that's better. fixed

test/container.test.ts
firecode9: ok, fixed

src/lib/container/index.ts
shaimendel: I also meant this to be done in the 3 other functions here :)

src/lib/container/index.ts
```diff
@@ -0,0 +1,33 @@
+import { ScannedProject } from '@snyk/cli-interface/legacy/common';
+
+export function isContainer(scannedProject: ScannedProject): boolean {
+  return scannedProject.meta?.imageName !== undefined;
```
shaimendel, line 4: will this be true also for dynamic scanning?

firecode9: It wont, afaik just the static code would return it

firecode9:
> another thing - don't forget that this is a public repo and external people read the release notes :)
> I would change `app+os | adding support for snyk monitor` to something more non-snyk friendly, like `feat: container manifest file scanning in --experimental flag` and might elaborate more in the commit message

noted.

src/lib/container/index.ts
firecode9: do you think it's bad/good?

src/lib/container/index.ts
firecode9: fixed

src/lib/container/index.ts
shaimendel: I guess it is ok for now, until we remove the old format, as long as the dynamic static work is still working :)

PR #1104, opened 2020-05-05
feat: support `snyk monitor --reachable-vulns`

darmalovan:
- [x] Ready for review

#### What does this PR do?
Support `snyk monitor --reachable-vulns` for maven behind a feature flag.

This change will swill send an additional payload in the form of
call graph to our backend.

https://snyksec.atlassian.net/browse/FLOW-207


src/cli/commands/monitor/index.ts
```diff
@@ -94,6 +95,11 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         packageManager = detect.detectPackageManager(path, options);
       }
 
+      await validateOptions(
+        options as Options & MonitorOptions,
```
anthogez, line 99: I think would be nice if you created a extended type in `src/cli/commands/monitor/types.ts` where you apply the `type intersection` between `Options & MonitorOptions` something like `GlobalOptions` or `ExtendedOptions`

darmalovan: I think `GlobalOptions` might not be so useful from a user perspective, as this is the method used to show that it's only `Options & MonitorOptions` You can also have `Options & TestOptions`
This convention is used across the codebase - I don't think I want to change this as part of this PR.

src/lib/options-validator.ts
```diff
@@ -0,0 +1,17 @@
+import { MonitorOptions, Options, TestOptions } from './types';
+import * as config from './config';
+import { SupportedPackageManagers } from './package-managers';
+import * as reachableVulns from './reachable-vulns';
+
+export async function validateOptions(
+  options: (Options & TestOptions) | (Options & MonitorOptions),
```
anthogez, line 7: Same here would be nice to have a type something like GlobalOptions?
`export type GlobalOptions = Options & MonitorOptions | Options & TestOptions;` ?
What do you think?

src/cli/commands/monitor/index.ts
anthogez: I understood your point, it's ok

Issue #1027, opened 2020-02-26
[🐛] json log can't be parsed when an exception happens during `snyk test`


jessehouwing:
+ snyk-to-html@1.7.2
+ snyk@1.297.1

- OS: (e.g. OSX, Linux, Windows, ...) Windows 
- Command run: (e.g. `snyk test --all-projects`, `snyk protect`, ...)

```
snyk test --all-projects --exclude=project.json --json
```
The project has both .NET Core and NPM files, at this time we're only interested in the .NET Core ones and are explicitly ignoring `project.json`.

### Expected behaviour
json log is actually of type JSON and can be parsed as such

### Actual behaviour

The report includes a stack trace and can't be parsed. See below:

![image](https://user-images.githubusercontent.com/4173387/75362630-5535ba80-58b9-11ea-849c-2a25d4f3914b.png)

[report-2020-02-26T15-12-45.zip](https://github.com/snyk/snyk/files/4256723/report-2020-02-26T15-12-45.zip)

jessehouwing:
I'm guessing the issue is due to the fact that only output is redirected and error isn't. err follows output if not explicitly redirected:

![image](https://user-images.githubusercontent.com/4173387/75364338-de4df100-58bb-11ea-8f0b-8ab619e5989a.png)

jessehouwing:
It also fails when `--debug` is enabled, since all debug data is added to the json as plain text.

jessehouwing:
It's probably useful to just redirect the err stream explicitly to `console.err` or `console.out`.

lili2311:
👋 We should we redirecting errors to err stream, will check why this is not happening in this case

jessehouwing:
Cause it's not specified explicitly. The Task SDK has a "useful" feature to redirect that to stdOut if not specified and stdOut *is* redirected.

maxjeffos:
Thanks @jessehouwing . The errors and debug info from the Snyk CLI are indeed being redirected  to stderr so this is a thing we will tackle in the Azure extension (as opposed to here in the Snyk CLI). I'll follow up with you there with a PR and then close the loop here once that's done.

lili2311:
Is this still an issue?

jessehouwing:
Yeah. The --json-file-output will fix this for sure.

maxjeffos:
I merged https://github.com/snyk/snyk/pull/1107 which adds the `--json-file-output=<file>` option. 
[...]

PR #1107, opened 2020-05-07
feat: add --json-file-output option for snyk test

maxjeffos:
- [x] Ready for review

#### What does this PR do?
This allows a user to directly save JSON format output to a file as specified by the --json-file-output=<file> parameter, regardless of whether they are using the human-readable or json output. The primary use-case of this is to enable users (and Snyk CI plugins) to simultaneously get the human readable output while saving a JSON format output to a file for other purposes (for example, generating a report).

Please take a look at the code and advise as to testing best-practices. I will then include appropriate tests for this PR.

Say you want to save the JSON output to a file called "snyk-output.json" in the current directory. You could then run snyk test --json-file-output=snyk-output.json which should do a normal snyk test and generate the human-readable output to stdout but at the same time save the JSON format output to snyk-output.json.

We need this for #1027 and snyk/snyk-azure-pipelines-task#30 as well as for another CI plugin we're currently working on.
This PR replaces a https://github.com/snyk/snyk/pull/1059 which I will close once this gets traction

- https://github.com/snyk/snyk/issues/1027
- https://github.com/snyk/snyk-azure-pipelines-task/issues/30

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -81,6 +81,10 @@ Options:
                        Upgradable fails when there is at least one vulnerability that can be upgraded.
                        Patchable fails when there is at least one vulnerability that can be patched.
                        If vulnerabilities do not have a fix and this option is being used tests will pass.
+  --json-file-output=<string>
+                       (test command only)
+                       Save test output in JSON format directly to the specified file, regardless of whether or not you use the `--json` option.
+                       This is especially useful if you want to display the human-readable test output via stdout and at the same time save the JSON format output to a file.
 
 Maven options:
   --scan-all-unmanaged
--- a/src/cli/commands/types.ts
+++ b/src/cli/commands/types.ts
@@ -1,0 +1,57 @@
+export class CommandResult {
+  result: string;
+  constructor(result: string) {
+    this.result = result;
+  }
+
+  public toString(): string {
+    return this.result;
+  }
+
+  public getDisplayResults() {
+    return this.result;
+  }
+}
+
+export abstract class TestCommandResult extends CommandResult {
+  protected jsonResult = '';
+  public getJsonResult(): string {
+    return this.jsonResult;
+  }
+
+  public static createHumanReadableTestCommandResult(
+    humanReadableResult: string,
+    jsonResult: string,
+  ): HumanReadableTestCommandResult {
+    return new HumanReadableTestCommandResult(humanReadableResult, jsonResult);
+  }
+
+  public static createJsonTestCommandResult(
+    jsonResult: string,
+  ): JsonTestCommandResult {
+    return new JsonTestCommandResult(jsonResult);
+  }
+}
+
+class HumanReadableTestCommandResult extends TestCommandResult {
+  protected jsonResult = '';
+
+  constructor(humanReadableResult: string, jsonResult: string) {
+    super(humanReadableResult);
+    this.jsonResult = jsonResult;
+  }
+
+  public getJsonResult(): string {
+    return this.jsonResult;
+  }
+}
+
+class JsonTestCommandResult extends TestCommandResult {
+  constructor(jsonResult: string) {
+    super(jsonResult);
+  }
+
+  public getJsonResult(): string {
+    return this.result;
+  }
+}
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -10,6 +10,7 @@ import * as analytics from '../lib/analytics';
 import * as alerts from '../lib/alerts';
 import * as sln from '../lib/sln';
 import { args as argsLib, Args } from './args';
+import { CommandResult, TestCommandResult } from './commands/types';
 import { copy } from './copy';
 import spinner = require('../lib/spinner');
 import errors = require('../lib/errors/legacy-errors');
@@ -26,6 +27,11 @@ import {
 import stripAnsi from 'strip-ansi';
 import { ExcludeFlagInvalidInputError } from '../lib/errors/exclude-flag-invalid-input';
 import { modeValidation } from './modes';
+import { JsonFileOutputBadInputError } from '../lib/errors/json-file-output-bad-input-error';
+import {
+  createDirectory,
+  writeContentsToFileSwallowingErrors,
+} from '../lib/json-file-output';
 
 const debug = Debug('snyk');
 const EXIT_CODES = {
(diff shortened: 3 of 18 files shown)


src/cli/commands/types.ts
```diff
@@ -0,0 +1,57 @@
...
+  public toString(): string {
+    return this.result;
+  }
+
+  public getDisplayResults() {
+    return this.result;
+  }
+}
+
+export abstract class TestCommandResult extends CommandResult {
```
dkontorovskyy, line 16: Should this be abstract? I don't see any abstract methods in this class

maxjeffos: @dkontorovskyy TestCommandResult is abstract to prevent its instantiation; one should use one of its static create methods to create instances of the appropriate subclass.

aarlaud:
@maxjeffos can you also make sure the output of --print-deps is in that file too if the option is specified. I'm using the depgraph for something and that'd be great to have in this function. Thanks!

src/cli/index.ts
```diff
@@ -120,6 +148,17 @@ async function handleError(args, error) {
...
 }
 
+function saveJsonResultsToFile(
+  stringifiedJson: string,
+  jsonOutputFile: string,
+) {
+  try {
+    writeFileSync(jsonOutputFile, stringifiedJson);
+  } catch (err) {
+    console.error(err);
```
gitphill, line 158: potentially 'ugly' for user

for example: `snyk test --json-file-output` (no path provided) produces:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received type boolean (true)
    at Object.openSync (fs.js:450:10)
    at Object.writeFileSync (fs.js:1279:35)
    at saveJsonResultsToFile (/Users/phill/code/snyk/src/cli/index.ts:156:5)
    at handleError (/Users/phill/code/snyk/src/cli/index.ts:111:5)
    at main (/Users/phill/code/snyk/src/cli/index.ts:273:28)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

gitphill: maybe some validation of argument passed?

package.json
```diff
@@ -72,6 +72,7 @@
     "diff": "^4.0.1",
     "git-url-parse": "11.1.2",
     "glob": "^7.1.3",
+    "mkdirp": "^1.0.4",
```
gitphill, line 75: lets use Node.js core lib - CLI already has enough dependencies :)

src/cli/index.ts
```diff
@@ -120,6 +153,44 @@ async function handleError(args, error) {
...
+  stringifiedJson: string,
+  jsonOutputFile: string,
+) {
+  if (!jsonOutputFile) {
+    console.error("empty jsonOutputFile");
+    return;
+  }
+  
+  if (jsonOutputFile.constructor.name !== String.name) {
+    console.error("jsonOutputFile is invalid type");
```
gitphill, line 175: "--json-output-file should be a filename path"

src/cli/index.ts
maxjeffos: This is resolved now and I added some tests for it. Thanks.

maxjeffos:
> @maxjeffos can you also make sure the output of --print-deps is in that file too if the option is specified. I'm using the depgraph for something and that'd be great to have in this function. Thanks!

@aarlaud I created a ticket for this as we will need to think more on this use-case and the implementation is outside the usual flow.

maxjeffos:
> can we update help text please?

I added help text

src/cli/index.ts
maxjeffos: 👍

help/help.txt
```diff
@@ -81,6 +81,10 @@ Options:
                        Upgradable fails when there is at least one vulnerability that can be upgraded.
                        Patchable fails when there is at least one vulnerability that can be patched.
                        If vulnerabilities do not have a fix and this option is being used tests will pass.
+  --json-file-output=<string>
+                       (test command only)
+                       Save test output in JSON format directly to the specified file, regardless of whether or not you use the `--json` option.
+                       This is especially useful if you want to display the human-readable test output via stdout and at the same time save the JSON format output to a file.
```
gitphill, line 87: 👍

test/cli-create-directory.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  // if node 8 verify that the deep folder was not created
+  const nodeVersion = process.version;
+  if (gte(nodeVersion, MIN_VERSION_FOR_MKDIR_RECURSIVE)) {
+    t.ok(success);
+    t.ok(directoryExists);
+  } else {
+    t.notOk(success);
+    t.notOk(directoryExists);
+  }
+});
```
gitphill, line 71: ❤️

src/lib/cli-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+      // nodeVersion is < 10.12.0
+      mkdirSync(newDirectoryFullPath);
+      return true;
+    }
+  } catch (err) {
+    console.error(err);
+    console.error(`could not create directory ${newDirectoryFullPath}`);
+    return false;
+  }
+}
```
gitphill, line 39: 👍 nice, think we should call it `src/lib/create-directory.ts` though so we don't encourage large 'util' files

src/lib/cli-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+      const options: any = { recursive: true }; // TODO: remove this after we drop support for node v8
+      mkdirSync(newDirectoryFullPath, options);
+      return true;
+    } else {
+      // nodeVersion is < 10.12.0
+      mkdirSync(newDirectoryFullPath);
+      return true;
+    }
+  } catch (err) {
+    console.error(err);
```
gitphill, line 35: do we need both of these error logs?

package.json
maxjeffos: done

src/lib/cli-utils.ts
maxjeffos: I renamed it to `/src/lib/json-file-output.ts` because I ended up adding another function to it (again, relating to the --json-file-output option). LMK if that works for you.

src/lib/cli-utils.ts
maxjeffos: I think the error itself would be useful to log so that a user knows why we can't write the file...

PR #1113, opened 2020-05-11
fix: updated test command so Circle will test properly

aviadhahami:
- [X] Ready for review

#### What does this PR do?
Updated the CircleCi step where it invoked `test` incorrectly;

[Slack thread](https://snyk.slack.com/archives/CFTAUCYQ7/p1589208414100000)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,7 @@ jobs:
           command: npm run snyk-auth
       - run:
           name: Run tests
-          command: npm run test:test
+          command: npm test
 
   release:
     <<: *defaults


.circleci/config.yml
```diff
@@ -143,7 +143,7 @@ jobs:
           command: npm run snyk-auth
       - run:
           name: Run tests
-          command: npm run test:test
+          command: npm test
```
anthogez, line 146: I think it should be `npm run test` 🤔

aviadhahami: it's the same, `npm t` === `npm test` === `npm run test`
just like `npm i`

aviadhahami: https://docs.npmjs.com/cli/test

anthogez: nvm what I said.

PR #1116, opened 2020-05-12
feat: add IPv6 detection to fix auth bug

dkontorovskyy:
- [X] Ready for review

#### What does this PR do?
Add support to clients who are using IPv6

Blocked until new release of `needle`
https://github.com/tomas/needle/pull/316#issuecomment-626929494

**NOTA BENE**: need to update needle version in this PR after it's release


src/cli/commands/auth/index.ts
```diff
@@ -61,21 +67,25 @@ async function webAuth(via: AuthCliCommands) {
...
   }
 }
 
-async function testAuthComplete(token: string): Promise<{ res; body }> {
-  const payload = {
+async function testAuthComplete(
+  token: string,
+  ipFamily?: number,
+): Promise<{ res; body }> {
+  const payload: any = {
```
anthogez, line 88: can we have some type here?

src/cli/commands/auth/index.ts
```diff
@@ -162,3 +177,18 @@ function errorForFailedAuthAttempt(res, body) {
     return error;
   }
 }
+
+async function getIpFamily(): Promise<6 | undefined> {
```
anthogez, line 181: could you apply a type instead of just 6 on `Promise<6 | undefined>`? Thanks!

src/cli/commands/auth/index.ts
```diff
@@ -162,3 +177,18 @@ function errorForFailedAuthAttempt(res, body) {
...
+  const IPv6Family = 6;
+
+  try {
+    const { address } = await dnsLookupPromise(apiUrl.hostname!, {
+      family: IPv6Family,
+    });
+    const res = parse(address) as IPv6;
```
anthogez, line 188: I would suggest to change this `const res = parse(address) as IPv6;` to `const res:  IPv6 = parse(address)`;

dkontorovskyy: This won't work in this case, cos parse return IPv4 | IPv6. But at this point I know address can be only IPv6

src/cli/commands/auth/index.ts
gitphill: is it 6 | 4? or must it be undefined when not 6?

PR #1118, opened 2020-05-12
Snyk test results contain wrong path if scanning a multi-module project

firecode9:
- [ X] Ready for review
- [ X] Reviewed by Snyk internal team

#### What does this PR do?
When we have a multi project the path we return back from the CLI is correct only for the first (master) project.
It came from an on-call ticket where the customer was trying to snyk-to-html the results from snyk test and he saw:
image

After the fix the page looks like that:
image

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -41,6 +41,7 @@ import {
   summariseReachableVulns,
   summariseVulnerableResults,
 } from './formatters';
+import * as utils from './utils';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
@@ -132,7 +133,13 @@ async function test(...args: MethodArgs): Promise<string> {
     const resArray: any[] = Array.isArray(res) ? res : [res];
 
     for (let i = 0; i < resArray.length; i++) {
-      results.push(_.assign(resArray[i], { path }));
+      const pathWithOptionalProjectName = utils.getPathWithOptionalProjectName(
+        path,
+        resArray[i],
+      );
+      results.push(
+        _.assign(resArray[i], { path: pathWithOptionalProjectName }),
+      );
       // currently testOpts are identical for each test result returned even if it's for multiple projects.
       // we want to return the project names, so will need to be crafty in a way that makes sense.
       if (!testOpts.projectNames) {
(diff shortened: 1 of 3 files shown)


lili2311:
@firecode9 please could you update the commit message so release notes read more like a sentence :)

firecode9:
@lili2311 Yes, fixed :-)

src/cli/commands/test/index.ts
```diff
@@ -132,7 +132,21 @@ async function test(...args: MethodArgs): Promise<string> {
     const resArray: any[] = Array.isArray(res) ? res : [res];
 
     for (let i = 0; i < resArray.length; i++) {
-      results.push(_.assign(resArray[i], { path }));
+      let projectName = resArray[i].projectName;
```
lili2311, line 135: please add a test for this?

firecode9: tests added

src/cli/commands/test/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  let projectName = testResult.projectName;
+  if (projectName) {
+    const index = projectName.indexOf('/');
+    if (index > -1) {
+      projectName = projectName.substr(index);
+    } else {
+      projectName = undefined;
+    }
+  }
+  if (projectName?.startsWith('/')) {
```
aviadhahami, line 16: can be united w/ the previous condition s.t.:
```ts
  let projectName = testResult.projectName;
  if (projectName) {
    const index = projectName.indexOf('/');
    if (index > -1) {
      projectName = projectName.substr(index + 1);
    } else {
      projectName = undefined;
    }
  }
```
due to the fact that your substring will always begin w/ `/`

firecode9: yep.fixed

aviadhahami:
the last thing is commit message (like @lili2311 mentioned). Write why and not what :)

anthogez:
> **Expected release notes (by @firecode9)**
> 
> _fixes_:
> wrong path to display better test results ([2ac2af4](https://github.com/snyk/snyk/commit/2ac2af42994c3c9d8929d41ce8800564cc39bd8f))
> 
> * [ ]  I hereby acknowledge these release notes are 🥙 AWESOME 🥙

What do you think about this as release note message?

fix: display tests full path results

PR #1120, opened 2020-05-13
fix: proposal to "container" command

agranado2k:
It's just a proposal about how to include the "container test" and "container monitor" sub command.

### DIFF
--- a/src/cli/modes.ts
+++ b/src/cli/modes.ts
@@ -1,0 +1,33 @@
+interface ModeData {
+  allowedSubCommands: Array<string>;
+  config: (args) => void;
+}
+
+const modes: Record<string, ModeData> = {
+  container: {
+    allowedSubCommands: ['test', 'monitor'],
+    config: (args): void => {
+      args['docker'] = true;
+    },
+  },
+};
+
+const isModesAllowed = (command) => Object.keys(modes).includes(command);
+
+const isSubCommandAllowed = (command, subCommand) =>
+  modes[command].allowedSubCommands.includes(subCommand);
+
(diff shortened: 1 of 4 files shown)


src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+const isModesAllowed = (command) => Object.keys(modes).includes(command);
+
+const isSubCommandAllowed = (command, subCommand) =>
+  modes[command].allowedSubCommands.includes(subCommand);
+
+const configArgs = (command, args) => modes[command].config(args);
+
+export default (command, args): string => {
```
orsagie, line 22: This should be called `mode`, as that is what we are checking.  Command in the CLI is reserved for test/monitor/protect/help/etc...

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
+interface ModeData {
+  allowedSubCommands: Array<string>;
```
orsagie, line 2: allowedSubCommands -> allowedCommands

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+  },
+};
+
+const isModesAllowed = (command) => Object.keys(modes).includes(command);
+
+const isSubCommandAllowed = (command, subCommand) =>
+  modes[command].allowedSubCommands.includes(subCommand);
+
+const configArgs = (command, args) => modes[command].config(args);
```
orsagie, line 20: we have a command called config, so prefer if you use the full configureArgs here. (also change to mode instead of command)

orsagie:
please change the commit message to "feat: ..."

PR #1124, opened 2020-05-14
feat: include command "container" as alias for option "--docker"

agranado2k:
This is the first step to create a "snyk container test" or "snyk container monitor" commands that doesn't require Docker engine running. Although it's not relevant to all users, more flexibility here is both more technically correct, future proofs our technology, and is important to several separate strategic partners.

- [x] Ready for review

#### What does this PR do?
 Include the command `container` that is an alias for the option `--docker`. The command accepts the sub commands `test` and `monitor`. 

Run the commands `snyk container test <image> ...` and `snyk container monitor <image> ...`. After check `src/cli/args.ts` where the `mode` concept was introduced. 

You should be able to run the command `snyk container test <image> ...`, and get the same result when running the command `snyk test --docker <image> ...`.
The same should work for `monitor`. Running the command `snyk container monitor <image> ...`, should lead to the same result that runs the command `snyk monitor --docker <image> ...`.

The implementation follows the idea create by the Cloud-Config team in this spike [PR](https://github.com/snyk/snyk/commit/9f2905696e5f555c46e36549aa2341f1c92c143b).

- [RUN-589](https://snyksec.atlassian.net/browse/RUN-589)
- [RUN-940](https://snyksec.atlassian.net/browse/RUN-940)

### DIFF
--- a/src/cli/modes.ts
+++ b/src/cli/modes.ts
@@ -1,0 +1,66 @@
+import * as abbrev from 'abbrev';
+import { UnsupportedOptionCombinationError, CustomError } from '../lib/errors';
+
+interface ModeData {
+  allowedCommands: Array<string>;
+  config: (args) => [];
+}
+
+const modes: Record<string, ModeData> = {
+  container: {
+    allowedCommands: ['test', 'monitor'],
+    config: (args): [] => {
+      args['docker'] = true;
+
+      return args;
+    },
+  },
+};
+
+export function parseMode(mode: string, args): string {
+  if (isValidMode(mode)) {
+    const command: string = args._[0];
+
+    if (isValidCommand(mode, command)) {
+      configArgs(mode, args);
+      mode = args._.shift();
+    }
+  }
+
+  return mode;
+}
+
+export function modeValidation(args: object) {
+  const mode = args['command'];
+  const commands: Array<string> = args['options']._;
+
+  if (isValidMode(mode) && commands.length <= 1) {
+    const allowed = modes[mode].allowedCommands
+      .join(', ')
+      .replace(/, ([^,]*)$/, ' or $1');
+    const message = `use snyk ${mode} with ${allowed}`;
+
+    throw new CustomError(message);
+  }
+
+  const command = commands[0];
+  if (isValidMode(mode) && !isValidCommand(mode, command)) {
+    const notSupported = [mode, command];
+
+    throw new UnsupportedOptionCombinationError(notSupported);
+  }
+}
+
+function isValidMode(mode: string): boolean {
+  return Object.keys(modes).includes(mode);
+}
+
+function isValidCommand(mode: string, command: string): boolean {
+  const aliases = abbrev(modes[mode].allowedCommands);
+
+  return Object.keys(aliases).includes(command);
+}
+
+function configArgs(mode: string, args): [] {
+  return modes[mode].config(args);
+}
--- a/test/modes.test.ts
+++ b/test/modes.test.ts
@@ -1,0 +1,212 @@
+import { test } from 'tap';
+import {
+  UnsupportedOptionCombinationError,
+  CustomError,
+} from '../src/lib/errors';
+import { parseMode, modeValidation } from '../src/cli/modes';
+
+test('when is missing command', (c) => {
+  c.test('should do nothing', (t) => {
+    const cliCommand = 'container';
+    const cliArgs = {
+      _: [],
+      'package-manager': 'pip',
+    };
+
+    const command = parseMode(cliCommand, cliArgs);
+
+    t.equal(command, cliCommand);
+    t.equal(cliArgs, cliArgs);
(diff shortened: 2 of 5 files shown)


src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+const isModesAllowed = (command) => Object.keys(modes).includes(command);
+
+const isSubCommandAllowed = (command, subCommand) =>
+  modes[command].allowedSubCommands.includes(subCommand);
+
+const configArgs = (command, args) => modes[command].config(args);
+
+export default (command, args): string => {
+  if (isModesAllowed(command)) {
+    const subCommand = args._[0] as string; // can actually be undefined
```
agatakrajewska, line 24: Why do we have to use type assertion here? Could we not have `const sunCommand: string | undefined = args._[0]`?

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+const isModesAllowed = (mode: string) => Object.keys(modes).includes(mode);
+
+const isCommandAllowed = (mode: string, command: string) =>
+  modes[mode].allowedCommands.includes(command);
+
+const configArgs = (mode: string, args) => modes[mode].config(args);
+
+export default (mode: string, args): string => {
```
gitphill, line 22: We prefer named functions (over anonymous functions) and named exports

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+const modes: Record<string, ModeData> = {
+  container: {
+    allowedCommands: ['test', 'monitor'],
+    config: (args): void => {
+      args['docker'] = true;
+    },
+  },
+};
+
+const isModesAllowed = (mode: string) => Object.keys(modes).includes(mode);
```
ivanstanev, line 15: This could be a function instead of a const variable that is assigned an anonymous function

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const isCommandAllowed = (mode: string, command: string) =>
```
ivanstanev, line 17: This could also be a function

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
...
+  modes[mode].allowedCommands.includes(command);
+
+const configArgs = (mode: string, args) => modes[mode].config(args);
```
ivanstanev, line 20: And this!

src/cli/modes.ts
```diff
@@ -0,0 +1,33 @@
+interface ModeData {
+  allowedCommands: Array<string>;
+  config: (args) => void;
+}
+
+const modes: Record<string, ModeData> = {
+  container: {
+    allowedCommands: ['test', 'monitor'],
+    config: (args): void => {
```
ivanstanev, line 9: Kind of odd how this mutates the object! I think it's very invisible to the rest of the code (line 27) and I had to go all the way to the implementation to figure this out, any chance this function could return `args` in the end? This way on line 27 you can do `args = configArgs(mode, args)` and it's a bit more obvious that args changes

agranado2k:
> without providing 'test' or 'monitor' I get:
> 
> snyk container
> Unknown command "container"
> 
> Maybe a better error message here is, use snyk container with test or monitor and a docker image, for example 'snyk container test node'

Hi @gitphill I addressed your points here. I included a better error message and support for abbreviations too.

test/modes.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        },
+      };
+
+      try {
+        modeValidation(args);
+      } catch (err) {
+        t.ok(err instanceof CustomError, 'should throw CustomError');
+        t.equal(
+          err.message,
+          'use snyk container with test or monitor',
```
gitphill, line 137: 🙏

src/cli/modes.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export function modeValidation(args: object) {
+  const mode = args['command'];
+  const commands: Array<string> = args['options']._;
+
+  if (commands.length <= 1) {
+    const allowed = modes[mode].allowedCommands
+      .join(', ')
+      .replace(/, ([^,]*)$/, ' or $1');
```
gitphill, line 40: fancy 😄

test/modes.test.ts
```diff
@@ -0,0 +1,212 @@
+import { test } from 'tap';
+import {
+  UnsupportedOptionCombinationError,
+  CustomError,
+} from '../src/lib/errors';
+import { parseMode, modeValidation } from '../src/cli/modes';
+
+test('when is missing command', (t) => {
+  t.test('should do nothing', (t) => {
```
gitphill, line 9: this overloads 't' - i tend to use something different at top level, i.e. 

```
test('when is missing command', (missing) => {
  missing.test('should do nothing', (t) => {
```

PR #1126, opened 2020-05-18
Remove git-url-parse dependency

JackuB:
#### What does this PR do?
Remove in favor of a native url parser. Added some basic tests

- [x] Ready for review


michael-go:
I think some more detail is needed in the PR/commit body:
* why are replacing `git-url-parse`
* does the replacement behaves the same as the `git-url-parse`, if not - where does it differ?
* why are we attaching `http://` prefix to all urls, even these that had a different scheme?

JackuB:
@michael-go 

1. continuing https://www.notion.so/snyk/980f13c7e0104a139479f9db211436c0 we are aiming to have less dependencies in the CLI and make it leaner (working on more PRs to remove other deps as well)

2. we could achieve the same functionality with `git-url-parse` by using it's [`.parse()` object output](https://github.com/IonicaBizau/git-url-parse#clipboard-example), but for the sake of point one, switched to native `url.parse()`

3. not happy about this part, but kept it for the consistency sake, as we use these project URLs as a hint how to visually group projects in the Snyk UI.

PR #1129, opened 2020-05-20
fix: do not ignore project name override for container projects

karniwl:
- [x] Ready for review

Run `snyk monitor --docker <image-name> --project-name=<override-name>`

A recent change introduced https://github.com/snyk/snyk/commit/183242cabec47f5506ca48208eea4fe78161624f#diff-cab76b228f009ba2ffcded98078566baR36 mistakenly ignored completely the override project name option using flag `--project-name`  for container projects. This commit fixes it and checks if `--project-name` is used.


src/lib/container/index.ts
```diff
@@ -12,18 +13,28 @@ export function getContainerTargetFile(
...
   }
 }
 
 export function getContainerProjectName(
   scannedProject: ScannedProject,
+  meta: MonitorMeta,
 ): string | undefined {
-  return scannedProject.meta?.imageName;
+  let name = scannedProject.meta?.imageName;
+  if (meta['project-name'] && meta['project-name'] !== '') {
```
aviadhahami, line 36: `if (meta['project-name']?.length)` will simplify this a bit :)

src/lib/container/index.ts
```diff
@@ -12,18 +13,28 @@ export function getContainerTargetFile(
 
 export function getContainerName(
   scannedProject: ScannedProject,
+  meta: MonitorMeta,
 ): string | undefined {
+  let name = scannedProject.meta?.imageName;
```
aviadhahami, line 18: what if there's no `imageName` (for some reason)?
maybe we should have a default value for rainy days?

karniwl: @firecode9 wdyt?

karniwl: that's how we validate it's a container project https://github.com/snyk/snyk/commit/183242cabec47f5506ca48208eea4fe78161624f#diff-2efe993c8fa18a460f73eb44bf5517c8R3

firecode9: afaik there must be an image name

firecode9: > that's how we validate it's a container project [183242c#diff-2efe993c8fa18a460f73eb44bf5517c8R3](https://github.com/snyk/snyk/commit/183242cabec47f5506ca48208eea4fe78161624f#diff-2efe993c8fa18a460f73eb44bf5517c8R3)

yes, thats what I meant 👍

aviadhahami: An empty string bypass these :smiling_imp:

firecode9: but the image name is given as a cli argument, so it must not be empty

aviadhahami: what if we also introduce this change:
```ts
export function isContainer(scannedProject: ScannedProject): boolean {
  return scannedProject.meta?.imageName.length;
}
```

aviadhahami: @firecode9 fair enough;
But may become problematic when you'll introduce API access(not sure if even planned)

firecode9: > what if we also introduce this change:
> 
> ```ts
> export function isContainer(scannedProject: ScannedProject): boolean {
>   return scannedProject.meta?.imageName.length;
> }
> ```

I think it wont do any harm to add this check

aviadhahami: All yours ;)

karniwl: adding :)

karniwl:
@aviadhahami sorry to ping you again before the tests pass, we missed fixing some tests so had to push again, so your review is needed once again

PR #1138, opened 2020-05-22
feat: add lib for dev count

maxjeffos:
- [x] Ready for review

#### What does this PR do?
Upon `snyk monitor`, run git log in the direcotry `snyk monitor` is run; this means that that we are assuming it is run from a directory which is in a git repo. We git log only the last 90 days. Look at the hash(author email) and the commit timestamps. We create a map of (hashed author email) -> (latest commit timestamp for that author). Send that along with the API call to Snyk that is already used in monitor.

This is for the "contributing developer" count feature.

HAMMER-36

#### Related PRs
https://github.com/snyk/registry/pull/13742

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -32,6 +32,14 @@ import { MultiProjectResultCustom } from '../../../lib/plugins/get-multi-plugin-
 import { convertMultiResultToMultiCustom } from '../../../lib/plugins/convert-multi-plugin-res-to-multi-custom';
 import { convertSingleResultToMultiCustom } from '../../../lib/plugins/convert-single-splugin-res-to-multi-custom';
 import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
+import {
+  CONTRIBUTING_DEVELOPER_PERIOD_DAYS,
+  getTimestampStartOfContributingDevTimeframe,
+  parseGitLog,
+  runGitLog,
+  GitRepoCommitStats,
+  execShell,
+} from '../../../lib/monitor/dev-count-analysis';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 const debug = Debug('snyk');
@@ -79,6 +87,27 @@ async function monitor(...args0: MethodArgs): Promise<any> {
   }
 
   apiTokenExists();
+
+  let contributors: { userId: string; lastCommitDate: string }[] = [];
+  if (!options.docker && analytics.allowAnalytics()) {
+    try {
+      const repoPath = process.cwd();
+      const dNow = new Date();
+      const timestampStartOfContributingDeveloperPeriod = getTimestampStartOfContributingDevTimeframe(
+        dNow,
+        CONTRIBUTING_DEVELOPER_PERIOD_DAYS,
+      );
+      const gitLogResults = await runGitLog(
+        timestampStartOfContributingDeveloperPeriod,
+        repoPath,
+        execShell,
+      );
+      const stats: GitRepoCommitStats = parseGitLog(gitLogResults);
+      contributors = stats.getRepoContributors();
+    } catch (err) {
+      debug('error getting repo contributors', err);
+    }
+  }
 
   // Part 1: every argument is a scan target; process them sequentially
   for (const path of args as string[]) {
@@ -175,6 +204,7 @@ async function monitor(...args0: MethodArgs): Promise<any> {
             options,
             projectDeps.plugin as PluginMetadata,
             targetFileRelativePath,
+            contributors,
           ),
           spinner.clear(postingMonitorSpinnerLabel),
         );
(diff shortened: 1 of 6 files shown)


src/lib/monitor/dev-count-analysis.ts
```diff
@@ -0,0 +1,154 @@
...
+  const commitTimestamp = unwrapQuotedString(quotedCommitTimestamp);
+  const commitInfo = new GitCommitInfo(authorEmail, commitTimestamp);
+  return commitInfo;
+}
+
+export function parseGitLog(gitLog: string): GitRepoCommitStats {
+  if (gitLog.trim() === '') {
+    return GitRepoCommitStats.empty();
+  }
+  const logLines = gitLog.trim().split('\n'); // previously used os.EOL but that didn't work right on Windows
```
JackuB, line 100: btw it's a configurable option to use different newline character https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace

So maybe you'll have to check both?

maxjeffos: Ya, good call. I also need to test it on a Windows machine to see what I get by default with real data.

maxjeffos: fixed

src/cli/commands/monitor/index.ts
```diff
@@ -80,6 +87,29 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
+        CONTRIBUTING_DEVELOPER_PERIOD_DAYS,
+      );
+      const gitLogResults = await runGitLog(
+        timestampStartOfContributingDeveloperPeriod,
+        repoPath,
+      );
+      const stats: GitRepoCommitStats = parseGitLog(gitLogResults);
+      repoContributors = stats.getUniqueHashedEmailsAndMostRecentCommitTimestamps();
+    } catch (err) {
+      console.error(err);
```
JackuB, line 109: I’d avoid outputting anything from here unless in debug. Console output (unrelated to monitor command they are trying to run) could confuse users.

I’d check if we have a way to post these errors to our logz.io

src/lib/monitor/dev-count-analysis.ts
```diff
@@ -0,0 +1,178 @@
+import * as crypto from 'crypto';
```
JackuB, line 1: I would add here a rather large comment to explain why we do it and what we collect.

src/lib/monitor/index.ts
```diff
@@ -48,6 +48,9 @@ interface MonitorBody {
   target: {};
   targetFileRelativePath: string;
   targetFile: string;
+  repoContributors?: {
```
JackuB, line 51: Registry is expecting a different format - array of hash and last commit timestamp. See my PR on registry for key names

src/lib/monitor/dev-count-analysis.ts
```diff
@@ -0,0 +1,178 @@
...
+  const lookbackTimespanSeconds = timespanInDays * ONE_DAY_IN_SECONDS;
+  const startOfPeriodEpochSeconds = nowUtcEpocS - lookbackTimespanSeconds;
+  return startOfPeriodEpochSeconds;
+}
+
+export async function runGitLog(
+  timestampEpochSecondsStartOfPeriod: number,
+  repoPath: string,
+): Promise<string> {
+  const gitLogCommand = `git --no-pager log --pretty=tformat:"'%H'${SERIOUS_DELIMITER}'%an'${SERIOUS_DELIMITER}'%ae'${SERIOUS_DELIMITER}'%aI'" --after="${timestampEpochSecondsStartOfPeriod}"`;
```
JackuB, line 167: Why even have the quotes if you are separating by delimiter?

test/dev-count-analysis.jest-test.ts
```diff
@@ -0,0 +1,198 @@
...
+  const line =
+    "'0bd4d3c394a54ba54f6c44705ac73d7d87b39525'_SNYK_SEPARATOR_'some-user'_SNYK_SEPARATOR_'someemail-1@somedomain.com'_SNYK_SEPARATOR_'2020-02-06T11:43:11+00:00'";
+  const commitInfo: GitCommitInfo = parseGitLogLine(line);
+  t.equal(
+    commitInfo.authorHashedEmail,
+    expectedEmailHashes['someemail-1@somedomain.com'],
+  );
+});
+
+test('can handle an empty git log', (t) => {
```
JackuB, line 77: Would also add a case where git is not available and command throws (?)

src/lib/monitor/index.ts
maxjeffos: fixed

src/lib/monitor/dev-count-analysis.ts
maxjeffos: I don't think you can even use the Snyk CLI unless you've authenticated it... so not sure the part about "authenticated users" is needed. What do you think?

src/lib/monitor/dev-count-analysis.ts
maxjeffos: took them out

test/dev-count-analysis.jest-test.ts
maxjeffos: done

src/cli/commands/monitor/index.ts
maxjeffos: Fixed the console.log; looking into the logzio part.

PR #1140, opened 2020-05-24
feat: add Cloud Config support

orkamara:
- [ ] Ready for review

#### What does this PR do?
Cloud Config support

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -21,6 +21,7 @@ import {
   LegacyVulnApiResult,
   SEVERITY,
   TestResult,
+  CloudConfigTestResult,
   VulnMetaData,
 } from '../../../lib/snyk-test/legacy';
 import {
@@ -42,6 +43,7 @@ import {
   summariseVulnerableResults,
 } from './formatters';
 import * as utils from './utils';
+import { getCloudConfigDisplayedOutput } from './cloud-config-output';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
@@ -340,12 +342,12 @@ function isVulnFixable(vuln) {
 }
 
 function displayResult(
-  res: TestResult,
+  res: TestResult | CloudConfigTestResult,
   options: Options & TestOptions,
   foundProjectCount?: number,
 ) {
   const meta = formatTestMeta(res, options);
-  const dockerAdvice = dockerRemediationForDisplay(res);
+  const dockerAdvice = dockerRemediationForDisplay(res as TestResult);
   const projectType =
     (res.packageManager as SupportedProjectTypes) || options.packageManager;
   const localPackageTest = isLocalFolder(options.path);
@@ -355,11 +357,16 @@ function displayResult(
   if (res instanceof Error) {
     return prefix + res.message;
   }
-  const issuesText = res.licensesPolicy ? 'issues' : 'vulnerabilities';
+  const issuesText =
+    //TODO(orka): remove (res as TestResult)
+    (res as TestResult).licensesPolicy || res.packageManager === 'k8sconfig'
+      ? 'issues'
+      : 'vulnerabilities';
   let pathOrDepsText = '';
 
   if (res.hasOwnProperty('dependencyCount')) {
-    pathOrDepsText += res.dependencyCount + ' dependencies';
+    //TODO(orka): remove (res as TestResult)
+    pathOrDepsText += (res as TestResult).dependencyCount + ' dependencies';
   } else {
     pathOrDepsText += options.path;
   }
@@ -377,7 +384,9 @@ function displayResult(
   }
 
   // OK  => no vulns found, return
-  if (res.ok && res.vulnerabilities.length === 0) {
+  //TODO(orka): handle res.ok
+  //TODO(orka): remove (res as TestResult)
+  if (res.ok && (res as TestResult).vulnerabilities.length === 0) {
     const vulnPathsText = options.showVulnPaths
       ? 'no vulnerable paths found.'
       : 'none were found.';
@@ -408,10 +417,21 @@ function displayResult(
     );
   }
 
+  if (res.packageManager === 'k8sconfig') {
+    return getCloudConfigDisplayedOutput(
+      res as CloudConfigTestResult,
+      options,
+      testedInfoText,
+      meta,
+      prefix,
+      multiProjAdvice,
+    );
+  }
+
   // NOT OK => We found some vulns, let's format the vulns info
 
   return getDisplayedOutput(
-    res,
+    res as TestResult,
     options,
     testedInfoText,
     localPackageTest,
--- a/src/lib/cloud-config/cloud-config-parser.ts
+++ b/src/lib/cloud-config/cloud-config-parser.ts
@@ -1,0 +1,102 @@
+import * as _ from 'lodash';
+
+import * as YAML from 'js-yaml';
+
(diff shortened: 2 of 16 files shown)


src/lib/snyk-test/legacy.ts
```diff
@@ -121,24 +122,36 @@ export interface IgnoreSettings {
   disregardFilesystemIgnores: boolean;
 }
 
-export interface LegacyVulnApiResult {
-  vulnerabilities: AnnotatedIssue[];
+interface BasicResultData {
```
Shesekino, line 125: doesn't that mean that we're doubling down on something referred to as legacy?

src/lib/snyk-test/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  projectNameOverride?: string;
+  hasDevDependencies?: boolean;
+  originalProjectName?: string; // used only for display
+  foundProjectCount?: number; // used only for display
+  docker?: any;
+  displayTargetFile?: string;
+  target?: GitTarget | ContainerTarget | null;
+}
+
+export interface CloudConfigPayloadBody {
```
Shesekino, line 20: aren't we trying to achieve a similar payload body interface in project-envelope?

Shesekino:
could you split this to several smaller commits with more precise changes please? it would be much easier to review in my opinion

src/lib/errors/invalid-cloud-config-file.ts
```diff
@@ -0,0 +1,38 @@
...
+  const error = new CustomError(errorMsg);
+  error.code = 422;
+  error.userMessage = errorMsg;
+  return error;
+}
+
+export function IlegalCloudConfigFileError(atLocations: string[]) {
+  const locationsStr = atLocations.join(', ');
+  const errorMsg =
+    'Ilegal supported Cloud Config target files in ' +
```
Shesekino, line 25: typo here I think

src/lib/errors/invalid-cloud-config-file.ts
```diff
@@ -0,0 +1,38 @@
+import chalk from 'chalk';
+import { CustomError } from './custom-error';
+
+export function NoSupportedCloudConfigFileError(atLocations: string[]) {
```
Shesekino, line 4: missing return type I think... also shouldn't functions start with a lower case letter?

src/lib/errors/invalid-cloud-config-file.ts
```diff
@@ -0,0 +1,38 @@
...
+    ) +
+    ' and make sure you are in the right directory.';
+
+  const error = new CustomError(errorMsg);
+  error.code = 422;
+  error.userMessage = errorMsg;
+  return error;
+}
+
+export function IlegalCloudConfigFileError(atLocations: string[]) {
```
Shesekino, line 22: typo here I think?

Shesekino: and missing return type

src/lib/snyk-test/legacy.ts
```diff
@@ -279,7 +294,7 @@ export interface RemediationChanges {
 function convertTestDepGraphResultToLegacy(
   res: TestDepGraphResponse,
   depGraph: depGraphLib.DepGraph,
-  packageManager: string,
+  packageManager: SupportedProjectTypes | undefined,
```
Shesekino, line 297: why are we allowing ```undefined``` now?

src/lib/snyk-test/legacy.ts
orkamara: WDYM?

src/lib/snyk-test/types.ts
orkamara: yes - that's only a temp change :) just to check the logic of this PR

src/lib/snyk-test/legacy.ts
orkamara: That's a good question, but it's possible that `depgraph` doesn't contain `PM` - https://github.com/snyk/snyk/blob/master/src/lib/snyk-test/run-test.ts#L97

src/lib/cloud-config/cloud-config-parser.ts
```diff
@@ -0,0 +1,102 @@
+import * as _ from 'lodash';
```
orsagie, line 1: The goal of this parser is to live outside of the CLI, right?

orkamara:
> could you split this to several smaller commits with more precise changes please? it would be much easier to review in my opinion

That's what I'm doing right now :)

src/lib/cloud-config/cloud-config-parser.ts
orkamara: Maybe we can create a lib - but let's wait with that? so there will be not too many changes right now?

src/lib/snyk-test/legacy.ts
orsagie: @Shesekino see below for the LegacyVulnApiResult type

Shesekino: what I mean is that instead of moving to new types, we're extracting some of the things considered Legacy from ```LegacyVulnApiResult``` and expanding over that, probably making future changes there harder, and I wonder if project-envelope does not take that into consideration (maybe it should)?

src/cli/commands/test/index.ts
```diff
@@ -431,10 +461,21 @@ function displayResult(
...
   }
 
+  if (res.packageManager === 'k8sconfig') {
+    return getIacDisplayedOutput(
+      (res as any) as IacTestResult,
+      options,
+      testedInfoText,
+      meta,
+      prefix,
+      multiProjAdvice,
```
orsagie, line 471: multiProjAdvice is only for gradle, don't think you need it here

src/lib/snyk-test/run-test.ts
```diff
@@ -49,76 +51,61 @@ import { getSubProjectCount } from '../plugins/get-sub-project-count';
...
         _.get(payload, 'body.projectNameOverride') ||
         _.get(payload, 'body.originalProjectName');
       const foundProjectCount = _.get(payload, 'body.foundProjectCount');
       const displayTargetFile = _.get(payload, 'body.displayTargetFile');
       let dockerfilePackages;
       if (
-        payload.body &&
-        payload.body.docker &&
-        payload.body.docker.dockerfilePackages
+        depGraphPayload &&
```
orsagie, line 101: any reason to rename this to be explicitly depGraph? Don't think this needs to be called out in the variable. You can call it payloadBody if you feel the need to rename

src/lib/snyk-test/run-test.ts
```diff
@@ -326,19 +326,27 @@ async function assembleLocalPayloads(
   options: Options & TestOptions & PolicyOptions,
 ): Promise<Payload[]> {
   // For --all-projects packageManager is yet undefined here. Use 'all'
-  const analysisType =
-    (options.docker ? 'docker' : options.packageManager) || 'all';
+
+  let analysisTypeText = options.packageManager + ' dependencies for ';
```
orsagie, line 330: The `'all'` got lost in translation here

src/cli/commands/test/index.ts
orkamara: done

src/lib/snyk-test/run-test.ts
orkamara: renamed

src/lib/snyk-test/run-test.ts
orkamara: done

PR #1155, opened 2020-05-31
chore: update nodejs-lockfile-parser that contain pre-yarn2 refactor

dkontorovskyy:
### What does this PR do?
Update nodejs-lockfile-parser that contain pre-yarn2 refactor

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "snyk-gradle-plugin": "3.3.4",
     "snyk-module": "3.1.0",
     "snyk-mvn-plugin": "2.15.2",
-    "snyk-nodejs-lockfile-parser": "1.22.0",
+    "snyk-nodejs-lockfile-parser": "1.23.0",
     "snyk-nuget-plugin": "1.18.1",
     "snyk-php-plugin": "1.9.0",
     "snyk-policy": "1.14.1",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,6 +8,7 @@
         ],
         "module": "commonjs",
         "allowJs": true,
+        "skipLibCheck": true,
         "sourceMap": true,
         "strict": true,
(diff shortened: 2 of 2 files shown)


tsconfig.json
```diff
@@ -8,6 +8,7 @@
         ],
         "module": "commonjs",
         "allowJs": true,
+        "skipLibCheck": true,
```
dkontorovskyy, line 11: Skip type declarations check for imported @types through package.json

anthogez: This way you are silence ts, will this be only temporary?

anthogez: with skipLibCheck you are applying a silence to TS on types, will this be only temporary? If not, why?

package.json
```diff
@@ -82,7 +82,7 @@
     "snyk-gradle-plugin": "3.3.4",
     "snyk-module": "3.1.0",
     "snyk-mvn-plugin": "2.15.2",
-    "snyk-nodejs-lockfile-parser": "1.22.0",
+    "snyk-nodejs-lockfile-parser": "1.23.0",
```
anthogez, line 85: I would suggest to change from chore to feat, since this change on `snyk-nodejs-lockfile-parser` apply the following changes:

- add support for yarn 2 lock files
- prepare for yarn v2 feature

dkontorovskyy: There is no functional change. Shouldn't be included in Release notes

tsconfig.json
dkontorovskyy: Checking with @arcanis in main PR https://github.com/snyk/nodejs-lockfile-parser/pull/57#issuecomment-636692497

anthogez: @dkontorovskyy thanks 👌

dkontorovskyy:
Closing in fare of 
https://github.com/snyk/snyk/pull/1254

PR #1157, opened 2020-06-01
feat: --reachable-vulns supports --all-projects flag in CLI

AlbertKogan:
#### What does this PR do?
Adding support of `--all-projects` flag altogether with `--reachable-vulns` for `snyk test` CLI command

Run in CLI `snyk test --all-projects --reachable-vulns`

### DIFF
--- a/src/lib/options-validator.ts
+++ b/src/lib/options-validator.ts
@@ -6,12 +6,13 @@ import * as reachableVulns from './reachable-vulns';
 export async function validateOptions(
   options: (Options & TestOptions) | (Options & MonitorOptions),
   packageManager?: SupportedPackageManagers,
-) {
+): Promise<void> {
   if (options.reachableVulns) {
-    if (!packageManager) {
+    // Throwing error only in case when both packageManager and allProjects not defined
+    if (!packageManager && !options.allProjects) {
       throw new Error('Could not determine package manager');
     }
     const org = options.org || config.org;
-    await reachableVulns.validatePayload(packageManager, org);
+    await reachableVulns.validatePayload(org, options, packageManager);
   }
 }
--- a/src/lib/reachable-vulns.ts
+++ b/src/lib/reachable-vulns.ts
@@ -11,6 +11,7 @@ import {
   FeatureNotSupportedByPackageManagerError,
   UnsupportedFeatureFlagError,
 } from './errors';
+import { MonitorOptions, Options, TestOptions } from './types';
 
 const featureFlag = 'reachableVulns';
 
@@ -29,10 +30,15 @@ export function serializeCallGraphWithMetrics(
 }
 
 export async function validatePayload(
-  packageManager: SupportedPackageManagers,
   org: any,
+  options: (Options & TestOptions) | (Options & MonitorOptions),
+  packageManager?: SupportedPackageManagers,
 ): Promise<boolean> {
-  if (!REACHABLE_VULNS_SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)) {
+  if (
+    packageManager &&
+    !options.allProjects &&
+    !REACHABLE_VULNS_SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)
+  ) {
     throw new FeatureNotSupportedByPackageManagerError(
       'Reachable vulns',
       packageManager,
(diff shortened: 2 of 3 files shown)


src/lib/options-validator.ts
```diff
@@ -6,12 +6,13 @@ import * as reachableVulns from './reachable-vulns';
 export async function validateOptions(
   options: (Options & TestOptions) | (Options & MonitorOptions),
   packageManager?: SupportedPackageManagers,
-) {
+): Promise<void> {
   if (options.reachableVulns) {
-    if (!packageManager) {
+    // Throwing error only in case when both packageManager and allProjects not defined
+    if (!packageManager && !options.allProjects) {
```
andremidea, line 12: The only reason why it wouldn't work was this validation?

andremidea: I asked that because I don't understand the implication of the `allProjects` flag

AlbertKogan: @andremidea This check prevented workflow and passing `reachableVulns` option to `snyk-mvn-plugin` and after to `java-call-graph-builder` module where actual work done. This option will be ignoring in other plugins.

src/lib/reachable-vulns.ts
```diff
@@ -29,10 +29,13 @@ export function serializeCallGraphWithMetrics(
 }
 
 export async function validatePayload(
-  packageManager: SupportedPackageManagers,
+  packageManager: SupportedPackageManagers | null,
   org: any,
 ): Promise<boolean> {
-  if (!REACHABLE_VULNS_SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)) {
+  if (
+    packageManager &&
```
darmalovan, line 36: I'm afraid of `null` package manager that are coming from other places bypassing this logic.
Maybe worth also passing options.allProjects as a parameter to the function?

src/lib/options-validator.ts
```diff
@@ -6,12 +6,13 @@ import * as reachableVulns from './reachable-vulns';
...
       throw new Error('Could not determine package manager');
     }
     const org = options.org || config.org;
-    await reachableVulns.validatePayload(packageManager, org);
+    await reachableVulns.validatePayload(packageManager ?? null, org);
```
darmalovan, line 16: maybe the `null` part can be set as a default value, e.g.
`packageManager: SupportedPackageManagers | null = null`

src/lib/reachable-vulns.ts
AlbertKogan: @darmalovan Tried to avoid changing of function's parameters. Can you elaborate a bit more about your concerns?

src/lib/options-validator.ts
```diff
@@ -5,10 +5,11 @@ import * as reachableVulns from './reachable-vulns';
 
 export async function validateOptions(
   options: (Options & TestOptions) | (Options & MonitorOptions),
-  packageManager?: SupportedPackageManagers,
-) {
+  packageManager: SupportedPackageManagers | null = null,
```
dkontorovskyy, line 8: `packageManager: SupportedPackageManagers | null = null` means packageManager can be null when before it was defaulting to `undefined`, cos of `?` why this change is crucial?

AlbertKogan: @daniel The intend here is to prevent changing order of parameters for this function `await reachableVulns.validatePayload(packageManager, org);`. So if you want to keep this order, you need to provide value, because TS requires non optional parameters first

src/lib/reachable-vulns.ts
darmalovan: My worry is that in the future a packageManager can be propagated here with a `null` value, without the allProjects parameter, and it will be considered valid.

AlbertKogan: Hmm, but why without `allProjects`? We not modifying `options` object, so this parameter should be there

AlbertKogan: Ah, you mean here in this function. Get it

AlbertKogan: But we have this check in `validateOptions` function
``` javascript
if (!packageManager && !options.allProjects) {
    throw new Error('Could not determine package manager');
}
```

darmalovan: `validatePayload` is a public function that can be used in the future from other callsites

AlbertKogan: Ok, so maybe it worth to pass `options` object?

darmalovan: Sounds good :)

PR #1165, opened 2020-06-04
feat/support depgraph deptree coexistence from plugins

anthogez:
- [x] Ready for review

This pr allows snyk-cli of accepts `depGraphs from snyk-plugins`, and keep the current available support to the rest of ecosystem by creating a coexistence path between depgraph and deptrees.

After this pr we will create another one where we will add tests and use
the `snyk-gradle-plugin` coming from this pr 
https://github.com/snyk/snyk-gradle-plugin/pull/127 🚧  
that will be the 1st `snyk-cli` plugin to be sending `depGraph`

The goal of divide this release in two steps is to keep things simple and safer

This PR depends of:
https://github.com/snyk/snyk-cli-interface/pull/34 ✅

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -181,6 +181,9 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
       // Post the project dependencies to the Registry
       for (const projectDeps of perProjectResult.scannedProjects) {
+        if (!projectDeps.depTree) {
+          continue;
+        }
         const extractedPackageManager = extractPackageManager(
           projectDeps,
           perProjectResult,
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -20,7 +20,7 @@ import * as spinner from '../../../lib/spinner';
 import * as analytics from '../../../lib/analytics';
 import { MethodArgs, ArgsOptions } from '../../args';
 import { apiTokenExists } from '../../../lib/api-token';
-import { maybePrintDeps } from '../../../lib/print-deps';
+import { maybePrintDepTree, maybePrintDepGraph } from '../../../lib/print-deps';
 import { monitor as snykMonitor } from '../../../lib/monitor';
 import { processJsonMonitorResponse } from './process-json-monitor';
 import snyk = require('../../../lib'); // TODO(kyegupov): fix import
@@ -40,6 +40,7 @@ import {
   GitRepoCommitStats,
   execShell,
 } from '../../../lib/monitor/dev-count-analysis';
+import { FailedToRunTestError } from '../../../lib/errors';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 const debug = Debug('snyk');
@@ -181,22 +182,42 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
       // Post the project dependencies to the Registry
       for (const projectDeps of perProjectResult.scannedProjects) {
-        if (!projectDeps.depTree) {
-          continue;
+        if (!projectDeps.depGraph && !projectDeps.depTree) {
+          debug(
+            'scannedProject is missing depGraph or depTree, cannot run test/monitor',
+          );
+          throw new FailedToRunTestError(
+            'Your monitor request could not be completed. Please email support@snyk.io',
+          );
         }
         const extractedPackageManager = extractPackageManager(
           projectDeps,
           perProjectResult,
           options as MonitorOptions & Options,
         );
-        analytics.add('packageManager', extractedPackageManager);
-        maybePrintDeps(options, projectDeps.depTree);
 
-        debug(`Processing ${projectDeps.depTree.name}...`);
+        analytics.add('packageManager', extractedPackageManager);
+
+        let projectName;
+
+        if (projectDeps.depGraph) {
+          debug(`Processing ${projectDeps.depGraph.rootPkg.name}...`);
+          maybePrintDepGraph(options, projectDeps.depGraph);
+          projectName = projectDeps.depGraph.rootPkg.name;
+        }
+
+        if (projectDeps.depTree) {
+          debug(`Processing ${projectDeps.depTree.name}...`);
+          maybePrintDepTree(options, projectDeps.depTree);
+          projectName = projectDeps.depTree.name;
+        }
+
         const tFile = projectDeps.targetFile || targetFile;
-        const targetFileRelativePath = tFile
-          ? pathUtil.join(pathUtil.resolve(path), tFile)
-          : '';
+        const targetFileRelativePath =
+          projectDeps.plugin.targetFile ||
+          (tFile && pathUtil.join(pathUtil.resolve(path), tFile)) ||
+          '';
+
         const res: MonitorResult = await promiseOrCleanup(
           snykMonitor(
             path,
@@ -211,8 +232,6 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         );
 
         res.path = path;
-        const projectName = projectDeps.depTree.name;
-
         const monOutput = formatMonitorOutput(
           extractedPackageManager,
           res,
@@ -220,9 +239,9 @@ async function monitor(...args0: MethodArgs): Promise<any> {
           projectName,
           getSubProjectCount(inspectResult),
         );
+        // push a good result
         results.push({ ok: true, data: monOutput, path, projectName });
       }
-      // push a good result
     } catch (err) {
       // push this error, the loop continues
       results.push({ ok: false, data: err, path });
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -85,6 +85,7 @@ export async function monitor(
   analytics.add('isDocker', !!meta.isDocker);
 
   if (GRAPH_SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)) {
+    // TODO @boost: remove the code below once 'experimental-dep-graph' is deleted
     const monitorGraphSupportedRes = await isFeatureFlagSupportedForOrg(
       _.camelCase('experimental-dep-graph'),
       options.org || config.org,
@@ -97,7 +98,7 @@ export async function monitor(
       );
     }
     if (monitorGraphSupportedRes.ok) {
-      return await monitorGraph(
+      return await experimentalMonitorDepGraph(
         root,
         meta,
         scannedProject,
@@ -264,7 +265,10 @@ async function monitorDepTree(
   });
 }
 
-export async function monitorGraph(
+// @deprecated: it will be deleted once experimentalDepGraph FF will be deleted
+// and npm, yarn, sbt and rubygems usage of `experimentalMonitorDepGraph`
+// will be replaced with `monitorDepGraph` method
+export async function experimentalMonitorDepGraph(
   root: string,
   meta: MonitorMeta,
   scannedProject: ScannedProject,
@@ -273,7 +277,7 @@ export async function monitorGraph(
   contributors?: { userId: string; lastCommitDate: string }[],
 ): Promise<MonitorResult> {
   const packageManager = meta.packageManager;
-  analytics.add('monitorGraph', true);
+  analytics.add('experimentalMonitorDepGraph', true);
 
   let treeMissingDeps: string[];
   let pkg = scannedProject.depTree;
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -85,6 +85,20 @@ export async function monitor(
   analytics.add('isDocker', !!meta.isDocker);
 
   if (GRAPH_SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)) {
+    const isGradlePkgManager: boolean = packageManager === 'gradle';
+
+    // TODO @boost: delete this condition once 'experimental-dep-graph' ff is deleted
+    if (isGradlePkgManager) {
+      return await monitorDepGraph(
+        root,
+        meta,
+        scannedProject,
+        pluginMeta,
+        targetFileRelativePath,
+        contributors,
+      );
+    }
+
     // TODO @boost: remove the code below once 'experimental-dep-graph' is deleted
     const monitorGraphSupportedRes = await isFeatureFlagSupportedForOrg(
       _.camelCase('experimental-dep-graph'),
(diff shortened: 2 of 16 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -181,16 +181,28 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
       // Post the project dependencies to the Registry
       for (const projectDeps of perProjectResult.scannedProjects) {
+        if (!projectDeps.depTree && !projectDeps.depGraph) {
+          continue;
```
gitphill, line 185: do we not want to throw? this should never happen

src/cli/commands/monitor/index.ts
```diff
@@ -181,16 +181,28 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
-        maybePrintDeps(options, projectDeps.depTree);
+        if (projectDeps.depGraph) {
+          maybePrintDepGraphs(options, projectDeps.depGraph);
+        }
+
+        if (projectDeps.depTree) {
+          analytics.add('packageManager', extractedPackageManager);
+          maybePrintDeps(options, projectDeps.depTree);
+          debug(`Processing ${projectDeps.depTree.name}...`);
+          maybePrintDeps(options, projectDeps.depTree);
```
gitphill, line 204: did we figure out why this prints twice?

src/lib/monitor/index.ts
```diff
@@ -264,7 +288,106 @@ async function monitorDepTree(
...
+  }
+
+  const policy = await snyk.policy.load(policyLocations, { loose: true });
+  const target = await projectMetadata.getInfo(scannedProject, meta);
+  if (isGitTarget(target) && target.branch) {
+    analytics.add('targetBranch', target.branch);
+  }
+
+  return new Promise((resolve, reject) => {
+    if (!depGraph) {
```
gitphill, line 325: already done this? line 305

src/lib/monitor/index.ts
```diff
@@ -85,6 +85,21 @@ export async function monitor(
   analytics.add('isDocker', !!meta.isDocker);
 
   if (GRAPH_SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)) {
+    const isGradlePkgManager: boolean = packageManager === 'gradle';
+
+    // TODO @boost: delete this condition once 'experimental-dep-graph' ff is deleted
+    if (isGradlePkgManager) {
```
gitphill, line 91: && they have new feature flag?

src/lib/monitor/index.ts
anthogez: Without this if the internal variables complains `Object is possibly 'undefined'`

src/cli/commands/monitor/index.ts
anthogez: Yes we did, but I prefered to apply small changes (clean up once the slightly basic changes are done)

src/cli/commands/monitor/index.ts
anthogez: This for will be handled different, and will have a custom error that is on the TODO list for when does not exists a depTree or depGraph

src/lib/monitor/index.ts
anthogez: The feature flag is not required for the moment, we will consider it again after the benchmarks tests with the big fixtures that I mentioned yesterday.

I am trying to keep it simple for now and slowly polishing

anthogez:
> we need some tests for the different paths through monitor now

Yes we will be creating those, once the depGraph path is done to double check that both are working properly. The path is still broken and under dev

src/lib/monitor/index.ts
anthogez: And.. I am considering (once the huge part of the job is done) to refactor it and apply a normal await async

src/lib/snyk-test/run-test.ts
```diff
@@ -354,15 +355,24 @@ async function assembleLocalPayloads(
     }
 
     for (const scannedProject of deps.scannedProjects) {
-      const pkg = scannedProject.depTree;
+      if (!scannedProject.depTree && !scannedProject.depGraph) {
+        throw new InvalidScannedProjectError();
+      }
+
+      let pkg: any = scannedProject.depTree;
```
anthogez, line 362: TODO: make sure to add missing types

src/lib/errors/invalid-scanned-project-error.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomError } from './custom-error';
+
+export class InvalidScannedProjectError extends CustomError {
+  private static ERROR_MESSAGE =
+    'Invalid ScannedProject. It must contains at least a depGraph or depTree';
```
lili2311, line 5: does it make sense to a user to see?

anthogez: This error is unlikely to happen to a user unless we have internal problems (in our plugins or any bad code). I think written this way will allow us, understand immediately the problem. But if you have any suggestion about how we can improve it I am all ears ^^

src/cli/commands/monitor/index.ts
```diff
@@ -181,21 +182,37 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
+        if (projectDeps.depGraph) {
+          debug(`Processing ${projectDeps.depGraph.rootPkg.name}...`);
+          maybePrintDepGraph(options, projectDeps.depGraph);
+          projectName = projectDeps.depGraph.rootPkg.name;
+          targetFileRelativePath = projectDeps.plugin.targetFile;
+        }
+
+        if (projectDeps.depTree) {
+          debug(`Processing ${projectDeps.depTree.name}...`);
+          maybePrintDepTree(options, projectDeps.depTree);
```
lili2311, line 208: i think it would be nicer to make `mayebprintDeps` handle both instead of the `if` statement here. Pushing teh complexity to the function instead?

src/lib/print-deps.ts
```diff
@@ -1,9 +1,20 @@
...
 import { DepDict, Options, MonitorOptions } from './types';
 import { legacyCommon as legacyApi } from '@snyk/cli-interface';
 
+export function maybePrintDepGraph(
+  options: Options | MonitorOptions,
+  depGraph: DepGraph,
+) {
+  if (options['print-deps']) {
+    // TODO @boost: add as output graphviz 'dot' file to visualize?
+    console.log(JSON.stringify(depGraph.toJSON(), null, 2));
```
lili2311, line 11: this output format is totally different to the other one, this will be confusing?

anthogez: It's ok because at the moment we cannot to print it in the same way that we did for the tree, we need to define a new model

src/cli/commands/monitor/index.ts
anthogez: I think the way it's now is simple, we current need those if anyway to define the projectName and the targetFile, once we finish of convert the rest of plugins to use depGraph we can get rid of this minimal complexity and have 0 ifs ^^

src/lib/errors/invalid-scanned-project-error.ts
orsagie: I think this type of error message should be a debug statement, and some more user friendly error should be thrown

src/lib/print-deps.ts
orsagie: shouldn't we also be checking for `options.json`? if options.json is not set, we can say we don't do prettyPrinting for graphs yet?

or we can [for now] convert back to tree to use the `printDeps` method.

Ideally we would write a printGraphDeps method that prints deps prettily.

src/cli/commands/monitor/index.ts
```diff
@@ -181,19 +182,37 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
-          ? pathUtil.join(pathUtil.resolve(path), tFile)
-          : '';
+        let projectName;
+        let targetFileRelativePath;
+
+        if (projectDeps.depGraph) {
+          debug(`Processing ${projectDeps.depGraph.rootPkg.name}...`);
+          maybePrintDepGraph(options, projectDeps.depGraph);
+          projectName = projectDeps.depGraph.rootPkg.name;
+          targetFileRelativePath = projectDeps.plugin.targetFile;
```
orsagie, line 203: is this correct? we do a path.resolve for the other targetFileRelativePath

anthogez: Are you suggesting to keep the same for both cases?

src/lib/monitor/index.ts
```diff
@@ -264,7 +281,112 @@ async function monitorDepTree(
...
+  targetFileRelativePath?: string,
+  contributors?: { userId: string; lastCommitDate: string }[],
+): Promise<MonitorResult> {
+  const packageManager = meta.packageManager;
+  analytics.add('monitorDepGraph', true);
+
+  let depGraph = scannedProject.depGraph;
+
+  if (!depGraph) {
+    throw new InvalidScannedProjectError();
```
orsagie, line 298: Maybe we should have one `monitor` method that knows to do these validation checks and sends to the correct underlying method (monitorDepTree|Graph`?

src/lib/monitor/index.ts
```diff
@@ -273,23 +395,28 @@ export async function monitorGraph(
   contributors?: { userId: string; lastCommitDate: string }[],
 ): Promise<MonitorResult> {
   const packageManager = meta.packageManager;
-  analytics.add('monitorGraph', true);
+  analytics.add('experimentalMonitorDepGraphFromDepTree', true);
```
orsagie, line 398: Is this going to affect any analytics or reports we look at?

src/lib/monitor/index.ts
```diff
@@ -315,13 +442,16 @@ export async function monitorGraph(
   }
 
   return new Promise((resolve, reject) => {
+    if (!depTree) {
```
orsagie, line 445: why would depTree disappear here? There is already a check for this above. 

(same question for the depGraph method above)

src/lib/print-deps.ts
anthogez: We can skip the prettyPrinting, because converting a big depGraph to depTree would break the CLI 😄

src/lib/snyk-test/run-test.ts
```diff
@@ -354,10 +355,23 @@ async function assembleLocalPayloads(
     }
 
     for (const scannedProject of deps.scannedProjects) {
-      const pkg = scannedProject.depTree;
+      if (!scannedProject.depTree && !scannedProject.depGraph) {
+        throw new InvalidScannedProjectError();
```
orsagie, line 359: Is this valuable to display to users? I don't this we should ever be in this situation, and if we are, the user needs to understand that this is a bug on our side, and they need to report it.

src/lib/print-deps.ts
orsagie: But we are not only solving this for bigTrees. We are solving this for everyone. If bigTree projects, that already fail, keep failing, but we also start failing or stopping a feature for all gradle projects, that is not OK.

src/lib/project-metadata/target-builders/git.ts
```diff
@@ -1,15 +1,11 @@
...
 import subProcess = require('../../sub-process');
-import { DepTree } from '../../types';
 import { GitTarget } from '../types';
-import { ScannedProject } from '@snyk/cli-interface/legacy/common';
 
 // for scp-like syntax [user@]server:project.git
 const originRegex = /(.+@)?(.+):(.+$)/;
 
 export async function getInfo(
-  scannedProject: ScannedProject,
```
orsagie, line 11: 👌

src/lib/monitor/index.ts
anthogez: I wanted to keep things as simple as possible and afterwards polish it to avoid breaking changes ^^

src/lib/print-deps.ts
anthogez: here I just kept the logic as it was to prevent breaking changes and add the support for depGraph and we are not breaking anything. I am just saying that if the format is not good we can just drop it for languages that will be using depGraph while we do not have a properly format for it on this case specific gradle.

src/lib/errors/invalid-scanned-project-error.ts
anthogez: What message would you suggest for this one?

src/lib/monitor/index.ts
anthogez: the idea is to drop this experimentalMonitorDepGraph once we are done with the implementation of this new depGraph path

src/lib/monitor/index.ts
anthogez: Out of scope and this does not recognize the depTree check, I will keep it like this till I create another pr that will implements async await

src/lib/snyk-test/run-test.ts
anthogez: this was answered upside

src/lib/errors/invalid-scanned-project-error.ts
anthogez: This has been solved :)

src/cli/commands/monitor/index.ts
```diff
@@ -181,19 +182,42 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
       // Post the project dependencies to the Registry
       for (const projectDeps of perProjectResult.scannedProjects) {
+        if (!projectDeps.depGraph && !projectDeps.depTree) {
+          debug(
+            'scannedProject is missing depGraph or depTree, cannot run test/monitor',
+          );
+          throw new FailedToRunTestError(
+            'scannedProject is missing depGraph or depTree, cannot run test/monitor',
```
lili2311, line 190: this message will be visible to the user? Not sure it makes sense to them `Test could not be completed successull contact support` or something is actionable. This message has no action a user could take alone

src/cli/commands/monitor/index.ts
```diff
@@ -181,19 +182,42 @@ async function monitor(...args0: MethodArgs): Promise<any> {
...
+          projectName = projectDeps.depGraph.rootPkg.name;
+        }
+
+        if (projectDeps.depTree) {
+          debug(`Processing ${projectDeps.depTree.name}...`);
+          maybePrintDepTree(options, projectDeps.depTree);
+          projectName = projectDeps.depTree.name;
+        }
+
         const tFile = projectDeps.targetFile || targetFile;
```
lili2311, line 215: maybe something other than `tFile` I know it was there before but it is hard on the eyes compare to other var names

src/lib/monitor/index.ts
```diff
@@ -111,11 +123,10 @@ export async function monitor(
     }
   }
 
-  return monitorDepTree(
+  return await monitorDepTree(
```
lili2311, line 126: does this await not have any strange consequences at call site? Surely call site already awaits this?

src/lib/monitor/index.ts
```diff
@@ -135,43 +145,50 @@ async function monitorDepTree(
...
 
-  let pkg = scannedProject.depTree;
+  let depTree = scannedProject.depTree;
+
+  if (!depTree) {
+    debug(
+      'scannedProject is missing depGraph or depTree, cannot run test/monitor',
+    );
+    throw new FailedToRunTestError(
+      'scannedProject is missing depGraph or depTree, cannot run test/monitor',
```
lili2311, line 155: same here, this is user facing but using internal language only we will understand

src/lib/monitor/index.ts
```diff
@@ -193,27 +210,39 @@ async function monitorDepTree(
 
   // TODO(kyegupov): async/await
   return new Promise((resolve, reject) => {
+    if (!depTree) {
+      debug(
+        'scannedProject is missing depGraph or depTree, cannot run test/monitor',
+      );
+      return reject(
+        new FailedToRunTestError(
+          'scannedProject is missing depGraph or depTree, cannot run test/monitor',
```
lili2311, line 219: same here :)

src/lib/print-deps.ts
lili2311: not a fan of two different formats from the same command, this is confusing

src/cli/commands/monitor/index.ts
anthogez: Will create a separated minor PR where I can fix your concern

src/lib/monitor/index.ts
anthogez: It's required and will not have any strange consequence ^^

src/cli/commands/monitor/index.ts
anthogez: I created this for snyk monitor
`Your monitor request could not be completed, contact support`
and this for snyk test
`Your test request could not be completed, contact support` :)

src/lib/print-deps.ts
orsagie:
```suggestion
if (count too big) {
  console.log(JSON.stringify(depGraph.toJSON(), null, 2));
  debug (pretty tree print notSupported for big gradle graphs);
} else {
  convertToTree
  maybePrintDepTree
}```

Comment (anthogez) [reply src/lib/print-deps.ts:11]: This solution that we discussed is good but this convertion and re-usage is not so straightforward 

`Argument of type '{ [x: string]: legacy.DepTree; }' is not assignable to parameter of type 'DepDict'.
  Index signatures are incompatible.`

Comment (orsagie) [reply src/lib/print-deps.ts:11]: `maybePrintDeps` takes a depTree. No need to call printDeps. 

For the debug message, `--print-deps option not yet supported for large projects. Displaying graph json output instead`

Comment (dkontorovskyy) [reply src/lib/print-deps.ts:11]: Yeap, we cannot break this contract! Clients sometimes depend on the output of `--print-deps` and `--json`. The crucial part of this PR and should be solved before merging

File: src/cli/commands/monitor/index.ts (hunk 16 above)
Comment (dkontorovskyy) [reply src/cli/commands/monitor/index.ts:215]: Please do it here.

File: src/lib/monitor/index.ts [hunk 20]
```diff
@@ -135,43 +145,50 @@ async function monitorDepTree(
...
+    .concat(pluckPolicies(depTree))
     .filter(Boolean);
   // docker doesn't have a policy as it can be run from anywhere
   if (!meta.isDocker || !policyLocations.length) {
     await snyk.policy.create();
   }
   const policy = await snyk.policy.load(policyLocations, { loose: true });
 
-  const target = await projectMetadata.getInfo(scannedProject, pkg, meta);
+  const target = await projectMetadata.getInfo(scannedProject, meta, depTree);
```
dkontorovskyy, line 185: why order of the variables changed?

anthogez: we moved optional depTree to the end, beyond that scannedProject and meta where 

not being used here:
https://github.com/snyk/snyk/pull/1165/commits/1f96517856e417f46c4852f4de37afbc2c0dc786#diff-c1de9d1e264d10b945610b8d84988eceL12

as you can see pointless

https://github.com/snyk/snyk/pull/1165/commits/1f96517856e417f46c4852f4de37afbc2c0dc786#diff-a376bec60e3deb50aaa63448c811c0dcL57

src/cli/commands/monitor/index.ts
anthogez: This pr already does too much.

src/lib/print-deps.ts
anthogez: We solved this issue, we are not breaking any contract.

anthogez: Just to be clear, this is not the crucial part of this PR, this is nothing related to the rest of the work that was done here.
And.. We solved this issue we are not breaking any contract.

src/cli/commands/monitor/index.ts
dkontorovskyy: Renaming one variable won't do harm :)

PR #1173, opened 2020-06-10
feat: support custom token from docker

RotemS:
- [ ] Ready for review

#### What does this PR do?

Draft PR for using a custom auth token (jwt) for docker.


src/lib/snyk-test/run-test.ts
```diff
@@ -531,12 +533,19 @@ async function assembleRemotePayloads(root, options): Promise<Payload[]> {
...
     },
   ];
 }
 
+function getAuthHeader() {
+  if (dockerIdExists()) {
+    return 'bearer ' + config.dockerSnykID;
+  }
+  return 'token ' + snyk.api;
+}
```
jurglic, line 547: The logic will likely need to be opposite - if api token is present use that one (normal auth), but if not, check if dockerJwt is present and use that one...
p.s. I like the extracted function 👍

RotemS: Yep realised that as well, will make adjustments today based on the new plans with the flag.

PR #1175, opened 2020-06-11
Show message when using a custom API endpoint and validate its value

JackuB:
Adds a info log that `endpoint` is set in config, and adds a URL validation.

This fixes 2 use cases:

- users having `endpoint` in snyk config, without even realising it's there: CLI now announces it's trying to connect somewhere non-standard
- user having invalid `endpoint` in snyk config, resulting in `$ snyk auth` URL being `null//null/login?token=<token>`

https://snyksec.atlassian.net/browse/HAMMER-40

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1,4 +1,5 @@
 import * as snykConfig from 'snyk-config';
+import { InvalidEndpointConfigError } from './errors/invalid-endpoint-config-error';
 import { config as userConfig } from './user-config';
 import * as url from 'url';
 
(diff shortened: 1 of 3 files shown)


src/lib/config.ts
```diff
@@ -21,9 +22,15 @@ const config = (snykConfig.loadConfig(
...
 ) as unknown) as Config;
 
-// allow user config override of the api end point
+// allow user config override of the API endpoint
 const endpoint = userConfig.get('endpoint');
-if (endpoint) {
+if (endpoint && endpoint !== config.API) {
+  const parsedEndpoint = url.parse(endpoint);
+  // Endpoint option must be a valid URL including protocol
+  if (!parsedEndpoint || !parsedEndpoint.protocol || !parsedEndpoint.host) {
```
maxjeffos, line 30: The endpoint needs to have a path of `/api` (or apparently `/api/v1`) - should we check for that here @JackuB ?

src/lib/errors/invalid-endpoint-config-error.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomError } from './custom-error';
+
+export class InvalidEndpointConfigError extends CustomError {
+  private static ERROR_MESSAGE =
+    "Invalid 'endpoint' config option. Endpoint must be a full and valid URL including protocol";
```
maxjeffos, line 5: Same idea here - should we mention the URL needs to have path `/api` or `/api/v1`?

src/lib/config.ts
JackuB: Disagree - you don't _need_ the `api` in the path. It just happens that our public API lives on that path. CLI will call whatever "endpoint" (not "host"!) that is specified.

src/lib/errors/invalid-endpoint-config-error.ts
JackuB: Updated the help text with `(tip: it should contain path to /api)` 👍

src/lib/config.ts
maxjeffos: Cool. It's just that I don't see this documented anywhere that's particularly obvious so it's good to surface it.

PR #1176, opened 2020-06-11
fix: prune depGraph paths accuracy

anthogez:
- [x] Ready for review

#### What does this PR do?
The goal of this pr is to remove the previously logic that did not make much sense,
and makes the way that we prune depGraphs more accurate

### DIFF
--- a/src/lib/prune.ts
+++ b/src/lib/prune.ts
@@ -4,6 +4,7 @@ import { DepGraph, legacy } from '@snyk/dep-graph';
 import { DepTree } from './types';
 import * as config from './config';
 import { TooManyVulnPaths } from './errors';
+import * as analytics from '../lib/analytics';
 import { SupportedPackageManagers } from './package-managers';
 
 const debug = _debug('snyk:prune');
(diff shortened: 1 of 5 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -452,7 +452,7 @@ async function assembleLocalPayloads(
           const prePruneDepCount = countPathsToGraphRoot(depGraph);
           debug('pre prunedPathsCount: ' + prePruneDepCount);
 
-          depGraph = await pruneGraph(depGraph, packageManager);
+          depGraph = await pruneGraph(depGraph, true, packageManager);
```
anthogez, line 455: This will be like this while we do not complete delete the `--prune-repeated-subdependencies` command

src/lib/prune.ts
```diff
@@ -18,27 +18,28 @@ export function countPathsToGraphRoot(graph: DepGraph): number {
 
 export async function pruneGraph(
   depGraph: DepGraph,
+  pruneIsRequired: boolean = false,
```
darscan, line 21: optional arguments should come last

src/lib/prune.ts
```diff
@@ -19,26 +20,28 @@ export function countPathsToGraphRoot(graph: DepGraph): number {
...
-    debug('prunedPathsCount: ' + count);
-
-    if (count < threshold) {
-      return prunedGraph;
+    const postPrunePathsCount = countPathsToGraphRoot(prunedGraph);
+    analytics.add('postPrunePathsCount', postPrunePathsCount);
+    debug('postPrunePathsCount' + postPrunePathsCount);
+    if (postPrunePathsCount > config.MAX_PATH_COUNT) {
+      debug('Too many vulnerable paths to process the project');
+      throw new TooManyVulnPaths();
```
JackuB, line 42: So before it failed pruning on graph being too big to process… and then returned you the unpruned graph that failed on some other place (in registry?)

anthogez: Exactly 😄

anthogez:
> A bit surprised it didn't break any tests 🤔 do you think it's worth adding something?

On this case no, but we will be monitoring the pre and post pruned paths to see how effectively we will be doing now :)

src/lib/prune.ts
anthogez: It was not failing on pruning, it simple skipped the graph pruned if it was too big and returned the original depGraph that was even more big and afterwards it failed on internal snyk service, which is definitely wrong

PR #1179, opened 2020-06-12
fix: make container commands use always static scan

agranado2k:
Continue the efforts to have "snyk container test" or "snyk container monitor" commands that doesn't require Docker engine running. In this step we make these commands always use the static scan. Although it's not relevant to all users, more flexibility here is both more technically correct, future proofs our technology, and is important to several separate strategic partners.

- [x] Ready for review

#### What does this PR do?
It makes the command `snyk container` always use the static scan.

The result for `snyk container test <image>...` has to be equal `snyk test --docker --experimental <image>...`

Run the following command and compare the output. It has to me the same:
 - `snyk container test <image>...`
 - `snyk test --docker --experimental <image>...`

 - [RUN-986](https://snyksec.atlassian.net/browse/RUN-986)
 - [RUN-1002](https://snyksec.atlassian.net/browse/RUN-1002)

#### Screenshots
Screenshot 2020-06-12 at 12 52 15

Screenshot 2020-06-12 at 12 52 21

### DIFF
--- a/src/cli/modes.ts
+++ b/src/cli/modes.ts
@@ -11,6 +11,7 @@ const modes: Record<string, ModeData> = {
     allowedCommands: ['test', 'monitor'],
     config: (args): [] => {
       args['docker'] = true;
+      args['experimental'] = true;
 
       return args;
     },
--- a/test/args.test.ts
+++ b/test/args.test.ts
@@ -160,6 +160,7 @@ test('test command line "container test"', (t) => {
   ];
   const result = args(cliArgs);
   t.ok(result.options.docker);
+  t.ok(result.options.experimental);
   t.end();
 });
 
@@ -172,6 +173,7 @@ test('test command line "container monitor"', (t) => {
   ];
   const result = args(cliArgs);
   t.ok(result.options.docker);
+  t.ok(result.options.experimental);
   t.end();
 });
 
@@ -184,6 +186,7 @@ test('test command line "container protect"', (t) => {
   ];
   const result = args(cliArgs);
   t.notOk(result.options.docker);
+  t.notOk(result.options.experimental);
   t.end();
 });
 
--- a/test/modes.test.ts
+++ b/test/modes.test.ts
@@ -135,6 +135,7 @@ test('when is a valid mode', (c) => {
         const expectedArgs = {
           _: [],
           docker: true,
+          experimental: true,
           'package-manager': 'pip',
         };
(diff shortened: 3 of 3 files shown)


agranado2k:
> We've reviewed this with @maxjeffos and we have few questions as we haven't seen these modes before.
> Do we want to keep the `--experimental` flag around? This make it seem like we are setting this flag as default? And is the `--no-experimental` supported? :) and should this mode be visible in the `snyk help`? Currently isn't?

Hi @JackuB and @maxjeffos! Good questions!

This module was built weeks ago, you can see more details in this [PR](https://github.com/snyk/snyk/pull/1124).
The idea is to create command allies for some options. For instance, in the PR that mentioned above, we've created the command `snyk container test <image>` that is an alias for   `snyk test --docker <image>. This is the first step to create a "snyk container test" or "snyk container monitor" commands that doesn't require Docker engine running. Although it's not relevant to all users, more flexibility here is both more technically correct, future proofs our technology, and is important to several separate strategic partners.

Also in this [PR](https://github.com/snyk/snyk/pull/1136) we've exposed as help specific for this command `snyk container`.

This current PR it'll make the options `--experimental` as default for `snyk container`. It's because all the containers scan will be static scan rather than dynamic scan, as requested by Docker itself.

If you need a better explanation I can make a call with both of you! :)

maxjeffos:
@agranado2k it looks like the modes thing - i.e. `snyk container...` is not yet official because it is not documented (which is fine). In that case it seems fine that you're always enabling the `--experimental` option. When the `snyk container` mode becomes official, is the plan to remove the hard coding of the `--experimental` option? (This is not blocking, we are just wondering what the plan is?)

agranado2k:
> @agranado2k it looks like the modes thing - i.e. `snyk container...` is not yet official because it is not documented (which is fine). In that case it seems fine that you're always enabling the `--experimental` option. When the `snyk container` mode becomes official, is the plan to remove the hard coding of the `--experimental` option? (This is not blocking, we are just wondering what the plan is?)

Hi @maxjeffos! That's a good question and I'm not sure about the answer. Actually you have two different things here:
 - the command `snyk container`. It's not official yet. As you said there is no documentation. You're using just internally. 
- the option `--experimental` it's to activate the static analysis, that I believe will become the default (maybe the only) to scan. However you need to convert with the problem to make sure what the plan is. @garethr is the best to talk about both points.

PR #1222, opened 2020-06-18
fix: allows scanning even if a single proj import fails

anthogez:
- [x] Ready for review

#### What does this PR do?

The scenario of this pr is the following one:

Imagine an user tries to scan a mono-repo and this repository possess 200 sub-projects.

What we normally do? If a single project import is unsuccessful we just abort everything without push forward the others that could be imported without issues.

So now what's gonna happen? If 1 or 5 projects of 200 fails, they will be just skipped and the import will continue, and this will be really helpful to import large scale-projects.

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -182,65 +182,71 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
       // Post the project dependencies to the Registry
       for (const projectDeps of perProjectResult.scannedProjects) {
-        if (!projectDeps.depGraph && !projectDeps.depTree) {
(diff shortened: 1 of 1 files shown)


JackuB:
Thinking about this use case, would it make sense to continue pushing projects, but return a non-zero exit code in the end?

anthogez:
> Thinking about this use case, would it make sense to continue pushing projects, but return a non-zero exit code in the end?

It already returns an error at the end :)

anthogez:
> Thinking about this use case, would it make sense to continue pushing projects, but return a non-zero exit code in the end?

It's already throwing an error :)

PR #1223, opened 2020-06-18
feat: unify policy handling & plucking

lili2311:
- [x] Ready for review

#### What does this PR do?

- Unify how we find & load policies for all the 4 places where we need to do the exact same thing.
- add some types
- update pluck policies to always return an string or string array

There is a policy handling but with `--all-project` , `gradle sub-projects` and `--file` where we look for policy in the root of where the command is run and not next to the manifest, which can be exposed in a test and fixed in a PR after this initial refactor so that the change is made in one place not 4.

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -37,7 +37,6 @@ import {
   TestOptions,
   SupportedProjectTypes,
   PolicyOptions,
-  PackageJson,
 } from '../types';
 import { pruneGraph } from '../prune';
 import { getDepsFromPlugin } from '../plugins/get-deps-from-plugin';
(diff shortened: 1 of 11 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -466,7 +454,7 @@ async function assembleLocalPayloads(
         targetFileRelativePath: `${targetFileRelativePath}`, // Forcing string
         projectNameOverride: options.projectName,
         originalProjectName,
-        policy: policy && policy.toString(),
+        policy: policy ? policy.toString() : undefined,
```
lili2311, line 457: make it the same as other instances

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -1531,8 +1531,7 @@ test('`monitor foo:latest --docker` doesnt send policy from cwd', async (t) => {
     'calls docker plugin with expected arguments',
   );
 
-  const emptyPolicy = await snykPolicy.create();
-  t.deepEqual(req.body.policy, emptyPolicy.toString(), 'empty policy is sent');
+  t.deepEqual(req.body.policy, undefined, 'no policy is sent');
```
lili2311, line 1534: make monitor & test behave the same way (preferring test)

PR #1224, opened 2020-06-19
fix: policies to be correctly picked up from nested scans

lili2311:
- [x] Ready for review

#### What does this PR do?
Show that when triggering a test with `--file` or `--all-project` the policy is not discovered as we expect to find it in the `root` which is where the command is run and it should instead be where the manifest was discovered.

While working on yarn workspaces feature test noticed the policy from the workspaces were not being used or sent.

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -109,7 +109,6 @@ async function runTest(
         _.get(payload, 'body.originalProjectName');
       const foundProjectCount = _.get(payload, 'body.foundProjectCount');
       const displayTargetFile = _.get(payload, 'body.displayTargetFile');
-
       let dockerfilePackages;
       if (
         payload.body &&
@@ -403,6 +402,22 @@ async function assembleLocalPayloads(
         }
       }
 
+      // todo: normalize what target file gets used across plugins and functions
+      const targetFile =
+        scannedProject.targetFile || deps.plugin.targetFile || options.file;
+
+      // Forcing options.path to be a string as pathUtil requires is to be stringified
+      const targetFileRelativePath = targetFile
+        ? pathUtil.join(pathUtil.resolve(`${options.path || root}`), targetFile)
+        : '';
+
+      let targetFileDir;
+
+      if (targetFileRelativePath) {
+        const { dir } = path.parse(targetFileRelativePath);
+        targetFileDir = dir;
+      }
+
       const policy = await findAndLoadPolicy(
         root,
         options.docker ? 'docker' : packageManager!,
@@ -410,6 +425,7 @@ async function assembleLocalPayloads(
         // TODO: fix this and send only send when we used resolve-deps for node
         // it should be a ExpandedPkgTree type instead
         pkg,
+        targetFileDir,
       );
 
       analytics.add('packageManager', packageManager);
@@ -421,15 +437,6 @@ async function assembleLocalPayloads(
         const depTree = pkg as DepTree;
         addPackageAnalytics(depTree.name!, depTree.version!);
       }
-
-      // todo: normalize what target file gets used across plugins and functions
-      const targetFile =
-        scannedProject.targetFile || deps.plugin.targetFile || options.file;
-
-      // Forcing options.path to be a string as pathUtil requires is to be stringified
-      const targetFileRelativePath = targetFile
-        ? pathUtil.join(pathUtil.resolve(`${options.path}`), targetFile)
-        : '';
 
       let target: GitTarget | ContainerTarget | null;
       if (scannedProject.depGraph) {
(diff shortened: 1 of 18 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -403,13 +402,30 @@ async function assembleLocalPayloads(
         }
       }
 
+      // todo: normalize what target file gets used across plugins and functions
+      const targetFile =
+        scannedProject.targetFile || deps.plugin.targetFile || options.file;
+
+      // Forcing options.path to be a string as pathUtil requires is to be stringified
+      const targetFileRelativePath = targetFile
+        ? pathUtil.join(pathUtil.resolve(`${options.path || root}`), targetFile)
```
lili2311, line 411: there is no `options.path` when coming via `snyk protect` flow

package.json
```diff
@@ -27,7 +27,7 @@
     "format": "prettier --write '{src,test,scripts}/**/*.{js,ts}'",
     "prepare": "npm run build",
     "test:common": "npm run check-tests && npm run lint && node --require ts-node/register src/cli test --org=snyk",
-    "test:acceptance": "tap test/acceptance/**/*.test.* test/acceptance/*.test.* -Rspec --timeout=300 --node-arg=-r --node-arg=ts-node/register",
+    "test:acceptance": "tap test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts -Rspec --timeout=300 --node-arg=-r --node-arg=ts-node/register",
```
anthogez, line 30: this ☝️ is current skipping all the acceptance tests

test/acceptance/workspaces/gradle-multi-project/.gradle/buildOutputCleanup/cache.properties
```diff
@@ -0,0 +1,2 @@
+#Fri Jun 19 12:31:38 BST 2020
```
anthogez, line 1: Could you gentle delete all files available in .gradle folder? e.g. cache.properties bin etc ☝️ thanks!

test/acceptance/cli-test/cli-test.gradle.spec.ts
```diff
@@ -112,6 +112,50 @@ export const GradleTests: AcceptanceTests = {
...
+      const requests = params.server.popRequests(2);
+      let policyCount = 0;
+      requests.forEach((req) => {
+        if (
+          req.body.displayTargetFile.endsWith('gradle-multi-project/subproj')
+        ) {
+          // TODO: this should return 1 policy when fixed
+          // uncomment then
+          // t.match(
+          //   req.body.policy,
```
anthogez, line 143: To make this work is required the changes on the snyk-gradle-plugin targetFile that we talked about, few days ago?

anthogez, line 143: What is current missing to make this work? Something that we already discussed about on snyk-gradle-plugin? Thanks!

lili2311: yes we must fix how targetFile is sent there to not just be `build.gradle` but relative path from root. currently all the scans look like they are from the same folder and only the projectName indicates otherwise.

lili2311: Please see the thread in Slack

test/acceptance/workspaces/gradle-multi-project/.gradle/buildOutputCleanup/cache.properties
lili2311: done

erwinpleo:
Eh, this change constitutes in fact a breaking change, which has caused us quite some time loss trying to figure out why Snyk suddenly ignored the `.snyk` file.

I guess we were unknowingly relying on a bug, but that is no excuse in my opinion. Changing which locations are accepted as a location for config files is a breaking change, so rather than releasing as a *bugfix* release, v1.138.1 should've been v2.0.0, a *major* release.

Also, why are the release notes so sparse? It's hardly possible to figure out what has actually changed, and `calculate the policy folder from targetFile` is *not* very descriptive unless you're very deep into Snyk.

lili2311:
👋 @erwinpleo  Apologies for this inconvenience, you are right the notes could be better I have gone ahead and updated these with more information. I have written up a detailed response here: https://github.com/snyk/snyk/issues/1237#issuecomment-651378812 I hope this helps to clarify the reason for the change and provides some advice. If you have further comments It may be beneficial to keep it on that thread in one place.

lili2311:
@erwinpleo just realised that perhaps you can achieve exactly what you had if you use `--policy-path=path/to/.snyk`, please let me know if this helps

PR #1233, opened 2020-06-24
feat: detect other files & suggest to use --all-projects

lili2311:
- [x] Ready for review

#### What does this PR do?
Suggest scanning with `--all-projects` when we detect other supported manifests up to 4 sub-directories deep. For both `test` and `monitor` to help discover the new feature.

#### Screenshots

Screen Shot 2020-06-24 at 22 33 55
Screen Shot 2020-06-24 at 22 34 06


src/cli/commands/monitor/formatters/format-monitor-response.ts
```diff
@@ -34,6 +35,12 @@ export function formatMonitorOutput(
             'use --all-sub-projects flag to scan all sub-projects.\n\n',
         )
       : '') +
+    (advertiseAllProjectsCount && foundProjectCount
+      ? chalk.bold.white(
+          `Tip: Detected multiple supported manifests (${foundProjectCount}), ` +
+            'use --all-projects --detection-depth=4 to scan all of them at once.\n\n',
```
gitphill, line 41: `--detection-depth=4` is this now the default?

src/lib/plugins/get-extra-project-count.ts
```diff
@@ -0,0 +1,29 @@
...
+    inspectResult.plugin.meta &&
+    inspectResult.plugin.meta.allSubProjectNames &&
+    inspectResult.plugin.meta.allSubProjectNames.length > 1
+  ) {
+    return inspectResult.plugin.meta.allSubProjectNames.length;
+  }
+  try {
+    const extraTargetFiles = await find(root, [], AUTO_DETECTABLE_FILES, 4);
+    const foundProjectsCount =
+      extraTargetFiles.length > 1 ? extraTargetFiles.length - 1 : undefined;
```
gitphill, line 24: is it more interesting to use which files (their path and name) we found? rather than just a count?

src/cli/commands/monitor/formatters/format-monitor-response.ts
lili2311: our default is 2 which basically finds not much, 4 seemed to find much more, also it looks more impactful rather than suggesting to scan for the sake of 1 more file. What are thoughts on looking deeper by default?

src/lib/plugins/get-extra-project-count.ts
lili2311: Interesting, how would you stop this list from getting too big? `file-a, file-b and 24 more...` or something?

gitphill: yea maybe truncate to a reasonable limit

lili2311: so changing both the message for gradle and this one? They both use the same ish wording & show how many more you could scan with an different option? This function handles both discovered gradle sub-proejcts (where we have the sub-project names) and now also extra manifests that you could scan with `--all-projects`

PR #1242, opened 2020-06-26
feat: send on original gradle project name

lili2311:
- [ ] Ready for review

#### What does this PR do?
- a while ago we started sending from the plugin original Gradle Project name, send it on so we can display it in the UI as meta

https://github.com/snyk/snyk-gradle-plugin/pull/95


src/lib/monitor/index.ts
```diff
@@ -272,6 +272,9 @@ async function monitorDepTree(
             versionBuildInfo: JSON.stringify(
               scannedProject.meta?.versionBuildInfo,
             ),
+            gradleProjectName: scannedProject.meta?.gradleProjectName
+              ? scannedProject.meta?.gradleProjectName
+              : undefined,
```
anthogez, line 277: I would suggest to delete `undefined` because when you are using `?` you are already handling the case 
Screen Shot 2020-06-30 at 15 02 52

src/lib/monitor/index.ts
```diff
@@ -393,6 +396,9 @@ export async function monitorDepGraph(
...
```
anthogez, line 401: same that I mentioned ☝️

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -927,13 +927,18 @@ test('`monitor gradle-app with dep-graph`', async (t) => {
...
   const plugin = {
     async inspect() {
       return {
-        plugin: { name: 'gradle' },
+        plugin: {
+          name: 'gradle',
+        },
         package: {},
+        meta: {
+          versionBuildInfo: { java: '8' },
```
anthogez, line 939: would be nice if you could add also the a gradle version e.g. `6.4` so we can cover not only the gradleProjectName but also the current gradleVersion (I understand that is not the scope of this pr)

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -952,6 +957,16 @@ test('`monitor gradle-app --all-sub-projects`', async (t) => {
...
     'sends version number',
   );
+  t.deepEqual(
+    req.body.meta.gradleProjectName,
+    'original-name',
+    'gradleProjectName passed',
+  );
+  t.deepEqual(
+    req.body.meta.versionBuildInfo,
+    '{"java":"8"}',
```
anthogez, line 967: Same that I mentioned ☝️

PR #1247, opened 2020-06-29
fix: allow IPv6 if supported

aviadhahami:
- [ ] Ready for review

#### What does this PR do?
Fixing a bug where clients were forced to IPv6 _iff_ their DNS clients could resolve an IPv6 address;
This is problematic as a client (or the client's ISP) may support IPv6 address resolving(as good as the DNS software client), but it's ISP won't support IPv6 connections; Given such case - `snyk` will fail w/ connection error

* Previous fix attempt: [BST-1107](https://snyksec.atlassian.net/browse/BST-1107)
* Residual bug forcing clients to IPv6 regardless of their support [BST-1606](https://snyksec.atlassian.net/browse/BST-1606)


src/cli/commands/auth/index.ts
```diff
@@ -21,8 +18,6 @@ import { getQueryParamsAsString } from '../../../lib/query-strings';
 
 export = auth;
 
-const dnsLookupPromise = util.promisify(dns.lookup);
```
JackuB, line 24: Why were we using these in the first place? And why don't we need to use it for other places where we make API requests? 🤔

aviadhahami: Usually we(and every high-level coder) "leave" these to the OS to decide;
The issue was (that resulted in this solution - which is buggy, see description) that clients had the browser communicate via IPv6 (IP1) but their client would connect w/ IPv4 (IP2). When registry auth handler compared the IPs, they are obviously not equal thus will fail CLI clients on the auth process;

aviadhahami: CC: @dkontorovskyy
> may elaborate more on the `dns.lookup` solution

src/cli/commands/auth/index.ts
```diff
@@ -179,16 +174,15 @@ function errorForFailedAuthAttempt(res, body) {
...
 
 async function getIpFamily(): Promise<6 | undefined> {
-  const IPv6Family = 6;
-
+  const family = 6;
   try {
-    const { address } = await dnsLookupPromise(apiUrl.hostname!, {
-      family: IPv6Family,
+    // Dispatch a FORCED IPv6 request to test client's ISP and network capability
+    await request({
```
gitphill, line 180: is this a POST (because GET is not a valid request)?

aviadhahami: It may be a post too (and I'll change it), but since I force ipv6 - if the client doesn't support it it won't be able to dispatch the request

PR #1252, opened 2020-06-29
feat: add support for yarn workspaces pattern in workspaces.packages

mhassan1:
fixes https://github.com/snyk/snyk/issues/1251

- [ ] Ready for review

#### What does this PR do?
Fixes https://github.com/snyk/snyk/issues/1251 by adding support for the alternative Yarn Workspaces config structure:
```
{
  "name": "my-workspace",
  "private": true,
  "workspaces": {
    "packages": [ "packages/*" ]
  }
}
```

There's only one changed line :)

Create a Yarn Workspaces setup using the alternate configuration

Snyk is awesome

https://github.com/snyk/snyk/issues/1251

### DIFF
--- a/src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
+++ b/src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
@@ -58,7 +58,7 @@ export async function processYarnWorkspaces(
       ...getWorkspacesMap(packageJson),
     };
     for (const workspaceRoot of Object.keys(yarnWorkspacesMap)) {
-      const workspaces = yarnWorkspacesMap[workspaceRoot].workspaces || [];
+      const workspaces = yarnWorkspacesMap[workspaceRoot].workspaces?.packages || yarnWorkspacesMap[workspaceRoot].workspaces || [];
       const match = workspaces
         .map((pattern) => {
           return packageJsonFileName.includes(pattern.replace(/\*/, ''));
--- a/test/acceptance/workspaces/yarn-workspace-out-of-sync/package.json
+++ b/test/acceptance/workspaces/yarn-workspace-out-of-sync/package.json
@@ -1,8 +1,10 @@
 {
   "private": true,
-  "workspaces": [
-    "packages/*"
-  ],
+  "workspaces": {
+    "packages": [
+      "packages/*"
+    ]
+  },
   "resolutions": {
     "node-fetch": "^2.3.0"
   },


src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
```diff
@@ -58,7 +58,7 @@ export async function processYarnWorkspaces(
       ...getWorkspacesMap(packageJson),
     };
     for (const workspaceRoot of Object.keys(yarnWorkspacesMap)) {
-      const workspaces = yarnWorkspacesMap[workspaceRoot].workspaces || [];
+      const workspaces = yarnWorkspacesMap[workspaceRoot].workspaces?.packages || yarnWorkspacesMap[workspaceRoot].workspaces || [];
```
lili2311, line 61: 👋 Thank you for the contribution, however the change would need to be in the parser that parses the workspaces: https://github.com/snyk/nodejs-lockfile-parser/pull/38 here is the original PR adding support. 

We can certainly extend this to also support the `workspaces.packages` if you are keen to add it in with a test please do! Just let me know so I don't do the same

mhassan1: Is `src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts` not the correct parser?

I added a test by modifying `test/acceptance/workspaces/yarn-workspace-out-of-sync/package.json`. I know it's re-purposing that acceptance test, but I wasn't sure if you would want another large acceptance test just for `workspaces.packages`.

lili2311: By the time we are here we have extracted the `workspaces` from the package.json. See line https://github.com/snyk/snyk/pull/1252/files#diff-2bee611aba1b36e848d4beddf5aa8346R141
Here we try to match any package.json that matches one of the workspace definitions. 

A fix & a dedicated test in the original repo would be enough https://github.com/snyk/snyk/pull/1252/files#diff-2bee611aba1b36e848d4beddf5aa8346R141 as the rest of the code in here should just work after the correct workspaces definitions are extracted from the package.json.

mhassan1: Now I get it! Thanks for the explanation. I've opened a PR there: https://github.com/snyk/nodejs-lockfile-parser/pull/76

PR #1256, opened 2020-07-01
chore: rename reachable paths field

muscar:
- [x] Ready for review

#### What does this PR do?

We renamed the field in the result returned by the server to avoid awkward looking code like `reachablePaths.reachablePaths`. This changes the type definition to use the new name.

### DIFF
--- a/src/cli/commands/constants.ts
+++ b/src/cli/commands/constants.ts
@@ -1,0 +1,12 @@
+// Separator used while displaying various paths (e.g. package paths, call
+// paths) to the user
+export const PATH_SEPARATOR = ' > ';
+
+// String used to signify hidden path elements e.g. for abbreviated paths
+export const PATH_HIDDEN_ELEMENTS = '...';
+
+// Number of function names to show in the beginning of an abbreviated call path
+export const CALL_PATH_LEADING_ELEMENTS = 2;
+
+// Number of function names to show in the end of an abbreviated call path
+export const CALL_PATH_TRAILING_ELEMENTS = 2;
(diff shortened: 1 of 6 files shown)


src/cli/commands/test/formatters/format-reachability.ts
```diff
@@ -99,9 +99,9 @@ export function formatReachablePaths(
 }
 
 export function formatReachablePath(path: CallPath): string {
-  const head = path.slice(0, LEADING_PATH_ELEMENTS).join('>');
+  const head = path.slice(0, LEADING_PATH_ELEMENTS).join(' > ');
   const tail = path
     .slice(path.length - TRAILING_PATH_ELEMENTS, path.length)
-    .join('>');
+    .join(' > ');
```
gitphill, line 105: no tests break? could create a const for ' > ' I've seen it through-out this code-base

muscar: This particular piece of code uses a closure to format the path, specifically to avoid breakages like these (the `template` argument to `formatReachablePaths`).

Re. the constant, sure, I can add one. Good point.

src/cli/commands/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+export const PATH_SEPARATOR = ' > ';
+
+// String used to signify hidden path elements e.g. for abbreviated paths
+export const PATH_HIDDEN_ELEMENTS = '...';
+
+// Number of function names to show in the beginning of an abbreviated call path
+export const CALL_PATH_LEADING_ELEMENTS = 2;
+
+// Number of function names to show in the end of an abbreviated call path
+export const CALL_PATH_TRAILING_ELEMENTS = 2;
```
gitphill, line 12: 👍

PR #1263, opened 2020-07-09
chore: add test to check if new version of snyk is not broken

dkontorovskyy:
- [ ] Ready for review

#### What does this PR do?
Add test to check if new version of snyk is not broken

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,6 @@ commands:
   install_deps:
     description: Install dependencies
     steps:
-      - checkout
       - run:
           name: Use snyk-main npmjs user
           command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -159,6 +157,29 @@ jobs:
...
+          name: Install only production dependencies
+          command: npm install
+          environment:
+            NODE_ENV: production
+      - run:
+          name: Run auth
+          command: node ~/snyk/dist/cli/index.js auth $SNYK_API_KEY
+      - run:
+          name: Regression test
+          command: node ~/snyk/dist/cli/index.js test test/acceptance/workspaces/npm-package-no-vulns
```
gitphill, line 181: is it possible to check more commands? monitor? wizard? etc

dkontorovskyy: If the smth is missing, `test` is enough to know that.

aviadhahami: actually, can we parallel `test` and `monitor`? 
@gitphill `wizard` will require us to mock user interaction, no?

PR #1272, opened 2020-07-14
fix: increase tool flexibility for container static analysis

agranado2k:
Increase the tool flexibility allowing the user choose a location on disk with the best storage characteristics for storage images during the static analysis process.

- [x] Ready for review

#### What does this PR do?
It captures the option `imageSavePath` from env var or configured via `snyk config` to allow the `container` static analysis flow to store the downloaded image in a custom path given by the user.

Take a look at `args.ts` file. 

1. Run the command `snyk container test nginx:latest` to check the tool behaviour without the new option.
2. `SNYK_IMAGE_SAVE_PATH=./your/custom/path snyk container test nginx:latest` and check if the result was the same as the previous command. Also the path `./your/custom/path` would be created and it has to be empty.

 - [RUN-582](https://snyksec.atlassian.net/browse/RUN-582)

### DIFF
--- a/src/cli/args.ts
+++ b/src/cli/args.ts
@@ -7,6 +7,7 @@ import {
   SupportedCliCommands,
   SupportedUserReachableFacingCliArgs,
 } from '../lib/types';
+import { getContainerImageSavePath } from '../lib/container';
 
 export declare interface Global extends NodeJS.Global {
   ignoreUnknownCA: boolean;
@@ -159,6 +160,13 @@ export function args(rawArgv: string[]): Args {
     // if we failed to find a command, then default to an error
     method = require('../lib/errors/legacy-errors');
     argv._.push(command);
+  }
+
+  // TODO: Once experimental flag became default this block should be
+  // moved to inside the parseModes function for container mode
+  const imageSavePath = getContainerImageSavePath();
+  if (imageSavePath) {
+    argv['imageSavePath'] = imageSavePath;
   }
 
   const commands: SupportedCliCommands[] = [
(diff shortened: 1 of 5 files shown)


src/cli/args.ts
```diff
@@ -161,6 +162,15 @@ export function args(rawArgv: string[]): Args {
     argv._.push(command);
   }
 
+  // TODO: Once experimental flag became default this block should be
+  // moved to inside the parseModes function for container mode
+  if (argv['docker'] && argv['experimental']) {
```
shaimendel, line 167: why do we need this condition? isn't it ok to always call `getContainerImageSavePath`? just so we don't need to change that once we remove the experimental flag

src/cli/args.ts
```diff
@@ -161,6 +162,15 @@ export function args(rawArgv: string[]): Args {
...
+    const imageSavePath = getContainerImageSavePath();
+    if (imageSavePath) {
+      argv['imageSavePath'] = imageSavePath;
```
shaimendel, line 170: same here maybe? is it ok to just write 
```
argv['imageSavePath'] = getContainerImageSavePath();
```
instead of all of this chunk?

src/cli/args.ts
agranado2k: I think I was too much defensively here. But it's true. I can always call `getContainerImageSavePath`.

src/cli/args.ts
agranado2k: The problem here is `argv['imageSavePath']` doesn't accept `undefined`, because it's [`ArgsOptions`](https://github.com/snyk/snyk/blob/f93d68863488241d23a084606ffc57913866b0c5/src/cli/args.ts#L55). Definitely I would not include this second `if` statement if I could not include it! :)

shaimendel:
why is it a fix and not a feat? :)

PR #1276, opened 2020-07-17
fix: improve reliability around call graph generation

muscar:
- [x] Ready for review

#### What does this PR do?

This is meant to improve the experience around reachable vulns by:
- giving control to the CLI on how long the call graph generator can run;
- running the user command (e.g. monitor, test) even if the call graph generation failed;
- allowing for better error messages is the call graph generation fails.

Merge after https://github.com/snyk/snyk-cli-interface/pull/36 and https://github.com/snyk/snyk-mvn-plugin/pull/73, and after updating the dependencies to the appropriate versions.

https://snyksec.atlassian.net/browse/FLOW-336

#### Screenshots

Running with a timeout:

step1

Running with a timeout and passing `-d`:

step2

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -90,6 +90,18 @@ Maven options:
   --scan-all-unmanaged
                        Auto detects maven jars and wars in given directory.
                        Individual testing can be done with --file=<jar-file-name>
+  --reachable
+                       (test and monitor commands only)
+                       Analyze your source code to find which vulnerable
+                       functions and packages are called.
+  --reachable-timeout=<number>
+                       (test and monitor commands only)
+                       The amount of time (in seconds) to wait for Snyk
+                       to gather reachability data. If it takes longer than
+                       the specified interval, Reachable Vulnerabilities are
+                       not reported. This does not affect regular test or
+                       monitor output.
+                       Defaults to waiting 300s (5 minutes).
 
 Gradle options:
   --sub-project=<string> (alias: --gradle-sub-project)
--- a/src/cli/args.ts
+++ b/src/cli/args.ts
@@ -36,6 +36,10 @@ const DEBUG_DEFAULT_NAMESPACES = [
   'snyk-mvn-plugin',
 ];
 
+// NOTE[muscar] This is accepted in seconds for UX reasons, the mavem plugin
+// turns it into milliseconds before calling the call graph generator
+const REACHABLE_VULNS_TIMEOUT = 5 * 60; // 5 min (in seconds)
+
 function dashToCamelCase(dash) {
   return dash.indexOf('-') < 0
     ? dash
@@ -192,11 +196,17 @@ export function args(rawArgv: string[]): Args {
     'all-projects',
     'yarn-workspaces',
     'detection-depth',
+    'reachable',
     'reachable-vulns',
+    'reachable-timeout',
+    'reachable-vulns-timeout',
   ];
   for (const dashedArg of argumentsToTransform) {
     if (argv[dashedArg]) {
       const camelCased = dashToCamelCase(dashedArg);
+      if (camelCased === dashedArg) {
+        continue;
+      }
       argv[camelCased] = argv[dashedArg];
       delete argv[dashedArg];
     }
@@ -218,10 +228,20 @@ export function args(rawArgv: string[]): Args {
     }
   }
 
+  if (
+    argv.reachableTimeout === undefined &&
+    argv.reachableVulnsTimeout === undefined
+  ) {
+    argv.reachableVulnsTimeout = REACHABLE_VULNS_TIMEOUT.toString();
+  }
+
   // Alias
   const aliases = {
     gradleSubProject: 'subProject',
     container: 'docker',
+    reachable: 'reachableVulns',
+    reachableTimeout: 'callGraphBuilderTimeout',
+    reachableVulnsTimeout: 'callGraphBuilderTimeout',
   };
   for (const argAlias in aliases) {
     if (argv[argAlias]) {
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -36,7 +36,11 @@ import { dropEmptyDeps } from './drop-empty-deps';
 import { pruneTree } from './prune-dep-tree';
 import { findAndLoadPolicy } from '../policy';
 import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
-import { CallGraph, ScannedProject } from '@snyk/cli-interface/legacy/common';
+import {
+  CallGraph,
+  CallGraphError,
+  ScannedProject,
+} from '@snyk/cli-interface/legacy/common';
 import { isGitTarget } from '../project-metadata/types';
 import { serializeCallGraphWithMetrics } from '../reachable-vulns';
 import {
@@ -46,6 +50,7 @@ import {
   getTargetFile,
 } from './utils';
 import { countPathsToGraphRoot } from '../utils';
+import * as alerts from '../alerts';
 
 const debug = Debug('snyk');
 
@@ -213,7 +218,17 @@ async function monitorDepTree(
   depTree = dropEmptyDeps(depTree);
 
   let callGraphPayload;
-  if (scannedProject.callGraph) {
+  if (options.reachableVulns && scannedProject.callGraph?.innerError) {
+    const err = scannedProject.callGraph as CallGraphError;
+    analytics.add('callGraphError', err.innerError.toString());
+    alerts.registerAlerts([
+      {
+        type: 'error',
+        name: 'missing-call-graph',
+        msg: err.message,
+      },
+    ]);
+  } else if (scannedProject.callGraph) {
     const { callGraph, nodeCount, edgeCount } = serializeCallGraphWithMetrics(
       scannedProject.callGraph,
     );
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -53,6 +53,8 @@ import { validateOptions } from '../options-validator';
 import { findAndLoadPolicy } from '../policy';
 import { assembleIacLocalPayloads, parseIacTestResult } from './run-iac-test';
 import { Payload, PayloadBody, DepTreeFromResolveDeps } from './types';
+import { CallGraphError } from '@snyk/cli-interface/legacy/common';
+import * as alerts from '../alerts';
 
 const debug = debugModule('snyk');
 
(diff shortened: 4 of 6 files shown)


src/lib/monitor/index.ts
```diff
@@ -213,7 +214,15 @@ async function monitorDepTree(
   depTree = dropEmptyDeps(depTree);
 
   let callGraphPayload;
-  if (scannedProject.callGraph) {
+  if (options.reachableVulns && (scannedProject.callGraph as CallGraphError).innerError) {
```
darmalovan, line 217: maybe can be shortened by `scannedProject.callGraph?.innerError` ?

src/lib/types.ts
```diff
@@ -91,6 +92,7 @@ export interface MonitorOptions {
   // Used with the Docker plugin only. Allows application scanning.
   'app-vulns'?: boolean;
   reachableVulns?: boolean;
+  callGraphBuilderTimeout?: number;
```
darmalovan, line 95: Nice!

darmalovan:
@muscar nice! how should the usage of the timeout from the user perspective should look like?

src/lib/types.ts
```diff
@@ -160,7 +162,8 @@ export type SupportedUserReachableFacingCliArgs =
   | 'yarn-workspaces'
   | 'detection-depth'
   | 'project-name'
-  | 'reachable-vulns';
+  | 'reachable-vulns'
+  | 'call-graph-builder-timeout';
```
darmalovan, line 166: I think it better not to expose call-graph to the user, can be `reachable-vulns-timeout` can use this to alias it
https://github.com/snyk/snyk/blob/f7c0abf36fff25a6984aa8263ec918cd57c0bacf/src/cli/args.ts#L213

darmalovan: If we are adding new command, can we add it to help text, together with --reachable-vulns?

src/lib/monitor/index.ts
```diff
@@ -213,7 +218,20 @@ async function monitorDepTree(
   depTree = dropEmptyDeps(depTree);
 
   let callGraphPayload;
-  if (scannedProject.callGraph) {
+  if (
+    options.reachableVulns &&
+    (scannedProject.callGraph as CallGraphError).innerError
```
darmalovan, line 223: Maybe add here for safety just `scannedProject.callGraph?.innerError`

src/lib/monitor/index.ts
```diff
@@ -46,6 +50,7 @@ import {
   getTargetFile,
 } from './utils';
 import { countPathsToGraphRoot } from '../utils';
+import alerts = require('../alerts');
```
darmalovan, line 53: Can we use the `from` syntax?

src/lib/snyk-test/run-test.ts
```diff
@@ -53,6 +53,8 @@ import { validateOptions } from '../options-validator';
 import { findAndLoadPolicy } from '../policy';
 import { assembleIacLocalPayloads, parseIacTestResult } from './run-iac-test';
 import { Payload, PayloadBody, DepTreeFromResolveDeps } from './types';
+import { CallGraphError } from '@snyk/cli-interface/legacy/common';
+import alerts = require('../alerts');
```
darmalovan, line 57: Can we use the `from` syntax?

darmalovan:
@muscar can you add screenshot on the pr description?

help/help.txt
```diff
@@ -90,6 +90,14 @@ Maven options:
   --scan-all-unmanaged
                        Auto detects maven jars and wars in given directory.
                        Individual testing can be done with --file=<jar-file-name>
+  --reachable-vulns    Try to analyze the application in order to finc which of
```
darmalovan, line 93: `finc` :)

help/help.txt
```diff
@@ -90,6 +90,14 @@ Maven options:
...
+                       the vulnerabe functions and packages are actually called
+                       by client code.
+  --reachable-vulns-timeout The amount of time (in seconds) to wait for the
+                       Maven plugin to discover the reachable vulns. If it
```
darmalovan, line 97: I think you can drop the drop `Maven plugin` - less implementation details :)

help/help.txt
```diff
@@ -90,6 +90,14 @@ Maven options:
...
+                       takes longer than the specified timeout, the plugin
+                       gives up, and no reachability info is reported.
+                       Defaults to waiting indefinitely.
```
darmalovan, line 100: Let's with this line? we gonna change it really soon

help/help.txt
muscar: Thanks!

src/lib/monitor/index.ts
```diff
@@ -213,7 +218,17 @@ async function monitorDepTree(
   depTree = dropEmptyDeps(depTree);
 
   let callGraphPayload;
-  if (scannedProject.callGraph) {
+  if (options.reachableVulns && scannedProject.callGraph?.innerError) {
+    const err = scannedProject.callGraph as CallGraphError;
```
darmalovan, line 222: I think the as ` as CallGraphError` is not needed now, but maybe help in reading?

help/help.txt
```diff
@@ -90,6 +90,13 @@ Maven options:
   --scan-all-unmanaged
                        Auto detects maven jars and wars in given directory.
                        Individual testing can be done with --file=<jar-file-name>
+  --reachable-vulns    Try to analyze the application in order to find which of
```
darmalovan, line 93: I see other flags have a new line after the flag

help/help.txt
```diff
@@ -90,6 +90,13 @@ Maven options:
...
+                       the vulnerabe functions and packages are actually called
+                       by client code.
+  --reachable-vulns-timeout The amount of time (in seconds) to wait for the
```
darmalovan, line 96: ` --reachable-vulns-timeout=<number>` ? (and new line)

help/help.txt
darmalovan: Also, can add like others
`(test and monitor command only)`

src/cli/args.ts
```diff
@@ -218,10 +221,15 @@ export function args(rawArgv: string[]): Args {
     }
   }
 
+  if (argv.reachableVulnsTimeout === undefined) {
+    argv.reachableVulnsTimeout = REACHABLE_VULNS_TIMEOUT.toString();
```
darmalovan, line 225: Why `toString` btw?

help/help.txt
darmalovan: Can you write that is only available for paid users?

src/cli/args.ts
```diff
@@ -36,6 +36,8 @@ const DEBUG_DEFAULT_NAMESPACES = [
   'snyk-mvn-plugin',
 ];
 
+const REACHABLE_VULNS_TIMEOUT = 5 * 60; // 5 min (in seconds)
```
darmalovan, line 39: Where is this translate to ms?

help/help.txt
orsagie: I am a little uneasy about adding a flag for paid users only. In the past we have checked against feature flags, but eventually released to all. Maybe instead display an error saying something like `This feature is not available for this org, please contact us if you wish to use it`?

help/help.txt
orsagie: Is there a case where users want different timeouts for different cases? Consider allowing to set a longer default in the config? (maybe not for now, but in future)

help/help.txt
darmalovan: This is static text - rather not to share this info in the help text?

src/lib/monitor/index.ts
orsagie: Can you elaborate on the different kinds of errors? If it's a timeout error, a permissions error, or something else. Each one will have a different action the user can take before contacting us

help/help.txt
orsagie: Then do it in both places. They might not be targeting the correct org?

src/cli/args.ts
muscar: @darmalovan Because WebStorm (and typescript) complains that the value can't be a number.

src/cli/args.ts
muscar: @darmalovan  In the maven plugin, before it calls the call graph builder. Yes, it's messy, and we could move it here, and have the maven plugin accept milliseconds. I'm open to that.

help/help.txt
muscar: @orsagie I don't know. I think this is highly project specific, so the users are in the best position to tweak it as needed. We provide a default just because we don't want it to block their flows for too long.

src/lib/monitor/index.ts
muscar: @orsagie We don't have enough context in the CLI right now to elaborate--because we only have a generic `CallGraphError` with a generic string message, and an inner error. Since string messages can change, I don't think inspecting those to glean more info is a good idea, because it leads to brittle code.

I do agree with your point that it's good to provide users with a more detailed reason for the failure so they can provide that info to support.

I see two ways of achieving this:

1. Change the error message to tell them to re-run the command with `-d`, and provide the output in their support query; or
2. Change the error message in the maven plugin to something less generic, and more error specific.

I would personally prefer the first approach as the code in the plugin doesn't have to change as new error cases appear, and we'll be getting almost the same info anyway. What do you think?

help/help.txt
muscar: I thought the plan is to roll this to free users at some point, right? And we already have a feature flag which we check for, and show a message. We could change the feature flag message slightly to say that this is a paid feature only (for now)?

cc/ @orsagie @darmalovan

src/cli/args.ts
darmalovan: Thanks @muscar can you just point to the place it happens?

help/help.txt
```diff
@@ -90,6 +90,18 @@ Maven options:
   --scan-all-unmanaged
                        Auto detects maven jars and wars in given directory.
                        Individual testing can be done with --file=<jar-file-name>
+  --reachable-vulns    
+                       (test and monitor commands only)
+                       Analyze your source code to find which vulnerabe
```
orsagie, line 95: vulnerabe -> vulnerable :)

src/lib/monitor/index.ts
orsagie: Just noticed I didn't comment in the right place, this comment's place is after the request to registry has been made, and regards how we handle those responses. 

This error handling is specifically for after a call to the plugins, and in that case I agree with #1. I think the error from maven plugin already asks to run with -d, can you double check? might not need to change anything here.

muscar: The maven plugin doesn't print the `-d` advice, but I will open a PR to make it do so.

src/lib/monitor/index.ts
```diff
@@ -213,7 +218,17 @@ async function monitorDepTree(
...
+    analytics.add('callGraphError', err.innerError);
```
darmalovan, line 223: @muscar I just verified locally seems that we need to do `err.innerError.toString()` in order for it be populated

PR #1279, opened 2020-07-22
fix: abridge call graph creation error messages in analytics

muscar:
- [x] Ready for review

#### What does this PR do?

Since the call graph error messages can get pretty big, abridge them to something more manageable. This keep the beginning, and end of the message since, presumably, that's where the most important details are.

### DIFF
--- a/src/lib/error-format.ts
+++ b/src/lib/error-format.ts
@@ -1,0 +1,12 @@
+export function abridgeErrorMessage(
+  msg: string,
+  maxLen: number,
+  ellipsis?: string,
+): string {
+  if (msg.length <= maxLen) {
+    return msg;
+  }
+  const e = ellipsis || ' ... ';
+  const toKeep = (maxLen - e.length) / 2;
+  return msg.slice(0, toKeep) + e + msg.slice(msg.length - toKeep, msg.length);
+}
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -51,8 +51,11 @@ import {
 } from './utils';
 import { countPathsToGraphRoot } from '../utils';
 import * as alerts from '../alerts';
+import { abridgeErrorMessage } from '../error-format';
 
 const debug = Debug('snyk');
+
+const ANALYTICS_PAYLOAD_MAX_LENGTH = 1024;
 
 // TODO(kyegupov): clean up the type, move to snyk-cli-interface repository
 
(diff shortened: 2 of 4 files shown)


src/lib/error-format.ts
```diff
@@ -0,0 +1,12 @@
+export function abridgeErrorMessage(
+  msg: string,
+  maxLen: number,
+  ellipsis?: string,
+): string {
+  if (msg.length <= maxLen) {
+    return msg;
+  }
+  const e = ellipsis || ' ... ';
+  const toKeep = (maxLen - e.length) / 2;
```
darmalovan, line 10: I think `toKeep` can be a not whole number

src/lib/error-format.ts
```diff
@@ -0,0 +1,12 @@
+export function abridgeErrorMessage(
+  msg: string,
+  maxLen: number,
+  ellipsis?: string,
+): string {
+  if (msg.length <= maxLen) {
+    return msg;
+  }
+  const e = ellipsis || ' ... ';
```
darmalovan, line 9: Maybe have it as an optional param on line 4?
`ellipsis = '...': string`,

src/lib/monitor/index.ts
```diff
@@ -220,7 +223,13 @@ async function monitorDepTree(
   let callGraphPayload;
   if (options.reachableVulns && scannedProject.callGraph?.innerError) {
     const err = scannedProject.callGraph as CallGraphError;
-    analytics.add('callGraphError', err.innerError.toString());
+    analytics.add(
+      'callGraphError',
+      abridgeErrorMessage(
```
darmalovan, line 228: This should go also to the test file?

src/lib/error-format.ts
muscar: I know. But it seems to round it when used in `slice`.

darmalovan: I think it would be better to round it before, less JS magic

src/lib/monitor/index.ts
darmalovan, line 228: I think it would be nice to say why we are doing this trimming with a comment

src/lib/error-format.ts
muscar: Is that supported by TypeScript?

test/error-format.test.ts
```diff
@@ -0,0 +1,22 @@
+import { test } from 'tap';
+import { abridgeErrorMessage } from '../src/lib/error-format';
+
+test('abridge empty string', async (t) => {
```
darmalovan, line 4: Maybe also a null test?

src/lib/monitor/index.ts
muscar: That is pretty self-explanatory from the code, no? E.g. `ANALYTICS_PAYLOAD_MAX_LENGTH`

src/lib/error-format.ts
darmalovan: Optional parameters? in JS and TS

src/lib/monitor/index.ts
muscar: Testing that method would be quite involved, and the functional tests for abridging the string should cover a reasonable number of cases.

test/error-format.test.ts
muscar: The string cannot be `null`.

darmalovan: Got it!

PR #1281, opened 2020-07-23
Node v14 binaries build

JackuB:
Tested [built v14 artifacts on CI](https://app.circleci.com/pipelines/github/snyk/snyk/1491/workflows/ccd98361-ceb3-40f6-bf74-c7358e20d4a1/jobs/7076/artifacts) on these platforms:

- [x] macOS
- [x] Windows (10)
- [x] alpine (in docker)
- [x] linux (in docker)

Can't reproduce the issues with previous update to v10 https://github.com/snyk/snyk/pull/672 - couldn't replicate the original reported issues with Alpine build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,6 +227,17 @@ workflows:
               ignore:
                 - master
       - test-windows:
+          name: Windows Tests for Node v14 support
+          context: nodejs-install
+          node_version: '14'
+          requires:
+            - Common
+            - Danger Zone
+          filters:
+            branches:
+              ignore:
+                - master
+      - test-windows:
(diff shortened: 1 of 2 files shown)


JackuB:
@dkontorovskyy don't think that's needed for this release. [Node v10 is still maintained](https://nodejs.org/en/about/releases/) and changing binaries is risky on its own.

dkontorovskyy:
@JackuB, so we are just changing on which version it's gonna be built?

JackuB:
Those binaries (snyk-win.exe / snyk-macos / …) are not connected to the `snyk/snyk-cli` docker images - we can keep docker using NodeJS v10 and just change the binaries

PR #1286, opened 2020-07-26
feat: Support using Docker JWT token for snyk test

RotemS:
- [ ] Ready for review

#### What does this PR do?

Adding support for using `snyk test --docker` without an API key, using a JWT generated by Docker.

Removing the existing api key and setting an env var `SNYK_DOCKER_TOKEN`.
An alternative test endpoint should be called - `docker-jwt/test-dep-graph`.
Note - the entire flow will not work until we deploy the changes to the relevant endpoint.

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -5,7 +5,7 @@ import chalk from 'chalk';
 import * as snyk from '../../../lib';
 import * as config from '../../../lib/config';
 import { isCI } from '../../../lib/is-ci';
-import { apiTokenExists } from '../../../lib/api-token';
+import { apiTokenExists, getDockerToken } from '../../../lib/api-token';
 import { FAIL_ON, FailOn, SEVERITIES } from '../../../lib/snyk-test/common';
 import * as Debug from 'debug';
 import {
@@ -91,7 +91,15 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
     return Promise.reject(chalk.red.bold(error.message));
   }
 
-  apiTokenExists();
+  try {
+    apiTokenExists();
+  } catch (err) {
+    if (options.docker && getDockerToken()) {
+      options.testDepGraphDockerEndpoint = '/docker-jwt/test-dep-graph';
+    } else {
+      throw err;
+    }
+  }
 
   // Promise waterfall to test all other paths sequentially
   for (const path of args as string[]) {
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -56,6 +56,7 @@ import { Payload, PayloadBody, DepTreeFromResolveDeps } from './types';
 import { CallGraphError } from '@snyk/cli-interface/legacy/common';
 import * as alerts from '../alerts';
 import { abridgeErrorMessage } from '../error-format';
+import { getDockerToken } from '../api-token';
 
 const debug = debugModule('snyk');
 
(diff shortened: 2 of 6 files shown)


test/acceptance/docker-token.test.ts
```diff
@@ -0,0 +1,148 @@
...
+const BASE_API = '/api/v1';
+process.env.SNYK_API = 'http://localhost:' + port + BASE_API;
+process.env.SNYK_HOST = 'http://localhost:' + port;
+process.env.LOG_LEVEL = '0';
+const apiKey = '123456789';
+let oldkey;
+let oldendpoint;
+const server = fakeServer(BASE_API, apiKey);
+
+import * as sinon from 'sinon';
```
aviadhahami, line 17: can we hoist imports?

test/acceptance/docker-token.test.ts
```diff
@@ -0,0 +1,148 @@
...
+import * as plugins from '../../src/lib/plugins/index';
+
+// @later: remove this config stuff.
```
aviadhahami, line 20: should these stay?

test/acceptance/docker-token.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      docker: false,
+    });
+    t.fail('expected MissingApiTokenError');
+  } catch (err) {
+    t.equal(err.name, 'MissingApiTokenError', 'should throw if not docker');
+  }
+});
+
+// @later: try and remove this config stuff
+// Was copied straight from ../src/cli-server.js
```
aviadhahami, line 103: same?

test/acceptance/docker-token.test.ts
RotemS: I can move the `sinon` one, but the `plugins` needs to be after the server, it causes the config to be set too soon and the test isn't using the correct test API, but the default snyk one

test/acceptance/docker-token.test.ts
RotemS: I don't have a strong opinion either way, saw it was left in other tests that used it as well. Can remove :)

test/acceptance/docker-token.test.ts
aviadhahami: maybe just add a comment then :)

test/acceptance/docker-token.test.ts
aviadhahami: IMO can be either; will probably be removed anyways when jest shows up; let them stay :D

test/acceptance/docker-token.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  try {
+    await cli.test('foo:latest', {
+      docker: false,
+    });
+    t.fail('expected MissingApiTokenError');
+  } catch (err) {
+    t.equal(err.name, 'MissingApiTokenError', 'should throw if not docker');
+  }
+});
+
```
jurglic, line 99: Should we also have a test when both api and docker token are set? That would be in the case when someone calls `docker scan` and he's already authenticated with snyk...

src/cli/commands/test/index.ts
```diff
@@ -91,7 +91,15 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
     return Promise.reject(chalk.red.bold(error.message));
   }
 
-  apiTokenExists();
+  try {
+    apiTokenExists();
```
JackuB, line 95: Couldn't we extend the `apiTokenExists` (and `api()`) to look for the docker token? Or we are doing this to keep it contained in the `test` command?

RotemS: To be contained and also since they can happen at the same time. There's a scenario where both exist (and then the api key should be used). But I'm open to improvements if you think it would be a better implementation

src/lib/snyk-test/run-test.ts
```diff
@@ -557,14 +558,19 @@ async function assembleLocalPayloads(
         });
         body.callGraph = callGraph;
       }
-
+      const authHeader = getAuthHeader();
+      const reqUrl =
+        config.API +
+        (options.vulnEndpoint ||
```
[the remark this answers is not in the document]
RotemS: I think you're right and it shouldn't. Though it will be used by Docker so I don't think it would be used. Will move

PR #1288, opened 2020-07-27
feat: support pipfile on containerized cli

andremidea:
- [x] Ready for review

#### What does this PR do?

It allows the user to specify a -f `Pipfile` when running the CLI in a containerized environment.
Also, fix issues with automatically detecting a `setup.py` file and installing its dependencies, but only if no `-f` was provided.

### DIFF
--- a/docker/docker-python-entrypoint.sh
+++ b/docker/docker-python-entrypoint.sh
@@ -49,6 +49,10 @@ if [ -n "${TARGET_FILE}" ]; then
         echo "Installing dependencies from setup.py"
         pip install -U -e "${PROJECT_PATH}"
         ;;
+    *Pipfile)
+        echo "Installing dependencies from Pipfile"
+        installPipfileDeps
+        ;;
     *)
         exitWithMsg "\"${PROJECT_PATH}/${TARGET_FILE}\" is not supported" 1
         ;;
--- a/docker/docker-python-entrypoint.sh
+++ b/docker/docker-python-entrypoint.sh
@@ -59,15 +59,17 @@ if [ -n "${TARGET_FILE}" ]; then
     esac
 fi
 
-if [ -f "${PROJECT_PATH}/requirements.txt" ]; then
-    echo "Found requirement.txt"
-    installRequirementsTxtDeps "${PROJECT_PATH}/requirements.txt"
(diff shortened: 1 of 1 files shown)


docker/docker-python-entrypoint.sh
```diff
@@ -49,17 +49,27 @@ if [ -n "${TARGET_FILE}" ]; then
         echo "Installing dependencies from setup.py"
         pip install -U -e "${PROJECT_PATH}"
         ;;
+    Pipfile)
```
odinn1984, line 52: @andremidea I really don't remember but is there a reason why there's a REGEX in the other cases and you put Pipfile here? does it have to be with a capital P? should we maybe make it less strict?

docker/docker-python-entrypoint.sh
```diff
@@ -49,17 +49,27 @@ if [ -n "${TARGET_FILE}" ]; then
...
+if [ -z "${TARGET_FILE}" ]; then
+    if [ -f "${PROJECT_PATH}/requirements.txt" ]; then
+        echo "Found requirement.txt"
+        installRequirementsTxtDeps "${PROJECT_PATH}/requirements.txt"
+    elif [ -f "${PROJECT_PATH}/setup.py" ]; then
+        echo "Found setup.py"
+        pip install -U -e "${PROJECT_PATH}"
+    elif [ -f "${PROJECT_PATH}/Pipfile" ]; then
+        echo "Found Pipfile"
+        installPipfileDeps
```
odinn1984, line 71: what is this?

andremidea: What do you mean? This was always executed, to install the python deps if those files are found

docker/docker-python-entrypoint.sh
andremidea: No, using pipenv the file is always `Pipenv`

docker/docker-python-entrypoint.sh
odinn1984: just making sure it's not something custom that i missed it's definition i didn't use pipfile beforehand :)

docker/docker-python-entrypoint.sh
andremidea: makes sense to have a `*Pipfile`, good catch!

PR #1301, opened 2020-07-31
fix: propagate failed scans with --all-projects

lili2311:
- [x] Ready for review

#### What does this PR do?

- Collect failed dependency extractions with `--all-projects` and propagate it higher up.
- For `monitor` these are now propagated all the way and will appear in output + json output. This may affect the existing error code as previously we silently skipped failure like these but not failures to get vulnerabilities during monitor or when a path test failed. So it was inconsistent.
- For test more work / refactor needed to get these to the output & json so for now warning the user & providing instructions how to see the error with `-d`
https://github.com/snyk/snyk/compare/fix/throw-erorr-if-all-tests-failed?expand=1#diff-e4ec2ebbae5004d31be7a5b680580a46R42

You can use `cd snyk/test/acceptance/workspaces && snyk test --all-projects` and `cd snyk/test/acceptance/workspaces && snyk monitor --all-projects` to see the output.

Sometimes users see a blank screen when 100% of `-all-projects` scans fail and we do not flag about other failures as it was too noisy unless you run with `-d`

#### Screenshots

Screen Shot 2020-07-31 at 14 27 22
Screen Shot 2020-07-31 at 14 42 06
Screen Shot 2020-07-31 at 14 49 03
Screen Shot 2020-07-31 at 16 45 33

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ key.pem
 
 # Diagnostic reports (https://nodejs.org/api/report.html)
 report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
+# test fixture build artifacts
+/test/acceptance/workspaces/**/project/
+/test/acceptance/workspaces/**/target/
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -42,7 +42,7 @@ import {
   GitRepoCommitStats,
   execShell,
 } from '../../../lib/monitor/dev-count-analysis';
-import { FailedToRunTestError } from '../../../lib/errors';
+import { FailedToRunTestError, MonitorError } from '../../../lib/errors';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 const debug = Debug('snyk');
@@ -166,12 +166,7 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         }),
         spinner.clear(analyzingDepsSpinnerLabel),
       );
-
       analytics.add('pluginName', inspectResult.plugin.name);
-
-      const postingMonitorSpinnerLabel =
-        'Posting monitor snapshot for ' + displayPath + ' ...';
-      await spinner(postingMonitorSpinnerLabel);
 
       // We send results from "all-sub-projects" scanning as different Monitor objects
       // multi result will become default, so start migrating code to always work with it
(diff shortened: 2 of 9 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -166,13 +166,8 @@ async function monitor(...args0: MethodArgs): Promise<any> {
         }),
         spinner.clear(analyzingDepsSpinnerLabel),
       );
-
       analytics.add('pluginName', inspectResult.plugin.name);
 
-      const postingMonitorSpinnerLabel =
```
lili2311, line 172: moved a little further down unchanged

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -171,10 +171,14 @@ export const AllProjectsTests: AcceptanceTests = {
       utils.chdirWorkspaces();
       const spyPlugin = sinon.spy(params.plugins, 'loadPlugin');
       t.teardown(spyPlugin.restore);
-
-      const result = await params.cli.monitor('monorepo-bad-project', {
-        allProjects: true,
-      });
+      let result;
+      try {
```
lili2311, line 175: these were not previously propagated up if dependencies failed to be detected, but we do fail if one of the paths fails to be scanned so this aligns that behaviour.

This may however start failing some pipelines that previously were skipping the failures. How can this be communicated & is this the desired behaviour right? You want to know something you monitored failed and it did not pass.

.gitignore
```diff
@@ -17,3 +17,5 @@ snyk_report.html
 !/docker/snyk_report.css
 cert.pem
 key.pem
+test/acceptance/workspaces/sbt-app/project/
```
orsagie, line 20: Can you do a catch all like on line 10, so we don't do this for every sbt app?

src/lib/plugins/get-multi-plugin-result.ts
```diff
@@ -20,9 +21,16 @@ export interface ScannedProjectCustom
   plugin: PluginMetadata;
   callGraph?: CallGraph;
 }
+
+export interface FailedProjectScan {
```
orsagie, line 25: why is this exported?

Can you rename to FailedProjectScanError?

PR #1304, opened 2020-08-03
Capture snyk integration name

JackuB:
This PR adds an integrationName and integrationVersion optional parameters to the Analytics.
  We are collecting Snyk CLI usage in our official integrations

  We distinguish them by either:
  - Setting integrationNameHeader or integrationVersionHeader in environment when CLI is run
  - passing an --integration-name or --integration-version flags on CLI invocation

  Integration name is validated with a list

### DIFF
--- a/src/lib/analytics-sources.ts
+++ b/src/lib/analytics-sources.ts
@@ -1,0 +1,65 @@
+/*
+  We are collecting Snyk CLI usage in our official integrations
+
+  We distinguish them by either:
+  - Setting integrationNameHeader or integrationVersionHeader in environment when CLI is run
+  - passing an --integration-name or --integration-version flags on CLI invocation
+
+  Integration name is validated with a list
+*/
+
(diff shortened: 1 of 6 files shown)


src/lib/analytics-sources.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // CI - tracked by passing envvar on CLI invocation
+  'JENKINS',
+  'TEAMCITY',
+  'BITBUCKET_PIPELINES',
+  'AZURE_PIPELINES',
+  'CIRCLECI_ORB',
+  'GITHUB_ACTIONS',
+  'NEXUS',
+  'ARTIFACTORY',
```
pavel-snyk, line 36: Do we need this? because both plugins uses API, so we'll not be able to run CLI inside.

JackuB: Removed them :+1:

PR #1307, opened 2020-08-04
feat: show original severity CLI

joaolemos:
- [x] Ready for review

https://snyksec.atlassian.net/browse/GOV-748

##### Before

Screenshot 2020-08-11 at 09 33 19

##### After

Screenshot 2020-08-11 at 09 33 01

### DIFF
--- a/src/cli/commands/test/formatters/legacy-format-issue.ts
+++ b/src/cli/commands/test/formatters/legacy-format-issue.ts
@@ -13,6 +13,7 @@ import {
   GroupedVuln,
   AnnotatedIssue,
   DockerIssue,
+  SEVERITY,
 } from '../../../../lib/snyk-test/legacy';
 import { formatLegalInstructions } from './legal-license-instructions';
 import { getReachabilityText } from './format-reachability';
@@ -35,12 +36,13 @@ export function formatIssues(
   ).join(', ');
 
   const vulnOutput = {
-    issueHeading: createSeverityBasedIssueHeading(
-      vuln.metadata.severity,
-      vuln.metadata.type,
-      vuln.metadata.name,
-      false,
-    ),
+    issueHeading: createSeverityBasedIssueHeading({
+      severity: vuln.metadata.severity,
+      originalSeverity: vuln.originalSeverity,
+      type: vuln.metadata.type,
+      packageName: vuln.metadata.name,
+      isNew: false,
+    }),
     introducedThrough: '  Introduced through: ' + uniquePackages,
     description: '  Description: ' + vuln.title,
     info: '  Info: ' + chalk.underline(config.ROOT + '/vuln/' + vulnID),
@@ -76,7 +78,21 @@ export function formatIssues(
   );
 }
 
-function createSeverityBasedIssueHeading(severity, type, packageName, isNew) {
+type CreateSeverityBasedIssueHeading = {
+  severity: SEVERITY;
+  originalSeverity?: SEVERITY;
+  type: string;
+  packageName: string;
+  isNew: boolean;
+};
+
+function createSeverityBasedIssueHeading({
+  severity,
+  originalSeverity,
+  type,
+  packageName,
+  isNew,
+}: CreateSeverityBasedIssueHeading) {
   // Example: ✗ Medium severity vulnerability found in xmldom
   const vulnTypeText = type === 'license' ? 'issue' : 'vulnerability';
   const severitiesColourMapping = {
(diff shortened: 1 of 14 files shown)


src/cli/commands/test/formatters/remediation-based-format-issues.ts
```diff
@@ -493,11 +499,16 @@ export function formatIssue(
     reachablePathsTemplate,
   );
 
+  let originalSeverityStr = '';
+  if (originalSeverity && originalSeverity !== severity) {
```
lili2311, line 503: there is also `test/formatters/legacy-format-issue.ts` for package managers that do not support remediation. You need to make the same changed there too.

joaolemos: Got it, ported these into legacy formatter and added some tests for it 👍

src/cli/commands/test/formatters/legacy-format-issue.ts
```diff
@@ -76,7 +77,13 @@ export function formatIssues(
   );
 }
 
-function createSeverityBasedIssueHeading(severity, type, packageName, isNew) {
+function createSeverityBasedIssueHeading(
```
odinn1984, line 80: did you make sure that you updated all the places that use this function?

src/cli/commands/test/formatters/legacy-format-issue.ts
```diff
@@ -96,11 +103,17 @@ function createSeverityBasedIssueHeading(severity, type, packageName, isNew) {
       },
     },
   };
+
+  let originalSeverityStr = '';
+  if (originalSeverity && originalSeverity !== severity) {
```
odinn1984, line 108: can we maybe type the params of the function to make sure we're always getting a string here?

src/cli/commands/test/formatters/legacy-format-issue.ts
```diff
@@ -96,11 +103,17 @@ function createSeverityBasedIssueHeading(severity, type, packageName, isNew) {
...
+    originalSeverityStr = ` (originally ${titleCaseText(originalSeverity)})`;
+  }
+
   return (
     severitiesColourMapping[severity].colorFunc(
       '✗ ' +
         titleCaseText(severity) +
-        ' severity ' +
+        ` severity${originalSeverityStr} ` +
```
odinn1984, line 116: shouldn't there b a space in this string before "$"?

src/cli/commands/test/formatters/legacy-format-issue.ts
joaolemos: It's a local function to that file with one place, though there is another with that name that is used for `snyk wizard`. I wonder if we need to change it too though... 🤔

mrded: Yeah, the function isn't exported anywhere. 

nit: However, I think originally 4 arguments was a lot already, and I would make it an object.

```suggestion
function createSeverityBasedIssueHeading({ severity, originalSeverity, type, packageName, isNew })
```

src/cli/commands/test/formatters/legacy-format-issue.ts
joaolemos: Original severity might be undefined, and it will have a single space after in that case.

src/cli/commands/test/formatters/remediation-based-format-issues.ts
```diff
@@ -493,11 +499,16 @@ export function formatIssue(
...
+    originalSeverityStr = ` (originally ${titleCaseText(originalSeverity)})`;
+  }
+
   return (
     severitiesColourMapping[severity].colorFunc(
       `  ✗ ${chalk.bold(title)}${newBadge} [${titleCaseText(
         severity,
-      )} Severity]`,
+      )} Severity${originalSeverityStr}]`,
```
mrded, line 511:
```suggestion
      )} Severity ${originalSeverityStr}]`,
```

src/cli/commands/test/formatters/legacy-format-issue.ts
mrded:
```suggestion
        ` severity${originalSeverityStr && ' ' + originalSeverityStr} ` +
```

mrded: Or perhaps something like this:
```
['✗ ', titleCaseText(severity), titleCaseText(severity), 'severity', originalSeverityStr].join(' ')
```

src/cli/commands/test/formatters/remediation-based-format-issues.ts
joaolemos: This will add an extra space when not needed, if the `originalSeverityStr`is not defined 🙂

src/cli/commands/test/formatters/legacy-format-issue.ts
joaolemos: As discussed, leaving it for now as string concatenation is quite spread out across the codebase 🙂

src/cli/commands/test/formatters/remediation-based-format-issues.ts
JackuB: PROTIP: you could use [`trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim)

PR #1323, opened 2020-08-11
chore: add VS launch.json config

JackuB:
It was removed few days ago https://github.com/snyk/snyk/pull/1308 - but actually it would be useful to have this shared for development.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,19 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Tap Current File",
+      "console": "integratedTerminal",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
pavel-snyk, line 4: Is it possible to have comments in JSON file?

JackuB: No… _but_
There are some supersets of JSON standard like [JSON5](https://json5.org) that supports it, so I guess that (or similar) is what VS Code is using

PR #1332, opened 2020-08-14
feat: Allow environment variables to override config values

mladkau:
- [x] Ready for review

#### What does this PR do?
Adding a wrapper around configstore to allow environment variables to override config values.

You can run the following to disable suggestions:
```
$ SNYK_CFG_DISABLESUGGESTIONS=true snyk test
```

https://snyksec.atlassian.net/browse/DC-881

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -52,7 +52,7 @@ if (!config.org && org) {
 // invalid (non-numeric) value will fallback to the default
 const timeout = userConfig.get('timeout');
 if (!config.timeout) {
-  config.timeout = +timeout ? +timeout : DEFAULT_TIMEOUT;
+  config.timeout = timeout && +timeout ? +timeout : DEFAULT_TIMEOUT;
 }
 
 // this is a bit of an assumption that our web site origin is the same
--- a/src/lib/user-config.ts
+++ b/src/lib/user-config.ts
@@ -1,3 +1,16 @@
 const Configstore = require('configstore');
 const pkg = require(__dirname + '/../../package.json');
-export const config = new Configstore(pkg.name);
+
+class ConfigStoreWithEnvironmentVariables extends Configstore {
+  constructor(id, defaults = undefined, options = {}) {
+    super(id, defaults, options);
+  }
+
+  public get(key: string): string | undefined {
+    const envKey = `SNYK_CFG_${key.replace(/-/g, '_').toUpperCase()}`;
+    const envValue = process.env[envKey];
+    return super.has(key) && !envValue ? String(super.get(key)) : envValue;
+  }
+}
+
+export const config = new ConfigStoreWithEnvironmentVariables(pkg.name);


src/lib/user-config.ts
```diff
@@ -1,3 +1,15 @@
 const Configstore = require('configstore');
 const pkg = require(__dirname + '/../../package.json');
-export const config = new Configstore(pkg.name);
+
+class ConfigStoreWithEnvironmentVariables extends Configstore {
+  constructor(id, defaults = undefined, options = {}) {
```
mladkau, line 5: Copied from configstore.

src/lib/user-config.ts
```diff
@@ -1,3 +1,15 @@
...
+    super(id, defaults, options)
+  }
+
+  public get(key: string) {
+    const env_key = `SNYK_${key.replace('-', '_').toUpperCase()}`
```
mladkau, line 10: We need to do a bit of string replace here so we can support things like `disable-analytics` which becomes SNYK_DISABLE_ANALYTICS.

garethr:
Is it `SNYK_DISABLE_ANALYTICS` or `SNYK_DISABLEANALYTICS`? One is used in the description and one in the comment?

mladkau:
The comment was outdated. It is `SNYK_CFG_DISABLE_ANALYTICS` for the config option `disable-analytics`

src/lib/user-config.ts
```diff
@@ -1,3 +1,15 @@
...
-export const config = new Configstore(pkg.name);
+
+class ConfigStoreWithEnvironmentVariables extends Configstore {
+  constructor(id, defaults = undefined, options = {}) {
+    super(id, defaults, options);
+  }
+
+  public get(key: string) {
+    const envKey = `SNYK_CFG_${key.replace(/-/g, '_').toUpperCase()}`;
+    return process.env[envKey] || super.get(key);
```
JackuB, line 11: Are we worried about types here?
Afaik, configstore can return types like bool or number as well, but process.env will always be a string?

mladkau: For disableSuggestions I can see the following code:
```
const disableSuggestions = userConfig.get('disableSuggestions');
if (disableSuggestions) {
  config.disableSuggestions = disableSuggestions;
}

...

  let dockerSuggestion = '';
  if (options.docker && config.disableSuggestions !== 'true') {
 ```
Which suggests we look for a string here but in other cases we might not. We could try to `eval` the environment value but this will have security implications. We could also try to do a parseInt, parseBool here and try to interpret integer and boolean values (else would be string). WDYT?

Comment (JackuB) [reply src/lib/user-config.ts:11]: Or the other way around, cast everything from configstore as String to be consistent? Currently it's not typed at all and on many places we are already doing the conversion in code…

**Comment by mladkau** [comment]
I think Boost having a firedrill might make the tests fail in the moment...

File: src/lib/user-config.ts (hunk 3 above)
Comment (mladkau) [reply src/lib/user-config.ts]: I've checked the places in the code which I could find and converting everything into a string or undefined should work ... (famous last words)

File: src/lib/config.ts [hunk 4]
```diff
@@ -50,9 +50,12 @@ if (!config.org && org) {
 // client request timeout
 // to change, set this config key to the desired value in seconds
 // invalid (non-numeric) value will fallback to the default
-const timeout = userConfig.get('timeout');
-if (!config.timeout) {
-  config.timeout = +timeout ? +timeout : DEFAULT_TIMEOUT;
+let timeoutString = userConfig.get('timeout');
+if (timeoutString) {
+  let timeout = parseInt(timeoutString);
```
mladkau, line 55: Wanted to make the string conversion here a bit more explicit - can remove if anyone disagrees ...

mladkau: (actually nevermind)

src/lib/config.ts
```diff
@@ -52,7 +52,7 @@ if (!config.org && org) {
 // invalid (non-numeric) value will fallback to the default
 const timeout = userConfig.get('timeout');
 if (!config.timeout) {
-  config.timeout = +timeout ? +timeout : DEFAULT_TIMEOUT;
+  config.timeout = timeout && +timeout ? +timeout : DEFAULT_TIMEOUT;
```
mladkau, line 55: I opted in the end for this as it doesn't require too much code change and seems to be ok with the now explicit typing.

PR #1348, opened 2020-08-19
test: ruby custom gemfile test

lili2311:
- [x] Ready for review

#### What does this PR do?

Include a test that shows targeting a custom Gemfile with --file will not be able to determine the lockfile name and asks the user to use the lockfile.

### DIFF
--- a/test/acceptance/cli-test/cli-test.ruby.spec.ts
+++ b/test/acceptance/cli-test/cli-test.ruby.spec.ts
@@ -77,6 +77,60 @@ export const RubyTests: AcceptanceTests = {
         'depGraph looks fine',
       );
     },
+    '`test ruby-app-custom-names --file=123.gemfile --package-manager=rubygems`': (
+      params,
+      utils,
+    ) => async (t) => {
+      utils.chdirWorkspaces();
+      await params.cli.test('ruby-app-custom-names', {
+        file: '123.gemfile',
+        packageManager: 'rubygems',
+      });
+
+      const req = params.server.popRequest();
+      t.equal(req.method, 'POST', 'makes POST request');
+      t.equal(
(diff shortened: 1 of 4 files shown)


test/acceptance/cli-test/cli-test.ruby.spec.ts
```diff
@@ -78,6 +78,24 @@ export const RubyTests: AcceptanceTests = {
...
+      utils.chdirWorkspaces();
+      try {
+        await params.cli.test('ruby-app-custom-names', {
+          file: '123.gemfile',
+          packageManager: 'rubygems',
+        });
+      } catch (e) {
+        t.match(
+          e.message,
+          'if this is a custom file name re-run with --file=path/to/custom.gemfile.lock --package-manager=rubygems',
```
anthogez, line 94: Shouldn’t the message say `if this is a custom file name re-run with --file=path/to/anything-food.gemfile --package-manager=rubygems`

without `.lock`?

lili2311: the message is correct, we require the `.lock` file only for a scan. We can auto detect it from a Gemfile normally if it follows a standard pattern, but id the pattern is custom we can't guess where `.lock` might be. We do by default try to append it to the end of the Gemfile name, let me check this again as it looks like in this case it should have worked & I need to change the example to a complex one where .lock is not at the end

lili2311: Okay updated!

anthogez: Thanks! 🎉

anthogez (approved):

PR #1352, opened 2020-08-19
feat: adding cta for docker users

RotemS:
- [x] Ready for review

#### What does this PR do?

Adding a CTA if a test is made without an API key and with a Docker JWT. 
There are 2 cases - 
* Scan successful
image
image

* Scan not successful - the user has reached the free test limit
image

Need to setup the env locally with JWT auth enabled and with a valid token.
In any `snyk test` where the API key doesn't exist and the `SNYK_DOCKER_TOKEN` env var is set, the CTA should be displayed.

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -49,6 +49,7 @@ import {
 import * as utils from './utils';
 import { getIacDisplayedOutput } from './iac-output';
 import { getEcosystem, testEcosystem } from '../../../lib/ecosystems';
+import { TestLimitReachedError } from '../../../lib/errors';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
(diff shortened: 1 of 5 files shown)


src/cli/commands/test/index.ts
```diff
@@ -263,6 +264,10 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
     // first one
     error.code = errorResults[0].code;
     error.userMessage = errorResults[0].userMessage;
+    if (error.userMessage === 'Test limit reached!' && options.isDockerUser) {
```
jurglic, line 267: That condition might be a bit fragile (if message changes), but probably we don't have an alternative check we could use?

jurglic (approved):
LGTM 👍

src/cli/commands/test/index.ts
RotemS: I didn't see any better way of doing this, but open to suggestions if anyone would have an idea

PR #1355, opened 2020-08-20
feat: Detect and log orphaned or errored gradle project scans by comparing detected files vs scanned.

lili2311:
- [x] Ready for review
### Depends on
- https://github.com/snyk/snyk-cli-interface/pull/38
- https://github.com/snyk/snyk-gradle-plugin/pull/142
#### What does this PR do?
Detect and log orphaned or errored gradle project scans by comparing detected files vs scanned.
This is useful to find any gradle files that looked like they are part of a multi-project build but were in fact not included or had an error being scanned. Since only the root level file is detected these errors can otherwise fo unnoticed.

https://github.com/snyk/snyk/pull/1355/files#diff-91534f3e6e36be30e5b2863969dc6417R82

Drop a gradle file in a folder in a multi-build project, this file is not included as a sub-project and will be detected but not scanned since the gradle build will not return it.

#### Screenshots
Screen Shot 2020-08-21 at 11 43 52

### DIFF
--- a/src/lib/find-files.ts
+++ b/src/lib/find-files.ts
@@ -47,16 +47,20 @@ export async function getStats(path: string): Promise<fs.Stats> {
  * @param filter (optional) file names to find. If not provided all files are returned.
  * @param levelsDeep (optional) how many levels deep to search, defaults to two, this path and one sub directory.
  */
+interface FindFilesRes {
+  files: string[];
+}
+
 export async function find(
   path: string,
   ignore: string[] = [],
   filter: string[] = [],
   levelsDeep = 4,
-): Promise<string[]> {
+): Promise<FindFilesRes> {
   const found: string[] = [];
   // ensure we ignore find against node_modules path.
   if (path.endsWith('node_modules')) {
-    return found;
+    return { files: found };
   }
   // ensure node_modules is always ignored
   if (!ignore.includes('node_modules')) {
@@ -64,13 +68,13 @@ export async function find(
   }
   try {
     if (levelsDeep < 0) {
-      return found;
+      return { files: found };
     } else {
       levelsDeep--;
     }
     const fileStats = await getStats(path);
     if (fileStats.isDirectory()) {
-      const files = await findInDirectory(path, ignore, filter, levelsDeep);
+      const { files } = await findInDirectory(path, ignore, filter, levelsDeep);
       found.push(...files);
     } else if (fileStats.isFile()) {
       const fileFound = findFile(path, filter);
@@ -78,8 +82,7 @@ export async function find(
         found.push(fileFound);
       }
     }
-
-    return filterForDefaultManifests(found);
+    return { files: filterForDefaultManifests(found) };
   } catch (err) {
     throw new Error(`Error finding files in path '${path}'.\n${err.message}`);
   }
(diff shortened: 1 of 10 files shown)


src/lib/plugins/get-deps-from-plugin.ts
```diff
@@ -105,3 +118,43 @@ export async function getDepsFromPlugin(
...
+
+function warnSomeGradleManifestsNotScanned(
+  scannedProjects: ScannedProject[],
+  allFilesFound: string[],
+  root: string,
+  optionsJson = false,
+): void {
+  const scannedGradleFiles = scannedProjects
+    .map((p) => {
+      const targetFile = p.meta?.targetFile || p.targetFile;
```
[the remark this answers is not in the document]
lili2311: I see what you mean ty! Will keep this in mind, in this particular PR we first auto detect any `build.gradle` files and then compare the detected vs scanned to alert if some were missed for any reason.

src/lib/find-files.ts
```diff
@@ -102,20 +123,30 @@ async function findInDirectory(
...
-        return [];
+        return { files: [], allFilesFound: [] };
       }
       return find(resolvedPath, ignore, filter, levelsDeep);
     });
+
   const found = await Promise.all(toFind);
-  return Array.prototype.concat.apply([], found);
+  return {
+    files: Array.prototype.concat.apply(
```
aviadhahami, line 141: using `apply` in combination w/ `map` will make this call allocate 3 arrays; imo we can simply return the mapped array:
```suggestion
    files: found.map(...)
```


> I'm not sure why this specific trick was previously used, as the `this` arg for `concat.apply` is an empty array so the "append" magic `concat.apply` can do is redundant;
 [`concat.apply` example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply#Using_apply_to_append_an_array_to_another)

src/lib/plugins/get-deps-from-plugin.ts
```diff
@@ -105,3 +121,40 @@ export async function getDepsFromPlugin(
...
+      return targetFile ? pathLib.resolve(root, targetFile) : null;
+    })
+    .filter(
+      (targetFile) =>
+        targetFile &&
+        (targetFile.endsWith('build.gradle') ||
+          targetFile.endsWith('build.gradle.kts')),
+    );
+  const detectedGradleFiles = allFilesFound.filter(
+    (targetFile) =>
```
aviadhahami, line 142: we need to assert `targetFile` as well no?

aviadhahami: Can we extract the following code?
```js
      (targetFile) =>
        targetFile &&
        (targetFile.endsWith('build.gradle') ||
          targetFile.endsWith('build.gradle.kts'))
```

something like:
```js
const gradleTargetFilesFilter = (targetFile)=> (...);
...
const scannedGradleFiles = scannedProjects.map(...).filter(gradleTargetFilesFilter)
const detectedGradleFiles = allFilesFound.filter(gradleTargetFilesFilter)

src/lib/find-files.ts
lili2311: The result is nested arrays, `concat.apply` flattens them all into 1 array.
Screen Shot 2020-08-26 at 17 14 08

lili2311: I think I can use `.flat()` will try it!

lili2311: @aviadatsnyk .flat() is not Node 8 compatible and couldn't get it working with TS with es2019 either :/ so reverted back for now until we drop Node 8

aviadhahami (approved):
Lgtm :)

aviadhahami (approved):
LGTM :checkered_flag:

PR #1357, opened 2020-08-21
Move Danger.js to GitHub Action

JackuB:
Danger.js is having issues running on CircleCI. This will also make it a bit faster.


.github/workflows/danger-zone.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    name: Danger JS
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
pavel-snyk, line 11: @JackuB , how about add a name for `checkout` action?

pavel-snyk (approved):

.github/workflows/danger-zone.yml
JackuB: Don't have a better name for `actions/checkout` then `Checkout` ¯\\\_(ツ)\_/¯

PR #1388, opened 2020-09-07
Use upstream deps

JackuB:
Screenshot 2020-09-09 at 13 57 04
We can drop our forked lodash dependency

HAMMER-99


package.json
```diff
@@ -53,14 +53,10 @@
...
   "dependencies": {
-    "@snyk/cli-interface": "2.9.0",
-    "@snyk/dep-graph": "1.19.3",
+    "@snyk/cli-interface": "2.9.1",
+    "@snyk/dep-graph": "1.19.4",
     "@snyk/gemfile": "1.2.0",
-    "@snyk/graphlib": "2.1.9-patch",
-    "@snyk/inquirer": "6.2.2-patch",
-    "@snyk/lodash": "^4.17.15-patch",
-    "@snyk/ruby-semver": "2.2.0",
```
pavel-snyk, line 62: @JackuB , `ruby-semver` is removed without updating: typo or expected?

JackuB: Ruby-semver is not being used in CLI, seems like a left over.

I also removed it from another unrelated snyk package and it's currently not in the CLI dependency tree

pavel-snyk: 👍

pavel-snyk (approved):

package.json
```diff
@@ -53,14 +53,10 @@
   "author": "snyk.io",
   "license": "Apache-2.0",
   "dependencies": {
-    "@snyk/cli-interface": "2.9.0",
-    "@snyk/dep-graph": "1.19.3",
+    "@snyk/cli-interface": "2.9.1",
```
maxjeffos, line 56: 2.9.2 is available now - should be functionally equivalent, but it's the one where we removed the extra types file

maxjeffos (approved):

aviadhahami (approved):

almog27 (approved):

PR #1416, opened 2020-09-16
Fix/sarif outputfile fixes

RotemS:
- [x] Ready for review

#### What does this PR do?
Fixes for `--sarif-file-output`
1. Not creating the file properly
2. Additional tests to cover e2e flow

### DIFF
--- a/src/cli/commands/types.ts
+++ b/src/cli/commands/types.ts
@@ -40,9 +40,11 @@ export abstract class TestCommandResult extends CommandResult {
   }
 
   public static createJsonTestCommandResult(
-    jsonResult: string,
+    stdout: string,
+    jsonResult?: string,
+    sarifResult?: string,
   ): JsonTestCommandResult {
-    return new JsonTestCommandResult(jsonResult);
+    return new JsonTestCommandResult(stdout, jsonResult, sarifResult);
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/cli/commands/types.ts
```diff
@@ -72,11 +74,17 @@ class HumanReadableTestCommandResult extends TestCommandResult {
 }
 
 class JsonTestCommandResult extends TestCommandResult {
-  constructor(jsonResult: string) {
-    super(jsonResult);
+  constructor(stdout: string, jsonResult?: string, sarifResult?: string) {
+    super(stdout);
+    this.jsonResult = jsonResult || '';
```
aviadhahami, line 79: Will we try to parse this somewhere on our flow?
If yes - `""` will throw

test/acceptance/fixtures/docker/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM node:10-slim
```
aviadhahami, line 1: Why do we care about this?

RotemS: To use `--sarif-file-output` we have to include a dockerfile. It's purely for the test to be able to run the entire flow.

aviadhahami: But we won't pull this image every test, right?

src/cli/commands/types.ts
RotemS: We're not, but that's a fair point. Changed it to just check and not have a default.

test/acceptance/fixtures/docker/Dockerfile
RotemS: It doesn't pull it, but actually realized I could just do from scratch since I don't mind about the base image info, so making it simpler.

test/acceptance/fixtures/docker/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM scratch
```
aviadhahami, line 1: YES!

aviadhahami (approved):

JackuB (approved):

PR #1419, opened 2020-09-21
feat: monitor dependencies with new ecosystem plugins

ivanstanev:
- [ ] Ready for review

#### What does this PR do?

Adds support for the new `monitor` flow used by the ecosystems plugin format ([#project-envelope](https://snyk.slack.com/archives/C01342FUD53)).

This currently uses a spike branch for `snyk-docker-plugin`, we still need to release a new version of the plugin and only then we can merge this PR.

[Jira ticket RUN-1181](https://snyksec.atlassian.net/browse/RUN-1181)

### DIFF
--- a/src/lib/ecosystems.ts
+++ b/src/lib/ecosystems.ts
@@ -1,93 +1,63 @@
-import * as cppPlugin from 'snyk-cpp-plugin';
-import { DepGraphData } from '@snyk/dep-graph';
+import { InspectResult } from '@snyk/cli-interface/legacy/plugin';
+import chalk from 'chalk';
+
 import * as snyk from './index';
 import * as config from './config';
 import { isCI } from './is-ci';
 import { makeRequest } from './request/promise';
-import { Options } from './types';
+import { MonitorResult, Options } from './types';
 import { TestCommandResult } from '../cli/commands/types';
 import * as spinner from '../lib/spinner';
+import { formatMonitorOutput } from '../cli/commands/monitor/formatters/format-monitor-response';
+import { GoodResult, BadResult } from '../cli/commands/monitor/types';
+import { MonitorError } from './errors';
+import { getExtraProjectCount } from './plugins/get-extra-project-count';
+import {
+  Ecosystem,
+  getPlugin,
+  ScanResult,
+  TestResult,
+} from './ecosystems-plugins';
 
-export interface PluginResponse {
-  scanResults: ScanResult[];
+const SEPARATOR = '\n-------------------------------------------------------\n';
+
+export interface EcosystemMonitorError {
+  error: string;
+  path: string;
+  scanResult: ScanResult;
 }
 
-export interface GitTarget {
-  remoteUrl: string;
-  branch: string;
+export interface MonitorDependenciesResponse {
+  ok: boolean;
+  org: string;
+  id: string;
+  isMonitored: boolean;
+  licensesPolicy: any;
+  uri: string;
+  trialStarted: boolean;
+  path: string;
+  projectName: string;
 }
 
-export interface ContainerTarget {
-  image: string;
+export interface EcosystemMonitorResult extends MonitorDependenciesResponse {
+  scanResult: ScanResult;
 }
 
-export interface ScanResult {
-  identity: Identity;
-  facts: Facts[];
-  name?: string;
+export interface MonitorDependenciesRequest {
+  scanResult: ScanResult;
+
+  /** If provided, overrides the default project name (usually equivalent to the root package). */
+  projectName?: string;
   policy?: string;
-  target?: GitTarget | ContainerTarget;
-}
-
-export interface Identity {
-  type: string;
-  targetFile?: string;
-  args?: { [key: string]: string };
-}
-
-export interface Facts {
-  type: string;
-  data: any;
-}
-
-export interface Issue {
-  pkgName: string;
-  pkgVersion?: string;
-  issueId: string;
-  fixInfo: {
-    nearestFixedInVersion?: string;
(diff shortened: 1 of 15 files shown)


src/lib/ecosystems.ts
```diff
@@ -44,8 +55,30 @@ export interface TestResult {
...
+  ok: boolean;
+  org: string;
+  id: string;
+  isMonitored: boolean;
+  licensesPolicy: any;
+  uri: string;
+  trialStarted: boolean;
+  path: string;
+  projectName: string;
+  projectType: string;
```
ivanstanev, line 73: These are new -- we need the API to return them because artifacts are resolved on Registry's side

src/lib/ecosystems.ts
```diff
@@ -44,8 +55,30 @@ export interface TestResult {
...
+}
+
 export interface EcosystemPlugin {
-  scan: (options: Options) => Promise<ScanResult[]>;
+  scan: (
+    root: string,
+    targetFile?: string,
```
ivanstanev, line 79: Are we looking to only have 1 parameter `options`? And `root` and `targetFile` are fields of `options`?

src/lib/ecosystems.ts
```diff
@@ -82,7 +119,8 @@ export async function testEcosystem(
   const scanResultsByPath: { [dir: string]: ScanResult[] } = {};
   for (const path of paths) {
     options.path = path;
-    const results = await plugin.scan(options);
+    // TODO
+    const results = await plugin.scan('', undefined, options);
```
ivanstanev, line 123: Undecided what to do with the first 2 params, that's why it's a TODO

src/lib/ecosystems.ts
```diff
@@ -140,3 +177,168 @@ export async function testDependencies(scans: {
...
+export async function monitorEcosystem(
+  ecosystem: Ecosystem,
+  paths: string[],
+  options: Options,
+): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
+  const plugin = getPlugin(ecosystem);
+  const scanResultsByPath: { [dir: string]: ScanResult[] } = {};
+  for (const path of paths) {
+    options.path = path;
+    // TODO: What if this throws?
```
ivanstanev, line 190: Also relevant for `testEcosystem()`

src/lib/ecosystems.ts
```diff
@@ -140,3 +177,168 @@ export async function testDependencies(scans: {
...
+      try {
+        const response = await makeRequest<EcosystemMonitorResult>(payload);
+        results.push({
+          ...response,
+          path,
+        });
+      } catch (error) {
+        if (error.code >= 400 && error.code < 500) {
+          throw new Error(error.message);
+        }
```
ivanstanev, line 275: Not yet sure how this is handled

src/lib/ecosystems.ts
gitphill: yep I think anything that is a CLI flag should be in options, which i think targetFile is

gitphill: i think we also put 'path' in options, which is equivalent to root

src/lib/ecosystems.ts
gitphill: we catch that higher up in the call stack https://github.com/snyk/snyk/blob/master/src/cli/commands/test/index.ts#L127

src/lib/ecosystems.ts
```diff
@@ -140,3 +177,168 @@ export async function testDependencies(scans: {
...
+  return [monitorResults, errors];
+}
+
+function generateMonitorMeta(options, packageManager?): MonitorMeta {
+  return {
+    method: 'cli',
+    packageManager,
+    'policy-path': options['policy-path'],
+    'project-name': options['project-name'] || config.PROJECT_NAME,
+    isDocker: !!options.docker,
```
gitphill, line 207: it's back - can we get rid of it 😄

src/lib/ecosystems.ts
```diff
@@ -140,3 +177,168 @@ export async function testDependencies(scans: {
...
+export async function monitorDependencies(
+  scans: {
+    [dir: string]: ScanResult[];
+  },
+  options: Options,
+): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
+  const results: EcosystemMonitorResult[] = [];
+  const errors: EcosystemMonitorError[] = [];
+  for (const [path, scanResults] of Object.entries(scans)) {
+    await spinner(`Testing dependencies in ${path}`);
```
gitphill, line 243: Monitoring dependencies in?

src/lib/project-metadata/index.ts
```diff
@@ -12,7 +13,7 @@ interface Options {
   isDocker?: boolean; // isDocker coming from MonitorMeta
 }
 export async function getInfo(
-  scannedProject: ScannedProject,
+  scannedProject: ScannedProject | ScanResult,
```
gitphill, line 16: not sure i like us mixing these two together

src/lib/ecosystems.ts
```diff
@@ -140,3 +177,168 @@ export async function testDependencies(scans: {
...
+        method: 'PUT',
+        url: `${config.API}/monitor-dependencies`,
+        json: true,
+        headers: {
+          'x-is-ci': isCI(),
+          authorization: 'token ' + snyk.api,
+        },
+        body: {
+          scanResult: scanResultWithSerializedArtifacts,
+          target,
```
gitphill, line 263: is target not something we can encapsulate in a scan result?

PR #1424, opened 2020-09-25
feat: update scan result test dependencies

gitphill:
- [X] Ready for review

#### What does this PR do?
Add facts and identity to the scan result interface used to test dependencies.

See https://github.com/snyk/snyk-cpp-plugin/pull/16

### DIFF
--- a/src/lib/ecosystems.ts
+++ b/src/lib/ecosystems.ts
@@ -8,17 +8,36 @@ import { Options } from './types';
 import { TestCommandResult } from '../cli/commands/types';
 import * as spinner from '../lib/spinner';
 
-export interface Artifact {
-  type: string;
-  data: any;
-  meta: { [key: string]: any };
+export interface PluginResponse {
+  scanResults: ScanResult[];
+}
+
+export interface GitTarget {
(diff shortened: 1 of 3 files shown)


src/lib/ecosystems.ts
```diff
@@ -122,8 +141,8 @@ export async function testDependencies(scans: {
           authorization: 'token ' + snyk.api,
         },
         body: {
-          artifacts: scanResult.artifacts,
-          meta: {},
+          facts,
+          identity,
         },
```
ivanstanev, line 146: I think we should send the whole ScanResult object as the request body, or at least something like:
```typescript
body: {
  ...scanResult,
},
```
so it allows us to add some optional things here (like the `method`).

The reason we need the whole ScanResult is because the `snyk-docker-plugin` will also return `target`

gitphill: yep ok 👍  will change

ivanstanev (approved):

anthogez (approved):

anthogez (commented):
Needs to update ecosystems.spec.ts

PR #1428, opened 2020-09-29
fix: sarif descirption change

RotemS:
- [x] Ready for review

#### What does this PR do?

Small change to SARIF output description, severity to look the same in IaC and containers.

### DIFF
--- a/src/cli/commands/test/sarif-output.ts
+++ b/src/cli/commands/test/sarif-output.ts
@@ -1,4 +1,5 @@
 import * as sarif from 'sarif';
+import { upperFirst } from 'lodash';
 
 export function createSarifOutputForContainers(testResult): sarif.Log {
   const sarifRes: sarif.Log = {
@@ -40,7 +41,9 @@ export function getTool(testResult): sarif.Tool {
       return {
         id: vuln.id,
         shortDescription: {
-          text: `${vuln.severity} severity ${vuln.title} vulnerability in ${vuln.packageName}`,
+          text: `${upperFirst(vuln.severity)} severity - ${
+            vuln.title
+          } vulnerability in ${vuln.packageName}`,
         },
         fullDescription: {
           text: cve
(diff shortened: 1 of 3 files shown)


benlaplanche (commented):
looks good to me.

src/cli/commands/test/sarif-output.ts
```diff
@@ -1,4 +1,5 @@
 import * as sarif from 'sarif';
+import upperFirst = require('lodash/upperFirst');
```
JackuB, line 2: You can't use import?
```suggestion
import upperFirst from 'lodash/upperFirst';
```

anthogez, line 2: I would like to recommend `import { upperFirst } from 'lodash';` instead of `import upperFirst = require('lodash/upperFirst');`

JackuB (approved):

src/cli/commands/test/sarif-output.ts
anthogez: Would be nice also to fix the commit message to `fix: sarif description change`

anthogez: Would be nice also to fix the commit message to something that is more related to what you have added into the description of your pr:

`small change to SARIF output description, severity to look the same in IaC and containers.`

To give more context to our users of what is happening

RotemS: This was actually a change requested on the PR that introduced this functionality, I originally used your suggestion :) Don't mind changing if needed

anthogez: We are reducing the usage of `require` giving the preference to `import`

anthogez: This solution would not help because `upperFirst` was exported as `export = upperFirst;`
and in order to allow it we need to change `tsconfig.json` **compilerOptions** and add a new property
`"esModuleInterop": true,`

RotemS: Was a request in the original PR. Considering the two comments on this now, changed 👍

anthogez (approved):

almog27 (approved):

PR #1421, opened 2020-09-22
Refactor/remove k8s kind filter

p0tr3c:
- [x] Ready for review
- [x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules
- [ ] Reviewed by Snyk internal team

#### What does this PR do?

Remove k8s object filter based on `kind` type
Only check if k8s object has all required fields

#### How should this be manually tested?

snyk iac test pod.yaml

#### Any background context you want to provide?

Instead of checking for specific object types, validate that the k8s object is valid 
according to upstream documentation(https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields)

This will remove requirement to update CLI if we want to introduce security checks for new 
type of k8s objects

PR #1430, opened 2020-09-29
feat: iac test directory support

almog27:
- [ ] Ready for review

#### What does this PR do?
This PR adds Directory support for `snyk iac test`.
Before this PR, only a single file was supported. This PR enables us also to scan a directory of IaC files (with various IaC types) and get their result in a similar way that `--all-projects` is working on.

Small note - we will need to merge this after #1421 will be merged - however it will not affect the functionality of this PR.

This PR using the Smoke tests functionality.

#### Screenshots
Screen Shot 2020-09-29 at 13 59 58


src/lib/detect-iac.ts
```diff
@@ -0,0 +1,121 @@
+import * as fs from 'fs';
```
almog27, line 1: In this file, we separated the IaC `detect` functionality from the rest of the `detect` functionality.
Most of the code here was moved from `src/lib/detect.ts` file

JackuB (changes requested):
We are too worried about this PR not updating any tests. We would like to see new Jest tests for the new behavior - e.g. detect-iac methods and something to check the return types.

Also CI tests are still failing on cli-args test 🙏

almog27:
> We are too worried about this PR not updating any tests. We would like to see new Jest tests for the new behavior - e.g. detect-iac methods and something to check the return types.
> 
> Also CI tests are still failing on cli-args test 🙏

@JackuB , Yes, you are right there are some tests that needed to be edited that causing the CI tests to fail.
Will update them after the other pr will be merged.

Regarding the jest test - as the Terraform, it more complicated to use the Jest but will check it again

rontalx (commented):
we're really forcing this code-path to work for directories...
if this isn't useful from a product perspective, following our discussion, let's see if we can hack it to work with full recursive directory scanning.

rontalx (approved):
Approving this just because we are tight on schedule with this feature.
Following our recent team discussions on the CLI refactor and our Tech-Roadmap, this PR marks a point where we stop adding new functionality to the CLI until we properly refactor our logic - as it's no longer reasonably maintainable, testable, and sustainable.
cc: @JackuB

JackuB:
Screenshot 2020-10-15 at 17 05 27

Ran into a small issue with error not being returned correctly `NotSupportedIacFileError is not a constructor`
When running `node ./dist/cli/index.js iac test ./test/fixtures/iac/terraform/ --all-projects` locally. Maybe it's not exported properly?

JackuB (approved):
Small issue with --all-projects is not blocking, as it's an unsupported flag combination.

Please squash or rename the commits (see Danger message) so it follows contributing guidelines https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md#commit-messages so that our Release notes make sense https://github.com/snyk/snyk/releases

PR #1432, opened 2020-09-30
Pass integration name and version to UTM parameters

JackuB:
This PR propagates integration names to the UTM parameters used for `$ snyk auth` command.

You can check the logic https://github.com/snyk/snyk/pull/1432/files#diff-a66b785903d0b694ec783784733b20a3, but in general it will result in this:

- UTM_MEDIUM = 'cli' 
- UTM_SOURCE = 'cli'
- UTM_CAMPAIGN = integrationName https://github.com/snyk/snyk/blob/master/src/lib/analytics-sources.ts#L14 e.g. `JENKINS` 
- UTM_CAMPAIGN_CONTENT = integrationVersion e.g. `2.12.1`

HAMMER-117

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -7,6 +7,7 @@ import { Spinner } from 'cli-spinner';
 import * as snyk from '../../../lib';
 import { verifyAPI } from './is-authed';
 import { isCI } from '../../../lib/is-ci';
+import { args as argsLib } from '../../args';
 import * as config from '../../../lib/config';
 import request = require('../../../lib/request');
 import { CustomError } from '../../../lib/errors';
@@ -37,7 +38,9 @@ async function webAuth(via: AuthCliCommands) {
 
   let urlStr = authUrl + '/login?token=' + token;
 
-  const utmParams = getQueryParamsAsString();
+  // It's not optimal, but I have to parse args again here. Alternative is reworking everything about how we parse args
+  const args = [argsLib(process.argv).options];
+  const utmParams = getQueryParamsAsString(args);
   if (utmParams) {
     urlStr += '&' + utmParams;
   }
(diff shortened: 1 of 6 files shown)


src/cli/commands/auth/index.ts
```diff
@@ -37,7 +38,9 @@ async function webAuth(via: AuthCliCommands) {
 
   let urlStr = authUrl + '/login?token=' + token;
 
-  const utmParams = getQueryParamsAsString();
+  // I have to do it this way. Alternative is reworking everything about how we parse args
```
maxjeffos, line 41: I think I get it but a better comment might include what "this" is and why it's perhaps not as ideal as you'd like.

test/analytics-sources.test.ts
```diff
@@ -48,18 +48,21 @@ test('Integration version is loaded from envvar', (t) => {
...
+  t.equal(
+    getIntegrationName([{ integrationName: 'homebrew' }] as any),
+    'HOMEBREW',
+  );
   t.end();
 });
 
 test('Integration name is loaded and validated from CLI flag', (t) => {
-  t.equal(getIntegrationName([{ integrationName: 'invalid' }]), '');
+  t.equal(getIntegrationName([{ integrationName: 'invalid' }] as any), '');
```
maxjeffos, line 59: Why the `as any` here?

test/query-strings.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+describe('Query strings', () => {
+  it('returns a string', () => {
+    expect(typeof getQueryParamsAsString([])).toBe('string');
+  });
+
+  it("returns a string that's a valid URL query string", () => {
+    // Object.fromEntries is not available for Node 8 and 10. Testing this in latest Node is enough
+    if (Object.fromEntries) {
+      expect(
+        Object.fromEntries(new url.URLSearchParams(getQueryParamsAsString([]))),
```
maxjeffos, line 14: Nice

maxjeffos (approved):
Looks great. Left a couple comments / questions.

test/analytics-sources.test.ts
JackuB: Our typing on params is not great, so I've added an extra `{
  _: [],
  _doubleDashArgs: [],
}` rather than the `any`

PR #1433, opened 2020-09-30
Cleaner Smoke test output

JackuB:
This PR clears the output of Smoke test commands, by silencing some before/after hooks. And fixes an issue on Windows.

### Before

Screenshot 2020-09-30 at 13 55 53

### After

Screenshot 2020-09-30 at 13 56 28

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -37,12 +37,12 @@ if (danger.github && danger.github.pr) {
   }
 
   // Forgotten tests check
-  const modifiedTest = danger.git.modified_files.some((f) =>
-    f.startsWith('test/'),
-  );
-  const modifiedSrc = danger.git.modified_files.some((f) =>
-    f.startsWith('src/'),
-  );
+  const modifiedTest =
+    danger.git.modified_files.some((f) => f.startsWith('test/')) ||
+    danger.git.created_files.some((f) => f.startsWith('test/'));
+  const modifiedSrc =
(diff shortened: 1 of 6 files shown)


dangerfile.js
```diff
@@ -54,6 +54,7 @@ if (danger.github && danger.github.pr) {
   // Smoke test modification check
   const modifiedSmokeTest =
     danger.git.modified_files.some((f) => f.startsWith('test/smoke/')) ||
+    danger.git.created_files.some((f) => f.startsWith('test/smoke/')) ||
     danger.git.modified_files.includes(SMOKE_TEST_WORKFLOW_FILE_PATH);
 
   const isOnSmokeTestBranch = danger.github.pr.head.ref === SMOKE_TEST_BRANCH;
```
maxjeffos, line 60: @JackuB what do you think of making this look for a branch that starts with `feat/smoke-test` as opposed to being equal to `feat/smoke-test`? This way we can support more than one open branch for working on smoke tests.

JackuB: Ah, didn't know you can have patterns in GH Actions triggers as well https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags

Hmm, we could make a new branch prefix `smoke/**`? WDYT?

JackuB: Added in latest commit

pavel-snyk (approved):

PR #1434, opened 2020-09-30
feat: updated snyk-docker-plugin to v3.26.2

ahmed-agabani-snyk:
- [x] Ready for review

#### What does this PR do?

updates `snyk-docker-plugin` dependency from `3.24.0` to `3.26.2`

* [v3.25.0](https://github.com/snyk/snyk-docker-plugin/releases/tag/v3.25.0) add platform info to CLI output
* [v3.26.0](https://github.com/snyk/snyk-docker-plugin/releases/tag/v3.26.0) display scan result as human readable format
* [v3.26.1](https://github.com/snyk/snyk-docker-plugin/releases/tag/v3.26.1) display scan result as human readable format
* [v3.26.2](https://github.com/snyk/snyk-docker-plugin/releases/tag/v3.26.2) fix --app-vulns flag parsing

#### Additional questions

1) should the commit message be a `fix` or a `feat`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "semver": "^6.0.0",
     "snyk-config": "3.1.1",
     "snyk-cpp-plugin": "2.0.0",
-    "snyk-docker-plugin": "3.25.0",
+    "snyk-docker-plugin": "3.26.1",
     "snyk-go-plugin": "1.16.2",
     "snyk-gradle-plugin": "3.7.0",
     "snyk-module": "3.1.0",
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "semver": "^6.0.0",
     "snyk-config": "3.1.1",
     "snyk-cpp-plugin": "2.0.0",
-    "snyk-docker-plugin": "3.24.0",
+    "snyk-docker-plugin": "3.25.0",
     "snyk-go-plugin": "1.16.2",
     "snyk-gradle-plugin": "3.7.0",
     "snyk-module": "3.1.0",
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "semver": "^6.0.0",
     "snyk-config": "3.1.1",
     "snyk-cpp-plugin": "2.0.0",
-    "snyk-docker-plugin": "3.26.1",
+    "snyk-docker-plugin": "3.26.2",
     "snyk-go-plugin": "1.16.2",
     "snyk-gradle-plugin": "3.7.0",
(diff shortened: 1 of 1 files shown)


pavel-snyk (approved):

ivanstanev:
@ahmed-agabani-snyk before you merge I have a request: we added the release notes that appear above ☝️ to remind us what our customers will see in the release notes for the new CLI. So imagine a user who reads the notes and sees "updated docker plugin to v x.y.z". What does it mean for them in terms of features? Why would they want to upgrade?
Instead a better message, if possible to find into a single line, would be to describe which features they are getting with this upgrade (e.g. "fixed container --app-vuln flag parsing")

ivanstanev:
@ahmed-agabani-snyk Rebasing should be OK for this repo, approvals will be preserved 🙏 

I'm not sure what's best in this case, we don't have guidelines on this yet I believe 😄  1 commit for every feature for example would be annoying when bringing in many features. 

Unfortunately we have a line limit (~72ish? characters) so a long message won't work and will be auto-trimmed. In this specific case I would suggest `feat: platform info to CLI output, fix --app-vulns flag parsing`. I've also pinged @Shesekino for an opinion 🙏

Shesekino:
👋
There's two main constraints here that I see:
1. Have a clear, informative commit history for engineers working on the CLI and trying to understand what each commit/change means.
2. Have clear, informative release notes for users consuming the CLI.

(1) is easily achieved even without a too-informative commit header: we can elaborate on the changes introduced in the commit message.
(2) is a bit more tricky because the release notes are built automatically by Semantic Release, which picks only the commit header.

Two viable options I can suggest:
a. Go with @ahmed-agabani-snyk 's suggestion https://github.com/snyk/snyk/pull/1434#issuecomment-702167334. It's a bit cumbersome, but could have great value if for example we discover that ```3.26.2``` is broken but ```3.26.1```, so it's easier to revert.
b. Keep using a single commit, with all the necessary details in the commit body. After Semantic Release takes place, manually edit the release notes with the proper information.

What do you think?

Shesekino:
> I've updated the commit history to reflect [#1434 (comment)](https://github.com/snyk/snyk/pull/1434#issuecomment-702167334).
> 
> I'm still trying to work out how to get the CI to run against every single commit without me having to push them one by one, anyone know how?

I don't think we do that. We use merge commits and merge all the changes with an all-or-nothing approach, within the context of a single pull request.
For multiple CI occurrences you can split it to multiple PRs :)

anthogez (approved):

PR #1439, opened 2020-10-05
feat: decrease to warning unsupported package manager error

SarahU:
when running with the --reachable flag on an unsupported
package manager, we would not stop the scan.

this is useful for having a global configuration that enables
this flag and want keep on testing with other package managers.

- [x] Ready for review

#### What does this PR do?
When using `snyk test --reachable` on a package manager that is not yet supported by ReachableVulns it used to stop running the cli and say Reachable Vulns is not supported... This has been changed to a warning message that displays at the bottom of the `snyk test` output.

Notice that the unit test is gone and these is a new acceptance test. The exception is no longer thrown. Instead an alert is saved for later display to the user. In the acceptance test, the alert is looked for 

run `snyk test --reachability` on an npm project. See warning at the end of cli output.

When one uses --reachable with snyk test, for a non-supported package manager, it used to throw an error and stop. This was a problem for automated CI when clients ran multiple projects through templated pipelines. This removes that stop and instead will run snyk test as normal and display a warning at the end that the particular package manager is not yet supported for reachability

https://snyksec.atlassian.net/browse/FLOW-435

#### Screenshots
Screenshot 2020-10-05 at 20 32 16


src/lib/errors/feature-not-supported-by-package-manager-error.ts
```diff
@@ -2,13 +2,16 @@ import { CustomError } from './custom-error';
 import { SupportedPackageManagers } from '../package-managers';
 
 export class FeatureNotSupportedByPackageManagerError extends CustomError {
+  public feature: string;
```
darmalovan, line 5: We can probably make this a readonly field?

test/acceptance/cli-test/cli-test.npm.spec.ts
```diff
@@ -251,5 +252,16 @@ export const NpmTests: AcceptanceTests = {
         'depGraph looks fine',
       );
     },
+
+    'test npm-package --reachable is not supported': (params, utils) => async (
```
darmalovan, line 256: What do you think for the test name `test --reachable is not supported for npm`

SarahU: Can change

darmalovan (approved):

src/lib/errors/feature-not-supported-by-package-manager-error.ts
SarahU: Have updated this to be readonly

test/acceptance/cli-test/cli-test.npm.spec.ts
SarahU: Have updated the name

pavel-snyk (approved):

PR #1447, opened 2020-10-06
fix: woof command

JackuB:
Woof command stopped working after #1441 

This PR fixes the functionality and introduces a woof smoke test

### DIFF
--- a/src/cli/commands/woof/getWoof.ts
+++ b/src/cli/commands/woof/getWoof.ts
@@ -1,0 +1,23 @@
+import { MethodArgs, ArgsOptions } from '../../args';
+
+const woofs = {
+  en: 'Woof!',
+  he: ' !הב ',
+  ru: ' Гав!',
+  es: 'Guau!',
+  cs: ' Haf!',
+};
+
+export default function getWoof(args: MethodArgs): string {
+  const options = args.pop() as ArgsOptions;
+  let lang = 'en';
+
+  if (
+    typeof options.language === 'string' &&
+    Object.keys(woofs).includes(options.language)
+  ) {
+    lang = options.language;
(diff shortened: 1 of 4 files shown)


Avishagp (approved):

src/cli/commands/woof/getWoof.ts
```diff
@@ -0,0 +1,23 @@
...
+  es: 'Guau!',
+  cs: ' Haf!',
+};
+
+export default function getWoof(args: MethodArgs): string {
+  const options = args.pop() as ArgsOptions;
+  let lang = 'en';
+
+  if (
+    typeof options.language === 'string' &&
```
maxjeffos, line 16: Because bike shedding is fun, I personally find this kind of hard to read.
It doesn't really matter, but to me something like this would be easier to follow:
```
  const options = args.pop() as ArgsOptions;
  const defaultLanguage = 'en';
  const language = (options.language as string) || defaultLanguage;
  
  if (language in woofs) {
    return woofs[language];
  } else {
    return woofs[defaultLanguage];
  }
```

This not a blocker for the PR, just me having an opinion.

maxjeffos (approved):
Nice. I especially like what you did for testing - having both some unit tests (in Jest) and Smoke tests. :+1:

src/cli/commands/woof/getWoof.ts
JackuB: Ha, that's some heavy bikeshedding!

Yours sets the default value twice!!1

giphy

I'll merge is so that Woof is fixed :)

maxjeffos: Good idea lol

PR #1448, opened 2020-10-07
fix: correct match check for package.json when not part of a workspace

lili2311:
- [x] Ready for review

#### What does this PR do?

When a workspace repo had both workspace & non workspace manifests the non-workspace manifests were not being filtered out because match is an `[]` not a boolean. Fixed this & had to update slightly the logic to separate rootPackageJson of a workspace vs a package in a workspace to use different logic to calculate where the lockfile is.

Added a orphaned package.json which would be picked up and processed without this fix so the existing tests cover this after adding this extra file.


test/acceptance/cli-test/cli-test.yarn-workspaces.spec.ts
```diff
@@ -35,94 +34,31 @@ export const YarnWorkspacesTests: AcceptanceTests = {
       utils,
     ) => async (t) => {
       utils.chdirWorkspaces();
-      await params.cli.test('yarn-workspace-out-of-sync', {
+      const result = await params.cli.test('yarn-workspace-out-of-sync', {
         dev: true,
         strictOutOfSync: false,
+        yarnWorkspaces: true,
```
lili2311, line 40: this was missing!

src/cli/args.ts
```diff
@@ -40,6 +40,7 @@ const DEBUG_DEFAULT_NAMESPACES = [
   'snyk-gradle-plugin',
   'snyk-sbt-plugin',
   'snyk-mvn-plugin',
+  'snyk:yarn-workspaces',
```
gitphill, line 43: whats the reason for colon here (and not hyphen like the others)?

lili2311: thats the string registered in the file I can update it 
https://github.com/snyk/snyk/pull/1448/files#diff-2bee611aba1b36e848d4beddf5aa8346R5

gitphill (approved):

PR #1471, opened 2020-10-14
Set integration name for Homebrew installs

maxjeffos:
- [X] Ready for review

#### What does this PR do?
Adds the integration name`HOMEBREW` when the running CLI instances is installed with Homebrew.

HAMMER-135

### DIFF
--- a/src/lib/analytics-sources.ts
+++ b/src/lib/analytics-sources.ts
@@ -11,6 +11,7 @@
 const debug = require('debug')('snyk');
 import * as fs from 'fs';
 import { ArgsOptions } from '../cli/args';
+import { join } from 'path';
 
 export const INTEGRATION_NAME_HEADER = 'SNYK_INTEGRATION_NAME';
 export const INTEGRATION_VERSION_HEADER = 'SNYK_INTEGRATION_VERSION';
@@ -50,10 +51,13 @@ enum TrackedIntegration {
 }
 
 export const getIntegrationName = (args: ArgsOptions[]): string => {
+  const maybeHomebrew = isHomebrew() ? 'HOMEBREW' : '';
   const maybeScoop = isScoop() ? 'SCOOP' : '';
+
   const integrationName = String(
     args[0]?.integrationName || // Integration details passed through CLI flag
       process.env[INTEGRATION_NAME_HEADER] ||
+      maybeHomebrew ||
       maybeScoop ||
       '',
   ).toUpperCase();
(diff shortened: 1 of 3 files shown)


src/lib/analytics-sources.ts
```diff
@@ -114,3 +118,27 @@ export function validateScoopManifestFile(snykExecutablePath: string): boolean {
   }
   return false;
 }
+
+export function isHomebrew(): boolean {
+  const currentProcessPath = process.execPath;
+  const isHomebrewPath = currentProcessPath.includes('Cellar/snyk');
```
JackuB, line 124: I'd consider extending the match

```suggestion
  const isHomebrewPath = currentProcessPath.includes('/Cellar/snyk/');
```

test/analytics-sources.spec.ts
```diff
@@ -47,6 +51,62 @@ describe('analytics-sources - scoop detection', () => {
   });
 });
 
+describe('analytics-sources - Homebrew detection', () => {
+  afterEach(() => {
+    // jest.resetAllMocks();
```
JackuB, line 56: Some leftovers?

test/analytics-sources.spec.ts
```diff
@@ -47,6 +51,62 @@ describe('analytics-sources - scoop detection', () => {
...
+      expect(isHomebrew()).toBe(false);
+      expect(fileExistsSpy).not.toHaveBeenCalled();
+    } finally {
+      process.execPath = originalExecPath;
+    }
+  });
+
+  it('returns false if formula files does not exist', () => {
+    const originalExecPath = process.execPath;
+    try {
```
JackuB, line 88: Is it worth try catching? If it throws, we would want the test to fail, no?
Same for the other tests that are using try/catch

JackuB (approved):
Small comment on the test format, but looks solid 👍

test/analytics-sources.spec.ts
maxjeffos: :green_tick:

src/lib/analytics-sources.ts
maxjeffos: good idea - done. thanks

test/analytics-sources.spec.ts
maxjeffos: Ya you're right. I thought it was nice to have the reset stuff separated into a finally like this to make it more obvious but effectively it's the same and just takes up more space.

PR #1473, opened 2020-10-15
Fix dangerfile tap test detection

JackuB:
We are getting false positives because `!![] => true`, so you need to check for the actual array length because `!!0 => false`


Avishagp (approved):
LGTM

PR #1479, opened 2020-10-16
Analytics: track integration environment

JackuB:
This will help us capture debug/analytics data if CLI is running e.g. in a specific IDE version.

### DIFF
--- a/test/analytics-sources.spec.ts
+++ b/test/analytics-sources.spec.ts
@@ -1,8 +1,12 @@
 import {
   getIntegrationName,
   getIntegrationVersion,
-  INTEGRATION_NAME_HEADER,
-  INTEGRATION_VERSION_HEADER,
+  getIntegrationEnvironment,
+  getIntegrationEnvironmentVersion,
+  INTEGRATION_NAME_ENVVAR,
+  INTEGRATION_VERSION_ENVVAR,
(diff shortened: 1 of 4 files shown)


test/analytics-sources.spec.ts
```diff
@@ -162,3 +168,41 @@ describe('analytics-sources - getIntegrationVersion', () => {
     ).toBe('1.2.3-Crystal');
   });
 });
+
+describe('analytics-sources - getIntegrationEnvironment', () => {
+  it('integration environment is empty by default', () => {
```
maxjeffos, line 173: Minor nitpick and I'm by no means perfect at this myself (naming is hard), but I understand that one should try to make the _describe_ + the _it_ sort of flow like a sentence. So you could make this 
```
it('returns empty environment by default', () => {
```
Then you'd have `getIntegrationEnvironment returns empty environment by default`.

Similarly for the others.

maxjeffos (approved):
Looks good other than needing to modify `test/analytics.test.ts` to add 
`integrationEnvironment` and `integrationEnvironmentVersion` to a few spots to fix the tests.

JackuB:
Rebased, simplified the function to always empty empty string instead of undefined, just to be consistent with the `integrationName` as well.

PR #1480, opened 2020-10-16
feat: paths always shown where available

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?
Currently paths for top-level dependencies that have a vuln are only shown if the `all` flag is passed to `--show-vulnerable-paths`. This can be a bit confusing and some people don't understand why they're not seeing an "introduced by:" line for these top-level deps (e.g. if you have a direct dependency of `adm-zip@0.4.7`)

This PR changes that so that whenever a path is available it will also be shown when the default of `some` is used for `show-vulnerable-paths`

### DIFF
--- a/src/cli/commands/test/formatters/remediation-based-format-issues.ts
+++ b/src/cli/commands/test/formatters/remediation-based-format-issues.ts
@@ -467,10 +467,11 @@ export function formatIssue(
   }
 
   let introducedBy = '';
+
   if (
     testOptions.showVulnPaths === 'some' &&
     paths &&
-    paths.find((p) => p.length > 2)
+    paths.find((p) => p.length > 1)
   ) {
     // In this mode, we show only one path by default, for compactness
     const pathStr = printPath(paths[0]);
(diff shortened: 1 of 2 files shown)


gitphill (approved):

src/cli/commands/test/formatters/remediation-based-format-issues.ts
```diff
@@ -467,10 +467,11 @@ export function formatIssue(
   }
 
   let introducedBy = '';
+
   if (
     testOptions.showVulnPaths === 'some' &&
     paths &&
-    paths.find((p) => p.length > 2)
+    paths.find((p) => p.length > 1)
```
JackuB, line 474: Just curios, should this affect JSON output as well?

JackuB (approved):

src/cli/commands/test/formatters/remediation-based-format-issues.ts
iamtmrobinson: @JackuB off the top of your head, do you know if the JSON output is affected by the `--show-vulnerable-paths` option?

PR #1485, opened 2020-10-21
Add metrics framework and network time metric

maxjeffos:
- [X] Ready for review

#### What does this PR do?
- adds a metrics collecting framework so that we can time things, etc
- collects total network time
- collects "cpu time" which is roughly estimated as total time - network time, understanding that it is not quite that simple

[Corresponding PR on registry](https://github.com/snyk/registry/pull/16616)

- HAMMER-189
- the motivation was to have some data for when we go and tackle other improvements/fixes, including HAMMER-189


src/lib/request/index.ts
```diff
@@ -21,5 +27,8 @@ export = async (
       return callback(error);
     }
     throw error;
+  } finally {
+    // call stop again here in case there's an error and the first .stop() is skipped over
+    totalNetworkTimeTimer.stop();
```
JackuB, line 32: Shouldn't a single `stop` call in `finally` be enough? You could say that even the response processing is part of the network

src/lib/analytics.js
```diff
@@ -79,6 +80,7 @@ function postAnalytics(data) {
 
       data.ci = isCI();
       data.durationMs = Date.now() - startTime;
+      data.metrics = MetricsCollector.getInstance().getAllMetricsData();
```
JackuB, line 83: I think I'd prefer we set these keys explicitly and prevent (accidental) dynamic fields loaded

src/lib/metrics.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+}
+
+export class NamedMetrics {
+  public static NETWORK_TIME: MetricId = new MetricId(
+    'network-time',
+    'amount of time in ms spent waiting on network requests',
+    METRIC_TYPE_TIMER,
+  );
+}
```
JackuB, line 208: I'm a bit worried that this file is 200 lines alone 🤔 is there something to simplify the interface? Maybe moving the Counter from standalone class?

src/lib/metrics.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (this.startTimeMs === 0) {
+      this.startTimeMs = Date.now();
+    } else {
+      MetricsCollector.getInstance().logImproperUse(
+        'Invalid Timer use: start() called when timer already stopped',
+      );
+    }
+  }
+
+  public stop() {
```
JackuB, line 175: I'd consider debug-logging the timer and its value as well

src/lib/request/index.ts
maxjeffos: I was just trying to get the .stop() as close to the network bits as possible and didn't to take those callbacks at line 22 into the timer time. But, sure you could consider processing them as part of the timer - would make it so we just call the timer in the finally which looks prettier.

src/lib/metrics.ts
maxjeffos: I'm not against breaking up the this metrics into multiple files. I think that would be best.

src/lib/analytics.js
maxjeffos: It's very explicit - all keys need to be hard coded in the NameMetrics class.

src/lib/metrics.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (this.endTimeMs === 0) {
+      this.endTimeMs = Date.now();
+      debug(
+        `Timer stopped at ${
```
JackuB, line 38: This wouldn't tell you which timer, no?

src/lib/metrics.ts
```diff
@@ -0,0 +1,142 @@
...
+        name: m.name,
+        type: m.metricType,
+        values: m.getValues(),
+        total: m.getTotal(),
+      };
+      res.push(nextEntry);
+    });
+    return res;
+  }
+}
```
JackuB, line 142: You could probably map these:

```suggestion
  public static getAllMetrics(): any[] {
    const metrics: Metric[] = [
      MetricsCollector.NETWORK_TIME,
      MetricsCollector.CPU_TIME,
    ];
    return metrics.map(m => ({
      name: m.name,
      type: m.metricType,
      values: m.getValues(),
      total: m.getTotal(),
    }));
  }
}
```

src/lib/metrics.ts
```diff
@@ -0,0 +1,142 @@
...
+    return sm;
+  }
+}
+
+export function logImproperUse(msg: string) {
+  debug(msg);
+}
+
+export class MetricsCollector {
+  public static NETWORK_TIME: TimerMetric = new TimerMetric(
```
JackuB, line 114: Never did too much OOP - what's the reason to have both `TimerMetric` and `TimerMetricInstance`?

src/lib/metrics.ts
maxjeffos: Thanks

src/lib/metrics.ts
maxjeffos: `TimerMetricInstance` is the individual thing where the `.start()` and `.stop()` and elapsed time calculation happen (in the `.getValue()`). `TimerMetric` is an abstraction that allows you to have _n_ `TimerMetricInstances`, thus allowing us to have multiple runs of the "same" metric (as we do for `NETWORK-TIME`), be able to get the total, and collect each individual value.

src/lib/metrics.ts
maxjeffos: Yes. Which is not ideal. I didn't want to kludge up (de-normalize, more specifically) the thing by passing the context and type right down into each individual TimerMetricInstance. But I think it's worth it to have it in the log and I thought of a less ugly way to do it. Will update. I'll also add a log for the start().

src/lib/metrics.ts
```diff
@@ -0,0 +1,144 @@
...
+   * Creates a new TimerMetricInstance
+   * @param metricTag used for logging to identify the metric
+   */
+  public constructor(metricTag: string) {
+    super();
+    this.metricTag = metricTag;
+  }
+
+  public getValue(): MetricValue {
+    if (this.startTimeMs != 0 && this.endTimeMs != 0) {
```
JackuB, line 27: Any reason to not use triple equal here?

JackuB (approved):

src/lib/metrics.ts
maxjeffos: no

PR #1491, opened 2020-10-26
feat: get slack alert when smoke tests fail

Avishagp:
#### What does this PR do?
Using GitHub Actions, check the 3 most recent Smoke Tests every hour. If all three of them failed, re-run them and send a Slack alert if the re-run fails as well.

Slack alert

### DIFF
--- a/.github/workflows/cli-alert.yml
+++ b/.github/workflows/cli-alert.yml
@@ -1,0 +1,21 @@
+name: CLI alert
+
+on:
+  schedule:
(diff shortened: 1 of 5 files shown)


JackuB:
- Cleaned up package.json a bit
- rebased
- changed `feat` to `chore` in commit message, because it's not an user-facing feature of a CLI, but a support tool for us

JackuB (approved):

PR #1500, opened 2020-11-02
test(smoke): update k8s assertion

JackuB:
Fix failing Smoke Tests because assertion changed

Are we planning to change those often? Should we only match the ID codes like `SNYK-CC-K8S-6`?


p0tr3c (approved):

p0tr3c:
We are not planning to change these often, but we have backlog ticket to avoid using these values in CLI/Registry tests.

test/smoke/spec/iac/snyk_test_k8s_spec.sh
```diff
@@ -14,7 +14,7 @@ Describe "Snyk iac test command"
       The output should include "Infrastructure as code issues:"
       The output should include "✗ Container is running in privileged mode [High Severity] [SNYK-CC-K8S-1] in Deployment"
       The output should include "  introduced by input > spec > containers[example] > securityContext > privileged"
-      The output should include "✗ Containers should ideally be explicit about required capabilities [Medium Severity] [SNYK-CC-K8S-6] in Deployment"
+      The output should include "✗ Container is running with default set of capabilities [Medium Severity] [SNYK-CC-K8S-6] in Deployment"
       The output should include "  introduced by input > spec > containers[example] > securityContext > capabilities > drop"
       The output should include "✗ Container can run as the root [Medium Severity] [SNYK-CC-K8S-10] in Deployment"
```
p0tr3c, line 19: This will change to `Container is running without root user control`

PR #1507, opened 2020-11-03
fix: support globs pattern in yarn workspaces definitions

lili2311:
- [x] Ready for review

#### What does this PR do?

Yarn workspaces patterns also support *globs* pattern, update the matching to cover this too.
> The workspaces field is an array containing the paths to each workspace. Since it might be tedious to keep track of each of them, this field also accepts glob patterns! For example, Babel reference all of their packages through a single packages/* directive.
https://classic.yarnpkg.com/en/docs/workspaces/

Also added a failing test to show a deep globs pattern in workspaces definitions was not supported until the fix

### DIFF
--- a/test/acceptance/workspaces/yarn-workspaces/package.json
+++ b/test/acceptance/workspaces/yarn-workspaces/package.json
@@ -1,7 +1,8 @@
 {
   "private": true,
   "workspaces": [
-    "packages/*"
+    "packages/*",
+    "libs/**/*"
   ],
   "resolutions": {
     "node-fetch": "^2.3.0"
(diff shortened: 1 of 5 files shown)


test/acceptance/workspaces/yarn-workspaces/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "private": true,
   "workspaces": [
-    "packages/*"
+    "packages/*",
+    "libs/**/*"
```
lili2311, line 5: globs pattern here!

ekbsnyk (approved):

PR #1510, opened 2020-11-05
Retry large object stringify without pretty print

maxjeffos:
- [X] Ready for review

#### What does this PR do?
Try/catch the JSON.stringify where we're creating the JSON/SARIF JSON outputs using pretty print and, if we catch an error, try again without pretty-print.

This is because in very rare cases, we saw calls to JSON.stringify throw an `Invalid string length` error and in some cases we can work around this by outputting raw (non-pretty-printed) JSON rather than pretty-printed JSON.

HAMMER-214

### DIFF
--- a/src/cli/commands/test/formatters/format-test-results.ts
+++ b/src/cli/commands/test/formatters/format-test-results.ts
@@ -26,6 +26,7 @@ import { formatDockerBinariesIssues } from './docker';
 import { createSarifOutputForContainers } from '../sarif-output';
 import { createSarifOutputForIac } from '../iac-output';
 import { isNewVuln, isVulnFixable } from '../vuln-helpers';
+import { jsonStringifyLargeObject } from '../../../../lib/json';
 
 export function formatJsonOutput(jsonData) {
   const jsonDataClone = _.cloneDeep(jsonData);
@@ -52,12 +53,10 @@ export function extractDataToSendFromResults(
       : createSarifOutputForIac(results);
   }
 
-  const stringifiedJsonData = JSON.stringify(
+  const stringifiedJsonData = jsonStringifyLargeObject(
     formatJsonOutput(jsonData),
-    null,
-    2,
   );
-  const stringifiedSarifData = JSON.stringify(sarifData, null, 2);
+  const stringifiedSarifData = jsonStringifyLargeObject(sarifData);
 
   const dataToSend = options.sarif ? sarifData : jsonData;
   const stringifiedData = options.sarif
--- a/src/lib/json.ts
+++ b/src/lib/json.ts
@@ -1,0 +1,20 @@
+const debug = require('debug')('snyk-json');
+
+/**
+ * Attempt to json-stringify an object which is potentially very large and might exceed the string limit.
+ * If it does exceed the string limit, try again without pretty-print to hopefully come out below the string limit.
+ * @param obj the object from which you want to get a JSON string
+ */
+export function jsonStringifyLargeObject(obj: any): string {
+  let res = '';
+  try {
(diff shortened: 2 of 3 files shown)


src/lib/json.ts
```diff
@@ -0,0 +1,20 @@
+const debug = require('debug')('snyk-json');
+
+/**
+ * Attempt to json-stringify an object which is potentially very large and might exceed the string limit.
+ * If it does exceed the string limit, try again without pretty-print to hopefully come out below the string limit.
+ * @param obj the object from which you want to get a JSON string
+ */
+export function jsonStringifyLargeObject(obj: any): string {
```
aviadhahami, line 8: we should use `Record<string, unknown>` and not `any`

src/lib/json.ts
```diff
@@ -0,0 +1,20 @@
...
+  let res = '';
+  try {
+    // first try pretty-print
+    res = JSON.stringify(obj, null, 2);
+    return res;
+  } catch (err) {
+    // if that doesn't work, try non-pretty print
+    debug('JSON.stringify failed - trying again without pretty print', err);
```
aviadhahami, line 16: We will also enter the `catch` if `obj` isn't:
* json
* formatted properly

maybe we should if/else these?

src/lib/json.ts
maxjeffos: The JSON.stringify function of the standard library which we call into takes an `any`.

aviadhahami: yup, but we can make it stricter to make sure we don't pass a string for example;
```ts
const f = (v: Record<string, unknown>) => JSON.stringify(v); // this is a valid `any`

Comment (maxjeffos) [reply src/lib/json.ts:8]: `Record<string, unknown>` would restrict arrays which convert to JSON with a top level array which is perfectly valid.

It is odd that JSON.stringify will accept either a number or a string and essentially pass it through directly (in other words, not generate JSON) but I wouldn't want to restrict it to the point where it can't accept arrays.

We could do something like:
```
export type JsonableObject = Record<string, unknown> | unknown[];
export function jsonStringifyLargeObject(obj: JsonableObject): string {
...
```

but I wonder if that's making it unnecessarily complex. What do you think?

File: src/lib/json.ts (hunk 2 above)
Comment (maxjeffos) [reply src/lib/json.ts:16]: The input object isn't json. It should be an object which can be turned into json. Can you give me an example of something that you mean? Perhaps if we implement a more restrictive input type (like you suggested above), we don't need this?

I'm not sure what you mean by formatted properly. How could you get that? Formatted properly sort of implies something that is already converted to a JSON string, not something that we're going to attempt to convert to a JSON string.

File: src/cli/commands/test/formatters/format-test-results.ts [hunk 3]
```diff
@@ -52,12 +53,10 @@ export function extractDataToSendFromResults(
       : createSarifOutputForIac(results);
   }
 
-  const stringifiedJsonData = JSON.stringify(
+  const stringifiedJsonData = jsonStringifyLargeObject(
```
JackuB, line 56: I wonder if it would make sense to have this JSON utility everywhere we are doing stringify?

JackuB (approved):
Once we are clear on the interface :+1:

src/cli/commands/test/formatters/format-test-results.ts
maxjeffos: It might. I thought about that and thought maybe start here where we really need it and see how it works and go from there.

maxjeffos:
I'm closing this PR and incorporating the change into https://github.com/snyk/snyk/pull/1516 instead.

PR #1514, opened 2020-11-05
feat: allow usage of -q or --quiet

aviadhahami:
- [x] Ready for review

#### What does this PR do?
We would like to allow the usage of `-q` (== `--quiet`) as a cli flag in order to allow users to ignore warnings; A classic example will be the print-deps warning for large graphs/trees;

### DIFF
--- a/src/lib/print-deps.ts
+++ b/src/lib/print-deps.ts
@@ -21,15 +21,19 @@ export async function maybePrintDepGraph(
   } else {
     if (options['print-deps']) {
       if (options.json) {
-        console.log(
-          '--print-deps --json option not yet supported for large projects. Displaying graph json output instead',
-        );
+        if (!options.quiet) {
+          console.log(
+            '--print-deps --json option not yet supported for large projects. Displaying graph json output instead',
+          );
+        }
         // TODO @boost: add as output graphviz 'dot' file to visualize?
         console.log(JSON.stringify(depGraph.toJSON(), null, 2));
       } else {
-        console.log(
-          '--print-deps option not yet supported for large projects. Try with --json.',
-        );
+        if (!options.quiet) {
+          console.log(
+            '--print-deps option not yet supported for large projects. Try with --json.',
+          );
+        }
       }
     }
   }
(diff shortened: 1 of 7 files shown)


src/lib/print-deps.ts
```diff
@@ -21,15 +21,19 @@ export async function maybePrintDepGraph(
...
+        }
         // TODO @boost: add as output graphviz 'dot' file to visualize?
         console.log(JSON.stringify(depGraph.toJSON(), null, 2));
       } else {
-        console.log(
-          '--print-deps option not yet supported for large projects. Try with --json.',
-        );
+        if (!options.quiet) {
+          console.log(
+            '--print-deps option not yet supported for large projects. Try with --json.',
```
anthogez, line 34: Please add some tests to validate this alias ^^

src/cli/args.ts
```diff
@@ -26,6 +26,7 @@ const alias = abbrev(
 );
 alias.d = 'debug'; // always make `-d` debug
 alias.t = 'test';
+alias.q = 'quiet';
```
JackuB, line 29: Welp, our docs says we already support it https://support.snyk.io/hc/en-us/articles/360003812578-CLI-reference 🤪

> -q, --quiet Silence all output.

Then this should disable stdout altogether?

JackuB: Screenshot 2020-11-08 at 12 19 29

Yep

aviadhahami: I need to enable stdout, but disable specific error logs.
Maybe `--silence-errors`?

JackuB:
> A classic example will be the print-deps warning for large graphs/trees;

> `--print-deps option not yet supported for large projects. Try with --json.`

Why do we treat CLI _not doing something I asked it to do_ as a warning that I want to silence? Isn't it a failure then? 🤔 Or if it's a case where a log message and not doing something is sensible, then why not fallback to `--json` as well, automatically?

aviadhahami:
Closing as we already support it

PR #1516, opened 2020-11-05
Don't create JSON if we don't need it

maxjeffos:
- [X] Ready for review

#### What does this PR do?
Two important things wrt generating the Snyk-format JSON or SARIF-format JSON output from the test command:
- don't do `JSON.stringify` at all if we don't need it
- produce only the JSON formats we need: if we need Snyk-format JSON then generate that; if we need SARIF-format JSON, then generate that. If we need both (ex if the user uses `--json` and `--sarif-file-output`), then produce both.

We're also try/catching the `JSON.stringify` for the test output and, if it throws, we'll automatically try again without pretty-print. This may help in some scenarios where JSON.stringify is failing with `RangeError: Invalid string length`. Other `JSON.stringify` calls are likely to be tiny by comparison so we don't really need this for those, but we could consider it in the future.

- @lili2311 brought this up because a customer was having an issue related to failing to stringify JSON and found that we're creating the JSON output even when we don't need it and now that we're (optionally) doing SARIF output we're creating JSON twice - once for the regular Snyk format and again for the SARIF format.

HAMMER-182

### DIFF
--- a/src/cli/commands/test/formatters/format-test-results.ts
+++ b/src/cli/commands/test/formatters/format-test-results.ts
@@ -26,6 +26,7 @@ import { formatDockerBinariesIssues } from './docker';
 import { createSarifOutputForContainers } from '../sarif-output';
 import { createSarifOutputForIac } from '../iac-output';
 import { isNewVuln, isVulnFixable } from '../vuln-helpers';
+import { jsonStringifyLargeObject } from '../../../../lib/json';
 
 export function formatJsonOutput(jsonData) {
   const jsonDataClone = _.cloneDeep(jsonData);
@@ -46,18 +47,18 @@ export function extractDataToSendFromResults(
   options: Options,
 ): OutputDataTypes {
   let sarifData = {};
+  let stringifiedSarifData = '';
   if (options.sarif || options['sarif-file-output']) {
     sarifData = !options.iac
       ? createSarifOutputForContainers(results)
       : createSarifOutputForIac(results);
+    stringifiedSarifData = jsonStringifyLargeObject(sarifData);
   }
 
-  const stringifiedJsonData = JSON.stringify(
-    formatJsonOutput(jsonData),
-    null,
-    2,
-  );
-  const stringifiedSarifData = JSON.stringify(sarifData, null, 2);
+  let stringifiedJsonData = '';
+  if (options.json || options['json-file-output']) {
+    stringifiedJsonData = jsonStringifyLargeObject(formatJsonOutput(jsonData));
+  }
 
   const dataToSend = options.sarif ? sarifData : jsonData;
   const stringifiedData = options.sarif
@@ -65,10 +66,10 @@ export function extractDataToSendFromResults(
     : stringifiedJsonData;
 
   return {
-    stdout: dataToSend,
-    stringifiedData,
-    stringifiedJsonData,
-    stringifiedSarifData,
+    stdout: dataToSend, // this is for the human-readable stdout output and is set (but not used) even if --json or --sarif is set
+    stringifiedData, // this will be used to display either the Snyk or SARIF format JSON to stdout if --json or --sarif is set
+    stringifiedJsonData, // this will be used for the --json-file-output=<file.json> option
+    stringifiedSarifData, // this will be used for the --sarif-file-output=<file.json> option
   };
 }
 
--- a/src/lib/json.ts
+++ b/src/lib/json.ts
@@ -1,0 +1,20 @@
+const debug = require('debug')('snyk-json');
+
+/**
+ * Attempt to json-stringify an object which is potentially very large and might exceed the string limit.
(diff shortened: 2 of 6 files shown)


test/format-test-results.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+describe('format-test-results', () => {
+  describe('extractDataToSendFromResults', () => {
+    afterEach(() => {
+      jest.restoreAllMocks();
+    });
+
+    it('should not create any JSON unless it is needed per options', () => {
+      const options = {} as Options;
+      const jsonStringifySpy = jest.spyOn(JSON, 'stringify');
+      extractDataToSendFromResults({}, {}, options);
```
JackuB, line 15: Would it make sense to also check that the un/expected return values are unset - `''`? Or that's covered by another test?

JackuB (approved):

test/json.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { jsonStringifyLargeObject } from '../src/lib/json';
+
+describe('jsonStringifyLargeObject', () => {
+  it('works normally with a small object', () => {
+    const smallObject = {
+      name: 'Brian',
```
orsagie, line 6:
```suggestion
      name: 'Mozart',
```

src/lib/json.ts
```diff
@@ -0,0 +1,20 @@
+const debug = require('debug')('snyk-json');
```
orsagie, line 1: Considering this utility is purely for formatting output, I think it should sit next to the rest of the formatters `src/cli/commands/test/formatters`

maxjeffos: We were thinking we'd probably use it for other parts of the code outside of formatters eventually, so I think it would be good to keep in its own file.

test/json.spec.ts
maxjeffos: Is there a small dog named Mozart @orsagie ?

test/format-test-results.spec.ts
maxjeffos: I updated the tests... had to add a couple fixtures to do this right. LMK what you think @JackuB I'll wait before merging.

src/cli/commands/test/formatters/format-test-results.ts
```diff
@@ -46,29 +47,34 @@ export function extractDataToSendFromResults(
...
+    stringifiedJsonData = jsonStringifyLargeObject(formatJsonOutput(jsonData));
+  }
 
   const dataToSend = options.sarif ? sarifData : jsonData;
-  const stringifiedData = options.sarif
-    ? stringifiedSarifData
-    : stringifiedJsonData;
+
+  let stringifiedData = '';
+  if (options.sarif || options.json) {
```
maxjeffos, line 66: @SarahU you were last in here, so just wanted to make sure you were aware of this change. I'm pretty sure we don't need any value in stringifiedData unless `--json` or `--sarif` are set. The other two outputs - `stringifiedJsonData` and `stringifiedSarifData` are for when we want to save the JSON directly to a file using either `--json-file-output` or `--sarif-file-output`, respectively. @orkamara you may be interested in this as well.
This doesn't change the output for any variants of the flags I mentioned, just makes the code a bit more logical and consistent between the sarif and json options.

SarahU: Hey Jeff. I'm not 100% sure. Hope it's covered by a test TBH but I assume the intention was to have it set as json as default rather that leaving it empty. I don't think that's a problem in particular unless you need it to be empty in any explicit scenario, you can probably leave it as defaulting to the Json data if you want? But if it's only used in the 2 scenarios you mention then probably fine as is. For me, I didn't want to create any changes I didn't need to as I'm not sure of all the combos of flag that can be used / are supported

test/format-test-results.spec.ts
```diff
@@ -0,0 +1,130 @@
+import * as jsonModule from '../src/lib/json';
+
+import { extractDataToSendFromResults } from '../src/cli/commands/test/formatters/format-test-results';
+import { Options } from '../src/lib/types';
+const fs = require('fs');
```
JackuB, line 5: Why not use import like the other files?

```suggestion
import * as fs from 'fs';
```

JackuB (approved):

test/format-test-results.spec.ts
maxjeffos: done

src/cli/commands/test/formatters/format-test-results.ts
maxjeffos: Ok, sounds good @SarahU. I'm fairly sure the change I did is right as I did some manual testing with a bunch of options variants (--json, --json-file-output, --sarif, --sarif-file-output), I added a bunch of unit tests to that function, and didn't break any other tests.

PR #1520, opened 2020-11-10
feat: monitor cpp projects

gitphill:
- [x] Ready for review

#### What does this PR do?
 Monitor C/C++ projects
 
Using new release of snyk-cpp-plugin that adds target and project name. Sending C/C++ fingerprints to Snyk to be saved as a project and monitored over time.


src/lib/plugins/get-extra-project-count.ts
```diff
@@ -8,7 +8,7 @@ export async function getExtraProjectCount(
   options: Options,
   inspectResult: pluginApi.InspectResult,
 ): Promise<number | undefined> {
-  if (options.docker) {
+  if (options.docker || options.source) {
```
anthogez, line 11: I think I will add a test also for `getExtraProjectCount` it's not being tested at all

src/lib/ecosystems/plugins.ts
```diff
@@ -5,7 +5,7 @@ import { Ecosystem, EcosystemPlugin } from './types';
 const EcosystemPlugins: {
   readonly [ecosystem in Ecosystem]: EcosystemPlugin;
 } = {
-  cpp: cppPlugin,
+  cpp: cppPlugin as EcosystemPlugin,
```
anthogez, line 8: I do not like this casting neither the as any below - we need decide how do we want proceed here also keep plugins in sync with `EcosystemPlugin interface` I believe worth move it to its own lib

anthogez, line 8: I have added this cast in order to avoid TS crying on ScanResult types - cpp-plugin expect only GitTarget while CLI an union of `GitTarget | ContainerTarget`.
To be honest I do not like this casting neither the `as any` below - we need decide how do we want proceed here also keep plugins in sync with `EcosystemPlugin interface` I believe worth move it to its own lib

src/lib/plugins/get-extra-project-count.ts
anthogez: ✅  added for this functionality

test/ecosystems.spec.ts
```diff
@@ -124,6 +124,11 @@ describe('ecosystems', () => {
               identity: {
                 type: 'cpp',
               },
+              name: 'snyk',
+              target: {
+                branch: 'feat/monitor-cpp-projects',
+                remoteUrl: 'git@github.com:snyk/snyk.git',
```
gitphill, line 130: we need to relax this assertion - so that it still works when we merge to master!

anthogez (approved):

test/ecosystems-test-cpp.spec.ts
```diff
@@ -0,0 +1,207 @@
+import * as fs from 'fs';
```
anthogez, line 1: Once the number of ecosystems increases we can create a sub-folder foreach

test/ecosystems-monitor-cpp.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+  osName()
+    .toLowerCase()
+    .indexOf('windows') === 0;
+
+describe('monitorEcosystem cpp', () => {
+  const fixturePath = path.join(__dirname, 'fixtures', 'cpp-project');
+  const cwd = process.cwd();
+
+  function readFixture(filename: string) {
+    if (isWindows) {
```
gitphill, line 24: is this windows check required? we don't use any txt files

PR #1521, opened 2020-11-10
feat: Implement --detection-depth flag in iac command

aron:
- [x] Ready for review

#### What does this PR do?

This PR implements the `--detection-depth` flag for the `snyk iac test` command to limit the depth of subdirectories that will be traversed when looking for files. For example:

```bash
% snyk iac test --detection-depth=3 # limit to only 3 subdirectories
```

The implementation forgoes the use of the existing `glob` module and instead walks the directory tree using the built in file system node module. This has several advantages:

1) The code is simpler and clearer\*. Rather than relying on the glob syntax and constructing an appropriate string we're just using a loop and recursive call allowing us to exit early when we reach the desired depth.
2) The use of the generator provides fine grained control over to the caller allowing them to use the `break`, `continue` & `return` constructs to control the loop.
3) Performance gains, as a rough example it takes ~600ms to run `snyk iac test` on this snyk repository containing ~37000 files with this PR vs. ~2300ms with the glob implementation. 

I've added some basic integration tests for the `makeDirectoryIterator()` function that verifies that nested directories are walked + that the `maxDepth` and `includeDotfiles` options are respected.

\* Simplicity and clarity is subjective, so I'm also open to considering any of the following:

* Using the existing [`find()`](https://github.com/snyk/snyk/blob/master/src/lib/find-files.ts#L55) function already in the library which uses a similar implementation just with a few manifest specific paths.
* Reworking the API to skip the use of generators and use a callback interface e.g. `walkDirectory(path, file => files.push(file))`. Though the downside here is lack of ability to return early, though we don't need it right now.
* Finding another external dependency that meets our needs, suggestions welcome.

**Additional Changes**

I had to make a few additional changes to the code outside of the IaC implementation:

1) I removed the `#!/usr/bin/env node` from the cli/index.ts as it's broken (Node doesn't understand TypeScript) and it's stripped out by the tsc compiler anyway. It also broke filetype detection in vim.
2) I had to update the args parser to coerce the `detectionDepth` argument into a number. Currently it's a string and our existing JavaScript code uses the `>` operator which performs the coercion. TypeScript isn't so happy about this so fixing it at the argument parser level felt correct.
 
Start in `getDirectoryFiles()` in detect-iac.ts then check out `makeDirectoryIterator()` and finally the changes to args.ts.

Build locally and run in a repo containing IaC files e.g. cloud-native-configs. Or any directory containing a number of subdirectories with .json, .yaml or .tf files. As you change the value of `--detection-depth=n` you should see fewer files matched in the output.

[CC-434]

#### Additional questions

**Q: I've currently matched the existing implementation which a) follows symlinked directories and b) ignores files beginning with `.` e.g. `.config`. Do we want to change this?**
_A: No, we want to continue to ignore dotfiles_

**Q: I've avoided setting the default depth for the moment. I can understand the logic when scanning for package manifests where they're likely to be in the top level. For IaC files where we can expect deeply nested directory structures it might cause confusion when running the tool for the first time if many files are missing. So I would appreciate thoughts around this.**
_A: As this is the current behaviour for the command and provides the best first run experience we'll stick with it for the moment. But we may want to review this in the future for either consistency or performance reasons_

[CC-434]: https://snyksec.atlassian.net/browse/CC-434

### DIFF
--- a/src/cli/args.ts
+++ b/src/cli/args.ts
@@ -68,7 +68,7 @@ export interface ArgsOptions {
   // (see the snyk-mvn-plugin or snyk-gradle-plugin)
   _doubleDashArgs: string[];
   _: MethodArgs;
-  [key: string]: boolean | string | MethodArgs | string[]; // The two last types are for compatibility only
+  [key: string]: boolean | string | number | MethodArgs | string[]; // The two last types are for compatibility only
 }
 
 export function args(rawArgv: string[]): Args {
@@ -225,6 +225,10 @@ export function args(rawArgv: string[]): Args {
       argv[camelCased] = argv[dashedArg];
       delete argv[dashedArg];
     }
+  }
+
+  if (argv.detectionDepth !== undefined) {
+    argv.detectionDepth = Number(argv.detectionDepth);
   }
 
   if (argv.skipUnresolved !== undefined) {
--- a/src/lib/iac/detect-iac.ts
+++ b/src/lib/iac/detect-iac.ts
@@ -99,7 +99,10 @@ async function getDirectoryFiles(
   });
 
   for (const filePath of directoryPaths) {
-    const fileType = pathLib.extname(filePath).substr(1) as IacFileTypes;
+    const fileType = pathLib
+      .extname(filePath)
+      .substr(1)
+      .toLowerCase() as IacFileTypes;
     if (!fileType || !supportedExtensions.has(fileType)) {
       continue;
     }
--- a/src/lib/iac/detect-iac.ts
+++ b/src/lib/iac/detect-iac.ts
@@ -1,7 +1,6 @@
 import * as fs from 'fs';
 import * as pathLib from 'path';
 import * as debugLib from 'debug';
-import * as glob from 'glob';
 import { isLocalFolder, localFileSuppliedButNotFound } from '../detect';
 import { CustomError } from '../errors';
 import { validateK8sFile, makeValidateTerraformRequest } from './iac-parser';
@@ -10,6 +9,7 @@ import {
   IacProjectType,
   IacFileTypes,
 } from './constants';
+import { makeDirectoryIterator } from './makeDirectoryIterator';
 import {
   SupportLocalFileOnlyIacError,
   UnsupportedOptionFileIacError,
@@ -38,7 +38,9 @@ export async function getProjectType(
     // scanning the projects - we need save the files we need to scan on the options
     // so we could create assembly payloads for the relevant files.
     // We are sending it as a `Multi IaC` project - and later assign the relevant type for each project
-    const directoryFiles = await getDirectoryFiles(root);
+    const directoryFiles = await getDirectoryFiles(root, {
+      maxDepth: options.detectionDepth,
+    });
     options.iacDirFiles = directoryFiles;
     return IacProjectType.MULTI_IAC;
   }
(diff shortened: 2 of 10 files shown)


src/lib/iac/detect-iac.ts
```diff
@@ -84,33 +85,36 @@ async function getProjectTypeForIacFile(filePath: string) {
...
 
-async function getDirectoryFiles(root: string) {
+async function getDirectoryFiles(
+  root: string,
+  options: { maxDepth?: number } = {},
+) {
   const iacFiles: IacFileInDirectory[] = [];
   const dirPath = pathLib.resolve(root, '.');
-  const files = glob.sync(
-    pathLib.join(dirPath, '/**/**/*.+(json|yaml|yml|tf)'),
```
aron, line 91: It's worth flagging here that this code was intended to only search three directories in depth, but in actuality will search the entire tree.

> ** If a "globstar" is alone in a path portion, then it matches zero or more directories and subdirectories searching for matches. It does not crawl symlinked directories.

- https://www.npmjs.com/package/glob

I'm not sure how you would (or even if you can) express directory depth with the glob syntax.

p0tr3c:
To answer additional questions:
1. We should skip any file or directory starting with `.` . These may contain cache files for terraform which we do not want to scan. Also generally these refer to hidden files. Terraform will follow symlinks so we should as well. 
2. Default should be unlimited traversal. We cannot make any sensible assumption on this.

src/lib/iac/detect-iac.ts
```diff
@@ -120,3 +124,55 @@ async function getDirectoryFiles(root: string) {
 
   return iacFiles;
 }
+
+/**
+ * Walk the provided directory tree, depth first, yielding the matched filename
+ * to the caller. An optional `maxDepth` argument can be provided to limit how
+ * deep in the file tree the search will go.
+ */
+function* walkDirectory(
```
rontalx, line 133: great implementation using generator functions.

src/lib/iac/detect-iac.ts
```diff
@@ -84,33 +85,36 @@ async function getProjectTypeForIacFile(filePath: string) {
...
-        .catch((err: CustomError) => {
-          iacFiles.push({
-            filePath,
-            fileType: ext as IacFileTypes,
-            failureReason: err.userMessage,
-          });
-        });
+  const supportedExtensions = new Set(Object.keys(projectTypeByFileType));
+
+  for (const filePath of walkDirectory(dirPath, {
```
rontalx, line 96: minor:
maybe we can rename the `walkDirectory` method to something like:
`yieldNextDirectoryFilepaths` / `getNextDirectoryFilepaths` / whatever ?
as it's slightly unclear in this part that the `walkDirectory` method plans to recursively yield all files paths for each directory level.

rontalx (commented):
overall looks good.
I really like the elegance of this solution using generator functions, nice work 👍 
Just please make sure you add some tests for this.

src/lib/iac/detect-iac.ts
aron: That's a good point, the MDN docs have a nice example where they use [`makeRangeIterator()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#Generator_functions), I like it because it highlights the returned value is unusual. Perhaps we should follow the pattern:

```suggestion
const directoryIterator = makeDirectoryIterator(dirpath, { ... });
for (const filePath of directoryIterator) {
```

Thoughts?

rontalx: looks good and much clearer! 💪

src/cli/args.ts
```diff
@@ -221,6 +221,14 @@ export function args(rawArgv: string[]): Args {
...
   }
 
+  if (argv.detectionDepth !== undefined) {
+    if (!/\d+/.test(String(argv.detectionDepth))) {
+      // TODO: Real error handling
+      throw new Error('Unsupported value for --detection-depth flag');
+    }
+    argv.detectionDepth = Number(argv.detectionDepth);
+  }
+
```
aron, line 231: I could do with some pointers on how to implement this in an ideomatic way. I notice that there is some form of error handling in line 172:

```
method = require('../lib/errors/legacy-errors');
```

@snyk/hammer would anyone have any thoughts here?

aron:
I'm going to open this up for review. I think I've addressed the comments raised so far. I've renamed the directory walker into a seperate file and renamed it to `makeDirectoryIterator()` and I've added some tests to assert the correct behaviour.

src/lib/iac/detect-iac.ts
```diff
@@ -84,33 +86,41 @@ async function getProjectTypeForIacFile(filePath: string) {
...
+async function getDirectoryFiles(
+  root: string,
+  options: { maxDepth?: number } = {},
+) {
   const iacFiles: IacFileInDirectory[] = [];
   const dirPath = pathLib.resolve(root, '.');
-  const files = glob.sync(
-    pathLib.join(dirPath, '/**/**/*.+(json|yaml|yml|tf)'),
-  );
+  const supportedExtensions = new Set(Object.keys(projectTypeByFileType));
```
rontalx, line 95: minor:
you can create this set once under `lib/iac/constants.ts` and re-use it instead of re-creating it from the same `projectTypeByFileType` every call.

rontalx (approved):
nice work 🚀 
loving the generator usage, and the new clean Jest tests you added 💪

help/iac.txt
```diff
@@ -11,6 +11,9 @@ Commands:
 Options:
 
   -h, --help
+  --detection-depth=<number>
+                       (test command only)
+                       Use to indicate how many sub-directories to search. By default all subdirectories will be searched.
```
ekbsnyk, line 16: I think this is slightly unclear - could it be clearer? E.g. "Use to indicate how many levels deep to search inside subdirectories", or something? Possibly even with a short example

ekbsnyk (approved):

help/iac.txt
```diff
@@ -11,6 +11,9 @@ Commands:
 Options:
 
   -h, --help
+  --detection-depth=<number>
```
JackuB, line 14: I've just merged https://github.com/snyk/snyk/pull/1527 - you can now move this to a markdown document in `help/commands-docs/iac.md` to resolve the conflict👌

help/iac.txt
aron: Thanks, I've reworded this to try and be clearer and added an example.

src/cli/args.ts
aron: Based on feedback from @JackuB I've split this into two parts, first the coercion still happens here. But I've now created a custom error subclass of `CustomError` which is thrown in cli/index.ts within the try/catch block _after_ the arguments are parsed. This ensures the args can still be parsed correctly without throwing and we get nice error output on the CLI.

PR #1531, opened 2020-11-19
docs: Document username/password inputs

michael-c-snyk:
Explain how to configure remote container registry credentials via environment variables and/or flags

### DIFF
--- a/help/commands-docs/_ENVIRONMENT.md
+++ b/help/commands-docs/_ENVIRONMENT.md
@@ -1,6 +1,6 @@
 ## ENVIRONMENT
 
-You can set these environment variables to change CLI run settings. These keys will override settings in your `snyk config`:
+You can set these environment variables to change CLI run settings.
 
 - `SNYK_TOKEN`:
   Snyk authorization token. Setting this envvar will override the token that may be available in your `snyk config` settings.
@@ -15,3 +15,12 @@ You can set these environment variables to change CLI run settings. These keys w
   Allows you to override any key that's also available as `snyk config` option.
 
   E.g. `SNYK_CFG_ORG`=myorg will override default org option in `config` with "myorg".
+
+- `SNYK_REGISTRY_USERNAME`:
+    Specify a username to use when connecting to a container registry. Note that using the `--username` flag will
+     override this value. This will be ignored in favour of local Docker binary credentials when Docker is present. 
+  
+- `SNYK_REGISTRY_PASSWORD`:
+    Specify a password to use when connecting to a container registry. Note that using the `--password` flag will
+     override this value. This will be ignored in favour of local Docker binary credentials when Docker is present. 
+    
--- a/help/commands-docs/container.md
+++ b/help/commands-docs/container.md
@@ -50,8 +50,14 @@ Find vulnerabilities in your container images.
 - `--project-name`=<PROJECT_NAME>:
   Specify a custom Snyk project name.
 
-- `--policy-path`=<PATH_TO_POLICY_FILE>`:
+- `--policy-path`=<PATH_TO_POLICY_FILE>:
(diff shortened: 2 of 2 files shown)


help/commands-docs/_ENVIRONMENT.md
```diff
@@ -15,3 +15,12 @@ You can set these environment variables to change CLI run settings. These keys w
   Allows you to override any key that's also available as `snyk config` option.
 
   E.g. `SNYK_CFG_ORG`=myorg will override default org option in `config` with "myorg".
+
+- `SNYK_REGISTRY_USERNAME`:
+    Specify a username to use when connecting to a container registry. Note that providing a `--username` flag will
+     override this value. This will be ignored in favour of local docker binary credentials when using docker socket.
```
garethr, line 21: Assuming the line break doesn't cause different rendering? I see the output breaks the lines at the moment based on character width. Just checking this doesn't introduce an extra linebreak or space?

help/commands-docs/_ENVIRONMENT.md
```diff
@@ -15,3 +15,12 @@ You can set these environment variables to change CLI run settings. These keys w
   Allows you to override any key that's also available as `snyk config` option.
 
   E.g. `SNYK_CFG_ORG`=myorg will override default org option in `config` with "myorg".
+
+- `SNYK_REGISTRY_USERNAME`:
+    Specify a username to use when connecting to a container registry. Note that providing a `--username` flag will
+     override this value. This will be ignored in favour of local docker binary credentials when using docker socket. 
+  
+- `SNYK_REGISTRY_PASSWORD`:
+    Specify a password to use when connecting to a container registry. Note that providing a `--password` flag will
```
garethr, line 24: I feel like you don't provide a flag, as much as provide a value with a flag, or use a flag. Suggest changing here and above to 

> using the `--password` flag will

help/commands-docs/container.md
```diff
@@ -50,8 +50,16 @@ Find vulnerabilities in your container images.
...
-- `--policy-path`=<PATH_TO_POLICY_FILE>`:
+- `--policy-path`=<PATH_TO_POLICY_FILE>:
   Manually pass a path to a snyk policy file.
 
 - `--severity-threshold`=low|medium|high:
   Only report vulnerabilities of provided level or higher.
+  
+- `--username`=<CONTAINER_REGISTRY_USERNAME>:
+  Specify a username to use when connecting to a container registry. This will be ignored in favour of local docker
+   binary credentials when using docker socket.
```
garethr, line 61: Probably better to say

> when Docker is present

Rather than "when using docker socket" as that's really an implementation detail that most users wouldn't know.

help/commands-docs/_ENVIRONMENT.md
```diff
@@ -15,3 +15,12 @@ You can set these environment variables to change CLI run settings. These keys w
   Allows you to override any key that's also available as `snyk config` option.
 
   E.g. `SNYK_CFG_ORG`=myorg will override default org option in `config` with "myorg".
+
+- `SNYK_REGISTRY_USERNAME`:
```
garethr, line 19: Should this and the below only be shown on the Snyk Container help, rather than everywhere? Would need a separate template file I presume.

help/commands-docs/_ENVIRONMENT.md
michael-c-snyk: Good spot! It looks OK rendered locally but will take the linebreak out to avoid the risk of other systems handling it differently 👍

help/commands-docs/_ENVIRONMENT.md
michael-c-snyk: <resolved in Slack>

anthogez (approved):

PR #1541, opened 2020-11-24
feat: use policy testing scan results

gitphill:
- [X] Ready for review

#### What does this PR do?
If the CLI plugin does not provide a policy, default to searching for a
.snyk file in the scan result path, and use that if found.

`snyk ignore` to create a `.snyk` policy file, then `snyk source test` or `snyk source monitor` to ensure the policy is being used.

### DIFF
--- a/src/lib/ecosystems/test.ts
+++ b/src/lib/ecosystems/test.ts
@@ -9,6 +9,7 @@ import { Ecosystem, ScanResult, TestResult } from './types';
 import { getPlugin } from './plugins';
 import { TestDependenciesResponse } from '../snyk-test/legacy';
 import { assembleQueryString } from '../snyk-test/common';
+import { findAndLoadPolicyForScanResult } from './policy';
 
 export async function testEcosystem(
   ecosystem: Ecosystem,
@@ -58,6 +59,10 @@ async function testDependencies(
   for (const [path, scanResults] of Object.entries(scans)) {
     await spinner(`Testing dependencies in ${path}`);
     for (const scanResult of scanResults) {
+      if (!scanResult.policy) {
+        // search for `.snyk` file if plugin doesn't provide a policy
+        scanResult.policy = await findPolicy(scanResult, options);
+      }
       const payload = {
         method: 'POST',
         url: `${config.API}/test-dependencies`,
@@ -88,4 +93,15 @@ async function testDependencies(
   }
   spinner.clearAll();
   return [results, errors];
+}
+
+async function findPolicy(
+  scanResult: ScanResult,
+  options: Options,
+): Promise<string | undefined> {
+  const policy = await findAndLoadPolicyForScanResult(scanResult, options);
+  if (policy !== undefined) {
+    return policy.toString();
+  }
+  return undefined;
 }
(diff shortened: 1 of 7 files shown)


src/lib/ecosystems/test.ts
```diff
@@ -58,6 +59,10 @@ async function testDependencies(
   for (const [path, scanResults] of Object.entries(scans)) {
     await spinner(`Testing dependencies in ${path}`);
     for (const scanResult of scanResults) {
+      if (!scanResult.policy) {
+        // search for `.snyk` file if plugin doesn't provide a policy
+        scanResult.policy = await findPolicy(scanResult, options);
```
razzmadaz, line 64: I find that this kind of 'cleaning', annotating or transforming data is best carried out at the boundaries of the sub-domain working in. 

For example, this might be the subdomain `ecosystem test` which happens to collaborate with other subdomains to carry out behaviour such as performing a scan.
We might model this in the port & adaptors style, where for the `ecosystem-test` component, from its perspective, it is sending a `performing scan (for a particular ecosystem)` message/command to some component meeting this `EcosystemScanner` contract (aka interface or `port`).
In reality, we use a `pluginScanner` as the underlying mechanism and "adapt" it to the `ecosystem-scanner` concept by wrapping its creation + call so that the caller is obvlivious to the fact that it is a pluginScanner, and only knows (and cares) about that it has an EcosystemScanner to send the msg to.

I find this simplifies the 'business' logic part of the code, carrying out the actual behaviour, from the part of the code that wires up the components and transforms data being shuttled between these different components.
This has a huge added advantage that testing can happen at the component level without needing to have a complicated test-setup to account for knowing the actual implementation details of the collaborating components. All you need to know is they meet the interface, and can fake them out accordingly.

```
type ScanResults = ScanResult[];
type EcosystemScanner = (path) => Promise<ScanResults>;
type EcosystemScannerFactory = (ecosystem, options) => EcosystemScanner;

async function ensureHasPolicy(scanResult: ScanResult, options: Options): Promise<ScanResult> {
  if (!scanResult.policy) {
    scanResult.policy = await findPolicy(scanResult, options);
  }
  return scanResult;
}

/*
  example impl of `adaptor` to port `EcosystemScanner`,
  for conceptual purposes (and missing types,etc).
 */
const pluginScanner: EcosystemScannerFactory = (ecosystem, options) => {
  const plugin = getPlugin(ecosystem);
  return async (path) => plugin.scan({...options, path})
      .then((result) => result.scanResults)
    .then ((results) => Promise.all(results.map((r) => ensureHasPolicy(r, options)));
};
```

and then simplifies the use site to focus on fewer responsbilities:
```
export async function testEcosystem(....): Promise<TestCommandResult> {
  const scan = scanner(ecosystem, options);
  const scanResultsByPath: { [dir: string]: ScanResults } = {};
  for (const path of paths) {
    await spinner(`Scanning dependencies in ${path}`);
    scanResultsByPath[path] = await scan(path);
  }
  ......
```
and then you don't need this guard clause to transform the data in the midst of performing the behaviour this function/module is actually responsible for.

razzmadaz (approved):
Made a comment on design and what I think leads to overall simpler components, that are easier to test, read implementation for, etc. 
I think there is also a good example of where having multiple commits on the PR would have made understanding what changes (in this case, to the test) have actually happened, since there seems to be a lot of noise fro the reformatting?

PR #1545, opened 2020-11-26
fix: only check latest 7 smoke tests

Avishagp:
#### What does this PR do?
CLI alert will now check the latest 7 smoke tests and would be less chatty as a result

### DIFF
--- a/test/cli-alert/src/index.ts
+++ b/test/cli-alert/src/index.ts
@@ -40,7 +40,7 @@ async function run(octokit: Octokit) {
       }
     }
 
-    // Get last 3 smoke tests workflows
+    // Get latest smoke tests workflows
     const workflows = (
       await octokit.actions.listWorkflowRuns({
         owner: 'snyk',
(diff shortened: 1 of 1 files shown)


test/cli-alert/src/index.ts
```diff
@@ -40,18 +40,18 @@ async function run(octokit: Octokit) {
...
         repo: 'snyk',
         branch: 'master',
         // eslint-disable-next-line @typescript-eslint/camelcase
         workflow_id: smokeTestsID,
-        // eslint-disable-next-line @typescript-eslint/camelcase
-        per_page: 6,
       })
     ).data;
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    workflows.workflow_runs = workflows.workflow_runs.slice(0, 7);
```
JackuB, line 54: You might consider slicing in the same place where you filter - modifying object like this is a bad practice. `workflows` is a `const` and coming from a 3rd party library so it shouldn't be modified. This is abusing a JS behavior where it'll allow you to deeply modify objects.

Avishagp: Thanks! pushed a fix for that

JackuB (approved):

PR #1552, opened 2020-12-02
feat: analytics- enable getting a command version

Avishagp:
#### What does this PR do?
Enable getting a command's version and checking if a software is installed on user's machine.

https://snyksec.atlassian.net/browse/HAMMER-186


JackuB (changes requested):
I'd like to see a test that verifies the analytics object contains the `environment.npmVersion` with a valid semver

src/lib/analytics-sources.ts
```diff
@@ -149,3 +151,55 @@ export function validateHomebrew(snykExecutablePath: string): boolean {
   }
   return false;
 }
+
+function runCommand(cmd: string): Promise<string> {
+  return new Promise((resolve) => {
+    exec(cmd, (error, stdout, stderr) => {
+      if (error) {
+        console.warn(error);
```
JackuB, line 159: Wondering whether this should log only in debug mode - especially since it'll only affected the analytics module

test/analytics.spec.ts
```diff
@@ -0,0 +1,150 @@
+const analytics = require('../src/lib/analytics');
+
+describe('Analytics basic testing', () => {
+  it('Has all analytics arguments', async () => {
+    analytics.add('foo', 'bar');
+    const data = { args: [], command: '__test__' };
+    const res = await analytics(data);
+    if (!res) {
+      fail('analytics creation failed!');
```
JackuB, line 9: Wouldn't the `await analytics` call throw in this case? Rather than returning `falsy` value?

`analytics()` returns a Promise, so if the Promise would `reject`, then it would also kill this test, no?

test/analytics.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      args: [{ integrationName: 'JENKINS', integrationVersion: '1.2.3' }],
+      command: '__test__',
+      org: '__snyk__',
+    };
+    const res = await analytics(data);
+    if (!res) {
+      fail('analytics creation failed!');
+    }
+    const vals = Object.values(data);
+    expect(vals).toContain('__test__');
```
JackuB, line 145: What is this test proving? `data` is something created few lines above?

src/lib/analytics.js
```diff
@@ -85,6 +86,12 @@ function postAnalytics(data) {
       }
 
       data.ci = isCI();
+
+      data.environment = {};
+      if (isStandalone) {
```
JackuB, line 91:
```suggestion
      if (!isStandalone) {
```

We should check the opposite - in standalone mode we ship (and control) node/npm version. We care about the non-standalone ones.

Check whether it affects logic in any test as well 🤞

JackuB (changes requested):
Some questions/suggestions - one blocking comment around the standalone

test/analytics.spec.ts
Avishagp: Data is changed during analytics creation. This test checks if the original values given weren't changed

JackuB (approved):

PR #1554, opened 2020-12-03
fix: Group issues should also group names

mladkau:
- [x] Ready for review

#### What does this PR do?

Small fix for the `--group-issues` flag. We should also group the name attribute along with the from attribute.

#### Screenshots

image

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -254,7 +254,11 @@ async function main() {
         '--group-issues is currently not supported for Snyk IaC.',
       ]);
     }
-    if (args.options['group-issues'] && !args.options['json']) {
+    if (
+      args.options['group-issues'] &&
+      !args.options['json'] &&
+      !args.options['json-file-output']
+    ) {
       throw new UnsupportedOptionCombinationError([
         'JSON output is required to use --group-issues, try adding --json.',
       ]);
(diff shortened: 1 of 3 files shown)


src/cli/index.ts
```diff
@@ -254,7 +254,11 @@ async function main() {
         '--group-issues is currently not supported for Snyk IaC.',
       ]);
     }
-    if (args.options['group-issues'] && !args.options['json']) {
+    if (
+      args.options['group-issues'] &&
+      !args.options['json'] &&
+      !args.options['json-file-output']
```
mladkau, line 260: Not strictly part of the PR but the `json-file-output` should also be allowed with the `group-issues` flag.

test/fixtures/basic-apk/jsonDataGrouped.json
```diff
@@ -103,7 +103,21 @@
       "upgradePath": [],
       "isUpgradable": false,
       "isPatchable": false,
-      "name": "musl/musl-utils",
+      "name": [
```
JackuB, line 106: Should we also make these unique? Or the array length matters?

JackuB (approved):
Small question

test/fixtures/basic-apk/jsonDataGrouped.json
mladkau: I would leave the array as is as each entry can be matched with an entry in the from list.

PR #1558, opened 2020-12-07
feat: Add impact and resolve fields in sarif output.

ipapast:
CC-517

- [x] Ready for review

#### What does this PR do?
This PR adds the 'resolve' and 'impact' fields in the results of the sarif output in order to give more useful information to the users and show that to the Github Security page (coming from the registry's AnnotatedIssues object).

Check the two new fields in `iac-output.ts. The resolution is added to the 'help' field and the impact is included in the full description.

- [ ] Run iac test --sarif and see the results

- [ ] Check Github Security

We are using the reportingDescriptor object, more info on the docs here [Github Security SARIF Docs](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/sarif-support-for-code-scanning#supported-sarif-output-file-properties)

[Jira CC-517](https://snyksec.atlassian.net/browse/CC-517)

#### Screenshots
**CLI Output:**
image

**Github Security:**
image

### DIFF
--- a/src/cli/commands/test/iac-output.ts
+++ b/src/cli/commands/test/iac-output.ts
@@ -222,8 +222,14 @@ export function extractReportingDescriptor(
         text: `${iacTypeToText[issue.type]} ${issue.subType}`,
       },
       help: {
-        text: '',
-        markdown: issue.description,
+        text: `The issue is... \n${issue.iacDescription.issue}\n\n The impact of this is... \n ${issue.iacDescription.impact}\n\n You can resolve this by... \n${issue.iacDescription.resolve}`.replace(
+          /^\s+/g,
+          '',
+        ),
+        markdown: `**The issue is...** \n${issue.iacDescription.issue}\n\n **The impact of this is...** \n ${issue.iacDescription.impact}\n\n **You can resolve this by...** \n${issue.iacDescription.resolve}`.replace(
+          /^\s+/g,
+          '',
+        ),
       },
       defaultConfiguration: {
         level: getIssueLevel(issue.severity),
--- a/src/lib/snyk-test/iac-test-result.ts
+++ b/src/lib/snyk-test/iac-test-result.ts
@@ -1,4 +1,4 @@
-import { BasicResultData, TestDepGraphMeta, SEVERITY } from './legacy';
+import { BasicResultData, SEVERITY, TestDepGraphMeta } from './legacy';
 
 export interface AnnotatedIacIssue {
   id: string;
@@ -14,6 +14,11 @@ export interface AnnotatedIacIssue {
   name?: string;
   from?: string[];
   lineNumber?: number;
+  iacDescription: {
+    issue: string;
+    impact: string;
+    resolve: string;
+  };
 }
 
 type FILTERED_OUT_FIELDS = 'cloudConfigPath' | 'name' | 'from';
--- a/test/acceptance/cli-test/iac/cli-test.iac-utils.ts
+++ b/test/acceptance/cli-test/iac/cli-test.iac-utils.ts
@@ -281,6 +281,7 @@ export function iacTestSarifAssertions(
 
 function generateDummyIssue(severity): AnnotatedIacIssue {
   return {
+    iacDescription: { issue: '', impact: '', resolve: '' },
     id: 'SNYK-CC-K8S-1',
     title: 'Reducing the admission of containers with dropped capabilities',
     name: 'Reducing the admission of containers with dropped capabilities',
@@ -303,7 +304,7 @@ function generateDummyIssue(severity): AnnotatedIacIssue {
 }
 
 function generateDummyTestData(
-  cloudConfigResults: Array<AnnotatedIacIssue>,
+  cloudConfigResults: AnnotatedIacIssue[],
 ): IacTestResponse {
   return {
     path: '',


src/lib/snyk-test/iac-test-result.ts
```diff
@@ -14,6 +14,10 @@ export interface AnnotatedIacIssue {
   name?: string;
   from?: string[];
   lineNumber?: number;
+  iacDescription: {
+    impact: string;
```
almog27, line 18: We don't want to present the issue too?
I think even if we don't intend to present it in this case - it will be better to add it here and remove the description.

test/acceptance/cli-test/iac/cli-test.iac-utils.ts
```diff
@@ -281,6 +281,7 @@ export function iacTestSarifAssertions(
 
 function generateDummyIssue(severity): AnnotatedIacIssue {
   return {
+    iacDescription: { impact: '', resolve: '' },
```
almog27, line 284: Same here

src/lib/snyk-test/iac-test-result.ts
ipapast: Issue is slightly different than the shortDescription. I am not sure which one is more valuable as a piece of information, they both give a similar phrase more or less in different words. 

❓ What's the difference in adding it here? Just to have the same object fields in iacDescription?

❗ Some examples: 
- shortDescription: "Container could be running with outdated image"
- iacDescription.issue: "The image policy does not prevent image reuse"

- shortDescription: "Container is running without AppArmor profile"
- iacDescription.issue: "The AppArmor profile is not set correctly"

aron: >❓ What's the difference in adding it here? Just to have the same object fields in iacDescription?

Exactly. 👍 for adding it to the type definition for completeness. This should represent the object we're having returned from the API even if we don't use it here.

> I am not sure which one is more valuable as a piece of information, they both give a similar phrase more or less in different words.

I think it's largely based on this piece of UI. So I would imagine the issue is intended to be displayed with the impact and resolution.

image

src/cli/commands/test/iac-output.ts
```diff
@@ -219,11 +219,12 @@ export function extractReportingDescriptor(
         text: `${upperFirst(issue.severity)} severity - ${issue.title}`,
       },
       fullDescription: {
-        text: `${iacTypeToText[issue.type]} ${issue.subType}`,
+        text: `Type: ${iacTypeToText[issue.type]} ${issue.subType}, impact: ${
```
aron, line 222: I think seeing this in context within the screenshot, it might not make sense to put this here. This "full description" doesn't really have much space.

Instead how about we extend the markdown in the help text to mimic the Snyk UI (e.g. in the screenshot below).

```ts
`
**The issue is...** 
${issue.iacDescription.issue}

**The impact of this is...**
${issue.iacDescription.impact}

**You can resolve this by...**
${issue.iacDescription.resolve}
`
```

---

**The issue is...** 
The image policy does not prevent image reuse

**The impact of this is...**
Blah blah blah.

**You can resolve this by...**
Adding foo to the bar.

aron (changes requested):
This is fantastic, I think now seeing the output in the GitHub UI, the description isn't the best place for the impact.

The help field where we show the resolution looks like a better spot to put all of this information and describe to the user the issue, impact and resolution as we do in the Snyk UI.

What do you think?

src/cli/commands/test/iac-output.ts
ipapast: image

ipapast: The changes now reflect the Snyk UI too.
You can see it also in my test repo https://github.com/ipapast/k8-test/security/code-scanning

src/cli/commands/test/iac-output.ts
```diff
@@ -222,8 +222,15 @@ export function extractReportingDescriptor(
...
-        text: '',
-        markdown: issue.description,
+        text: `The issue is...
+        ${issue.iacDescription.issue}
+
+        The impact of this is...
+        ${issue.iacDescription.impact}
+        
+        You can resolve this by...
+        ${issue.iacDescription.resolve}`,
```
aron, line 232: We're going to have the same issue here as we had with the markdown. The string is going to look like:

```
The issue is...
        ${issue.iacDescription.issue}

        The impact of this is...
        ${issue.iacDescription.impact}
        
        You can resolve this by...
        ${issue.iacDescription.resolve}
```

aron (approved):
Nice, looks fantastic. I think it might be easier to keep the nice template literals but to trim the leading whitespace from each line with a `replace()` function.

src/cli/commands/test/iac-output.ts
ipapast: Correct me if I'm wrong, but my understanding is, that if the markdown field exists, then the sarif output will display that.
The text field is only mandatory if the markdown field does not exist. In this case, I think we can even have an empty string as a text field and the output would still be the same.
In summary, I think the best thing is to do is to display text: ' ' and then markdown as we have already formmated it.

aron: As you have it is absolutely fine, we want both as it means that the content can be output anywhere as plaintext but also enhanced as markdown if the client understands it. All I meant by my comment was that the plain text version is going to be indented after the first line. Imagine if this Sarif file was printed by a command line tool as plain text to the terminal, it would look odd if the first line of subsequent sections were indented.

Lets chat if this still doesn't make sense :)

ipapast: Sorry for the back and forth - I was talking for the specific purposes of Github Security display and you were talking about the case of the plain output. I get you now :) I'll do that change now

src/lib/snyk-test/iac-test-result.ts
almog27: The `description` is the old field that we should remove soon.
We split the description into 3 fields - `issue`, `impact` and `resolve` which we put under those under the iacDescription

almog27 (approved):

src/cli/commands/test/iac-output.ts
ipapast: Updated the original PR comment with the most updated screenshots for any future reference.
Added a replace function to the help fields.
Squashed the commits and will merge as soon as the checks passed.

PR #1565, opened 2020-12-13
Resolve type annotation changes required to enable `composite` in tsconfig

maxjeffos:
- [x] Ready for review

#### What does this PR do?
Fixes some typing issues that are needed for the upcoming modular CLI init PR (https://github.com/snyk/snyk/pull/1564).

For the modular CLI init / monorepo PR, we will be turning on `"composite": true` in the tsconfig which requires certain typing things be addressed beforehand.

Why not just do this as part of the aforementioned PR (https://github.com/snyk/snyk/pull/1564)? Because that PR is going to touch a lot of things and I would prefer to keep that PR as small as reasonably possible and since this is just about improving some typing (and refactoring analytics into TypeScript) I think it is best kept on its own.

HAMMER-237

### DIFF
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -1,6 +1,3 @@
-module.exports = analytics;
-module.exports.single = postAnalytics;
-
 const snyk = require('../lib');
 const config = require('./config');
 const version = require('./version');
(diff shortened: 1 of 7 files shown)


src/lib/analytics.ts
```diff
@@ -82,7 +94,8 @@ function postAnalytics(data) {
 
       const headers = {};
       if (snyk.api) {
-        headers.authorization = 'token ' + snyk.api;
+        // headers.authorization = 'token ' + snyk.api;
```
orsagie, line 97: Any reason for this comment to be here?

orsagie (approved):
small comment, but LG

src/lib/analytics.ts
maxjeffos: Nope, I'll remove it. Thanks!

JackuB (approved):

PR #1567, opened 2020-12-15
chore: check jobs’ success instead of workflow

Avishagp:
#### What does this PR do?
Check latest 2 smoke tests' jobs. If the same jobs failed 2 times in a row, try to re-run the latest smoke test and check the job's conclusion. 

This ensures CLI alerts become less chatty and more focused on urgent problems Hammer should take care of.

### DIFF
--- a/test/cli-alert/src/index.ts
+++ b/test/cli-alert/src/index.ts
@@ -10,20 +10,121 @@ if (!process.env.USER_GITHUB_TOKEN || !process.env.SLACK_WEBHOOK_URL) {
 const GITHUB_TOKEN = process.env.USER_GITHUB_TOKEN;
 const SLACK_WEBHOOK_URL = process.env.SLACK_WEBHOOK_URL;
 
-const octokitInstance = new Octokit({
+// Map from job id to it's name and conclusion
+const failedJobs: string[] = [];
+
+const octokit = new Octokit({
   auth: GITHUB_TOKEN,
 });
 const slackWebhook = new IncomingWebhook(SLACK_WEBHOOK_URL);
 
-function filterWorkflows(workflow) {
-  // Keep only workflows didn't succeed, and either completed or were triggered by release
-  return (
-    workflow.conclusion !== 'success' &&
-    (workflow.status === 'completed' || workflow.event === 'release')
-  );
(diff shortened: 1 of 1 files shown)


test/cli-alert/src/index.ts
```diff
@@ -10,20 +10,110 @@ if (!process.env.USER_GITHUB_TOKEN || !process.env.SLACK_WEBHOOK_URL) {
...
+  for (let i = 0; i < jobs1.length; i++) {
+    if (
+      jobs1[i].conclusion === 'failure' &&
+      jobs1[i].conclusion === jobs2[i].conclusion
+    ) {
+      console.log(`Found a job that failed 2 times in a row: ${jobs1[i].name}`);
+      failedJobs.push(jobs1[i].name);
+    }
+  }
+}
```
JackuB, line 51: Comparing by index might be inaccurate here. Don't believe GH guarantees order of the checks. Checking and comparing by names might be safer in this case.

```suggestion
for (const jobName of jobs1.map({name} => name)) {
  if ('failure' === jobs1.find(j => j.name === jobName).conclusion === jobs2.find(j => j.name === jobName).conclusion) {
    failedJobs.push(jobName);
  }
}
```

test/cli-alert/src/index.ts
```diff
@@ -10,20 +10,110 @@ if (!process.env.USER_GITHUB_TOKEN || !process.env.SLACK_WEBHOOK_URL) {
...
+
+async function sendSlackAlert() {
+  console.log('Jobs failed again. Sending Slack alert...');
+  const args: IncomingWebhookDefaultArguments = {
+    username: 'Hammer Alerts',
+    text:
+      'A Smoke Tests Job failed more than 2 times in a row. \n <https://github.com/snyk/snyk/actions?query=workflow%3A%22Smoke+Tests%22|Smoke Tests Results>',
```
JackuB, line 58: Idea: might be handy to also include the name of failed tests in the message. E.g. seeing that all `npm` tests failed could be helpful. I guess you'd have to pass `failedJobs.join(', ')` here

JackuB (approved):
Nothing blocking, approving

maxjeffos (approved):
Looks good. Maybe a silly question, but how do we know any of this works?

PR #1572, opened 2020-12-24
docs: fix config md

didier-durand:
- [ x] Ready for review
- [ x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules

#### What does this PR do?

This PR just fixes some markdown issues with < and > in config.md: for example, <KEY> is confused by the markdown processor with a HTML tag (non-existing in this case) and suppressed as it doesn't exist

Just agree on the issue and accept the solution : replace with &lt; and &gt;

Just preview new versions of the file

No

None

#### Screenshots

None

### DIFF
--- a/help/commands-docs/config.md
+++ b/help/commands-docs/config.md
@@ -2,7 +2,7 @@
 
 ## SYNOPSIS
 
-`snyk` `config` `get|set|clear` \[<KEY>\[=<VALUE>\]\] \[<OPTIONS>\]
+`snyk` `config` `get|set|clear` \[&lt;KEY&gt;\[=&lt;VALUE&gt;\]\] \[&lt;OPTIONS&gt;\]
 
 ## DESCRIPTION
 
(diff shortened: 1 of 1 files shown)


JackuB:
Hi, thanks a lot for the PR!
Did you try to parse the markdown files for something?

Since we are using these markdown files to generate a [roff](https://en.wikipedia.org/wiki/Roff_(software)) manpage files with [ronn-ng](https://github.com/apjanke/ronn-ng), the `<KEYWORD>` syntax is a markdown extension https://github.com/snyk/snyk/tree/master/help#keyword-extension

didier-durand:
@JackuB : Thanks for those details that I didn't know. The reason why I submitted this PR is because there is a rendering problem : the text with <XXXX> do not render if you look at the page. So, the command syntax is improperly truncated and misleading when you read.  So, should I NOT read those pages directly on Github project but somewhere else? Didier

JackuB:
@didier-durand we expose the CLI docs on our public support site: https://support.snyk.io/hc/en-us/articles/360003812578-CLI-reference

We don't yet have the HTML pages exported&linked for the CLI sub commands. Plan is to have those sub command pages exposed publicly and linked from the main support page. Currently, the canonical way to check docs is either `snyk config --help` or with a manpage `man snyk-config`.

I'll close the PR now, as this change on its own will not solve this and we want to keep the Markdown as it is now 🙏 Thank you!

PR #1576, opened 2021-01-08
Update snyk iac cli to issue parallel requests

ipapast:
- [ ] Ready for review

#### What does this PR do?

Currently for a call to snyk iac test our CLI process makes the following sequential requests to our backend API:
- One request per file to validate each Terraform file in the directory tree (or provided via arguments)
- One request pre file to test each matching IaC file in the directory tree (or provided via arguments)

1. As part of cloud-config's work on making tthe cli more performant, we are changing the requests above to **run in parallel.** Various tests were made to determine how this affects the timings, with results showing an average decrease of 8m30s. This is a small win which makes the cli run much faster.
2. In order not to throttle the system in the case of a big number of files being tested, we also introduced two small changes. By default, HTTP connections close after each request. The keepAlive property will allow a persistent connection. For rate limiting, a queue was implemented for both detecting the iac files and testing them. 

The results look very promising. From the original 10m22sec when run locally, it went down to 1min4s with a queue.

This is a table with some tests. It was run against cloud-config-goof, which contains 249 IaC files.

_Tested 247 projects, 201 contained issues. Failed to test 4 projects._

The goal of these experiments was to see what the "optimal" number of concurrent requests is for our case.
- The "before" run is by running it locally without any changes from the master branch. 
- The "after" run is by running it locally against the proposed changes.

***table to be updated asap***
Scenario | user | system | cpu | Total
-- | -- | -- | -- | --
Before | 14.32s | 36.63s | 8% | 10m28s18ms
After: |  
with promises running in parallel | 12.00s | 31.08s | 52% | 1m22s71ms
with promises running in parallel & Queue | 28.00s | 84.08s | 92% | 1m12s71ms
with  rest && https.agent {keepAlive: true} | 14.49s | 43.21s | 51% | 1m52s74ms

- Run the code on the master branch:
`time node snyk iac test`
- Then go to this branch and run `time node ~/snyk-repos/snyk/dist/cli iac test` .
Note the 'total' time in the end. It should have been improved.

There were [more performance improvements](https://snyksec.atlassian.net/jira/software/projects/CC/boards/93?selectedIssue=CC-519) done on the cloud-config-policy-engine where a new endpoint and a feature flag were added. 
As part of the current task, we also use those changes in combination. 
To do this, we needed to enable the _iacEvaluatePolicyEndpoint_ flag in the local snyk admin area when running locally. 

https://snyksec.atlassian.net/browse/CC-594

### DIFF
--- a/src/lib/iac/detect-iac.ts
+++ b/src/lib/iac/detect-iac.ts
@@ -90,7 +90,6 @@ async function getDirectoryFiles(
   root: string,
   options: { maxDepth?: number } = {},
 ) {
-  const iacFiles: IacFileInDirectory[] = [];
   const dirPath = pathLib.resolve(root, '.');
   const supportedExtensions = new Set(Object.keys(projectTypeByFileType));
 
@@ -98,6 +97,7 @@ async function getDirectoryFiles(
     maxDepth: options.maxDepth,
   });
 
+  const promises: Promise<IacFileInDirectory>[] = [];
   for (const filePath of directoryPaths) {
     const fileType = pathLib
       .extname(filePath)
@@ -107,23 +107,28 @@ async function getDirectoryFiles(
       continue;
     }
 
-    try {
-      const projectType = await getProjectTypeForIacFile(filePath);
-      iacFiles.push({
-        filePath,
-        projectType,
-        fileType,
-      });
-    } catch (err) {
-      iacFiles.push({
-        filePath,
-        fileType,
-        failureReason:
-          err instanceof CustomError ? err.userMessage : 'Unhandled Error',
-      });
-    }
+    promises.push(
+      (async () => {
+        try {
+          const projectType = await getProjectTypeForIacFile(filePath);
+          return {
+            filePath,
+            projectType,
+            fileType,
+          };
+        } catch (err) {
+          return {
+            filePath,
+            fileType,
+            failureReason:
+              err instanceof CustomError ? err.userMessage : 'Unhandled Error',
+          };
+        }
+      })(),
+    );
   }
 
+  const iacFiles = await Promise.all(promises);
   if (iacFiles.length === 0) {
     throw IacDirectoryWithoutAnyIacFileError();
   }
--- a/src/lib/iac/detect-iac.ts
+++ b/src/lib/iac/detect-iac.ts
@@ -22,7 +22,6 @@ import {
 } from '../errors/invalid-iac-file';
 import { Options, TestOptions, IacFileInDirectory } from '../types';
 import * as Queue from 'promise-queue';
-import * as util from 'util';
 
 const debug = debugLib('snyk-detect-iac');
 
@@ -100,8 +99,10 @@ async function getDirectoryFiles(
   });
 
   const iacFiles: IacFileInDirectory[] = [];
-  const queue = new Queue(50);
+  const maxConcurrent = 50;
+  const queue = new Queue(maxConcurrent);
 
+  const promises: Array<Promise<IacFileInDirectory>> = [];
   for (const filePath of directoryPaths) {
     const fileType = pathLib
       .extname(filePath)
@@ -110,30 +111,31 @@ async function getDirectoryFiles(
     if (!fileType || !supportedExtensions.has(fileType)) {
       continue;
     }
-
-    queue.add(async () => {
-      try {
-        const projectType = await getProjectTypeForIacFile(filePath);
-        iacFiles.push({
-          filePath,
-          projectType,
-          fileType,
-        });
-      } catch (err) {
-        iacFiles.push({
-          filePath,
-          fileType,
-          failureReason:
-            err instanceof CustomError ? err.userMessage : 'Unhandled Error',
-        });
-      }
-    });
+    promises.push(
+      queue.add(async () => {
+        try {
+          const projectType = await getProjectTypeForIacFile(filePath);
+          iacFiles.push({
+            filePath,
+            projectType,
+            fileType,
+          });
+        } catch (err) {
+          iacFiles.push({
+            filePath,
+            fileType,
+            failureReason:
+              err instanceof CustomError ? err.userMessage : 'Unhandled Error',
+          });
+        }
+      }),
+    );
   }
 
-  const setTimeoutAsync = util.promisify(setTimeout);
-  while (queue.getQueueLength() + queue.getPendingLength() > 0) {
-    await setTimeoutAsync(100);
+  if (promises.length > 0) {
+    await Promise.all(promises);
   }
+
   if (iacFiles.length === 0) {
     throw IacDirectoryWithoutAnyIacFileError();
   }
--- a/src/lib/iac/detect-iac.ts
+++ b/src/lib/iac/detect-iac.ts
@@ -99,10 +99,9 @@ async function getDirectoryFiles(
   });
 
   const iacFiles: IacFileInDirectory[] = [];
-  const maxConcurrent = 50;
+  const maxConcurrent = 25;
   const queue = new Queue(maxConcurrent);
 
-  const promises: Array<Promise<IacFileInDirectory>> = [];
   for (const filePath of directoryPaths) {
     const fileType = pathLib
       .extname(filePath)
@@ -111,33 +110,29 @@ async function getDirectoryFiles(
     if (!fileType || !supportedExtensions.has(fileType)) {
       continue;
     }
-    promises.push(
+    iacFiles.push(
       queue.add(async () => {
         try {
           const projectType = await getProjectTypeForIacFile(filePath);
-          iacFiles.push({
+          return {
             filePath,
             projectType,
             fileType,
-          });
+          };
         } catch (err) {
-          iacFiles.push({
+          return {
             filePath,
             fileType,
             failureReason:
               err instanceof CustomError ? err.userMessage : 'Unhandled Error',
-          });
+          };
         }
       }),
     );
   }
 
-  if (promises.length > 0) {
-    await Promise.all(promises);
-  }
-
   if (iacFiles.length === 0) {
     throw IacDirectoryWithoutAnyIacFileError();
   }
-  return iacFiles;
+  return Promise.all(iacFiles);
 }
--- a/src/lib/iac/detect-iac.ts
+++ b/src/lib/iac/detect-iac.ts
@@ -21,6 +21,8 @@ import {
   IllegalTerraformFileError,
 } from '../errors/invalid-iac-file';
 import { Options, TestOptions, IacFileInDirectory } from '../types';
+import * as Queue from 'promise-queue';
+import * as util from 'util';
 
 const debug = debugLib('snyk-detect-iac');
 
@@ -97,7 +99,9 @@ async function getDirectoryFiles(
     maxDepth: options.maxDepth,
   });
 
-  const promises: Promise<IacFileInDirectory>[] = [];
+  const iacFiles: IacFileInDirectory[] = [];
+  const queue = new Queue(50);
+
   for (const filePath of directoryPaths) {
     const fileType = pathLib
       .extname(filePath)
@@ -107,31 +111,31 @@ async function getDirectoryFiles(
       continue;
     }
 
-    promises.push(
-      (async () => {
-        try {
-          const projectType = await getProjectTypeForIacFile(filePath);
-          return {
-            filePath,
-            projectType,
-            fileType,
-          };
-        } catch (err) {
-          return {
-            filePath,
-            fileType,
-            failureReason:
-              err instanceof CustomError ? err.userMessage : 'Unhandled Error',
-          };
-        }
-      })(),
-    );
+    queue.add(async () => {
+      try {
+        const projectType = await getProjectTypeForIacFile(filePath);
+        iacFiles.push({
+          filePath,
+          projectType,
+          fileType,
+        });
+      } catch (err) {
+        iacFiles.push({
+          filePath,
+          fileType,
+          failureReason:
+            err instanceof CustomError ? err.userMessage : 'Unhandled Error',
+        });
+      }
+    });
   }
 
-  const iacFiles = await Promise.all(promises);
+  const setTimeoutAsync = util.promisify(setTimeout);
+  while (queue.getQueueLength() + queue.getPendingLength() > 0) {
+    await setTimeoutAsync(100);
+  }
   if (iacFiles.length === 0) {
     throw IacDirectoryWithoutAnyIacFileError();
   }
-
   return iacFiles;
 }
--- a/src/lib/request/request.ts
+++ b/src/lib/request/request.ts
@@ -112,13 +112,14 @@ export = function makeRequest(
         const proxyUri = getProxyForUrl(url);
         if (proxyUri) {
           snykDebug('using proxy:', proxyUri);
+          // proxyAgent type is an EventEmitter and not an http Agent
           options.agent = (new ProxyAgent(proxyUri) as unknown) as http.Agent;
         } else {
           snykDebug('not using proxy');
         }
 
         if (global.ignoreUnknownCA) {
-          debug('Using insecure mode (ignore unkown certificate authority)');
+          debug('Using insecure mode (ignore unknown certificate authority)');
           options.rejectUnauthorized = false;
         }
 
--- a/src/lib/request/request.ts
+++ b/src/lib/request/request.ts
@@ -8,6 +8,7 @@ import { getProxyForUrl } from 'proxy-from-env';
 import * as ProxyAgent from 'proxy-agent';
 import * as analytics from '../analytics';
 import { Agent } from 'http';
+import * as https from 'https';
 import { Global } from '../../cli/args';
 import { Payload } from './types';
 import { getVersion } from '../version';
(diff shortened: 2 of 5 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -292,6 +298,11 @@ async function sendAndParseResults(
       });
     }
   }
+
+  const setTimeoutAsync = util.promisify(setTimeout);
+  while (queue.getQueueLength() + queue.getPendingLength() > 0) {
+    await setTimeoutAsync(100);
+  }
```
ipapast, line 305: The library doesn't provide any built in function yet, so this workaround was used instead. Inspired by an [open issue](https://github.com/promise-queue/promise-queue/issues/23)

src/lib/iac/detect-iac.ts
```diff
@@ -107,31 +111,31 @@ async function getDirectoryFiles(
...
+        });
+      }
+    });
   }
 
-  const iacFiles = await Promise.all(promises);
+  const setTimeoutAsync = util.promisify(setTimeout);
+  while (queue.getQueueLength() + queue.getPendingLength() > 0) {
+    await setTimeoutAsync(100);
+  }
```
ipapast, line 136: The library doesn't provide any built in function yet, so this workaround was used instead.

src/lib/iac/detect-iac.ts
```diff
@@ -90,14 +92,16 @@ async function getDirectoryFiles(
...
-  const iacFiles: IacFileInDirectory[] = [];
   const dirPath = pathLib.resolve(root, '.');
   const supportedExtensions = new Set(Object.keys(projectTypeByFileType));
 
   const directoryPaths = makeDirectoryIterator(dirPath, {
     maxDepth: options.maxDepth,
   });
 
+  const iacFiles: IacFileInDirectory[] = [];
+  const queue = new Queue(50);
```
almog27, line 103: Worth explaining what is 50 by adding some const - my first thought was a queue max size and not the max concurrent

src/lib/iac/detect-iac.ts
almog27: If so - not worth considering something that do have the async ability? such as `async-promise-queue`?

aron: In this instance we know all the promises that we are trying to make requests for. The `add()` method returns a promise that resolves to the result of the enqueued function so we can make use of that.

```
// In the loop...
promises.push(queue.add(() => ...));

// Here:
if (promises.length > 0) {
  return await Promise.all(promises);
}
```

In case that doesn't make sense :) I've put a demo together here: https://runkit.com/aron/promisequeue-example

src/lib/snyk-test/run-test.ts
```diff
@@ -226,23 +226,36 @@ async function sendAndParseResults(
...
+
+          const projectName =
+            iacScan.projectNameOverride || iacScan.originalProjectName;
+          const result = await parseIacTestResult(
+            res,
+            iacScan.targetFile,
+            projectName,
+            options.severityThreshold,
+          );
+          results.push(result);
```
aron, line 256: The results are now captured in the promises array so we could just return `result` here and drop the additional `results` array.

src/lib/snyk-test/run-test.ts
```diff
@@ -226,23 +226,36 @@ async function sendAndParseResults(
...
+  // we introduced a concurrent requests limit of 25. With that, we will only process MAX 25 requests at the same time.
+  // In the future, we would probably want to introduce a RATE_LIMIT specific for IaC
+  const maxConcurrent = 25;
+  const queue = new Queue(maxConcurrent);
+
   for (const payload of payloads) {
     await spinner.clear<void>(spinnerLbl)();
     await spinner(spinnerLbl);
     if (options.iac) {
-      const iacScan: IacScan = payload.body as IacScan;
```
aron, line 233: A couple of extra observations here now.

1) Previously this would reset the spinner once per iteration. As we're now batching things this is not going to work as expected...
1) I think we could simplify this by handling the IaC condition first, otherwise handling the other paths rather than doing it inside the loop. This would mean that we could declare `promises` for IaC only and `results` for the primary path. This would aid refactoring in the future allowing us to extract the iac path into a seperate function.

Example:

```ts
// IaC path...
if (options.iac) {
  const promises: Promise<TestResult>[] = []
  const maxConcurrent = 25;
  const queue = new Queue(maxConcurrent);

  await spinner(spinnerLbl);
  for (const payload of payloads) {
    promises.push(queue.add(...));
  }

  return Promise.all(promises)
}

// Standard path...
const results: TestResult[] = []

...

for (const payload of payloads) {
    await spinner.clear<void>(spinnerLbl)();
    await spinner(spinnerLbl);
    ...
}

return results
```

aron (commented):
This is looking really good from my side, loving those time improvements ❤️ . A couple of final comments around typings and code structure. Also regarding the commit messages/release notes the only one that is really a "feature" is "run validate & test in parallel" can we update the other two to use the `refactor:` label. And let's include "IaC" in the othere one for clarity e.g. "Run the IaC queries in validate & test in parallel"

src/lib/snyk-test/run-test.ts
```diff
@@ -226,23 +226,36 @@ async function sendAndParseResults(
   options: Options & TestOptions,
 ): Promise<TestResult[]> {
   const results: TestResult[] = [];
+  const promises: any[] = [];
```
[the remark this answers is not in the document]
ipapast: updated to `Array<Promise<TestResult>> = [];`

test/request.test.ts
```diff
@@ -49,7 +51,7 @@ test('request calls needle as expected and returns status code and body', (t) =>
             follow_max: 5, // default
             timeout: 300000, // default
             json: undefined, // default
-            agent: undefined, // should not be set when not using proxy
+            agent: new https.Agent({ keepAlive: true }), // should not be set when not using proxy
```
aron, line 54: nit: comment is out of date.

src/lib/request/request.ts
```diff
@@ -94,19 +95,24 @@ export = function makeRequest(
...
           follow_max: 5,
           family: payload.family,
+          agent,
         };
 
         const proxyUri = getProxyForUrl(url);
         if (proxyUri) {
           snykDebug('using proxy:', proxyUri);
-          options.agent = (new ProxyAgent(proxyUri) as unknown) as Agent;
+          options.agent = (new ProxyAgent(proxyUri) as unknown) as http.Agent;
```
aron, line 115: Lets also install the types fro ProxyAgent via `npm install @types/http-proxy-agent` and avoid this casting.

src/lib/iac/detect-iac.ts
```diff
@@ -90,14 +91,18 @@ async function getDirectoryFiles(
...
 
   const directoryPaths = makeDirectoryIterator(dirPath, {
     maxDepth: options.maxDepth,
   });
 
+  const iacFiles: IacFileInDirectory[] = [];
+  const maxConcurrent = 50;
+  const queue = new Queue(maxConcurrent);
+
+  const promises: Array<Promise<IacFileInDirectory>> = [];
```
aron, line 105: nit: The `type[]` is a more common ideom now for expressing an array.

```suggestion
  const promises: Promise<IacFileInDirectory>[] = [];
```

src/lib/iac/detect-iac.ts
```diff
@@ -106,27 +111,33 @@ async function getDirectoryFiles(
     if (!fileType || !supportedExtensions.has(fileType)) {
       continue;
     }
+    promises.push(
+      queue.add(async () => {
+        try {
+          const projectType = await getProjectTypeForIacFile(filePath);
+          iacFiles.push({
```
aron, line 118: Same cleanup applies here, we no longer need both `promises` and `iacFiles` we can just return the object here and resolve promises later on to get the files.

src/lib/iac/detect-iac.ts
```diff
@@ -106,27 +111,33 @@ async function getDirectoryFiles(
...
-      });
-    } catch (err) {
-      iacFiles.push({
-        filePath,
-        fileType,
-        failureReason:
-          err instanceof CustomError ? err.userMessage : 'Unhandled Error',
-      });
-    }
+  if (promises.length > 0) {
```
aron, line 135: Suggestion:

```ts
if (promises.length === 0) {
  throw IacDirectoryWithoutAnyIacFileError();
}
return Promise.all(promises);
```

src/lib/snyk-test/run-test.ts
```diff
@@ -225,72 +225,90 @@ async function sendAndParseResults(
...
+  // Note for the IaC Test Flow:
+  // There is a RATE_LIMIT setup for network requests within a time period.
+  // To support this limit and avoid getting back 502 errors from registry,
+  // we introduced a concurrent requests limit of 25. With that, we will only process MAX 25 requests at the same time.
+  // In the future, we would probably want to introduce a RATE_LIMIT specific for IaC
+
+  if (options.iac) {
+    const maxConcurrent = 25;
+    const queue = new Queue(maxConcurrent);
+    const iacResults: Array<Promise<TestResult>> = [];
```
aron, line 237: nit:

```suggestion
    const iacResults: Promise<TestResult>[] = [];
```

aron (approved):
👏 This looks great, a huge improvement on the previous test times. I've a few minor comments below where I think things would benefit from a little more cleanup but otherwise I think this is ready for review from the wider team.

src/lib/iac/detect-iac.ts
ipapast: Ah, this was my auto TSLint fix.
Apparently it is giving an: 
`TSLint: Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead.(array-type)` [link](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md)

From a bit of googling, I understand that **[ ]** and **Array** are the same, except for the times we are using a `readonly` type. [reference](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#caveats)

I'm happy to put it back as it was since this is the preference, just wondering if I should be adding an ignore rule in TSLint too:
`"rules": {
    "array-type": false
}`

src/lib/iac/detect-iac.ts
ipapast: 🤦 thanks

src/lib/snyk-test/run-test.ts
ipapast: ^^I'm bringing it back too

src/lib/iac/detect-iac.ts
aron: Oh my apologies, leave it as is.

src/lib/request/request.ts
```diff
@@ -5,12 +5,13 @@ import * as querystring from 'querystring';
 import * as zlib from 'zlib';
 import * as config from '../config';
 import { getProxyForUrl } from 'proxy-from-env';
-import * as ProxyAgent from 'proxy-agent';
+import HttpProxyAgent = require('http-proxy-agent');
```
JackuB, line 8: Two things here:

1. `proxy-agent` is not used anymore, so it may be removed from the package.json?
2. `http-proxy-agent` has types defined in package.json, but not the package itself?

And out of curiosity, why the different import/require statement?

src/lib/snyk-test/run-test.ts
```diff
@@ -225,72 +225,90 @@ async function sendAndParseResults(
...
-        displayTargetFile,
-        dockerfilePackages,
-        platform,
-      } = prepareResponseForParsing(
-        payloadCopy,
-        res as TestDependenciesResponse,
-        options,
-      );
+    /** sendTestPayload() deletes the request.body from the payload once completed. */
+    const payloadCopy = Object.assign({}, payload);
```
JackuB, line 267: Is it problematic here that `Object.assign` is not creating a deep clone?

src/lib/iac/detect-iac.ts
```diff
@@ -90,14 +91,17 @@ async function getDirectoryFiles(
...
-  const iacFiles: IacFileInDirectory[] = [];
   const dirPath = pathLib.resolve(root, '.');
   const supportedExtensions = new Set(Object.keys(projectTypeByFileType));
 
   const directoryPaths = makeDirectoryIterator(dirPath, {
     maxDepth: options.maxDepth,
   });
 
+  const iacFiles: IacFileInDirectory[] = [];
+  const maxConcurrent = 50;
```
JackuB, line 102: Why do we have `maxConcurrent` set as 50 here and 25 in the other file?

JackuB (changes requested):
Looking good, I'd like to clear the questions about swapping of proxy dependencies first.

Also, we do collect some performance metrics from the CLI now, so we should be able to see real-world impact from this change once it's released.

src/lib/request/request.ts
ipapast: 1. good catch, removed 👍 

About the import/require: 
HttpProxyAgent is exported like this in @types/http-proxy-agent:
`export = HttpProxyAgent;`
And when I was trying a traditional i`mport {HttpProxyAgent} from '...'` I was experiencing errors, so I found in the documentation that the `import HttpProxyAgent = require('http-proxy-agent');` is what we should be using when something is exported as `export =`.  You can see it [here](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require)

src/lib/snyk-test/run-test.ts
ipapast: hm not sure, but maybe @almog27 can help. do we need a deep clone of the payload or is it ok as it is?

src/lib/iac/detect-iac.ts
ipapast: there is a rate limit configured in registry, and we experienced some 502s coming back when we issued too many requests in parallel. We were basically playing with the numbers and running it locally.

So the goal here was to avoid any errors caused by rate limits, just _**slowly**_ work through them.

50 for detecting the files in the directory (we were also considering CPU usage), 25 mostly because of the RATE_LIMIT since we are hitting the endpoint to sendTestPayload() in the for...loop of all the payloads

src/lib/snyk-test/run-test.ts
aron: Worth noting that this is existing behavior and only shows up because it's now outside of the `else` block.

src/lib/iac/detect-iac.ts
aron: I would agree that we should keep it consistent, let's use 25. I don't think it's a big deal as the difference will be small, the main thing is that it is batched and doesn't try to process _all_ the files at once.

src/lib/request/request.ts
aron: Oh this is my fault I think. @ipapast [`proxy-agent`](https://www.npmjs.com/package/proxy-agent) and [`http-proxy-agent`](https://www.npmjs.com/package/http-proxy-agent) are two different libraries that do the same thing maintained by the same person. `proxy-agent` supports more than just HTTP. Because I don't have a full understanding of how this is used downstream I think we need to keep using `proxy-agent` which also seems to bundle types but it may be that they're incompatible with `http.Agent`. So I think the original typing issue may be resolved with just an upgrade, or it might be that we need to keep the awkward casting.

ipapast: So I've been trying to make it work with `http-proxy-agent`. IT WON'T. 😢

Since the **http-proxy-agent** module uses old-school imports like: 
```
import net from 'net';
import tls from 'tls';
```

When we run the build,  it will fail with errors of style 
```
Module '"net"' has no default export.
Module '"tls"' has no default export.
```
Unfortunately, it's a third party module so we can't really edit these imports. The only way to make it pass was by adding `"allowSyntheticDefaultImports": true` in the tsconfig.json

src/lib/iac/detect-iac.ts
ipapast: I think we had a conversation about the files and laptops/local configurations limitations, but happy to change it for consistency and re-consider in the future if needed.

src/lib/request/request.ts
ipapast: Upgrading proxy-agent to 4.0.1 (latest) leads to very similar errors which are coming from a chained import of the `agent-base`.
```
 Module '"net"' has no default export.
 Module '"http"' has no default export.
  Module '"https"' has no default export.
```

src/lib/iac/detect-iac.ts
JackuB: 50 concurrent requests sounds a bit hardcore for Node.js, no?

ipapast: **25, apologies

src/lib/request/request.ts
ipapast: Reverted back to http-proxy-agent and the weird casting. Added comment: `proxyAgent type is an EventEmitter and not an http Agent`

ipapast:
> Looking good, I'd like to clear the questions about swapping of proxy dependencies first.
> 
> Also, we do collect some performance metrics from the CLI now, so we should be able to see real-world impact from this change once it's released.

Hi @JackuB, we reverted back to http-proxy-agent and added a comment to explain.
Also went back to 25 concurrent requests in the queue for both detection and running the tests.

Could you check it one final time please? 🙇 
Would you also be able to point me on where to find these performance metrics from the CLI you mentioned?

JackuB (approved):

PR #1578, opened 2021-01-10
fix: alert msg, constant failure of CLI alerts

Avishagp:
- Alert only for jobs that failed again (not the whole array of failedJobs)
- Wait for current run to finish before attempting re-run (fixes failures of CLI alerts)

CLI alerts failed a lot recently, as a result of trying to re-run smoke tests that are already running. This PR fixes this problem by waiting for the current run to finish before attempting a re-run.

### DIFF
--- a/test/cli-alert/src/index.ts
+++ b/test/cli-alert/src/index.ts
@@ -15,46 +15,46 @@ const octokit = new Octokit({
 });
 const slackWebhook = new IncomingWebhook(SLACK_WEBHOOK_URL);
 
-async function checkJobs(
-  smokeTest1: number,
-  smokeTest2: number,
+async function discoverConsecutiveFailures(
+  firstWorkflowRun: number,
+  secondWorkflowRun: number,
 ): Promise<string[]> {
   // Get all jobs of the latest 2 smoke tests
-  const jobs1 = (
+  const firstWorkflowRunJobs = (
     await octokit.actions.listJobsForWorkflowRun({
       owner: 'snyk',
       repo: 'snyk',
       // eslint-disable-next-line @typescript-eslint/camelcase
-      run_id: smokeTest1,
+      run_id: firstWorkflowRun,
     })
   ).data.jobs;
 
-  const jobs2 = (
+  const secondWorkflowRunJobs = (
     await octokit.actions.listJobsForWorkflowRun({
       owner: 'snyk',
       repo: 'snyk',
       // eslint-disable-next-line @typescript-eslint/camelcase
-      run_id: smokeTest2,
+      run_id: secondWorkflowRun,
     })
   ).data.jobs;
 
   const failedJobs: string[] = [];
 
   // If the same job failed in both smoke tests, it has been failing for 2 hours now. Save job to re-run later.
-  for (const jobName of jobs1.map((j) => j.name)) {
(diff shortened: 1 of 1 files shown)


test/cli-alert/src/index.ts
```diff
@@ -101,7 +101,10 @@ async function waitForConclusion(runID: number): Promise<boolean> {
       }`,
     );
   }
+  console.log('Finished run.');
+}
 
+async function checkJobConclusion(runID: number): Promise<boolean> {
   // Get conclusions of the jobs that failed before
   const smokeTetsJobs = (
```
maxjeffos, line 109: There's a typo here you could consider fixing.
```suggestion
  const smokeTestJobs = (
```

test/cli-alert/src/index.ts
```diff
@@ -117,11 +120,13 @@ async function waitForConclusion(runID: number): Promise<boolean> {
     failedJobs.includes(job.name),
   );
   for (const job of rerunJobs) {
-    if (job.conclusion === 'failure') {
-      return false;
+    if (job.conclusion !== 'failure') {
+      // Remove succeeding jobs from failedJobs
+      failedJobs.splice(failedJobs.indexOf(job.name), 1);
```
maxjeffos, line 125: mutating `failedJobs` while iterating over is really not a good idea.
I'd suggest something like:

```
  const failedAgain = rerunJobs.rerunJobs.filter(job => job.conclusion === 'failure');
  return failedAgain.length > 0;
```
assuming you don't need to update that `failedJobs` global. Speaking of which, using a global like this kind of non-idiomatic to JS/TS.

Avishagp: Got it. Thank you! 🙏

test/cli-alert/src/index.ts
```diff
@@ -10,15 +10,15 @@ if (!process.env.USER_GITHUB_TOKEN || !process.env.SLACK_WEBHOOK_URL) {
...
 
-// Map from job id to it's name and conclusion
-const failedJobs: string[] = [];
-
 const octokit = new Octokit({
   auth: GITHUB_TOKEN,
 });
 const slackWebhook = new IncomingWebhook(SLACK_WEBHOOK_URL);
 
-async function checkJobs(smokeTest1: number, smokeTest2: number) {
```
maxjeffos, line 21: `checkJobs` could be more descriptive or you could have a function header comment. I don't know (though I might be able to guess) what it does without having to read the code.

test/cli-alert/src/index.ts
```diff
@@ -38,6 +38,8 @@ async function checkJobs(smokeTest1: number, smokeTest2: number) {
     })
   ).data.jobs;
 
+  const failedJobs: string[] = [];
+
   // If the same job failed in both smoke tests, it has been failing for 2 hours now. Save job to re-run later.
```
maxjeffos, line 43: Wondering if we could be more clear with the terminology. Like maybe `both smoke tests` might be `both smoke test workflows runs` as that would coincide with the GA terminology.

maxjeffos (approved):
Also per our talk, that stuff around L150 could just be
```
const smokeTestsID = allWorkflows.workflows.find((w) => w.name === 'Smoke Tests');
```

PR #1584, opened 2021-01-17
fix: iac k8s validation helm file

almog27:
- [x] Ready for review

#### What does this PR do?
This PR fixes the fact the for k8s files `iac test` is not doing helm validation and fail on internal error instead of showing a proper message that indicates this file type is not supported.

As part of an investigation that happens in Cloud Config team, we understood that in the UI we have validation for Helm files as part of the k8s file scanning, while in the CLI there is no such validation.
This PR align between those

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,8 @@
     "tempfile": "^2.0.0",
     "update-notifier": "^4.1.0",
     "uuid": "^3.3.2",
-    "wrap-ansi": "^5.1.0"
+    "wrap-ansi": "^5.1.0",
+    "yamljs": "^0.3.0"
   },
   "devDependencies": {
     "@types/agent-base": "^4.2.1",
--- a/src/lib/iac/iac-parser.ts
+++ b/src/lib/iac/iac-parser.ts
@@ -1,7 +1,12 @@
 //TODO(orka): take out into a new lib
 import * as YAML from 'js-yaml';
+import * as yamljs from 'yamljs';
 import * as debugLib from 'debug';
-import { IllegalIacFileErrorMsg, NotSupportedIacFileErrorMsg } from '../errors';
+import {
+  IllegalIacFileErrorMsg,
+  NotSupportedIacFileErrorMsg,
+  InvalidYamlFileErrorMsg,
+} from '../errors';
 import request = require('../request');
 import { api as getApiToken } from '../api-token';
 import * as config from './../config';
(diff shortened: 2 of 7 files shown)


src/lib/iac/iac-parser.ts
```diff
@@ -45,6 +45,10 @@ function parseYamlOrJson(fileContent: string, filePath: string): any {
   return undefined;
 }
 
+export function isHelm(fileContent: string) {
+  return fileContent.includes('{{');
```
p0tr3c, line 49: This is not enough to identify file as only helm.
if `{{` is used outside of quotes or string literals then this is fine.
But for configMaps these characters may be included as part of configuration file for a service.

p0tr3c (approved):

package.json
```diff
@@ -101,7 +101,8 @@
     "tempfile": "^2.0.0",
     "update-notifier": "^4.1.0",
     "uuid": "^3.3.2",
-    "wrap-ansi": "^5.1.0"
+    "wrap-ansi": "^5.1.0",
+    "yamljs": "^0.3.0"
```
JackuB, line 105: We already have a different yaml parsing library in the snyk/policy package: https://github.com/snyk/policy/blob/master/package.json#L30

src/lib/iac/iac-parser.ts
```diff
@@ -45,6 +50,19 @@ function parseYamlOrJson(fileContent: string, filePath: string): any {
   return undefined;
 }
 
+export function isInvalidYaml(filePath: string, fileContent: string) {
```
JackuB, line 53: Wouldn't it make sense to have the check be the opposite: `isValidYaml`?

JackuB (changes requested):
Let's swap the dependency 👍

package.json
almog27: The `js-yaml` library does not do the required validation, but a parsing to JSON

JackuB: Calling a safeLoad function from it will validate it, no?

almog27: @JackuB I checked the posibility to use it instead, while js-yaml parse well multiple docs, yamljs only return the first one.
And yamljs do the validation while js-yaml not.
So we need both of them.

JackuB: Got some examples of those invalid docs? 🤔

almog27: On the `js-yaml` it writes on the load 

> NOTE: JS-YAML does not support schema-specific tag resolution restrictions. So, the JSON schema is not as strictly defined in the YAML specification. It allows numbers in any notation, use Null and NULL as null, etc. The core schema also has no such restrictions. It allows binary notation for integers.
https://github.com/nodeca/js-yaml#load-string---options-

As a result - when it finds invalid it changes it to null and not throw an error

`yamljs` readme only talks on a single yaml file an not talking on multidoc yaml.
However - I did tested it locally - and it only return the first one.
https://github.com/jeremyfa/yaml.js#readme

almog27:
Closing this PR till further investigation

PR #1597, opened 2021-01-29
fix(test): consider presence of fixedIn field as fixableVuln

JackuB:
This extends the `isVulnUpgradable` check to consider `fixedIn` array. 
Should resolve the issue with `--fail-on` flag - currently option `all` won't consider vulns where we know `fixedIn`, but don't have any `upgradePath`.

### DIFF
--- a/src/cli/commands/test/vuln-helpers.ts
+++ b/src/cli/commands/test/vuln-helpers.ts
@@ -46,7 +46,7 @@ export function hasPatches(testResults: any[]): boolean {
 }
 
 export function isVulnUpgradable(vuln) {
-  return vuln.isUpgradable || vuln.isPinnable;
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/vuln-helpers.ts
```diff
@@ -46,7 +46,7 @@ export function hasPatches(testResults: any[]): boolean {
 }
 
 export function isVulnUpgradable(vuln) {
-  return vuln.isUpgradable || vuln.isPinnable;
+  return vuln.isUpgradable || vuln.isPinnable || (vuln.fixedIn || []).length;
```
lili2311, line 49: there is also `nearestFixedIn` for some project types in case thats useful here

JackuB:
Closing as it's stale

PR #1600, opened 2021-01-30
Make fn async and use await to avoid race condition in tests causing test flakeyness

maxjeffos:
- [X] Ready for review

#### What does this PR do?
Updates the `writeContentsToFileSwallowingErrors` function to use be async and wait for the `finish` event to resolve and then use await when calling this function. This should eliminate the flakeyness of the corresponding tests which manifests intermittently with `test/acceptance/cli-args.test.ts` and `test/json-file-output.test.ts`

This test has been flakey for a while.
Example of test flaking: https://app.circleci.com/pipelines/github/snyk/snyk/3423/workflows/73c644ca-dd4b-4158-9510-ad4efa5b8eb7/jobs/19088
The issue is that the fn that the test is flaking was sometimes not done its IO before returning.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -76,9 +76,9 @@ async function runCommand(args: Args) {
   // also save the json (in error.json) to file if option is set
   if (args.command === 'test') {
     const jsonResults = (commandResult as TestCommandResult).getJsonResult();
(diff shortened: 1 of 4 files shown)


Avishagp (approved):
lgtm!

PR #1617, opened 2021-02-04
fix: generate sha256 with correct path

andremidea:
- [X] Ready for review

#### What does this PR do?

It changes the directory to `binary-releases` before generating a sha256, so that it has the correct file name to be verified.

This regression was introduced with this change https://github.com/snyk/snyk/pull/1564, which caused the `sha256` file to be generated with `f9245dbba7ccd1e8d4b5eebc7885f445a9af465fc6ef68cc89afa4964575ee36  binary-releases/snyk-macos` where the file path should be `snyk-macos`.

With this change, it will generate without the `binary-releases` prefix.


maxjeffos (approved):
Thanks @andremidea looks good. I checked the generated SHAs here: https://app.circleci.com/pipelines/github/snyk/snyk/3519/workflows/ba9d364a-727c-450b-98dd-5bba0a4e516d/jobs/19705/artifacts

PR #1618, opened 2021-02-04
Validate binary release checksums

maxjeffos:
- [x] Ready for review

#### What does this PR do?
Validates the computed checksums for the binary releases and fail build if they are not ok.

https://snyk.slack.com/archives/CH6616JU8/p1612466243049700

ZD 8834, 8836

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,6 +307,11 @@ jobs:
             ls -la ./binary-releases
       - store_artifacts:
           path: ./binary-releases
+      - run:
+          name: Validate binary release checksums
+          working_directory: ~/snyk/binary-releases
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -307,6 +307,11 @@ jobs:
             ls -la ./binary-releases
       - store_artifacts:
           path: ./binary-releases
+      - run:
```
maxjeffos, line 310: I think it makes sense to validate the checksums after storing the artifacts (in CCI) so that we can investigate the generated artifacts to understand what went wrong if the do not validate ok.

JackuB (approved):

PR #1620, opened 2021-02-05
Add dangerjs rule to check for bad dep targets

maxjeffos:
- [ ] Ready for review

#### What does this PR do?
Adds dangerjs rules to check for dependency/devDependency targets that we want to prevent being added.
If a dependency or devDependency starts with:
`git+` -> we error
`http://` -> we warn
`https://` -> we warn

Included tests so we can test the DangerJS rules going forward (in addition to this one).

We can't be having dependencies that start with `git+` as it will cause problems installing `snyk` if a user does not have git installed locally. `http://` we don't want be cause it's not secure. `https://` is probably ok but we'll warn on it to be sure.


dangerfile.js
```diff
@@ -1,9 +1,51 @@
...
+
+function dangerCheckPackageJsonDependencyTargets(packageJsonFilePath) {
+  const allDeps = getDependenciesList(packageJsonFilePath);
+  for (const [depName, depTarget] of Object.entries(allDeps)) {
+    const depTargetContainsGit = checkDepedencyTarget(depTarget, dependencyTargetWithGit);
+    const depTargetContainsHttp = checkDepedencyTarget(depTarget, dependencyTargetWithHttp);
+    const depTargetContainsHttps = checkDepedencyTarget(depTarget, dependencyTargetWithHttps);
+
+    if (depTargetContainsGit) {
+      fail(`The package.json file ${packageJsonFilePath} contains a dependency or devDependency \`${depName}\` with a \`git+\` target`);
```
JackuB, line 38: I'd also explain why it is a reason for a failure

maxjeffos: @JackuB I updated it.

JackuB (approved):

JackuB (changes requested):
Just realised that this PR won't save us from a transitive dependency using git/https 🤔 I think we should rethink our approach here. Maaybe using `npm ls` or something similar

PR #1621, opened 2021-02-08
re-release iac experimental single k8s file - with NPM dependency

rontalx:
This PR is an exact replica of this PR which was merged & reverted:
https://github.com/snyk/snyk/pull/1601
it was reverted due to an incident with the git url dependency (#incident-cli-git-url-dependency).
The only change is that now it uses a public npm version instead of the git one.


aron (approved):
Even better!

dtrunley-snyk (approved):

PR #1624, opened 2021-02-09
test: add pagerduty, check if test is running before attemmpting rerun

Avishagp:
- [X] Ready for review

#### What does this PR do?
* Add PagerDuty to CLI alerts
* Check if latest Smoke Test workflow is still running **before** checking conclusion and attempting re-run

### DIFF
--- a/test/cli-alert/src/index.ts
+++ b/test/cli-alert/src/index.ts
@@ -1,14 +1,22 @@
 import { Octokit } from '@octokit/rest';
 import { IncomingWebhook } from '@slack/webhook';
 import { IncomingWebhookDefaultArguments } from '@slack/webhook';
+import { event } from '@pagerduty/pdjs';
 
-if (!process.env.USER_GITHUB_TOKEN || !process.env.SLACK_WEBHOOK_URL) {
-  console.error('Missing USER_GITHUB_TOKEN or SLACK_WEBHOOK_URL');
+if (
+  !process.env.USER_GITHUB_TOKEN ||
+  !process.env.SLACK_WEBHOOK_URL ||
+  !process.env.PD_ROUTING_KEY
+) {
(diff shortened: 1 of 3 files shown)


test/cli-alert/src/index.ts
```diff
@@ -64,6 +72,24 @@ async function discoverConsecutiveFailures(
...
+      event_action: 'trigger',
+      payload: {
+        summary: 'CLI Alert. Smoke tests failing',
+        source: 'Snyk CLI Smoke tests',
+        severity: 'warning',
+      },
+    },
+  })
+    .then(console.log)
+    .catch(console.error);
```
JackuB, line 90: If you are `await`ing, then it's easier to reason about it with `try/catch` and 

```js
try {
  const pdResponse = await event() 
  console.log(pdResponse)
} catch (error) {
  console.error(error)
}
```

Or maybe not even catching it and just let it throw to kill the run?

test/cli-alert/src/index.ts
```diff
@@ -202,6 +213,21 @@ async function run() {
       await waitForConclusion(id);
     }
 
+    console.log('Checking smoke tests jobs...');
+    const failedWorkflows = await discoverConsecutiveFailures(
+      latestWorkflowRuns[0].id,
+      latestWorkflowRuns[1].id,
+    );
+
+    if (!failedWorkflows.length || failedWorkflows.length < 1) {
```
JackuB, line 222: `failedWorkflows.length < 1` would mean that `failedWorkflows` is `0`, and `!!0 = false` so it's already checked by the first condition.

```suggestion
    if (!failedWorkflows.length) {
```

JackuB (changes requested):

JackuB (approved):

PR #1627, opened 2021-02-10
feat: update critical sev colour

karenyavine:
- [ ] Ready for review

#### What does this PR do?
* Adds colors for future support for critical severity
* Adds default fallback color for unknown severities
* Single place for color definitions

### DIFF
--- a/src/cli/commands/protect/prompts.ts
+++ b/src/cli/commands/protect/prompts.ts
@@ -20,6 +20,10 @@ import * as config from '../../../lib/config';
 import * as snykPolicy from 'snyk-policy';
 import chalk from 'chalk';
 import { AnnotatedIssue, SEVERITY } from '../../../lib/snyk-test/legacy';
+import {
+  legacySeveritiesColourMapping,
+  defaultSeverityColor,
+} from '../../../lib/snyk-test/common';
 import { titleCaseText } from '../test/formatters/legacy-format-issue';
 
 const debug = debugModule('snyk');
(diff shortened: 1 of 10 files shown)


src/cli/commands/protect/prompts.ts
```diff
@@ -45,24 +49,10 @@ function sort(prop) {
...
-    high: {
-      colorFunc(text) {
-        return chalk.bold.red(text);
-      },
-    },
-  };
-  return severitiesColourMapping[severity].colorFunc(msg);
+  const severityColor = legacySeveritiesColourMapping[severity]
+    ? legacySeveritiesColourMapping[severity]
+    : defaultSeverityColor;
```
gitphill, line 54: `legacySeveritiesColourMapping[severity] || defaultSeverityColor`?

src/cli/commands/test/iac-output.ts
```diff
@@ -48,9 +35,12 @@ function formatIacIssue(
 
   const description = extractOverview(issue.description).trim();
   const descriptionLine = `\n    ${description}\n`;
+  const severityColor = severitiesColourMapping[issue.severity]
+    ? severitiesColourMapping[issue.severity]
+    : defaultSeverityColor;
```
aron, line 40: Instead of exporting both `severitiesColourMapping` and `defaultSeverityColor` and having this logic in each file, how about just exporting a single function (maybe: `severityColor(sev: SEVERITY)`?) that does this for you and returns the `colorFunc()` function? Would clean up all the call sites.

aron (approved):

src/lib/snyk-test/common.ts
```diff
@@ -24,6 +25,7 @@ export enum SEVERITY {
   LOW = 'low',
   MEDIUM = 'medium',
   HIGH = 'high',
+  CRITICAL = 'critical',
```
karenyavine, line 28: @JackuB Any reason we have severities enum defined twice? once here and once in `src/lib/snyk-test/legacy.ts`

JackuB: Don't think it's intended

src/lib/snyk-test/common.ts
```diff
@@ -41,8 +43,72 @@ export const SEVERITIES: Array<{
...
+  },
+};
+
+export const defaultSeverityColor = {
+  colorFunc(text) {
+    return chalk.grey(text);
+  },
+};
+
+export function getSeveritiesColour(severity) {
```
JackuB, line 104: Screen Shot 2021-02-16 at 13 46 39

Shouldn't this file be typed since we are introducing these functions?

src/cli/commands/test/iac-output.ts
```diff
@@ -48,9 +32,10 @@ function formatIacIssue(
 
   const description = extractOverview(issue.description).trim();
   const descriptionLine = `\n    ${description}\n`;
+  const severityColor = getSeveritiesColour(issue.severity);
```
JackuB, line 35: If I'm reading this right, we are planning to introduce the `critical` severity in a `issue.severity` field, this means you went for a breaking change on the API level? Are you worried about older clients in that case?

JackuB (changes requested):
Should this also be documented in the CLI help?

Left some minor questions before approving

src/cli/commands/test/iac-output.ts
karenyavine: Yes it will be a breaking change, but will be sent only to clients using CLI version greater than or equal to whatever version this gets shipped in

JackuB (approved):

PR #1645, opened 2021-02-17
feat: Don't open browser auth if inside container

jurglic:
#### What does this PR do?

Allows running auth command when inside a docker container by:
* not trying to open a browser via xdg open (breaks if not present or in container)
* providing a better auth prompt/message
* doubling the auth timeout period, so users have more time open browser
  manually

This makes it possible to use auth command via "docker scan" on soon to be released scan plugin for docker linux.

Run "snyk auth" inside a container.

https://snyksec.atlassian.net/browse/DC-1077

### DIFF
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -7,6 +7,7 @@ import { Spinner } from 'cli-spinner';
 import * as snyk from '../../../lib';
 import { verifyAPI } from './is-authed';
 import { isCI } from '../../../lib/is-ci';
+import { isDocker } from '../../../lib/is-docker';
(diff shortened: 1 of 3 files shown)


src/cli/commands/auth/index.ts
```diff
@@ -25,7 +26,7 @@ const debug = Debug('snyk-auth');
 let attemptsLeft = 0;
 
 function resetAttempts() {
-  attemptsLeft = 30;
+  attemptsLeft = isDocker() ? 60 : 30;
```
JackuB, line 29: 👍

JackuB (approved):

PR #1646, opened 2021-02-18
feat: execute all tests with lerna to test all packages

lili2311:
- [x] Ready for review

#### What does this PR do?
Run commands that need to be run on all packages with Lerna, and use `npm` for the rest.
Lerna can execute all package tests found in the repo
https://github.com/lerna/lerna/tree/main/commands/run#readme

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ commands:
     steps:
       - run:
           name: Build
-          command: npm run build
+          command: lerna run build --stream
       - persist_to_workspace:
           root: .
           paths:
@@ -163,6 +163,9 @@ jobs:
           name: Install npm@7
           command: |
             sudo npm install -g npm@7
+      - run:
+          name: Install Lerna
+          command: sudo npm install -g lerna
       - show_node_version
       - update_local_npmrc_linux
       - install_deps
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -226,9 +232,11 @@ jobs:
           name: Update local .npmrc file (Windows)
           command: echo "//registry.npmjs.org/:_authToken=$env:NPM_TOKEN" >> .npmrc
       - run:
-          name: Install dependencies (Windows)
-          command: npm install --ignore-scripts # ignore-scripts required because of postinstall script in snyk-resolve-deps package
-      - build_ts
+          name: Install dependencies @snyk/snyk
+          command: npm install
+      - run:
```
JackuB, line 237: Doesn't this happen automatically with npm@7? I tried adding `axios` as a local devDep to the protect package and it got installed after `npm i`

Screen Shot 2021-02-18 at 16 49 23

lili2311: removed, you are right. Forgot about npm 7 😅

lili2311:
@JackuB please 👀

test/analytics.spec.ts
```diff
@@ -128,7 +128,7 @@ describe('Analytics basic testing', () => {
         'integrationEnvironmentVersion',
       ].sort(),
     );
-  });
+  }, 10000);
```
lili2311, line 131: these were super flaky and not completing in a default given time

.circleci/config.yml
```diff
@@ -241,25 +247,25 @@ jobs:
           steps:
             - run:
                 name: Run "Root" tap tests
-                command: npm run test:test
+                command: lerna run test:test --stream
```
JackuB, line 250: So is the plan that each package will have a task called `test:test` and `test:jest`? I think it's just a limitation of the current CLI that we have tests split like this.

How about removing the `test` script from the root package.json (as we can't/don't use it now) and just have a `lerna run test` that will test packages? We might even have a single container dedicated for that now

PR #1654, opened 2021-02-22
feat: iac experimental terraform support

rontalx:
## What does this PR do?

Adds Single Terraform  file scan support to the: `snyk iac test --experimental` command.

## Running Locally

In order to run locally, checkout the branch, build and run the local built version.
make sure to download the `.iac-data` from [here](https://drive.google.com/drive/u/0/folders/1-9im-HH0E5rcaMC0G7YE1tYiWh3kZpsM) and place it in your CWD root folder where you're running the CLI from.
e.g: 
```
snyk iac test ./terraform/some_file.tf --experimental
```

**TODO**
- [x] Add  relevant `.wasm` file for terraform regression tests.
- [x] Decide how to proceed with `hcl-to-json` lib and HCL2 syntax validations.
- [x] Double check with new wasm supporting multiple CSP from this PR: https://github.com/snyk/cloud-config-opa-policies/pull/21


src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function isLocalCacheExists(): boolean {
+  return REQUIRED_LOCAL_CACHE_FILES.every(fs.existsSync);
+}
+
+export function getLocalCachePath(engineType: EngineType) {
+  switch (engineType) {
+    case EngineType.Kubernetes:
+      return [
+        `${process.cwd()}/${KUBERNETES_POLICY_ENGINE_WASM_PATH}`,
```
aron, line 27: I think it would be worth even at this stage pulling this out into a variable at the top level, e.g. `cacheDir` and passing it down. This would make it trivial in future to turn it into a command line flag e.g. `--cache-dir` where the default for `iac test` is `path.join(process.cwd(), ".iac-data")`

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -62,25 +64,21 @@ async function parseFileContentsForPolicyEngine(
...
+    parsedFileData.push(...parsedFiles);
   }
 
   return parsedFileData;
 }
+
+async function scanFilesForIssues(
+  parsedFiles: Array<IacFileData>,
+): Promise<IacFileScanResult[]> {
+  return Promise.all(
```
aron, line 73: Want to add a TODO here to do this in some kind of queue? Like we do in the current implementation so this doesn't choke when given 1000 files?

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -62,25 +60,21 @@ async function parseFileContentsForPolicyEngine(
...
```
aron, line 73: Likewise should we add a TODO to gracefully handle errors? Currently this will fail the test if any of the provided files has an error.

src/cli/commands/test/iac-local-execution/parsers.ts
```diff
@@ -0,0 +1,66 @@
...
+    ) {
+      return {
+        ...fileMetadata,
+        fileContent: fileContent,
+        jsonContent: parsedYamlDocument,
+        engineType: EngineType.Kubernetes,
+        docId,
+      };
+    } else {
+      throw new Error('Invalid K8s File!');
```
aron, line 49: Would be really nice to include the missing fields here.

src/cli/commands/test/iac-local-execution/parsers.ts
```diff
@@ -0,0 +1,66 @@
...
+): Array<IacFileData> {
+  switch (fileMetadata.fileType) {
+    case 'yaml':
+    case 'yml':
+    case 'json':
+      try {
+        return tryParsingKubernetesFile(fileContent, fileMetadata);
+      } catch (err) {
+        // TODO: if it's also not TF file then we should error
+        return [tryParsingTerraformFile(fileContent, fileMetadata)];
```
aron, line 19: I think your comment covers this but this will mask the above error and look like a failed Terraform file right? Perhaps we should do this a) only if `fileType === "json"` and b) rethrow the original error if `tryParsingTerraformFile` also errors.

src/cli/commands/test/iac-local-execution/policy-engine.ts
```diff
@@ -3,18 +3,34 @@ import {
...
 
-const LOCAL_POLICY_ENGINE_DIR = `.iac-data`;
-const LOCAL_POLICY_ENGINE_WASM_PATH = `${LOCAL_POLICY_ENGINE_DIR}${path.sep}policy.wasm`;
-const LOCAL_POLICY_ENGINE_DATA_PATH = `${LOCAL_POLICY_ENGINE_DIR}${path.sep}data.json`;
+export async function getPolicyEngine(
+  engineType: EngineType,
+): Promise<PolicyEngine> {
+  return (
+    policyEngineCache[engineType] ||
+    (policyEngineCache[engineType] = await buildPolicyEngine(engineType))
```
aron, line 17: Suggestion: To make this optional assignment more explicit
```suggestion
    if (policyEngineCache[engineType]) {
      return policyEngineCache[engineType]
    }
    policyEngineCache[engineType] = await buildPolicyEngine(engineType)
    return policyEngineCache[engineType]
```

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -16,6 +17,10 @@ export interface OpaWasmInstance {
   setData: (data: Record<string, any>) => void;
 }
 
+export enum EngineType {
+  Kubernetes = 'kubernetes',
```
aron, line 21: I much prefer `Kubernetes` as the key. But we also have `IacProjectType.K8S` so perhaps we should be consistent in our references.

src/cli/commands/test/iac-local-execution/local-cache.ts
rontalx: I'm intentionally keeping things separated and incrementally adding features as this is a huge project.
We have a specific story to tackle the `cacheDir` support later on towards the beta release.

src/cli/commands/test/iac-local-execution/index.ts
rontalx: will take care of it as part of the dir support task.
let's please keep the PR's based on the relevant story in progress and deliver them in incremental steps.

src/cli/commands/test/iac-local-execution/types.ts
rontalx: `kubernetes ` ,it's a left-over, revering back to `Kubernetes`.

>re: But we also have IacProjectType.K8S so perhaps we should be consistent in our references.

I'm trying to avoid as much as possible using the older types/enums/methods as it's soon to be deprecated and we have an opportunity to do things differently now.

src/cli/commands/test/iac-local-execution/policy-engine.ts
rontalx: I like shorthand syntax and minimum amount of return statements  🤓 
but ok, changing

rontalx:
@aron 
> re: Are you planning to add unit tests for this PR

I'm planning to add the unit-tests together with the dir support story (next one), as there are still lots changes until we reach that core-feature parity state in the experimental version.
I am taking that into account.

dtrunley-snyk (approved):

package.json
```diff
@@ -77,6 +77,7 @@
     "configstore": "^5.0.1",
     "debug": "^4.1.1",
     "diff": "^4.0.1",
+    "hcl-to-json": "^0.1.1",
```
JackuB, line 80: >  Decide how to proceed with hcl-to-json lib and HCL2 syntax validations.

https://github.com/gokmen/hcl-to-json
Seeing it's an archived and unmaintained project, is there an alternative?

rontalx: you are correct, it was an informed decision to start with it as a temporary "unblocker" solution.
as generally there is no good HCL-2-JSON parser npm package out there.

we are aware of it's limitations and it was an informed decision to start with it with the "good-enough" approach for the dog-fooding internal releases.

For the public beta and long-term solution, we will probably do one of the following:
- generate a Golang HCL to JSON parser into WASM and use it.
OR
- include HCL to JSON parsing inside our Terraform scanning wasm file, so you can just provide it with HCL.

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -48,4 +48,48 @@ Describe "Snyk iac test --experimental command"
       The result of function check_valid_json should be success
     End
   End
+
+  Describe "terraform single file scan"
```
JackuB, line 52: Would you run this only in the Regression test and not in the smoke test? I don't think we need hourly checks of this.

With this after the `Describe` like above:

```sh
Skip if "execute only in regression test" check_if_regression_test
```

rontalx: yes will add it as a regression 👍

aron (approved):
👍 Looks good, nice work.

src/cli/commands/test/iac-local-execution/parsers.ts
```diff
@@ -0,0 +1,65 @@
...
+): Array<IacFileData> {
+  switch (fileMetadata.fileType) {
+    case 'yaml':
+    case 'yml':
+    case 'json':
+      return tryParsingKubernetesFile(fileContent, fileMetadata);
+    case 'tf':
+      return [tryParsingTerraformFile(fileContent, fileMetadata)];
+    default:
+      throw new Error('Invalid IaC file');
```
JackuB, line 19: Could have a more descriptive message e.g.

```suggestion
      throw new Error(`Unsupported filetype ${fileMetadata.fileType }`);
```

JackuB (approved):

src/cli/commands/test/iac-local-execution/parsers.ts
rontalx: going to do a bit of an overhaul to the error handling in the next pr for Dir-Support, will take care of it over there.
tnx 👍

PR #1655, opened 2021-02-22
fix: avoid ansi escape codes in non-tty and no color output

jahed-snyk:
- [x] Ready for review

#### What does this PR do?

When the terminal doesn't support colour (either NO_COLOR env variable or not TTY as detected by NodeJS), it will remove every ANSI escape code when printing the `help` output. This avoids printing them in plain text and obfuscating the help text.

Only a few files to check.

Run `snyk help | cat` will print ANSI codes in plain text before this change. After it should print plain unformatted text.

Originally noticed in some Powershell sessions.

None.

#### Screenshots

None.

#### Additional questions

`strip-ansi` is a part of `chalk` so it's not added as a direct dependency, but as best practice, maybe it should be as we use it directly?

### DIFF
--- a/src/cli/commands/help.ts
+++ b/src/cli/commands/help.ts
@@ -1,7 +1,16 @@
 import * as fs from 'fs';
 import * as path from 'path';
+import stripAnsi from 'strip-ansi';
 
 const DEFAULT_HELP = 'snyk';
+
+function readHelpFile(filename: string): string {
+  const file = fs.readFileSync(filename, 'utf8');
+  if (typeof process.env.NO_COLOR !== 'undefined' || !process.stdout.isTTY) {
+    return stripAnsi(file);
+  }
+  return file;
+}
 
 export = async function help(item: string | boolean) {
   if (!item || item === true || typeof item !== 'string' || item === 'help') {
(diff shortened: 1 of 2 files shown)


src/cli/commands/help.ts
```diff
@@ -1,8 +1,17 @@
 import * as fs from 'fs';
 import * as path from 'path';
+import stripAnsi from 'strip-ansi';
 
 const DEFAULT_HELP = 'snyk';
 
+const readHelpFile = (filename: string): string => {
```
maxjeffos, line 7: Don't think we have a policy on this, but this kind of function definition is not particularly helpful unless it is required (i.e. when using as a closure) and does not appear to be inline with the norm.

maxjeffos (changes requested):
The PR description could use some more info, too (like "why"). Is there a reason you didn't use at least part of the PR description template?

JackuB:
Ha, my bad, told @jahed-snyk to not fill it :)

maxjeffos (approved):

src/cli/commands/help.ts
jahed-snyk: I've changed it to use plain old functions.

PR #1669, opened 2021-03-02
fix: don't fail on request's big payload

admons:
- [x] Ready for review

#### What does this PR do?

Today we're serialising the request's payload to debug it. this fails when the payload is too be (see ticket https://snyk.zendesk.com/agent/tickets/6712).
This will skip and log another message in this scenario.

### DIFF
--- a/src/lib/request/request.ts
+++ b/src/lib/request/request.ts
@@ -83,7 +83,11 @@ export = function makeRequest(
           payload.timeout = config.timeout * 1000; // s -> ms
         }
 
-        debug('request payload: ', JSON.stringify(payload));
+        try {
+          debug('request payload: ', JSON.stringify(payload));
+        } catch (e) {
+          debug('request payload is too big to log', e);
+        }
 
         const method = (
           payload.method || 'get'


src/lib/request/request.ts
```diff
@@ -83,7 +83,11 @@ export = function makeRequest(
           payload.timeout = config.timeout * 1000; // s -> ms
         }
 
-        debug('request payload: ', JSON.stringify(payload));
+        try {
+          debug('request payload: ', JSON.stringify(payload));
+        } catch (e) {
+          debug('request payload is too big to log');
```
robcresswell, line 89: Log the error here :)

admons: The only error can be with serialising
1. The payload variable exists (or it would fail way before)
2. If the debug has problem this line will also fail

robcresswell: > The only error can be with serialising

Sure, but it could be a bunch of different reasons. Maybe object size, maybe malformed payload.

> If the debug has problem this line will also fail

Why? I don't think the error object would contain the payload. Just a message and trace.

admons: Ok, I changed it

robcresswell (approved):

dtrunley-snyk (approved):

PR #1676, opened 2021-03-04
Use npm release as source of truth for version

JackuB:
Switches the source of truth for releases to npm package

### DIFF
--- a/release-scripts/update-versions.sh
+++ b/release-scripts/update-versions.sh
@@ -1,8 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-LATEST_PACKAGE_VERSION=$(gh api "repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/releases/latest" | jq .tag_name -r)
-LATEST_PACKAGE_VERSION=${LATEST_PACKAGE_VERSION:1}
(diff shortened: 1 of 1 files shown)


release-scripts/update-versions.sh
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-LATEST_PACKAGE_VERSION=$(gh api "repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/releases/latest" | jq .tag_name -r)
-LATEST_PACKAGE_VERSION=${LATEST_PACKAGE_VERSION:1}
+LATEST_PACKAGE_VERSION=$(npm view snyk version)
```
Avishagp, line 4: Wouldn't this be problematic as npm takes longer to publish than GH?

JackuB: This runs before we trigger an npm release:

https://github.com/snyk/snyk/blob/5f598386084c47701a67a13b0970c72f0047c602/.circleci/config.yml#L413-L426

It could cause issues when running 2 releases back-to-back. But re-running it again should resolve it, which is not possible with the current version afaik.

Avishagp (approved):

PR #1678, opened 2021-03-04
chore: Convert to TS, cleanup code

robcresswell:
Continuing to migrate code. The aim is zero functional changes, but code that is easier to reason about.


src/cli/commands/config.ts
```diff
@@ -0,0 +1,66 @@
+import * as snyk from '../../lib';
+
+const validMethods = ['set', 'get', 'unset', 'clear'];
+
+export = async function config(
```
JackuB, line 5: Didn't you just say this:

> General small comment about import / export. Please prefer named exports i.e. export { foo } or export function foo, instead of default (export default foo) or namespace exports (export = foo).

Don't we want to follow it a different export to avoid `import x = require('x')`?

```suggestion
export async function config(
```

robcresswell: I did! The problem here is that config is hot-required, so it needs to use older mechanisms, as far as I can tell. I think this may actually need to end up as a default export.

src/cli/commands/config.ts
```diff
@@ -0,0 +1,66 @@
...
+    // ensure we update the live library
+    (snyk as any).api = null;
+    return 'config cleared';
+  }
+
+  return Object.keys(snyk.config.all)
+    .sort((a, b) => Number(a.toLowerCase() < b.toLowerCase()))
+    .map((configKey) => `${configKey}: ${snyk.config.all[configKey]}`)
+    .join('\n')
+    .trim();
```
jahed-snyk, line 65: `trim` shouldn't be needed here with a `join`.

jahed-snyk (approved):
Left a bunch of comments, mostly existing behaviour, nothing blocking.

src/cli/commands/config.ts
robcresswell: Why? I'm not sure I follow.

```ts
[' foo', 'bar '].join('');
// ' foobar '

['foo', 'bar'].join('');
// 'foobar'
```

src/cli/commands/config.ts
```diff
@@ -0,0 +1,66 @@
...
+      snyk.config.set(key, val);
+      res += key + ' updated\n';
+
+      // ensure we update the live library
+      if (key === 'api') {
+        (snyk as any).api = val;
+      }
+    });
+
+    return res;
```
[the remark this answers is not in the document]
robcresswell: I could, but I tend to avoid using `reduce()`. In most cases its harder to read / understand than the equivalent loop.

src/cli/commands/config.ts
jahed-snyk: Depends on the expected inputs and outputs, sadly the tests aren't very thorough. I assume each arg doesn't have whitespace so they don't need trimming. Otherwise we should be trimming them individually before the `sort` too so that they are sorted and rendered consistently on every line and not just trimmed at the start and end.

src/cli/commands/config.ts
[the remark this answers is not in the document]
jahed-snyk: Readability is a contentious topic 😬  `forEach` vs `for...of`, `await/try-catch` vs promise chaining, etc.

In this case, I try to avoid `let` where possible. `reduce` to me is more readable as it reduces the scope of modifications to a single iterated function. A lot like how `map` is often preferred instead of creating an empty array and inserting values into it using a for loop.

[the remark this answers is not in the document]
robcresswell: (Does anyone really try and argue that promise chains are more readable than async await?)

I disagree with the comparison to `map`; I think `reduce` is rather uniquely poor in comparison to its simple loop counterparts.

My argument here is that JS's lack of immutability makes arguments along the lines of reducing side effects relatively obsolete, while the comprehension of the loop is almost always higher (better?).

> `reduce` to me is more readable as it reduces the scope of modifications to a single iterated function

I can't really argue with an opinion. I would contend that _to most developers_ a `reduce` will at best be _as readable as_ but not surpass simple loops, especially for newer devs. If we put a series of loops and a series of reduces in front of inexperienced developers, I would put money on the for loops being explained to me correctly, and faster, which I think has immense value.

src/cli/commands/config.ts
robcresswell: Sure; I think what I've got here is functionally identical to what we had before, which is the aim.

PR #1690, opened 2021-03-06
[IaC] feat: fetch policy bundle from S3 for local scans [Beta]- CC-627

ipapast:
#### What does this PR do?
_This is the basic e2e flow that we need for the Beta release. It will be incrementally enhanced._
Everytime a user runs `snyk iac --experimental`, the wasm policies will be available locally.
What happens in the background: 
- A tarball bundle.tar.gz is downloaded from S3, is extracted and placed in the `.iac-data` directory of your cwd. 

This way, the user will have the policies locally in a wasm file, giving them the ability to scan their files locally against them and not send any files over the network.

**Regarding the smoke tests and the env variable:** 
these are run against a subset of fixtures, but now that we introduce this new flow, the policies would get updated every time. Because of that, we provided an env variable temporarily named `SNYK_IAC_SKIP_BUNDLE_DOWNLOAD` to shellspec to skip downloading the bundle.

- Run `snyk iac test --experimental` as usual to check the full flow.
- Run `SNYK_IAC_SKIP_BUNDLE_DOWNLOAD=true snyk-dev iac test filename --experimental` to skip the download of the bundle from S3, for that you will need to have the `.iac-data` folder with the relevant files locally, or you will get an error.

There is [this decision document](https://www.notion.so/snyk/Policy-WASM-download-mechanism-8b25b39dceab4520b56f9783e6cd7ecf) in Notion that can be shared internally.

[CC-627](https://snyksec.atlassian.net/browse/CC-627)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,7 +207,7 @@ jobs:
             echo "Checkout the README in test/smoke folder for more details about this step"
             unset SNYK_API
             unset SNYK_API_KEY
-            shellspec -f d -e REGRESSION_TEST=1
+            shellspec -f d -e REGRESSION_TEST=1 -e SNYK_IAC_SKIP_BUNDLE_DOWNLOAD=true
 
   test-windows:
     <<: *defaults
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,12 +7,8 @@ module.exports = {
   coverageReporters: ['text-summary', 'html'],
   testMatch: [
     '<rootDir>/test/*.spec.ts',
-    '<rootDir>/test/iac-unit-tests/*.spec.ts',
-    '<rootDir>\\test\\*.spec.ts', // for Windows,
-    '<rootDir>\\test\\iac-unit-tests\\*.spec.ts', // for Windows
     '<rootDir>\\test\\*.spec.ts', // for Windows
     '<rootDir>/test/iac-unit-tests/*.spec.ts',
-    '<rootDir>\\test\\*.spec.ts', // for Windows,
     '<rootDir>\\test\\iac-unit-tests\\*.spec.ts', // for Windows
     '<rootDir>/packages/**/test/*.spec.ts',
     '<rootDir>/packages/**/test/**/*.spec.ts',
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,6 +7,9 @@ module.exports = {
   coverageReporters: ['text-summary', 'html'],
   testMatch: [
     '<rootDir>/test/*.spec.ts',
+    '<rootDir>/test/iac-unit-tests/*.spec.ts',
+    '<rootDir>\\test\\*.spec.ts', // for Windows,
+    '<rootDir>\\test\\iac-unit-tests\\*.spec.ts', // for Windows
     '<rootDir>\\test\\*.spec.ts', // for Windows
     '<rootDir>/test/iac-unit-tests/*.spec.ts',
     '<rootDir>\\test\\*.spec.ts', // for Windows,
--- a/src/cli/commands/test/iac-local-execution/file-utils.ts
+++ b/src/cli/commands/test/iac-local-execution/file-utils.ts
@@ -3,6 +3,7 @@ import * as tar from 'tar';
 import * as path from 'path';
 
 export function createIacDir(): void {
+  // this path will be able to be customised by the user in the future
   const iacPath: fs.PathLike = path.join('.iac-data/');
   try {
     if (!fs.existsSync(iacPath)) {
--- a/src/cli/commands/test/iac-local-execution/file-utils.ts
+++ b/src/cli/commands/test/iac-local-execution/file-utils.ts
@@ -1,0 +1,36 @@
+import * as fs from 'fs';
+import * as tar from 'tar';
+import * as path from 'path';
+
+export function createIacDir(): void {
+  const iacPath: fs.PathLike = path.join('.iac-data/');
+  try {
+    if (!fs.existsSync(iacPath)) {
+      fs.mkdirSync(iacPath, '700');
+    }
+    fs.accessSync(iacPath, fs.constants.W_OK);
+  } catch {
+    throw Error(
+      'The .iac-data directory can not be created. ' +
+        'Please make sure that the current working directory has write permissions',
+    );
+  }
+}
+
+export function extractBundle(
+  response,
+  bundlePath: fs.PathLike,
+): Promise<void> {
+  return new Promise((resolve, reject) => {
+    fs.createWriteStream(bundlePath).on('error', (err) => reject(err));
+
+    response
+      .pipe(
+        tar.x({
+          C: path.join('.iac-data'),
+        }),
+      )
+      .on('finish', resolve)
+      .on('error', reject);
+  });
+}
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -1,23 +1,17 @@
-import { isLocalCacheExists, REQUIRED_LOCAL_CACHE_FILES } from './local-cache';
 import { loadFiles } from './file-loader';
 import { parseFiles } from './file-parser';
 import { scanFiles } from './file-scanner';
 import { formatScanResults } from './results-formatter';
 import { isLocalFolder } from '../../../../lib/detect';
 import { IacOptionFlags } from './types';
+import { initLocalCache } from './local-cache';
 
 // this method executes the local processing engine and then formats the results to adapt with the CLI output.
 // the current version is dependent on files to be present locally which are not part of the source code.
 // without these files this method would fail.
 // if you're interested in trying out the experimental local execution model for IaC scanning, please reach-out.
 export async function test(pathToScan: string, options: IacOptionFlags) {
-  if (!isLocalCacheExists())
-    throw Error(
-      `Missing IaC local cache data, please validate you have: \n${REQUIRED_LOCAL_CACHE_FILES.join(
-        '\n',
-      )}`,
-    );
-
+  await initLocalCache();
   const filesToParse = await loadFiles(pathToScan);
   const { parsedFiles, failedFiles } = await parseFiles(filesToParse);
   const scannedFiles = await scanFiles(parsedFiles);
--- a/src/cli/commands/test/iac-local-execution/local-cache.ts
+++ b/src/cli/commands/test/iac-local-execution/local-cache.ts
@@ -30,8 +30,7 @@ export const REQUIRED_LOCAL_CACHE_FILES = [
   TERRAFORM_POLICY_ENGINE_WASM_PATH,
   TERRAFORM_POLICY_ENGINE_DATA_PATH,
 ];
-
-export function isLocalCacheExists(): boolean {
+function doesLocalCacheExist(): boolean {
   return REQUIRED_LOCAL_CACHE_FILES.every(fs.existsSync);
 }
 
@@ -58,10 +57,10 @@ export async function initLocalCache(): Promise<void> {
     const fileName: fs.PathLike = path.join('.iac-data/bundle.tar.gz');
 
     createIacDir();
-    const response: ReadableStream = await needle.get(preSignedUrl);
+    const response: ReadableStream = needle.get(preSignedUrl);
     await extractBundle(response, fileName);
   }
-  if (!isLocalCacheExists()) {
+  if (!doesLocalCacheExist()) {
     throw Error(
       `Missing IaC local cache data, please validate you have: \n${REQUIRED_LOCAL_CACHE_FILES.join(
         '\n',
--- a/src/cli/commands/test/iac-local-execution/local-cache.ts
+++ b/src/cli/commands/test/iac-local-execution/local-cache.ts
@@ -1,8 +1,11 @@
 import * as path from 'path';
 import * as fs from 'fs';
 import { EngineType } from './types';
+import * as needle from 'needle';
+import { createIacDir, extractBundle } from './file-utils';
+import ReadableStream = NodeJS.ReadableStream;
 
-export const LOCAL_POLICY_ENGINE_DIR = `.iac-data`;
+export const LOCAL_POLICY_ENGINE_DIR = '.iac-data';
 
 const KUBERNETES_POLICY_ENGINE_WASM_PATH = path.join(
   LOCAL_POLICY_ENGINE_DIR,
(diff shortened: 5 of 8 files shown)


jest.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
   coverageReporters: ['text-summary', 'html'],
   testMatch: [
     '<rootDir>/test/*.spec.ts',
+    '<rootDir>/test/iac-unit-tests/*.spec.ts',
+    '<rootDir>\\test\\*.spec.ts', // for Windows,
+    '<rootDir>\\test\\iac-unit-tests\\*.spec.ts', // for Windows
```
rontalx, line 12: should be able to rebase from master and have these now.

.github/CODEOWNERS
```diff
@@ -14,6 +14,7 @@ test/acceptance/cli-test/iac/ @snyk/cloudconfig
 test/fixtures/iac/ @snyk/cloudconfig
 test/smoke/spec/iac/ @snyk/cloudconfig
 test/smoke/.iac-data/ @snyk/cloudconfig
+test/iac-unit-tests/ @snyk/cloudconfig
```
rontalx, line 17: should be able to rebase from master and have these now.

src/lib/types.ts
```diff
@@ -78,6 +78,7 @@ export interface Options {
   debug?: boolean;
   sarif?: boolean;
   'group-issues'?: boolean;
+  cacheDir?: boolean;
```
rontalx, line 81: nit:
got a comment from Ben to rename this to: `cache-directory`

src/cli/commands/test/iac-local-execution/fetch-wasm-bundle.ts
```diff
@@ -0,0 +1,54 @@
+import * as fs from 'fs';
```
rontalx, line 1: I thought of bundling all of this code inside the `local-cache.ts` module, so it handles all logic related to the local cache and downloading.
Maybe we can rename it to something like `policy-provider.ts` or so.
WDYT?

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -1,21 +1,31 @@
...
+    await fetchBundle(preSignedUrl);
+  }
+  if (!isLocalCacheExists()) {
     throw Error(
       `Missing IaC local cache data, please validate you have: \n${REQUIRED_LOCAL_CACHE_FILES.join(
         '\n',
       )}`,
     );
+  }
+
```
rontalx, line 28: Following:
https://github.com/snyk/snyk/pull/1690/files#r589434889
Can we move these lines into the unified module?
so all cache/download logic is encapsulated there.
Then also the test command will keep a cool neat convention, for example something like that:
```
  import { initPolicyProvider } from './local-cache.ts'
  ....
  await initPolicyProvider(options);     //
  const filesToParse = await loadFiles(pathToScan);
  const { parsedFiles, failedFiles } = await parseFiles(filesToParse);
  const formattedResults = formatScanResults(scannedFiles, options);
```

src/cli/commands/test/iac-local-execution/fetch-wasm-bundle.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Readable } from 'stream';
+
+export async function fetchBundle(preSignedUrl: string) {
+  createIacDir();
+  const bundlePath: fs.PathLike = path.join('.iac-data/bundle.tar.gz');
+
+  https.get(preSignedUrl, (response) => {
+    streamToFile(response, bundlePath);
+  });
+  return extractBundle(bundlePath);
```
rontalx, line 14: Although this approach with is cool, I'd prefer to avoid this stream & callback implementation.
it's hard to comprehend from first look, and I don't think it adds value due to the synchronous nature of this flow.
I'd suggest checking if you can promisify the `https.get` (I know we use `needle` in the CLI so maybe that can be utilised).
so the flow eventually is as simple as:
```
// await on downloading the file
// synchronously unzipping it with tar
// synchronously write it to the directory.
```
it's easier to read and the readers doesn't need to wrap their heads around asynchronous streams/pipes and callbacks, just to realise it's a simple synchronous flow. 
on other scenarios it can be great, I think its less convenient here.
WDYT?

src/cli/commands/test/iac-local-execution/fetch-wasm-bundle.ts
```diff
@@ -0,0 +1,54 @@
...
+  const iacPath: fs.PathLike = path.join('.iac-data/');
+  try {
+    if (!fs.existsSync(iacPath)) {
+      fs.mkdirSync(iacPath, '700');
+    }
+    fs.accessSync(iacPath, fs.constants.W_OK);
+  } catch {
+    throw Error(
+      'The .iac-data directory can not be created or written. ' +
+        'Please check that the current directory has write permissions',
```
rontalx, line 27: Looking good, I'd double check with Ben for the specific wording.

test/iac-unit-tests/fetch-wasm-bundle.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { streamToFile } from '../../src/cli/commands/test/iac-local-execution/fetch-wasm-bundle';
```
rontalx, line 1: you mentioned that you're adding tests, 
double checking you're planning to test the public/exposable method of the module `fetchBundle`?

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -1,21 +1,31 @@
...
+import { fetchBundle } from './fetch-wasm-bundle';
 
 // this method executes the local processing engine and then formats the results to adapt with the CLI output.
 // the current version is dependent on files to be present locally which are not part of the source code.
 // without these files this method would fail.
 // if you're interested in trying out the experimental local execution model for IaC scanning, please reach-out.
+
 export async function test(pathToScan: string, options) {
-  if (!isLocalCacheExists())
+  if (!options.cacheDir) {
```
rontalx, line 15: @benlaplanche @ipapast 
Following our sync today, double checking we're aligned...
For the beta-release:
- the `--cache-directory` can be provided, which will avoid the forced downloading.
- in this release the `--cache-directory` serves merely as a boolean flag to allow avoid the forced download every time. emphasising it is still not supportive of providing a custom directory, this will be added for the GA. 
- open question: how would we expect the `--cache-directory` flag to work if the bundle wasn't downloaded yet, download it? or error out and expect a first iteration without it to run in order to do the initial bundle fetch.
I'd say, `--cache-directory` should download a bundle if there's non present (will happen for the first run with that flag.).

Although an edge-case, I think it can be impactful in terms of experience.
WDYT?

rontalx:
great work! left some comments.

src/cli/commands/test/iac-local-execution/fetch-wasm-bundle.ts
ipapast: Thanks, I changed this part to be synchronous as you suggested, I hope it reads better now

src/cli/commands/test/iac-local-execution/file-utils.ts
```diff
@@ -0,0 +1,47 @@
+import * as fs from 'fs';
+import { WriteStream } from 'fs';
+import * as tar from 'tar';
+import * as path from 'path';
+import ReadableStream = NodeJS.ReadableStream;
+
+export function createIacDir(): void {
+  const iacPath: fs.PathLike = path.join('.iac-data/');
```
rontalx, line 8: mind adding small `TODO` notes on the places it uses `'.iac-data/'` that in the future it would be custom?

src/cli/commands/test/iac-local-execution/index.ts
ipapast: reverting this flag as discussed

aron: Not for now, but my understanding of this flag, in the GA state based on experience with other tools is as follows.

1) Without the flag Snyk will default to `$(pwd)/.snyk-iac`.
  1) If the .snyk-iac directory exists the CLI will check if the contained policies are present, valid and up-to-date. And if stale either prompt to download or download automatically in the background.
2) With the `--cache-directory` flag the CLI will replace the `$(pwd)/.snyk-iac` path with that provided by the user. The same steps as above apply.

I don't think the flag should have any impact on the download behaviour, it should simply be used to configure a path.

aron: > instead of just running with the flag, and have it download the files to the provided directory if it's empty.

This is what I intended to say. I've updated the comment to hopefully clarify.

rontalx: so we are talking on the same thing 🤓 
Anyway for now let's extract this flag from this PR for simplicity.
We can add it in the coming sprint.

src/cli/commands/test/iac-local-execution/file-utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    inputStream
+      .pipe(fileWriteStream)
+      .on('finish', resolve)
+      .on('error', reject);
+  });
+}
+
+export function extractBundle(bundlePath: fs.PathLike): Promise<void> {
+  return new Promise((resolve, reject) => {
+    fs.createReadStream(bundlePath)
```
aron, line 37: Not for now, but why write the bundle to disk when we can pipe the HTTP stream directly into `tar.x()`?

.circleci/config.yml
```diff
@@ -207,7 +207,7 @@ jobs:
             echo "Checkout the README in test/smoke folder for more details about this step"
             unset SNYK_API
             unset SNYK_API_KEY
-            shellspec -f d -e REGRESSION_TEST=1
+            shellspec -f d -e REGRESSION_TEST=1 -e SNYK_IAC_SKIP_BUNDLE_DOWNLOAD=true
```
ipapast, line 210: This is to provide an env var for IaC smoke tests -we use local fixtures in the tests but now the implementation downloads new policies everytime. We wanted to skip the download for now by adding this var.

src/cli/commands/test/iac-local-execution/file-utils.ts
ipapast: changed this, writing is removed now

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -1,21 +1,17 @@
...
 // if you're interested in trying out the experimental local execution model for IaC scanning, please reach-out.
+
 export async function test(pathToScan: string, options) {
-  if (!isLocalCacheExists())
-    throw Error(
-      `Missing IaC local cache data, please validate you have: \n${REQUIRED_LOCAL_CACHE_FILES.join(
-        '\n',
-      )}`,
-    );
+  await initLocalCache();
```
rontalx, line 14: yasss ✨

src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -46,3 +49,23 @@ export function getLocalCachePath(engineType: EngineType) {
       ];
   }
 }
+
+export async function initLocalCache(): Promise<void> {
+  // temporarily use an ENV var to skip downloading of the bundles if we need to - e.g. smoke tests
+  if (!process.env.SNYK_IAC_SKIP_BUNDLE_DOWNLOAD) {
+    const preSignedUrl =
+      'https://cloud-config-policy-bundles.s3-eu-west-1.amazonaws.com/bundle.tar.gz';
```
rontalx, line 57: nit:
maybe worth adding a small TODO/comment that it will be replaced with snyk . io domain soon

rontalx (approved):
great stuff 👏

jest.config.js
ipapast: Those were there already. Line 13 has an extra comma in the end of the comment and didn't cause conflicts. I can remove the line.

src/cli/commands/test/iac-local-execution/local-cache.ts
ipapast: We haven't taken a decision officially with what it would be replaced exactly so I think I'll skip this

src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -46,3 +49,23 @@ export function getLocalCachePath(engineType: EngineType) {
       ];
   }
 }
+
+export async function initLocalCache(): Promise<void> {
+  // temporarily use an ENV var to skip downloading of the bundles if we need to - e.g. smoke tests
+  if (!process.env.SNYK_IAC_SKIP_BUNDLE_DOWNLOAD) {
```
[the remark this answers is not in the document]
ipapast: Just personal preference I guess 😅 I thought it would be much easier to clean afterwards

src/cli/commands/test/iac-local-execution/file-utils.ts
ipapast: done

jest.config.js
```diff
@@ -7,10 +7,12 @@ module.exports = {
   coverageReporters: ['text-summary', 'html'],
   testMatch: [
     '<rootDir>/test/*.spec.ts',
-    '<rootDir>\\test\\*.spec.ts', // for Windows
     '<rootDir>/test/iac-unit-tests/*.spec.ts',
     '<rootDir>\\test\\*.spec.ts', // for Windows,
     '<rootDir>\\test\\iac-unit-tests\\*.spec.ts', // for Windows
+    '<rootDir>\\test\\*.spec.ts', // for Windows
+    '<rootDir>/test/iac-unit-tests/*.spec.ts',
+    '<rootDir>\\test\\iac-unit-tests\\*.spec.ts', // for Windows
```
jahed-snyk, line 15: It looks like Line 10 has been moving to Line 13 and Line 14 and 15 are duplicates of Line 11 and 13. Rolling back all changes to `jest.config.js` in this PR should fix this.

.circleci/config.yml
jahed-snyk: I'm not sure about this change. The goal of smoke tests is to capture production issues so skipping a production step kind of defeats its purpose. Mentioning @JackuB as I don't have full context of smoke tests.

jahed-snyk (changes requested):

.circleci/config.yml
ipapast: Hi @jahed-snyk , @JackuB, I just want to clarify that we are not skipping the tests but the download of an artifact. The download step now overwrites the local fixtures we are using in the smoke tests on every download, so the expected results change. 
We introduced this var temporarily, we will think of how to do this better and according to the next steps of this flow, but for now we want the tests to run against the fixed set of fixtures.

Does this make more sense?

jahed-snyk: Thanks for clarifying. That makes sense.

Bear with me for a bit as I get some feedback on the rest of the changes from the team (I don't have enough context).

src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -46,3 +48,23 @@ export function getLocalCachePath(engineType: EngineType) {
...
+    const preSignedUrl =
+      'https://cloud-config-policy-bundles.s3-eu-west-1.amazonaws.com/bundle.tar.gz';
```
rontalx, line 56: following our conversation and drift of the bundle,
let's make sure to use here an explicit bundle name with a tagged version so we keep track of what is being used., and we know is working.
as we're not covering the download scenario of specific bundles in tests yet.
❗️🚨

rontalx: ok after what we discussed now, ignore this :)

PR #1698, opened 2021-03-08
feat: convert args into paths & options

lili2311:
- [x] Ready for review

#### What does this PR do?
Extract command args processing that generate the test paths & options from input which can be a variety of shapes coming in. All in preparation for a new command that will need to re-use most of snyk test flow.

### DIFF
--- a/src/cli/commands/process-command-args.ts
+++ b/src/cli/commands/process-command-args.ts
@@ -1,0 +1,17 @@
+import { Options } from '../../lib/types';
+
+export function processCommandArgs<CommandOptions>(
+  ...args
+): { paths: string[]; options: Options & CommandOptions } {
+  let options = ({} as any) as Options & CommandOptions;
+
+  if (typeof args[args.length - 1] === 'object') {
+    options = (args.pop() as any) as Options & CommandOptions;
+  }
+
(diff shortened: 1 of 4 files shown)


src/cli/commands/process-command-args.ts
```diff
@@ -0,0 +1,19 @@
+import { Options } from '../../lib/types';
```
JackuB, line 1: This module should have an unit test. Ideally a jest test

@maxjeffos @jahed-snyk we had plans to setup unit testing folder, right?

JackuB (changes requested):

src/cli/commands/process-command-args.ts
lili2311: we seem to only have `tap acceptance` + `tap system` + `tap other` and `jest all in top level directory`

This would be a new test so it would be jest and has somewhere to go, however I agree it would be great if we can start managing the test structure and move away from tap. We want to be able to put the tests in the same directory structure as the code being tested.

lili2311: Added a test

JackuB (approved):

PR #1704, opened 2021-03-10
refactor: set default test options

lili2311:
- [x] Ready for review

#### What does this PR do?
refactor settings default test options so it can be re-used in another command that relies on test running first

### DIFF
--- a/src/cli/commands/test/set-default-test-options.ts
+++ b/src/cli/commands/test/set-default-test-options.ts
@@ -1,0 +1,24 @@
+import * as config from '../../../lib/config';
+import { Options, ShowVulnPaths, TestOptions } from '../../../lib/types';
+
+export function setDefaultTestOptions(
+  options: Options & TestOptions,
+): Options & TestOptions {
+  const svpSupplied = (options['show-vulnerable-paths'] || '').toLowerCase();
+
+  return {
+    ...options,
+    // org fallback to config unless specified
+    org: options.org || config.org,
+    // making `show-vulnerable-paths` 'some' by default.
+    showVulnPaths: showVulnPathsMapping[svpSupplied] || 'some',
+  };
+}
+
+const showVulnPathsMapping: Record<string, ShowVulnPaths> = {
+  false: 'none',
+  none: 'none',
+  true: 'some',
+  some: 'some',
+  all: 'all',
+};
--- a/src/cli/commands/test/set-default-test-options.ts
+++ b/src/cli/commands/test/set-default-test-options.ts
@@ -1,11 +1,12 @@
 import * as config from '../../../lib/config';
 import { Options, ShowVulnPaths, TestOptions } from '../../../lib/types';
 
-export function setDefaultTestOptions(
-  options: Options & TestOptions,
-): Options & TestOptions {
-  const svpSupplied = (options['show-vulnerable-paths'] || '').toLowerCase();
+export function setDefaultTestOptions(options: Options): Options & TestOptions {
+  const svpSupplied = (options['show-vulnerable-paths'] || '')
+    .toString()
+    .toLowerCase();
 
+  delete options['show-vulnerable-paths'];
   return {
     ...options,
     // org fallback to config unless specified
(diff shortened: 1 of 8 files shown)


aviadhahami (approved):
lgtm :)

src/cli/commands/test/set-default-test-options.ts
```diff
@@ -0,0 +1,24 @@
+import * as config from '../../../lib/config';
+import { Options, ShowVulnPaths, TestOptions } from '../../../lib/types';
+
+export function setDefaultTestOptions(
```
JackuB, line 4: I believe we should keep adding jest-unit test for these new modules

JackuB (changes requested):

src/cli/commands/test/set-default-test-options.ts
lili2311: how are these structured? Right now I am dumping all new tests into top level test directory because of the jest pattern restrictions 😢 can we update this to allow us to add tests in correct folder structures instead?
- code: `src/lib/index.ts`
- test: `test/lib/index.spec.ts`? Maybe for now `test/jest/unit/lib/index.spec.ts` until we fully migrate to Jest?

It will be hard to separate jest tests later from system/acceptance/unit etc

src/cli/commands/test/set-default-test-options.ts
```diff
@@ -0,0 +1,25 @@
+import * as config from '../../../lib/config';
+import { Options, ShowVulnPaths, TestOptions } from '../../../lib/types';
+
+export function setDefaultTestOptions(options: Options): Options & TestOptions {
+  const svpSupplied = (options['show-vulnerable-paths'] || '')
+    .toString()
```
lili2311, line 6: this essentially adds support for `--show-vulnerable-paths` to be treated as `--show-vulnerable-paths=true` WDYT?

Otherwise if you run test with `snyk test --show-vulnerable-paths` you get a bad error that says
`toLowerCase is not a function` because it is not a string. Or alternative is to add validation for this options and throw outside of this function somewhere probably

src/lib/types.ts
```diff
@@ -13,8 +13,7 @@ export type DepTree = legacyApi.DepTree;
 export type ShowVulnPaths = 'none' | 'some' | 'all';
 
 export interface TestOptions {
-  traverseNodeModules: boolean;
-  interactive: boolean;
```
lili2311, line 17: this is a protect option for wizard only

src/cli/commands/test/set-default-test-options.ts
lili2311: Test added, please advise if it can move somewhere better than top level of test directory

src/cli/commands/test/set-default-test-options.ts
JackuB: :+1: this sounds in line with how arg parsing libraries should treat this

test/set-default-test-options.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { setDefaultTestOptions } from '../src/cli/commands/test/set-default-test-options';
+
+describe('setDefaultTestOptions', () => {
+  it('defaults to show-vulnerable-paths:some & org from config when no options passed in', () => {
+    const updated = setDefaultTestOptions({ path: '/' });
+    expect(updated).toEqual({
+      org: undefined,
```
JackuB, line 7: Just curious, in the jest's equals logic, is `org: undefined` same as not having `org` key at all?

src/cli/commands/test/set-default-test-options.ts
lili2311: @gitphill while @aviadhahami is off, could you help 👀 ?

JackuB (approved):

test/set-default-test-options.spec.ts
lili2311: it does indeed!

miiila (approved):

JackuB:
Just a nit, noticed a typo in the commit message, which would propagate it to the release notes:

```diff
features:
- delete vuln paths option once transofrmed + tests
+ delete vuln paths option once transformed
```

PR #1706, opened 2021-03-10
fix: terraform plan parse bug

rontalx:
#### What does this PR do?
Adds meaningful error messages to Terraform plan scanning and fixes cases where the child modules array is missing from the data structure.
The bug was in the reducing function which assumed `child_modules` is always present and didn't have safe defaults.

### DIFF
--- a/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
@@ -30,14 +30,14 @@ function terraformPlanReducer(
 function extractRootModuleResources(
   terraformPlanJson: TerraformPlanJson,
 ): Array<TerraformPlanResource> {
-  return terraformPlanJson?.planned_values?.root_module?.resources;
+  return terraformPlanJson?.planned_values?.root_module?.resources || [];
 }
 
 function extractChildModulesResources(
   terraformPlanJson: TerraformPlanJson,
 ): Array<TerraformPlanResource> {
   const childModules =
-    terraformPlanJson?.planned_values?.root_module?.child_modules;
+    terraformPlanJson?.planned_values?.root_module?.child_modules || [];
   const extractedChildModuleResources = childModules.flatMap(
     (childModule) => childModule.resources,
   );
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -30,14 +30,14 @@ function terraformPlanReducer(
 function extractRootModuleResources(
   terraformPlanJson: TerraformPlanJson,
 ): Array<TerraformPlanResource> {
-  return terraformPlanJson?.planned_values?.root_module?.resources;
+  return terraformPlanJson?.planned_values?.root_module?.resources || [];
```
aron, line 33: Nit:
```suggestion
  return terraformPlanJson?.planned_values?.root_module?.resources ?? [];
```

p0tr3c (approved):

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -47,23 +47,35 @@ function extractChildModulesResources(
...
-  const terraformPlanJson = JSON.parse(
-    terraformPlanFile.fileContent,
-  ) as TerraformPlanJson;
+  let terraformPlanJson;
+  try {
+    terraformPlanJson = JSON.parse(
+      terraformPlanFile.fileContent,
+    ) as TerraformPlanJson;
+  } catch (err) {
+    throw new Error('Failed to parse Terraform plan JSON file.');
```
aron, line 56: Likely not an issue now but `JSON.parse()` throws a SyntaxError so you can catch explicit parse errors and re-throw anything else that happens to be caught, like type/reference errors.

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -47,23 +47,35 @@ function extractChildModulesResources(
...
+      {
+        ...terraformPlanFile,
+        jsonContent: parsedInput,
+        engineType: EngineType.Terraform,
+      },
+    ];
+  } catch (err) {
+    throw new Error(
+      'Failed to extract resources from Terraform plan JSON file.',
+    );
```
aron, line 79: Ditto here, we probably want to throw a specific Error subclass and detect here and then rethrow any other errors. Does the CLI have a logger we can run in debug mode to log the underlying error here to help users debug the issue?

aron (approved):
Some suggestions for future error handling inline, we'll likely want to get these in for beta.

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
rontalx: agreed.
I want to include that work in the `CLI interface overhaul` ticket, to also map all of our potential errors and fine-tuning them with proper classes and messages.
not sure about the debug logger, worth checking.

PR #1707, opened 2021-03-10
Feat: snyk fix v1 (Python) in behind FF

lili2311:
- [x] Ready for review

#### What does this PR do?
Introduce `snyk fix` command behind FF as beta version. Supports only fixing `requirements.txt` projects that do not include a `-r` or `-c` directive.

The command first runs a `snyk test` command and then uses the results to automatically apply remediation, similar to wizard but in an automated fashion.

*Note*
- no help text for this command as it is behind a flag & will be in beta for a while.
- partial conversion to `TestResult` & `ScanResult` to only grab what is needed for now, this will evolve if needed but ideally we migrate away to new EcoSystem flow where this data will be available without a conversion.

`src/cli/commands/fix/index.ts`

`snyk fix --all-projects --org=org-with-FF-enabled`

Relies on https://github.com/snyk/snyk/pull/1716

#### Screenshots

snykFixGif1
snykFixGif2
CleanShot 2021-03-16 at 15 08 44@2x


package.json
```diff
@@ -101,6 +101,7 @@
     "micromatch": "4.0.2",
     "needle": "2.6.0",
     "open": "^7.0.3",
+    "ora": "5.3.0",
```
lili2311, line 104: `snyk fix` & `@snyk/fix` uses Ora, intention is for this to be the future and only spinner.

src/lib/plugins/get-deps-from-plugin.ts
```diff
@@ -85,7 +85,7 @@ export async function getDepsFromPlugin(
       root,
     );
 
-    if (!options.json && userWarningMessage) {
+    if (!options.json && !options.quiet && userWarningMessage) {
```
lili2311, line 88: `snyk fix` calls `snyk test`, we want to control the `snhyk fix` output to only be what the command spinner wants to display. So calling `snyk fix` in --quiet mode, however it looks like while we support the param we did not honour it.

src/lib/snyk-test/assemble-payloads.ts
```diff
@@ -31,7 +31,9 @@ export async function assembleEcosystemPayloads(
       path.relative('..', '.') + ' project dir');
 
   spinner.clear<void>(spinnerLbl)();
-  await spinner(spinnerLbl);
+  if (!options.quiet) {
+    await spinner(spinnerLbl);
+  }
```
lili2311, line 36: `snyk fix` calls `snyk test`, we want to control the `snhyk fix` output to only be what the command spinner wants to display. So calling `snyk fix` in --quiet mode, however it looks like while we support the param we did not honour it.

src/lib/snyk-test/run-test.ts
```diff
@@ -240,7 +240,9 @@ async function sendAndParseResults(
     const iacResults: Promise<TestResult>[] = [];
 
     await spinner.clear<void>(spinnerLbl)();
-    await spinner(spinnerLbl);
+    if (!options.quiet) {
```
lili2311, line 243: same here

src/lib/types.ts
```diff
@@ -79,6 +79,7 @@ export interface Options {
   debug?: boolean;
   sarif?: boolean;
   'group-issues'?: boolean;
+  quiet?: boolean;
```
lili2311, line 82: already supported  (see code & docs) but was not documented here.

package.json
JackuB: I'm trying to figure out why is nothing failing, if the `@snyk/fix` is not defined in package.json 🤔 I guess we need a new CI check to make sure that no `npm link`ed dependencies are used

lili2311: I just added it now via `npm i --save @snyk/fix` is the version as expected?

JackuB (approved):
Extensively discussed this within Hammer and we are happy to approve this 👍

src/cli/commands/fix/validate-fix-command-is-supported.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export async function validateFixCommandIsSupported(
+  options: Options & TestOptions,
+): Promise<boolean> {
+  if (options.docker) {
+    throw new FeatureNotSupportedByEcosystemError('snyk fix', 'docker');
+  }
+
+  const ecosystem = getEcosystemForTest(options);
+  if (ecosystem) {
```
MegaBean, line 21: Am I missing something here? We will always throw FeatureNotSupportedByEcosystemError unless the ecosystem is null in which case we do support it?

test/validate-fix-command-is-supported.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+      .mockResolvedValue({ ok: true });
+    const options = {
+      path: '/',
+      showVulnPaths: 'all' as ShowVulnPaths,
+      docker: true,
+    };
+    expect(validateFixCommandIsSupported(options)).rejects.toThrowError('');
+  });
+
+  it('fix is NOT supported for --code + enabled FF', () => {
```
MegaBean, line 49: Should these be checking for FeatureNotSupportedByEcosystemError which is the specific error thrown?

lili2311: pushing an update

src/cli/commands/fix/validate-fix-command-is-supported.ts
lili2311: Thats correct, `snyk fix` v1 doesn't support any ecosystems only old style test for now. So only if this something with a packageManager via old test flow

MegaBean: Ok. So a `null` ecosystem means that we are using the old test flow? Maybe a comment would help to make this clearer?

lili2311: thats the same as in `snyk test` flow, add the comment to both? https://github.com/snyk/snyk/blob/e0e1b30a10c80e55b18eb985f1e185a6a23503c2/src/cli/commands/test/index.ts#L70

test/cli-fix.system.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+        },
+      );
+    },
+    testTimeout,
+  );
+
+  /* this command is different
+   * it shows help text not an error when command is not supported
+   */
+  it.skip(
```
gitphill, line 149: what's the plan here?

gitphill (approved):

test/cli-fix.system.spec.ts
lili2311: I think let's expect the help text here as the team confirmed thats the expected behaviour

src/cli/modes.ts
```diff
@@ -8,7 +8,7 @@ interface ModeData {
 
 const modes: Record<string, ModeData> = {
   source: {
-    allowedCommands: ['test', 'monitor'],
+    allowedCommands: ['test', 'monitor', 'fix'],
```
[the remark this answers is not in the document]
lili2311: a mistake, removing. good catch! 🦅 👁️

test/cli-fix.functional.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+    ),
+    'utf8',
+  ),
+);
+
+describe('snyk fix (functional tests)', () => {
+  beforeAll(async () => {
+    jest
+      .spyOn(validateFixSupported, 'validateFixCommandIsSupported')
+      .mockResolvedValue(true);
```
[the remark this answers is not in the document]
lili2311: updated, makes sense yep!

test/cli-fix.system.spec.ts
lili2311: ah yes we can't test this as the tests are not setup to run container tests out of the box, I remember now :) Deleting this test until we can
`Received string:    "An unknown error occurred. Please run with `-d` and include full trace when reporting to Snyk·`

lili2311: This works locally otherwise

PR #1709, opened 2021-03-11
refactor: Remove writeable stream to stop writing tarball to the disk [IaC]

ipapast:
#### What does this PR do?
On a previous PR, we introduced the download of a tarball with policies and the extraction of its contents.

The code was unnecessarily writing the bundle.tar.gz to the disk before extracting the files too. This file is not needed, we only need its contents to be written to the disk, so this PR removes it.

Run `snyk iac test file.tf --experimental` and check that there is no bundle.tar.gz into your `.iac-data` folder

Related to CC-627


rontalx (approved):
👍 
notice it will trigger a new release

PR #1713, opened 2021-03-11
feat: iac cli flag validation

aron:
- [x] Ready for review

#### What does this PR do?

Adds a validation step to the `snyk iac` code path to validate that the flags provided to the command line via `process.argv` match the supported subset that the iac command expects. We test the raw `argv` array rather than the `options` object because by the time the `options` object reaches the IaC point the defaults have been filled in for all the other flags and it's hard to differentiate between the defaults and the user input.

I've added an `IaCTestOptions` object to try and ensure the keys we use in the `assertIaCOptionsFlags()` function are valid and changes are picked up in future by the type checker, though this won't help us if new flags are added to the list, but these will fail locally when run.

I would happily take suggestions for a better point to insert this logic. I also think that the CLI in general would benefit from strictly validating the user inputs provided but I'm sure that's already been discussed. We specifically need this functionality now so that we don't have users accidentally sending data to the back-end should they mistype the `--experimental` flag, I would love to discuss how we could do this in a more ecosystem friendly fashion going forward.

EDIT: On feedback from @JackuB I've moved all the iac test logic into a new `iac-test-shim.ts` file that now deals with validating the arguments and determining if we're in the new local or legacy remote flow. So the index.ts file now only contains a single call to this function that is under @snyk/cloudconfig ownership.

This PR also removes the `options.iacDirFiles` mutation on the `options` object and instead uses a local variable to track the parse failures and render the results. One side effect from this is that we will now always use the basename of the file in the CLI text output rather than the path when scanning a single file and the filename when scanning a directory.

Start in src/cli/commands/test/index.ts, then src/cli/commands/test/iac-test-shim.ts and work down to src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts. I've added unit tests for the assertions as well as a smoke test to catch the primary use case. There are some code changes in the iac-local-execution/index.ts file as well to remove the object mutation of the provided options object.

Build the pr and run the following command:

    % node ./dist/cli iac test --experimenta terraform2/aws_api_gateway_stage_logging.tf

Also run it with a combination of valid flags available via `snyk iac test --help`.

Validates the command line flags passed to the `snyk iac test` command. The current argument parsing is very permissive and allows unknown flags to be provided without validation. For snyk iac we need to explicitly validate the flags to avoid misconfigurations and typos. For example, if the `--experimental` flag were to be mis-spelled we would end up sending the client data to our backend rather than running it locally as intended.

#### Relevant tickets
[CC-739]

#### Screenshots
image

[CC-739]: https://snyksec.atlassian.net/browse/CC-739

### DIFF
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -1,11 +1,10 @@
 import { CustomError } from '../../../../lib/errors';
-import { IaCTestOptions } from '../../../../lib/types';
 import { args } from '../../../args';
+import { IaCTestFlags } from './types';
 
-const keys: (keyof IaCTestOptions)[] = [
+const keys: (keyof IaCTestFlags)[] = [
   'debug',
   'insecure',
-  'debug',
   'experimental',
   'detectionDepth',
   'severityThreshold',
@@ -27,7 +26,7 @@ function camelcaseToDash(key: string) {
 }
 
 class FlagError extends CustomError {
-  constructor(key) {
+  constructor(key: string) {
     const dashes = key.length === 1 ? '-' : '--';
     const flag = camelcaseToDash(key);
     const msg = `Unsupported flag "${dashes}${flag}" provided. Run snyk iac test --help for supported flags.`;
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -1,0 +1,61 @@
+import { CustomError } from '../../../../lib/errors';
+import { IaCTestOptions } from '../../../../lib/types';
+import { args } from '../../../args';
+
+const keys: (keyof IaCTestOptions)[] = [
+  'debug',
+  'insecure',
+  'debug',
+  'experimental',
+  'detectionDepth',
+  'severityThreshold',
+  'json',
+  'sarif',
+  'json-file-output',
+  'sarif-file-output',
+  'v',
+  'version',
+  'h',
+  'help',
+  'q',
+  'quiet',
+];
+const allowed = new Set<string>(keys);
+
+function camelcaseToDash(key: string) {
+  return key.replace(/[A-Z]/g, (m) => '-' + m.toLowerCase());
+}
+
+class FlagError extends CustomError {
+  constructor(key) {
+    const dashes = key.length === 1 ? '-' : '--';
+    const flag = camelcaseToDash(key);
+    const msg = `Unsupported flag "${dashes}${flag}" provided. Run snyk iac test --help for supported flags.`;
+    super(msg);
+    this.userMessage = msg;
+  }
+}
+
+/**
+ * Validates the command line flags passed to the snyk iac test
+ * command. The current argument parsing is very permissive and
+ * allows unknown flags to be provided without valdiation.
+ *
+ * For snyk iac we need to explictly validate the flags to avoid
+ * misconfigurations and typos. For example, if the --experimental
+ * flag were to be mis-spelled we would end up sending the client
+ * data to our backend rather than running it locally as intended.
+ * @param argv command line args passed to the process
+ */
+export function assertIaCOptionsFlags(argv: string[]) {
+  // We process the process.argv so we don't get default values.
+  const parsed = args(argv);
+  for (const key of Object.keys(parsed.options)) {
+    // The _ property is a special case that contains non
+    // flag strings passed to the command line (usually files)
+    // and `iac` is the command provided.
+    if (key !== '_' && key !== 'iac' && !allowed.has(key)) {
+      throw new FlagError(key);
+    }
+  }
+}
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -48,13 +48,12 @@ import {
   getDisplayedOutput,
 } from './formatters/format-test-results';
 
-import * as iacLocalExecution from './iac-local-execution';
+import { test as iacTest } from './iac-test';
 import { validateCredentials } from './validate-credentials';
 import { generateSnykTestError } from './generate-snyk-test-error';
 import { validateTestOptions } from './validate-test-options';
 import { setDefaultTestOptions } from './set-default-test-options';
 import { processCommandArgs } from '../process-command-args';
-import { assertIaCOptionsFlags } from './iac-local-execution/assert-iac-options-flag';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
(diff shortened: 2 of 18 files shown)


test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -5,6 +5,14 @@ Describe "Snyk iac test --experimental command"
   Before snyk_login
   After snyk_logout
 
+  Describe "basic usage"
+    It "outputs an error if the --experimental flag is mistyped"
+      When run snyk iac test ../fixtures/iac/kubernetes/pod-invalid.yaml --experimentl
+      The status should be failure
+      The output should include "Unsupported flag"
+    En
```
aron, line 13:
```suggestion
    End
```

src/cli/commands/test/index.ts
```diff
@@ -95,7 +96,14 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
...
     try {
-      if (options.iac && options.experimental) {
+      if (options.iac) {
+        // Ensure that all flags are correct. We do this to ensure that the
+        // caller doesn't accidentally mistype --experimental and send their
+        // configuration files to our backend by accident.
+        assertIaCOptionsFlags(process.argv);
+      }
+
+      if (options.iac && testOpts.experimental) {
```
JackuB, line 106: Shouldn't this be nested under a single `if iac`? Or moved to a specific IAC module?

aron: That makes the `else` clause for this `if` on 106 tricky.

```ts
if (options.iac) {
  assert();
  if (options.experimental) {
    // Do experimental path...
  } // else normal flow....
} else {
  // normal flow
}
```

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
```diff
@@ -0,0 +1,61 @@
+import { CustomError } from '../../../../lib/errors';
+import { IaCTestOptions } from '../../../../lib/types';
+import { args } from '../../../args';
+
+const keys: (keyof IaCTestOptions)[] = [
+  'debug',
+  'insecure',
+  'debug',
```
rontalx, line 8: nit:
duplicate debug key

src/cli/commands/test/index.ts
rontalx: agree with Aron,
though, maybe we can do something like:
```
if (options.iac && assertIaCOptionsFlags(process.argv) && testOpts.experimental)
```
and have `assertIaCOptionsFlags` return a `true` if the assertion passes successfully.
really negligible IMO, but will keep our integration point in this file a 1 liner.

src/lib/types.ts
```diff
@@ -26,6 +26,29 @@ export interface TestOptions {
...
+  'json-file-output'?: string;
+  'sarif-file-output'?: string;
+  v?: boolean;
+  version?: boolean;
+  h?: boolean;
+  help?: 'help';
+  q?: boolean;
+  quiet?: boolean;
+};
+
```
rontalx, line 51: maybe move it to:
`iac-local-execution/types.ts`
?
we also have `IacOptionFlags` over there, worth doing some merge between these two.

test/acceptance/cli-args.test.ts
```diff
@@ -81,9 +81,8 @@ test('snyk test command should fail when iac --file is specified', (t) => {
       }
       t.match(
         stdout.trim(),
-        'Not a recognised option, did you mean "snyk iac test ./test/acceptance/workspaces/iac-kubernetes/multi-file.yaml"? ' +
-          'Check other options by running snyk iac --help',
-        'correct error output',
+        'Unsupported flag',
+        'failed with Unsupported flag error',
```
rontalx, line 85: fyi, all these tap acceptance tests will soon be removed 🎉

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -6,6 +6,14 @@ Describe "Snyk iac test --experimental command"
   Before snyk_login
   After snyk_logout
 
+  Describe "basic usage"
+    It "outputs an error if the --experimental flag is mistyped"
+      When run snyk iac test ../fixtures/iac/kubernetes/pod-invalid.yaml --experimentl
```
rontalx, line 11: let's get shirts for the whole team with a print:
`--experimentl`

rontalx (approved):
few nit comments,
otherwise I'd just pull over the new options type you created to be inside `iac-local-execution/types.ts`, so we don't create an external reviewer dependancy in case that changes.

src/cli/commands/test/index.ts
JackuB: Yea, I'm not worried about the actual code structure in this case, but I thought that your goal was to keep everything detached from this file and keep it together with the rest of IaC stuff. This just introduces IaC code back to the commands/test/index

aron: Hmm, I guess we could duplicate the `snyk.test()` on line 111. @rontalx how do you feel about moving all this into an iacTest() wrapper? where the code path becomes:

```ts
if (options.iac) {
  res = await testIaC(path, testOpts);
} else {
  res = await snyk.test(path, testOpts);
}

async function testIaC(path: string, opts) {
  assertIaCOptionsFlags(process.argv)
  return opts.experimental ? snyk.test(path, testOpts) : iacLocalExecution.test(path, testOpts);
} 
```

I assume there shouldn't be too much drift on the `snyk.test()` interface in the coming months?

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
aron: YSSS

src/lib/types.ts
aron: Tried to merge it but `IacOptionFlags` contains fields that are not flags. So I might rename and make it clearer.

src/cli/commands/test/index.ts
aron: I've made this change and added a new iac-test.ts file in the root test directory, this now holds a shim that will do the flag validation and call either the new local exec `test()` function or the legacy `snyk.test()` path. I was also able to refactor a small part of the awkwardness of the options mutation to add the `iacDirectoryFiles` property out of the new flow.

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
```diff
@@ -0,0 +1,60 @@
+import { CustomError } from '../../../../lib/errors';
+import { args } from '../../../args';
+import { IaCTestFlags } from './types';
+
+const keys: (keyof IaCTestFlags)[] = [
```
JackuB, line 5: FYI this changes the behavior we have in other parts of the CLI, where we use `abbrev` to allow incomplete commands and flags

https://github.com/snyk/snyk/search?q=abbrev

I'd check how is that interacting with this implementation just to be sure it's not getting to a weird state

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
```diff
@@ -0,0 +1,60 @@
+import { CustomError } from '../../../../lib/errors';
+import { args } from '../../../args';
```
JackuB, line 2: On a slightly unrelated note, would you be interested in getting rid of this `args` parsing and switching to something like `minimist` for option parsing? Afaik now you don't rely on anything from the `args`, right?

JackuB (approved):
Nothing blocking

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
aron: YES, YES PLEASE :)

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
aron: Thanks for the heads up, will take a pass over the instances.

PR #1715, opened 2021-03-12
fix: filter out file content from iac test --experimental analytics

aron:
- [x] Ready for review

#### What does this PR do?

We have a minor bug in our Snyk CLI that when the `snyk iac test --experimental` command is run with a directory instead of a single file we include the `iacDirFiles` on the `options` object. This options object is then logged as part of the analytics flow and sent to Big Query. We do not want to be storing user file content in any part of our system so this needs to be filtered out here. 

This PR filters the File content and JSON output from the `iacDirFiles` property added to the `options` object in the test command and adds a smoke test and unit test to assert that the content is not present.

This is not a long term strategy for solving this issue but fixes the immediate problem that prevents users from adopting the beta. In future we'll want to decouple the file + parsed content from any metadata and ensure that the file content is only passed where needed and discarded when used.  We should also look into using a whitelist for the logger to only allow specific arguments to be logged.

Start at the smoke tests then the `test()` function and finish up with the unit tests. The unit tests are really a temporary measure I think to ensure that the `iacDirFiles` object is clean, we can remove the tests when the flow has been refactored.

Run the following and verify that the "args" property under analytics is clean of file content.

    node ./dist/cli iac test -d --experimental ./test/fixtures/iac/file-logging

### DIFF
--- a/packages/snyk-fix/src/index.ts
+++ b/packages/snyk-fix/src/index.ts
@@ -2,9 +2,8 @@ import * as debugLib from 'debug';
 import * as pMap from 'p-map';
 import * as ora from 'ora';
 import * as chalk from 'chalk';
-import stripAnsi = require('strip-ansi');
 
-import * as outputFormatter from './lib/output-formatters/show-results-summary';
+import { showResultsSummary } from './lib/output-formatters/show-results-summary';
 import { loadPlugin } from './plugins/load-plugin';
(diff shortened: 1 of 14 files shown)


rontalx:
I mistakenly messed up the commits, re-opening PR here:
https://github.com/snyk/snyk/pull/1719

PR #1724, opened 2021-03-15
feat: add sarif support for snyk code

Spoor2709:
- [x] Ready for review

#### What does this PR do?
This pr adds sarif support to snyk code

### DIFF
--- a/src/lib/plugins/sast/index.ts
+++ b/src/lib/plugins/sast/index.ts
@@ -9,6 +9,7 @@ import {
 } from './format/output-format';
 import { EcosystemPlugin } from '../../ecosystems/types';
 import { FailedToRunTestError } from '../../errors/failed-to-run-test-error';
+import { jsonStringifyLargeObject } from '../../json';
 
 const debug = debugLib('snyk-code-test');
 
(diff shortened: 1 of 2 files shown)


src/lib/plugins/sast/index.ts
```diff
@@ -30,6 +30,9 @@ export const codePlugin: EcosystemPlugin = {
         path,
         options,
       );
+      if(options.sarif || options.json){
+        return { readableResult: JSON.stringify(sarifTypedResult, null, 2)}
```
ArturSnyk, line 34: 1. [suggestion]: we have a lib function for stringifying json `jsonStringifyLargeObject` it just has a fallback if we fail doing it this way. 
2. The json and the sarif output are the same?

test/snyk-code-test.spec.ts
```diff
@@ -118,6 +118,28 @@ describe('Test snyk code', () => {
...
+      code: true,
+      sarif: true
+    };
+
+    analyzeFoldersMock.mockResolvedValue(sampleAnalyzeFoldersResponse);
+    isFeatureFlagSupportedForOrgSpy.mockResolvedValueOnce({ ok: true });
+
+    try {
+      await ecosystems.testEcosystem('code', ['some/path'], options);
+    } catch (error) {
```
ArturSnyk, line 135: why are we failing this call?

src/lib/plugins/sast/index.ts
Spoor2709: 1. Will implement
2. Yes the json and sarif output are the same.

ArturSnyk (approved):

PR #1726, opened 2021-03-15
fix: Error messages when Terraform validation fails due to bad credentials

aron:
- [x] Ready for review

#### What does this PR do?

Fixes CC-751

Handles 401 authentication errors when making a validation request to Terraform, previously we would always return a result regardless of the HTTP response, this resulted in misleading error messages.

This PR also fixes the type definition of the `request()` function so the caller no longer has to unwrap the returned response from the promise. The return type for the non-callback approach is now `Promise<{res: NeedleRes, body: any}>` instead of `Promise<void | {res: NeedleRes, body: any}>`

All the code is in src/lib/iac/iac-parser.ts, we have no unit tests for this code path :/ and it will all soon be replaced by the `--experimental` branch.

    1. Invalidate the API token e.g. `snyk config set api=foo`
    2. Test a Terraform file e.g `node ./dist/cli iac test ./test/fixtures/iac/terraform/sg_open_ssh.tf`

https://snyksec.atlassian.net/browse/CC-751

#### Screenshots

Before

image

After 
image

### DIFF
--- a/src/lib/iac/iac-parser.ts
+++ b/src/lib/iac/iac-parser.ts
@@ -1,7 +1,12 @@
 //TODO(orka): take out into a new lib
 import * as YAML from 'js-yaml';
 import * as debugLib from 'debug';
-import { IllegalIacFileErrorMsg, NotSupportedIacFileErrorMsg } from '../errors';
+import {
+  IllegalIacFileErrorMsg,
+  InternalServerError,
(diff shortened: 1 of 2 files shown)


JackuB (approved):

ipapast (approved):
Ran it locally with invalid token. works with a nice message  :)

PR #1737, opened 2021-03-17
feat: refactor to make it easier to pin/upgrade or both

lili2311:
- [ ] Ready for review

#### What does this PR do?
Refactor remediation logic to make it easier to separate applying
pins/upgrades to a manifest. This is key to applying changes to
multiple related manifests down the line.

Previously the file was mutated as it went through upgrade & pin logic. We need these separated to work well across applying fixed for multiple files.


packages/snyk-fix/src/plugins/python/handlers/pip-requirements/update-dependencies/generate-pins.ts
```diff
@@ -0,0 +1,56 @@
+import { DependencyPins, FixChangesSummary } from '../../../../../types';
+import { isDefined } from './is-defined';
+import { Requirement } from './requirements-file-parser';
+
+export function generatePins(
+  requirements: Requirement[],
+  updates: DependencyPins,
+): { pinnedRequirements: string[]; changes: FixChangesSummary[] } {
```
lili2311, line 8: the same logic mostly as before.

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/update-dependencies/generate-pins.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const changes: FixChangesSummary[] = [];
+  const pinnedRequirements = Object.keys(lowerCasedPins)
+    .map((pkgNameAtVersion) => {
+      const [pkgName, version] = pkgNameAtVersion.split('@');
+
+      // Pinning is only for non top level deps
+      if (topLevelDeps.indexOf(pkgName) >= 0) {
+        return;
```
lili2311, line 39: no longer returning original line, only returning updates

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/update-dependencies/generate-upgrades.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (!upgrade) {
+        return;
+      }
+      const newVersion = lowerCasedUpgrades[upgrade].split('@')[1];
+      const updatedRequirement = `${originalName}${versionComparator}${newVersion}`;
+      changes.push({
+        success: true,
+        userMessage: `Upgraded ${originalName} from ${version} to ${newVersion}`,
+      });
+      updatedRequirements[originalText] = `${updatedRequirement}${
```
lili2311, line 65: for upgrades we need to know which original line was updated to apply this back to the manifest down the line.

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/update-dependencies/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  );
+  debug('Finished generating upgrades to apply');
+
+  const { pinnedRequirements, changes: pinChanges } = generatePins(
+    parsedRequirementsData,
+    updates,
+  );
+  debug('Finished generating pins to apply');
+
+  let updatedManifest = [
```
lili2311, line 46: updating the manifest is now a separate activity from either pins or upgrades being applied. So we can apply only upgrades or only pins or neither and it will return merged changes correctly

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/update-dependencies/requirements-file-parser.ts
```diff
@@ -34,7 +34,7 @@ export function parseRequirementsFile(requirementsFile: string): Requirement[] {
 function extractDependencyDataFromLine(
   requirementText: string,
   line: number,
-): Requirement | false {
+): Requirement | void {
```
lili2311, line 37: fixing return type

packages/snyk-fix/test/unit/fix.spec.ts
```diff
@@ -270,17 +270,17 @@ describe('groupEntitiesPerScanType', () => {
     const txtProdProjectTestResult = generateEntityToFix(
       'pip',
       'prod.txt',
-      JSON.stringify({}),
+      'django===1.6.1',
```
lili2311, line 273: this test no longer works with fake `{}` manifest and it should not have done before either really. Because the manifest parsing check is now tightened and throws if the manifest is empty. So needed to give this test real manifest content to work with now.

aarlaud (approved):

PR #1739, opened 2021-03-18
feat: add elixir support

admons:
- [x] Ready for review

#### What does this PR do?

Adds support for elixir ecosystem


src/lib/plugins/index.ts
```diff
@@ -55,6 +56,9 @@ export function loadPlugin(
     case 'cocoapods': {
       return cocoapodsPlugin;
     }
+    case 'elixir': {
```
lili2311, line 59: would a new language not be added via the new ecosystems flow now? cc @darscan is that the expected new path?

darscan: Yep, this should be using the new flow. @admons let's chat

src/lib/detect.ts
```diff
@@ -31,6 +31,7 @@ const DETECTABLE_FILES: string[] = [
   'Podfile.lock',
   'pyproject.toml',
   'poetry.lock',
+  // 'mix.exs', // todo: remove when hex is going GA
```
dtrunley-snyk, line 34: Out of interest do we not need mix.lock here too?

src/lib/package-managers.ts
```diff
@@ -49,6 +51,7 @@ export const GRAPH_SUPPORTED_PACKAGE_MANAGERS: SupportedPackageManagers[] = [
   'yarn',
   'rubygems',
   'poetry',
+  'hex',
```
dtrunley-snyk, line 54: Had a chat with @anthogez as I wanted to flag that this might be required for Gradle (I had incorrectly recalled something about Gradle moving to dep-graphs but it was actually around introducing support cycles). 
It could definitely be better named but we don't need to add hex here if we build the dep graph from scratch (so it sounds like poetry also shouldn't be there). This variable is for package managers that build a dep tree and convert to dep graph.

src/lib/detect.ts
admons: Seems like we will need it, but this is irrelevant at the moment (as we're not auto detecting elixir projects)

dtrunley-snyk: Of course, was just interested if it was a new approach or not 👍

package.json
```diff
@@ -78,6 +78,7 @@
     "@snyk/graphlib": "^2.1.9-patch.3",
     "@snyk/inquirer": "^7.3.3-patch",
     "@snyk/snyk-cocoapods-plugin": "2.5.2",
+    "@snyk/snyk-hex-plugin": "^1.0.0",
```
anthogez, line 81: please remove `^` from `"@snyk/snyk-hex-plugin": "^1.0.0"`

admons: You're right, I'm used to projects with lock file

dtrunley-snyk (approved):

PR #1742, opened 2021-03-18
fix: stringify json output to promise valid json

Avishagp:
- [X] Ready for review

#### What does this PR do?
Call internal stringify functions on `snyk monitor` json output as done in `test` command to promise a valid json output.

Before:
Screen Shot 2021-03-18 at 14 16 24

After:
Screen Shot 2021-03-18 at 14 16 11

[Jira ticet](https://snyksec.atlassian.net/browse/HAMMER-250)
[ZenDesk ticket](https://snyk.zendesk.com/agent/tickets/9120)

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -253,7 +253,7 @@ async function monitor(...args0: MethodArgs): Promise<any> {
   }
   // Part 2: process the output from the Registry
   if (options.json) {
-    return processJsonMonitorResponse(results);
+    return processJsonMonitorResponse(results, options);
   }
 
   const output = results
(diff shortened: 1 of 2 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -253,7 +253,7 @@ async function monitor(...args0: MethodArgs): Promise<any> {
   }
   // Part 2: process the output from the Registry
   if (options.json) {
-    return processJsonMonitorResponse(results);
+    return processJsonMonitorResponse(results, options);
```
JackuB, line 256: You already have baked in the `options.json` in the check above, wouldn't it be better to just tell it to the function rather than to pass the object again?

Avishagp: options is a needed argument in the formatJsonOutput function, not sure I understand what you mean here

PR #1753, opened 2021-03-21
feat: scan terraform plan delta, experimental

rontalx:
#### What does this PR do?
Enhances the experimental terraform-plan scanning functionality to do delta scans as a default, instead of a full-state scan as a default.

Changes to `terraform-plan-parser.ts` contain the new logic to extract resource changes.
Then check the new unit-tests `terraform-plan-parser.spec.ts`.
And then the rest.
Please not some renaming was done to differentiate between the full & delta scans.

`snyk iac test ./tf-plan.json --experimental`

After some discussions on the topic, we reached consensus that the default should be a delta scan, and not a full scan.
Soon, a flag will be added to allow customers to control which scan they want with a flag.

https://snyksec.atlassian.net/browse/CC-758

#### Screenshots
image

### DIFF
--- a/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
@@ -66,7 +66,7 @@ function extractChildModulesResources(
 function extractResourceChanges(
   terraformPlanJson: TerraformPlanJson,
 ): Array<TerraformPlanResourceChange> {
-  return terraformPlanJson?.resource_changes || [];
+  return terraformPlanJson.resource_changes || [];
 }
 
 function extractResourcesForFullScan(
@@ -92,7 +92,7 @@ function extractResourcesForDeltaScan(
 
 export function tryParsingTerraformPlan(
   terraformPlanFile: IacFileData,
-  isFullScan = false,
+  { isFullScan }: { isFullScan: boolean } = { isFullScan: false },
 ): Array<IacFileParsed> {
   let terraformPlanJson;
   try {
--- a/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
@@ -4,7 +4,9 @@ import {
   IacFileParsed,
   TerraformPlanJson,
   TerraformPlanResource,
+  VALID_RESOURCE_ACTIONS,
   TerraformScanInput,
+  TerraformPlanResourceChange,
 } from '../types';
 
 function terraformPlanReducer(
@@ -22,6 +24,20 @@ function terraformPlanReducer(
   } else {
     // add a new resource type
     scanInput[inputKey][type] = { [name]: values };
+  }
+
+  return scanInput;
+}
+
+function resourceChangeReducer(
+  scanInput: TerraformScanInput,
+  resource: TerraformPlanResourceChange,
+): TerraformScanInput {
+  // TODO: investigate if we need to adress also `after_unknown` field.
+  const { actions, after } = resource.change || { actions: [], after: {} };
+  if (actions.some((action) => VALID_RESOURCE_ACTIONS.includes(action))) {
+    const resourceForReduction = { ...resource, values: after || {} };
+    return terraformPlanReducer(scanInput, resourceForReduction);
   }
 
   return scanInput;
@@ -47,8 +63,36 @@ function extractChildModulesResources(
   return extractedChildModuleResources;
 }
 
+function extractResourceChanges(
+  terraformPlanJson: TerraformPlanJson,
+): Array<TerraformPlanResourceChange> {
+  return terraformPlanJson?.resource_changes || [];
+}
+
+function extractResourcesForFullScan(
+  terraformPlanJson: TerraformPlanJson,
+): TerraformScanInput {
+  const rootModuleResources = extractRootModuleResources(terraformPlanJson);
+  const childModuleResources = extractChildModulesResources(terraformPlanJson);
+  return [
+    ...rootModuleResources,
+    ...childModuleResources,
+  ].reduce(terraformPlanReducer, { resource: {}, data: {} });
+}
+
+function extractResourcesForDeltaScan(
+  terraformPlanJson: TerraformPlanJson,
+): TerraformScanInput {
+  const resourceChanges = extractResourceChanges(terraformPlanJson);
+  return resourceChanges.reduce(resourceChangeReducer, {
+    resource: {},
+    data: {},
+  });
+}
+
 export function tryParsingTerraformPlan(
   terraformPlanFile: IacFileData,
+  isFullScan = false,
 ): Array<IacFileParsed> {
   let terraformPlanJson;
   try {
@@ -58,21 +102,15 @@ export function tryParsingTerraformPlan(
   } catch (err) {
     throw new Error('Failed to parse Terraform plan JSON file.');
   }
-
   try {
-    const rootModuleResources = extractRootModuleResources(terraformPlanJson);
-    const childModuleResources = extractChildModulesResources(
-      terraformPlanJson,
-    );
-    const parsedInput = [
-      ...rootModuleResources,
-      ...childModuleResources,
-    ].reduce(terraformPlanReducer, { resource: {}, data: {} });
+    const scannableInput = isFullScan
+      ? extractResourcesForFullScan(terraformPlanJson)
+      : extractResourcesForDeltaScan(terraformPlanJson);
 
     return [
       {
         ...terraformPlanFile,
-        jsonContent: parsedInput,
+        jsonContent: scannableInput,
         engineType: EngineType.Terraform,
       },
     ];
--- a/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
@@ -4,6 +4,7 @@ import {
   IacFileParsed,
   TerraformPlanJson,
   TerraformPlanResource,
+  ResourceActions,
   VALID_RESOURCE_ACTIONS,
   TerraformScanInput,
   TerraformPlanResourceChange,
@@ -35,7 +36,7 @@ function resourceChangeReducer(
 ): TerraformScanInput {
   // TODO: investigate if we need to adress also `after_unknown` field.
   const { actions, after } = resource.change || { actions: [], after: {} };
-  if (actions.some((action) => VALID_RESOURCE_ACTIONS.includes(action))) {
+  if (isValidResourceActions(actions)) {
     const resourceForReduction = { ...resource, values: after || {} };
     return terraformPlanReducer(scanInput, resourceForReduction);
   }
@@ -43,10 +44,19 @@ function resourceChangeReducer(
   return scanInput;
 }
 
+function isValidResourceActions(action: ResourceActions): boolean {
+  return VALID_RESOURCE_ACTIONS.some((validAction: string[]) => {
+    if (action.length !== validAction.length) return false;
+    return validAction.every(
+      (field: string, idx: number) => action[idx] === field,
+    );
+  });
+}
+
 function extractRootModuleResources(
   terraformPlanJson: TerraformPlanJson,
 ): Array<TerraformPlanResource> {
-  return terraformPlanJson?.planned_values?.root_module?.resources || [];
+  return terraformPlanJson.planned_values?.root_module?.resources || [];
 }
 
 function extractChildModulesResources(
--- a/src/cli/commands/test/iac-local-execution/types.ts
+++ b/src/cli/commands/test/iac-local-execution/types.ts
@@ -85,10 +85,17 @@ export interface TerraformPlanResource {
   mode: string; // "managed",
   type: string; // "aws_cloudwatch_log_group",
   name: string; // "terra_ci",
-  provider_name: string; // "registry.terraform.io/hashicorp/aws",
-  schema_version: number;
   values: Record<string, unknown>; // the values in the resource
   index: number;
+}
+
+export interface TerraformPlanResourceChange
+  extends Omit<TerraformPlanResource, 'values'> {
+  change: {
+    actions: ResourceActions;
+    before: Record<string, unknown> | null; // will be null when the action is `create`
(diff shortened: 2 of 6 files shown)


src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -47,8 +63,36 @@ function extractChildModulesResources(
...
+function extractResourcesForFullScan(
+  terraformPlanJson: TerraformPlanJson,
+): TerraformScanInput {
+  const rootModuleResources = extractRootModuleResources(terraformPlanJson);
+  const childModuleResources = extractChildModulesResources(terraformPlanJson);
+  return [
+    ...rootModuleResources,
+    ...childModuleResources,
+  ].reduce(terraformPlanReducer, { resource: {}, data: {} });
+}
```
rontalx, line 81: just an extraction to a method of the previous logic.

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -27,6 +29,20 @@ function terraformPlanReducer(
   return scanInput;
 }
 
+function resourceChangeReducer(
+  scanInput: TerraformScanInput,
+  resource: TerraformPlanResourceChange,
+): TerraformScanInput {
+  // TODO: investigate if we need to adress also `after_unknown` field.
```
p0tr3c, line 36: These would be fields that are generated by service provider once resource is created/updated. I don't expect us to deal with these until we have deployed environment feature fully operational. None of our policy content evaluates these fields.

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -99,15 +106,23 @@ export interface TerraformPlanJson {
...
-  name: string;
-  values: Record<string, unknown>;
-}
+// taken from: https://www.terraform.io/docs/internals/json-format.html#change-representation
+type ResourceActions =
+  | ['no-op']
+  | ['create']
+  | ['read']
+  | ['update']
+  | ['delete', 'create'] // not sure what this pair does
```
p0tr3c, line 123: sometimes there is no way to update a resource. In this cases the provider will delete existing one and create a new one.

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -99,15 +106,23 @@ export interface TerraformPlanJson {
...
+  | ['create', 'delete'] // not sure what this pair does
```
p0tr3c, line 124: this one I have no clue, makes little sense atm

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -99,15 +106,23 @@ export interface TerraformPlanJson {
...
+  | ['delete'];
+
+// we will be scanning the `create` & `update` actions only.
```
p0tr3c, line 127: we need to at least cover `delete | create` too.

p0tr3c (approved):
Approved, but note that we will need the `delete|create` tuple action as well.

src/cli/commands/test/iac-local-execution/types.ts
rontalx: cool, then this is taken care of in the implementation I did.

src/cli/commands/test/iac-local-execution/types.ts
rontalx: will be taken care of as the logic is:
"if either `update` or `create` is in the actions array AND the `after` object is not null"

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -47,8 +63,36 @@ function extractChildModulesResources(
   return extractedChildModuleResources;
 }
 
+function extractResourceChanges(
+  terraformPlanJson: TerraformPlanJson,
+): Array<TerraformPlanResourceChange> {
+  return terraformPlanJson?.resource_changes || [];
```
aron, line 69: nit: why is this `?` 
```suggestion
  return terraformPlanJson.resource_changes || [];
```

rontalx: being graceful as possible, so any case someone provides a tf plan without `resource_changes` (like we had with the child_modules), then it defaults to an empty array, instead of assuming it's always there.

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
```diff
@@ -47,8 +63,36 @@ function extractChildModulesResources(
...
+  const resourceChanges = extractResourceChanges(terraformPlanJson);
+  return resourceChanges.reduce(resourceChangeReducer, {
+    resource: {},
+    data: {},
+  });
+}
+
 export function tryParsingTerraformPlan(
   terraformPlanFile: IacFileData,
+  isFullScan = false,
```
aron, line 95: nit: I prefer to handle these kinds of arguments as an options object for readability at the call site.

```ts
export function tryParsingTerraformPlan(
  terraformPlanFile: IacFileData,
  {isFullScan = false}: {isFullScan: boolean} = {},
)

// call site
tryParsingTerraformPlan(data, { isFullScan: true });

// vs current
tryParsingTerraformPlan(data, true);
```

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -99,15 +106,23 @@ export interface TerraformPlanJson {
...
   resource: Record<string, Record<string, unknown>>;
   data: Record<string, Record<string, unknown>>;
 }
 
-export interface TerraformPlanResource {
-  type: string;
-  name: string;
-  values: Record<string, unknown>;
-}
+// taken from: https://www.terraform.io/docs/internals/json-format.html#change-representation
```
aron, line 117: 👍 nice comment

src/cli/commands/test/iac-local-execution/types.ts
aron: @rontalx it feels like this kind of complex behaviour should be explicitly handled for readability. I would suggest making the logic handle the ResourceActions arrays as distinct identifiers rather than relying on the contents of them. Also worth ensuring that this path is covered in the unit tests, I skimmed the fixture but I don't think we cover this right now?

rontalx: reason I didn’t stick to explicit identifiers, is to keep it intentionally more permissive. 
so in case that there are action combinations that we’re not aware of yet, then it will handle it more gracefully in an attempt to scan them as well.

src/cli/commands/test/iac-local-execution/parsers/terraform-plan-parser.ts
aron: Ah I see, that part is covered by `resource_changes || []`. The `terraformPlanJson?` is checking whether `terraformPlanJson` is `null` or `undefined` which is not currently allowed via the signature and also feels like it would be an invariant?

rontalx: woops, you're correct.
my brain mistakenly interpreted it as optional chaining for the `resources_changes` field

src/cli/commands/test/iac-local-execution/types.ts
aron: Okay, I would strongly document this assumption here. My personal feelings towards parsing arbitrary complex data structures is to be as strict as possible to avoid surprises. This to me gives the following advantages:

1) The next person maintaining the code can reason explicitly about what the code is going to do and what it supported at the time of writing. They can then reason about how to extend it or whether a behavior is a bug or a feature.
2) The behaviour is known, I love that we want a permissive parser but we don't know what kind of actions might be added in the future and whether for any action containing "create" or "update" our code will currently do the right thing now or in 6 months time after the logic has been extended with the assumption that the above subset are all we support. I might be overthinking this, but can we say that this is the case right now? If no, do we want to take on that debt? This gives me very XML vs. HTML vibes :)

rontalx: TBH, I'm fine with both approaches, and your concerns are absolutely valid.
So if you feel strongly about it, I can tackle that quickly to be explicit.

p0tr3c: I think we need to include the case in the fixture, following is an example. You can easily trigger this case by `tainting` a resource.
```
{
      "address": "aws_iam_user_group_membership.p0tr3c",
      "mode": "managed",
      "type": "aws_iam_user_group_membership",
      "name": "p0tr3c",
      "provider_name": "registry.terraform.io/hashicorp/aws",
      "change": {
        "actions": [
          "delete",
          "create"
        ],
        "before": {
          "groups": [
            "organization-administrators"
          ],
          "id": "terraform-20210210215502257100000001",
          "user": "p0tr3c"
        },
        "after": {
          "groups": [
            "organization-administrators"
          ],
          "user": "p0tr3c"
        },
        "after_unknown": {
          "groups": [
            false
          ],
          "id": true
        }
      }
    }
    ```

Comment (p0tr3c) [reply src/cli/commands/test/iac-local-execution/types.ts:127]: `create|delete` is used for zero-downtime updates. You can trigger it with `lifecycle`. Example plan output 
```
{
    "address": "aws_instance.terra_ci_public",
    "mode": "managed",
    "type": "aws_instance",
    "name": "terra_ci_public",
    "provider_name": "registry.terraform.io/hashicorp/aws",
    "change": {
      "actions": [
        "create",
        "delete"
      ],
      "before": {
        "ami": "ami-096a4ffcb13cbde38",
        "arn": "arn:aws:ec2:eu-west-1:155403611144:instance/i-00091bc3e2b173598",
        "associate_public_ip_address": true,
        "availability_zone": "eu-west-1a",
        "cpu_core_count": 1,
        "cpu_threads_per_core": 2,
        "credit_specification": [
          {
            "cpu_credits": "unlimited"
          }
        ],
        "disable_api_termination": false,
        "ebs_block_device": [],
        "ebs_optimized": false,
        "enclave_options": [
          {
            "enabled": false
          }
        ],
        "ephemeral_block_device": [],
        "get_password_data": false,
        "hibernation": false,
        "host_id": null,
        "iam_instance_profile": "",
        "id": "i-00091bc3e2b173598",
        "instance_initiated_shutdown_behavior": null,
        "instance_state": "running",
        "instance_type": "t3.micro",
        "ipv6_address_count": 0,
        "ipv6_addresses": [],
        "key_name": "terra-ci",
        "metadata_options": [
          {
            "http_endpoint": "enabled",
            "http_put_response_hop_limit": 1,
            "http_tokens": "optional"
          }
        ],
        "monitoring": false,
        "network_interface": [],
        "outpost_arn": "",
        "password_data": "",
        "placement_group": "",
        "primary_network_interface_id": "eni-0f558bc9a8b96e4b6",
        "private_dns": "ip-10-0-101-235.eu-west-1.compute.internal",
        "private_ip": "10.0.101.235",
        "public_dns": "",
        "public_ip": "18.200.191.145",
        "root_block_device": [
          {
            "delete_on_termination": true,
            "device_name": "/dev/sda1",
            "encrypted": false,
            "iops": 100,
            "kms_key_id": "",
            "tags": {},
            "throughput": 0,
            "volume_id": "vol-00d3e8dfaab6572b3",
            "volume_size": 8,
            "volume_type": "gp2"
          }
        ],
        "secondary_private_ips": [],
        "security_groups": [],
        "source_dest_check": true,
        "subnet_id": "subnet-050411f3bb5f9dbe4",
        "tags": {},
        "tenancy": "default",
        "timeouts": null,
        "user_data": null,
        "user_data_base64": null,
        "volume_tags": null,
        "vpc_security_group_ids": [
          "sg-0455cfe38c670b7b3"
        ]
      },
      "after": {
        "ami": "ami-0831162e2eb204b16",
        "associate_public_ip_address": true,
        "credit_specification": [],
        "disable_api_termination": null,
        "ebs_optimized": null,
        "get_password_data": false,
        "hibernation": null,
        "iam_instance_profile": null,
        "instance_initiated_shutdown_behavior": null,
        "instance_type": "t3.micro",
        "key_name": "terra-ci",
        "monitoring": null,
        "source_dest_check": true,
        "subnet_id": "subnet-050411f3bb5f9dbe4",
        "tags": null,
        "timeouts": null,
        "user_data": null,
        "user_data_base64": null,
        "volume_tags": null,
        "vpc_security_group_ids": [
          "sg-0455cfe38c670b7b3"
        ]
      },
      "after_unknown": {
        "arn": true,
        "availability_zone": true,
        "cpu_core_count": true,
        "cpu_threads_per_core": true,
        "credit_specification": [],
        "ebs_block_device": true,
        "enclave_options": true,
        "ephemeral_block_device": true,
        "host_id": true,
        "id": true,
        "instance_state": true,
        "ipv6_address_count": true,
        "ipv6_addresses": true,
        "metadata_options": true,
        "network_interface": true,
        "outpost_arn": true,
        "password_data": true,
        "placement_group": true,
        "primary_network_interface_id": true,
        "private_dns": true,
        "private_ip": true,
        "public_dns": true,
        "public_ip": true,
        "root_block_device": true,
        "secondary_private_ips": true,
        "security_groups": true,
        "tenancy": true,
        "vpc_security_group_ids": [
          false
        ]
      }
    }
  }
```

Comment (aron) [reply src/cli/commands/test/iac-local-execution/types.ts:127]: Added the fixtures and updated the tests to treat the actions as opaque identifiers.

File: test/smoke/spec/iac/snyk_test_local_exec_spec.sh [hunk 9]
```diff
@@ -146,11 +146,31 @@ Describe "Snyk iac test --experimental command"
...
+      The output should include "  introduced by resource > aws_security_group[some_created_resource] > ingress"
+      # Child modules
+      The output should include "✗ Security Group allows open ingress [Medium Severity] [SNYK-CC-TF-1] in Security Group"
+      The output should include "  introduced by resource > aws_security_group[some_updated_resource] > ingress"
+
+      The output should include "../fixtures/iac/terraform-plan/tf-plan.json for known issues, found 2 issues"
+    End
+
+    # The test below should be enabled once we add the full scan flag
+    xIt "finds issues in a Terraform plan file - full scan flag"
```
p0tr3c, line 169: `xIt` typo?

p0tr3c:
Looks good to me

PR #1758, opened 2021-03-23
refactor: remove SNYK_IAC_SKIP_BUNDLE_DOWNLOAD env var from code

ipapast:
#### What does this PR do?

This PR removes the use of SNYK_IAC_SKIP_BUNDLE_DOWNLOAD env var which we use to skip the download of the bundle in smoke tests. This was because tests were very explicit and we now have changed them in [PR](https://github.com/snyk/snyk/pull/1765) to be more vague.
It also refactors the unit tests which depended on that flag.

Run tests locally.

### DIFF
--- a/test/smoke/spec/iac/snyk_test_local_exec_spec.sh
+++ b/test/smoke/spec/iac/snyk_test_local_exec_spec.sh
@@ -20,7 +20,7 @@ Describe "Snyk iac test --experimental command"
       When run snyk iac test ../fixtures/iac/file-logging -d --experimental
       # We expect the output, specifically the analytics block not to include
       # the following text from the file.
-      The status should be success
+      The status should be failure  # issues found
       The output should not include "PRIVATE_FILE_CONTENT_CHECK"
       The error should not include "PRIVATE_FILE_CONTENT_CHECK"
     End
--- a/src/cli/commands/test/iac-local-execution/local-cache.ts
+++ b/src/cli/commands/test/iac-local-execution/local-cache.ts
@@ -50,15 +50,13 @@ export function getLocalCachePath(engineType: EngineType) {
 }
 
 export async function initLocalCache(): Promise<void> {
-  // temporarily use an ENV var to skip downloading of the bundles if we need to - e.g. smoke tests
-  if (!process.env.SNYK_IAC_SKIP_BUNDLE_DOWNLOAD) {
-    const preSignedUrl =
-      'https://cloud-config-policy-bundles.s3-eu-west-1.amazonaws.com/bundle.tar.gz';
+  const preSignedUrl =
+    'https://cloud-config-policy-bundles.s3-eu-west-1.amazonaws.com/bundle.tar.gz';
 
-    createIacDir();
-    const response: ReadableStream = needle.get(preSignedUrl);
-    await extractBundle(response);
-  }
+  createIacDir();
+  const response: ReadableStream = needle.get(preSignedUrl);
+  await extractBundle(response);
+
   if (!doesLocalCacheExist()) {
     throw Error(
       `Missing IaC local cache data, please validate you have: \n${REQUIRED_LOCAL_CACHE_FILES.join(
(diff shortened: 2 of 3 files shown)


test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -88,7 +88,7 @@ Describe "Snyk iac test --experimental command"
       The output should include "Testing sg_open_ssh.tf..."
 
       The output should include "Infrastructure as code issues:"
-      The output should include "Tested sg_open_ssh.tf for known issues, found 0 issues"
```
aron, line 91: Can we use a regexp here? Fine for the moment though.

aron (approved):
Nice, I wonder if we take it one step further and just look for lines starting with `✗ ` and exclude all policy information entirely.

rontalx:
@aron @ipapast 
I'm not so confident with the changes to the regression tests.
The whole point of migrating our old tap-tests to these was to have a similar last-line of sanity-test defence, which validates the test in various scenarios runs fully E2E with a limited fixture, and outputs an expected somewhat-deterministic result.
These tests help a LOT catching bugs along the way & during development, exactly because of them being a bit strict - even at the price of being a bit annoying some times.
I know that things have changed and now with the experimental flag these work E2E with the production WASM bundle which can yield dynamic results.
But because we don't control yet our own output mechanism, it's the only place we can gain some. minimal confidence that the expected outcome is printed.
Not saying the change is bad, it's good, but IMO it loosens up a bit some coverage which we might miss easily as we don't own the output mechanism.

ipapast:
> @aron @ipapast
> I'm not so confident with the changes to the regression tests assertions.
> The whole point of migrating our old tap-tests to these was to have a similar last-line of sanity-test defence, which validates the test in various scenarios runs fully E2E with a limited fixture, and outputs an expected somewhat-deterministic result.
> These tests help a LOT catching bugs along the way & during development, exactly because of them being a bit strict - even at the price of being a bit annoying some times.
> I know that things have changed and now with the experimental flag these work E2E with the production WASM bundle which can yield dynamic results.
> But because we don't control yet our own output mechanism, it's the only place we can gain some. minimal confidence that the expected outcome is printed.
> Not saying the change is bad, it's good, but IMO it loosens up a bit some coverage which we might miss easily as we don't own the output mechanism.

We still detect issues, just not the exact wording. The difference here is that we will often have updated data, so we can't rely on data like number of issues because we might find more for example. 

Which parts of these change do you think we should not be doing? Open to suggestions. 
From previous discussions in the team and with team hammer we were chatting about validating the main functionality in smoke tests, and strengthening our unit tests more.

aron:
> From previous discussions in the team and with team hammer we were chatting about validating the main functionality in smoke tests, and strengthening our unit tests more.

+100 

> Not saying the change is bad, it's good, but IMO it loosens up a bit some coverage which we might miss easily as we don't own the output mechanism.

@rontalx I totally agree it gave us a good safety net but it was also an abuse of these smoke tests. Jacob suggested that we could create a separate suite of tests using the shellspec suite, this seems to me a good way to achieve what you're after. As an interim fix we could:

1. Make the bundle (S3) url configurable via ENV variable
2. Copy the original version of these shellspec files with all the details into a new directory that provide a local bundle url.
3. Start a webserver and serve the fixtures bundle

But longer term we'd be better off extracting the formatter into a separate file and investing in unit tests for that part of the code which would benefit everyone.

p0tr3c:
Just note from content side of things. Any data that is retrieved from `metadata` information should be considered as dynamic. It will change from time to time. So things such as severities, descriptions, groups, resolve statements will change and should not really be breaking smoke tests. Only good assumption we can promise is that these fields will not be `null`

rontalx:
@ipapast 
> We still detect issues, just not the exact wording.

Exact wording is not important. 
But validating that our core user-flows are resilient, due to the fact we don't control the output logic yet, and don't have other tests that test the CLI E2E.
if something now breaks the `x > y > z` structure, and/or the amount of issues printed/filtered, and/or with severities -
then these tests wont surface it anymore, or any , and these have broken already in the past.

@aron 
I agree with the near/long term suggestions and I think the same. 
Though tbh I'm not seeing why you're calling it an "abuse" based on the circumstances and current state of the CLI, check out the previous Tap tests :)
Anyway, until we reach a point where we own our output module and can test the full E2E flows properly, we should have at least one suite of tests that is using a fixture that covers the output in a slightly stricter approach.
For unblocking the tests with the dynamic download mechanism let's go this way, but have in mind that we still have tech-debt in that aspect that leaves us slightly uncovered in terms of output assertions that our current unit-tests don't cover.

**just to clarify, this is not a blocker for this.**

ipapast:
The actual change of the smoke tests was pushed urgently as a fix that blocked the Pipeline. Relevant PR: https://github.com/snyk/snyk/pull/1765
This PR is now only about removing the use of the SNYK_IAC_SKIP_BUNDLE_DOWNLOAD variable in code.

PR #1762, opened 2021-03-24
feat: Delete local .iac-data directory after scan is finished [CC-765]

ipapast:

#### What does this PR do?

In a previous task, we introduced the download and extraction of files in a local cache directory, called '.iac-data'. 

We do not need the files anymore after the scan/test, so we decided to clean up that local directory after the tests finishes.

We also move the dev dependency `rimraf` from dev deps to prod deps. When we support Node >= v12.10.0 we can switch to the supported rmdirSync with recursive options.
https://snyk.slack.com/archives/C0127HWU0E7/p1616688051070900

- Run `npm run build && node iac test file.tf --experimental`
- Run ` ls .iac-data` and see that there is no such directory in the filesystem anymore.

The package [rimraf](https://www.npmjs.com/package/rimraf) was used for the deletion of the directory. This was already in our dev dependencies. 

_Option 2:_
In our version of node fs.rmdirSync is not supported, so in order to delete a non empty directory we would need to make our own recursive function, to check for files in the directory, unlink them and then delete them one by one. Rimraf does this in one line. That function would look like this:

```
export const cleanLocalCache = (directoryPath) => {
  if (fs.existsSync(directoryPath)) {
    fs.readdirSync(directoryPath).forEach((file) => {
      const curPath = path.join(directoryPath, file);
      fs.unlinkSync(curPath);
    });
    fs.rmdirSync(directoryPath);
  } else {
    console.log('Directory path not found');
  }
};
```

[CC-765]

[CC-765]: https://snyksec.atlassian.net/browse/CC-765


src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -67,3 +68,15 @@ export async function initLocalCache(): Promise<void> {
...
 }
+
+export function cleanLocalCache() {
+  // path to delete is hardcoded for now
+  const iacPath: fs.PathLike = path.join(`${process.cwd()}`, '.iac-data');
+  if (fs.existsSync(iacPath) && !process.env.SNYK_IAC_SKIP_BUNDLE_DOWNLOAD) {
+    try {
+      rimraf.sync(iacPath);
+    } catch (e) {
+      throw Error('The local cache directory could not be deleted');
```
rontalx, line 79: I don't think it's worth breaking the flow in that case, if from whatever reason it might have not succeeded deleting. 
as if there were results, it will completely not show them :\
when we add debug logs we can add here some extra info to learn on such cases potentially.
maybe we should gracefully handle it? 
as in case of a failed deletion, the files being left on disk will be idempotent for the next re-run.

src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -2,6 +2,7 @@ import * as path from 'path';
 import * as fs from 'fs';
 import { EngineType } from './types';
 import * as needle from 'needle';
+import * as rimraf from 'rimraf';
```
aron, line 5: This requires us to move the rimraf package into production `dependencies` which I think should require a review from Hammer as it has an impact on bundle size and potentially distribution. Is there a way we can do this using just the `fs` api?

ipapast: ^^ @aron  documented on PR description

src/cli/commands/test/iac-local-execution/local-cache.ts
aron: We have the `debug()` function we could use here, but yes I agree it shouldn't result in a critical failure.

src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -67,3 +68,15 @@ export async function initLocalCache(): Promise<void> {
     );
   }
 }
+
+export function cleanLocalCache() {
+  // path to delete is hardcoded for now
+  const iacPath: fs.PathLike = path.join(`${process.cwd()}`, '.iac-data');
+  if (fs.existsSync(iacPath) && !process.env.SNYK_IAC_SKIP_BUNDLE_DOWNLOAD) {
```
aron, line 75: This line is only here until we land the more permissive smoke tests right?

src/cli/commands/test/iac-local-execution/local-cache.ts
aron: My suggestion would be to use your code snippet from the description here instead.

src/cli/commands/test/iac-local-execution/local-cache.ts
ipapast: Chatted with Aron, we have rebased into master and got the permissive tests.
Also removed the flag so now we download and remove the cache every time.

src/cli/commands/test/iac-local-execution/local-cache.ts
ipapast: Sounds good, replaced it with a debug line

aron (approved):
👍 Very nice, shall we add a note above the rimraf import/call site that says it can be removed when minimum node version is above 12.x?

ipapast:
> 👍 Very nice, shall we add a note above the rimraf import/call site that says it can be removed when minimum node version is above 12.x?

Good call, I'll add that now

src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -65,3 +70,17 @@ export async function initLocalCache(): Promise<void> {
     );
   }
 }
+
+export function cleanLocalCache() {
+  // path to delete is hardcoded for now
+  const iacPath: fs.PathLike = path.join(`${process.cwd()}`, '.iac-data');
+  if (fs.existsSync(iacPath) && fs.lstatSync(iacPath).isDirectory()) {
```
jahed-snyk, line 77: Not blocking, but this `if` guard shouldn't be needed as rimraf already checks and succeeds if the directory doesn't exist.

> If the file doesn't exist, rimraf will return successfully, since your desired outcome is already the case.

Source: https://github.com/isaacs/rimraf

jahed-snyk (approved):

src/cli/commands/test/iac-local-execution/local-cache.ts
ipapast: Good call, I'll merge it right after this. thank you!

PR #1766, opened 2021-03-25
fix: support package.json missing name property

JamesPatrickGill:

#### What does this PR do?

This is to fix a bug where package.json without name property would throw an error.

### DIFF
--- a/src/lib/plugins/nodejs-plugin/npm-modules-parser.ts
+++ b/src/lib/plugins/nodejs-plugin/npm-modules-parser.ts
@@ -34,6 +34,7 @@ export async function parse(
     const packageJson = JSON.parse(
       getFileContents(root, packageJsonFileName).content,
     );
+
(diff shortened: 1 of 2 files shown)


src/lib/plugins/nodejs-plugin/npm-modules-parser.ts
```diff
@@ -34,14 +34,15 @@ export async function parse(
...
+
     let dependencies = packageJson.dependencies;
     if (options.dev) {
       dependencies = { ...dependencies, ...packageJson.devDependencies };
     }
     if (isEmpty(dependencies)) {
       return new Promise((resolve) =>
         resolve({
-          name: packageJson.name,
+          name: packageJson.name || 'unnamed-package',
```
gitphill, line 45: is this the default name we use elsewhere in Snyk?
could you write a test to assert this fallback behaviour?

gitphill (approved):

PR #1777, opened 2021-03-29
Feat: v1 fix with version provenance for requirements.txt projects

lili2311:

#### What does this PR do?
- stop skipping files with `-r` directive (continue not support `-c` for now)
- extract all relevant files referenced via `-r` and apply fixes with this context in mind
- sometimes we may have already fixed a certain file that snyk test will send to be fixed because it was included by another file.
- `snyk test` does not today skip testing files that have already been tested because they are essentially included as part of other tests, however `snyk fix` needs to separate this so we do not try to apply fixes again to a fixed file.
- call out previously fixed in the output as we bail to apply changes again there.

- `git rebase feat/snyk-fix` https://github.com/snyk/snyk/pull/1707
- `cd packages/snyk-fix/test/acceptance/plugins/python/update-dependencies/workspaces/app-with-already-fixed && snyk fix --all-projects --org=org-with-ff` 

The logic behind the fix of grouped files is as follows:
- for each sorted directory
- fix all entities in the current directory
- for each entity extract provenance information (find and parse all files that are included via `-r`)
- apply all upgrades first across the related files
- apply the remaining fixes as pins in the entry file (aka the file that was tested via `snyk test` first)
- keep track of files that we applied fixes in and skip fixing it if previously fixed

#### Screenshots
CleanShot 2021-03-29 at 14 09 22@2x


packages/snyk-fix/src/plugins/python/handlers/pip-requirements/index.ts
```diff
@@ -26,56 +35,207 @@ export async function pipRequirementsTxt(
...
+        handlerResult.succeeded.push({
+          original: entity,
+          changes: [{ success: true, userMessage: 'Previously fixed' }],
+        });
+        continue;
+      }
+      const { changes, fixedFiles } = await applyAllFixes(entity, options);
+      if (!changes.length) {
+        debug('Manifest has not changed!');
+        throw new NoFixesCouldBeAppliedError();
```
aarlaud, line 106: Is it necessarily an error? maybe one of the manifest file just doesn't need a change?

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/index.ts
```diff
@@ -26,56 +35,207 @@ export async function pipRequirementsTxt(
...
+  return { changes: [...upgradeChanges, ...pinnedChanges], fixedFiles };
+}
+
+function filterOutAppliedUpgrades(
+  remediation: RemediationChanges,
+  appliedRemediation: string[],
+): RemediationChanges {
+  const pinRemediation: RemediationChanges = {
+    ...remediation,
+    pin: {}, // delete the pin remediation so we can collect un-applied remediation
```
aarlaud, line 208: Why do we need to do that?

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/index.ts
lili2311: `applyAllFixes` is called on an entity that has remediation data available so expect that at least 1 change is applied. That entity may end up fixing multiple other files (the changes are still attached to this entity), and if we see a file was fixed we still do create a changes item to say `Previously fixed` so at this point no fixes is unexpected.

packages/snyk-fix/src/plugins/python/handlers/pip-requirements/index.ts
lili2311: The logic behind the fix of grouped files is as follows:

- for each sorted directory
- fix all entities in the current directory
- for each entity extract provenance information (find and parse all files that are included via -r)
- apply all upgrades first across the related files
- apply the remaining fixes as pins in the entry file (aka the file that was tested via snyk test first)
- keep track of files that we applied fixes in and skip fixing it if previously fixed


>apply the remaining fixes as pins
Here I am returning remediation data and selecting only the pins (for Python these are upgrades & pins) that have not already been applied as part of the apply upgrades flow. So all remainders are applied as pins in the root level scanned file. Because during testing we do not consider `-r` some fixes look at transitive but they are actually upgrades in another file. So first step is to apply all changes to all files where the dependency is defined, then everything else can be assumed to be a pin.

aarlaud: meaning the pin reset to {}

aarlaud (approved):

PR #1778, opened 2021-03-29
feat: introduce envvar to control HTTP-HTTPS upgrade behavior

JackuB:
Follow up to the #1771 

This PR adds an optional ennvar `SNYK_HTTP_PROTOCOL_UPGRADE` to disable the behavior to upgrade `http` endpoints to `https`.

Fixes the behavior about query strings, that could've lacked the `/?` characters

It also updates Help documentation about `HTTP(S)_PROXY` envvar.

### DIFF
--- a/help/commands-docs/_ENVIRONMENT.md
+++ b/help/commands-docs/_ENVIRONMENT.md
@@ -29,7 +29,7 @@ By default Snyk CLI will connect to `https://snyk.io/api/v1`.
   Sets API host to use for Snyk requests. Useful for on-premise instances and configuring proxies. If set with `http` protocol CLI will upgrade the requests to `https`. Unless `SNYK_HTTP_PROTOCOL_UPGRADE` is set to `0`.
 
 - `SNYK_HTTP_PROTOCOL_UPGRADE`=0:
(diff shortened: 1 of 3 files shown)


help/commands-docs/_ENVIRONMENT.md
```diff
@@ -8,19 +8,28 @@ You can set these environment variables to change CLI run settings.
...
+
+## Connecting to Snyk API
+
+By default Snyk CLI will connect to `https://snyk.io/api/v1`.
+
+- `SNYK_API`:
+  Sets API host to use for Snyk requests. Useful for on-premise instances and configuring proxies. If set with `http` protocol CLI will upgrade the requests to `https`. Unless `SNYK_HTTP_PROTOCOL_UPGRADE` is set to `0`.
+
+- `SNYK_HTTP_PROTOCOL_UPGRADE`=0:
+  If set to the value of `0` the API requests aimed at `http` URLs will not be upgraded to `https`. Useful e.g., for reverse proxies.
```
maxjeffos, line 32: Could be more clear that the default behavior is to upgrade the protocol.

```suggestion
  If set to the value of `0`, API requests aimed at `http` URLs will not be upgraded to `https`. If not set, the default behavior will be to upgrade these requests from `http` to `https`. Useful e.g., for reverse proxies.
```

maxjeffos (approved):

PR #1781, opened 2021-03-30
fix: replace vulnerable proxy dependency

JackuB:
Replaces `proxy-agent` with `global-agent` to remove a vulnerable `netmask` dependency


test/jest/acceptance/proxy-behavior.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+              SNYK_API: SNYK_API_HTTPS,
+            },
+          },
+          (err, stdout, stderr) => {
+            if (err) {
+              throw err;
+            }
+
+            expect(stderr).toContain(
+              `Error: connect ECONNREFUSED 127.0.0.1:${
```
maxjeffos, line 29: Some comments in this test file about how these work and why you're expecting an error would probably be nice for the next person in here.

maxjeffos (approved):

test/jest/acceptance/proxy-behavior.spec.ts
JackuB: Added 👍

PR #1788, opened 2021-03-31
feat: adds oauth token support for the test command

gablaxian:

#### What does this PR do?

Adds support for providing an oauth token as authorization using an environment variable, `SNYK_OAUTH_TOKEN`.

- Get hold of an OAuth access token for your user on the Dev server (Extensibility group can help with that).
- Set an environment variable with your access token, e.g. `export SNYK_OAUTH_TOKEN=mytoken`
- Point the CLI to our oauth endpoint, e.g: `snyk-dev set config endpoint=`snyk-dev config set endpoint=https://api.dev.snyk.io/v1`
- Run a test against a project, e.g: `snyk-dev test`

We are in the process of providing an OAuth flow into Snyk for third party integrations and partners. Adding OAuth token support to the CLI allows better integration into these workflows so that they can use the same mechanism to connect their systems to Snyk and then perform various commands on projects. We are limiting the implementation to the `test` command for now as this is all that is needed in the short-term and because there are many areas to update the header in the code.

### DIFF
--- a/src/cli/commands/test/validate-credentials.ts
+++ b/src/cli/commands/test/validate-credentials.ts
@@ -1,11 +1,17 @@
-import { apiTokenExists, getDockerToken } from '../../../lib/api-token';
+import {
+  apiTokenExists,
+  getOAuthToken,
+  getDockerToken,
+} from '../../../lib/api-token';
 import { TestOptions, Options } from '../../../lib/types';
 
-export function validateCredentials(options: Options & TestOptions) {
+export function validateCredentials(options: Options & TestOptions): void {
   try {
     apiTokenExists();
   } catch (err) {
-    if (options.docker && getDockerToken()) {
+    if (getOAuthToken()) {
+      return;
+    } else if (options.docker && getDockerToken()) {
       options.testDepGraphDockerEndpoint = '/docker-jwt/test-dependencies';
       options.isDockerUser = true;
     } else {
--- a/src/lib/api-token.ts
+++ b/src/lib/api-token.ts
@@ -8,6 +8,10 @@ export function api() {
   return config.api || config.TOKEN || userConfig.get('api');
 }
 
+export function getOAuthToken(): string | undefined {
+  return process.env.SNYK_OAUTH_TOKEN;
+}
+
 export function getDockerToken(): string | undefined {
   return process.env.SNYK_DOCKER_TOKEN;
 }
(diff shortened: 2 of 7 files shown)


davidharrigan (approved):
🎉

src/lib/api-token.ts
```diff
@@ -20,10 +24,15 @@ export function apiTokenExists() {
...
 }
 
-export function authHeaderWithApiTokenOrDockerJWT() {
-  const dockerToken = getDockerToken();
+export function getAuthHeader(): string {
+  const oauthToken: string | undefined = getOAuthToken();
+  const dockerToken: string | undefined = getDockerToken();
+
+  if (oauthToken) {
+    return `Bearer ${oauthToken}`;
```
jahed-snyk, line 32: Was `Bearer` and `bearer` used to differentiate between OAuth and Docker? Seems a bit confusing. According to IANA we can (should?) use `OAuth` as the scheme type here.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

src/cli/commands/test/validate-credentials.ts
```diff
@@ -1,11 +1,17 @@
...
+} from '../../../lib/api-token';
 import { TestOptions, Options } from '../../../lib/types';
 
 export function validateCredentials(options: Options & TestOptions) {
   try {
     apiTokenExists();
   } catch (err) {
-    if (options.docker && getDockerToken()) {
+    if (getOAuthToken()) {
+      return true;
```
jahed-snyk, line 13: This should just be a `return` without a value, there's no return type in the other branches.

test/acceptance/oauth-token.test.ts
```diff
@@ -0,0 +1,102 @@
+import * as tap from 'tap';
```
jahed-snyk, line 1: We are moving away from `tap` for new tests. If possible, can you make this a `jest` test under `test/jest/acceptance`? Is there any missing tooling?

jahed-snyk (changes requested):

src/lib/api-token.ts
gablaxian: No, `Bearer` is technically the correct header, I just didn't want to change Docker in case there was a limitation there.

test/acceptance/oauth-token.test.ts
gablaxian: I didn't know there was an active shift to Jest happening. I will convert over and let you know if any tooling is missing 😄

jahed-snyk: Thanks. I'll talk to the team about communicating that better.

src/lib/api-token.ts
jahed-snyk: Docker-wise looks fine to change it. CLI passes it through as-is. All the tests in Registry use `Bearer` for Docker tests. CLI tests it once but it's just a mock server assertion. I'll leave decision to you. Won't block on this.

https://github.com/snyk/registry/blob/c3b38a5aaecdda74875a16812cf813bd9d6ca827/src/web/middleware/index.ts#L1143

gablaxian: Updated Docker to `Bearer`

src/cli/commands/test/validate-credentials.ts
gablaxian: Fixed

test/acceptance/oauth-token.test.ts
gablaxian: Test converted to Jest

jahed-snyk (approved):

PR #1793, opened 2021-04-01
refactor: Scanning Helm files for iac test fails [CC-768]

ipapast:

#### What does this PR do?

This PR fixes the issue where YAML Helm files were scanned incorrectly. 
Instead, we now check if a yaml file contains a template and throw a `HelmFileNotSupportedError` error on that case.

We also add a new smoke test case and a new unit test case.

In the `parser.ts`:
-  we now have a validateYamlFile function for the 'yml' and 'yaml' cases. 
- This function will get the fileContent, split it by line and create an array of strings (every line is a new string).
- Then, we check if there is a template, line by line. A template is considered a line that has both `{{` and `}}` chars.
- The moment we detect a template, we throw a `HelmFileNotSupportedError` with a useful user error message and a new error code.

- Run `node ~/snyk-repos/snyk/dist/cli iac test test/fixtures/iac/kubernetes/helm-unsupported.yaml --experimental` to see a Helm file failing.

We do not support scanning of Helm files via the CLI, thus it should fail gracefully.

[CC-768]
#### Screenshots

image

[CC-768]: https://snyksec.atlassian.net/browse/CC-768

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/file-parser.ts
@@ -16,8 +16,8 @@ import { CustomError } from '../../../../lib/errors';
 export async function parseFiles(
   filesData: IacFileData[],
 ): Promise<ParsingResults> {
-  const parsedFiles: Array<IacFileParsed> = [];
-  const failedFiles: Array<IacFileParseFailure> = [];
+  const parsedFiles: IacFileParsed[] = [];
+  const failedFiles: IacFileParseFailure[] = [];
   for (const fileData of filesData) {
     try {
       parsedFiles.push(...tryParseIacFile(fileData));
@@ -52,11 +52,12 @@ function generateFailedParsedFile(
 
 const TF_PLAN_NAME = 'tf-plan.json';
 
-export function tryParseIacFile(fileData: IacFileData): Array<IacFileParsed> {
+export function tryParseIacFile(fileData: IacFileData): IacFileParsed[] {
   analytics.add('iac-terraform-plan', false);
   switch (fileData.fileType) {
     case 'yaml':
     case 'yml':
+      return tryParsingKubernetesFile(fileData);
     case 'json':
       // TODO: this is a temporary approach for the internal release only
       if (path.basename(fileData.filePath) === TF_PLAN_NAME) {
--- a/src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
@@ -2,17 +2,30 @@ import * as YAML from 'js-yaml';
 import { CustomError } from '../../../../../lib/errors';
 import {
   EngineType,
-  IacFileParsed,
-  IacFileData,
   IaCErrorCodes,
+  IacFileData,
+  IacFileParsed,
 } from '../types';
 
 const REQUIRED_K8S_FIELDS = ['apiVersion', 'kind', 'metadata'];
+
+export function assertHelmAndThrow(fileData: IacFileData) {
+  const lines: string[] = fileData.fileContent.split(/\r\n|\r|\n/);
+
+  lines.forEach((line) => {
+    const isHelmFile = line.includes('{{') && line.includes('}}');
+    if (isHelmFile) {
+      throw new HelmFileNotSupportedError(fileData.filePath);
+    }
+  });
+}
 
 export function tryParsingKubernetesFile(
   fileData: IacFileData,
 ): IacFileParsed[] {
   let yamlDocuments;
+
+  assertHelmAndThrow(fileData);
   try {
     yamlDocuments = YAML.safeLoadAll(fileData.fileContent);
   } catch (e) {
@@ -44,6 +57,15 @@ class FailedToParseKubernetesYamlError extends CustomError {
     this.userMessage = `We were unable to parse the YAML file "${filename}". Please ensure that it contains properly structured YAML`;
   }
 }
+
+export class HelmFileNotSupportedError extends CustomError {
+  constructor(filename: string) {
+    super('Failed to parse Helm file');
+    this.code = IaCErrorCodes.FailedToParseHelmError;
+    this.userMessage = `We were unable to parse the YAML file "${filename}" as we currently do not support scanning of Helm files. More information can be found through our documentation:\nhttps://support.snyk.io/hc/en-us/articles/360012429477-Test-your-Kubernetes-files-with-our-CLI-tool`;
+  }
+}
+
 export class MissingRequiredFieldsInKubernetesYamlError extends CustomError {
   constructor(filename: string) {
     super('Failed to detect Kubernetes file, missing required fields');
--- a/src/cli/commands/test/iac-local-execution/types.ts
+++ b/src/cli/commands/test/iac-local-execution/types.ts
@@ -180,6 +180,7 @@ export enum IaCErrorCodes {
   // kubernetes-parser errors
   FailedToParseKubernetesYamlError = 1030,
   MissingRequiredFieldsInKubernetesYamlError = 1031,
+  FailedToParseHelmError = 1032,
 
   // terraform-file-parser errors
   FailedToParseTerraformFileError = 1040,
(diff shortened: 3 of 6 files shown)


src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -16,13 +19,15 @@ import { CustomError } from './../../../../lib/errors/custom-error';
 export async function parseFiles(
   filesData: IacFileData[],
 ): Promise<ParsingResults> {
-  const parsedFiles: Array<IacFileParsed> = [];
-  const failedFiles: Array<IacFileParseFailure> = [];
+  const parsedFiles: IacFileParsed[] = [];
```
ipapast, line 22: Recommended: TSLint: Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.(array-type)
https://palantir.github.io/tslint/rules/array-type/

src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -16,13 +19,15 @@ import { CustomError } from './../../../../lib/errors/custom-error';
...
+  const failedFiles: IacFileParseFailure[] = [];
```
ipapast, line 23: Recommended: TSLint: Array type using 'Array' is forbidden for simple types. Use 'T[]' instead.(array-type)
https://palantir.github.io/tslint/rules/array-type/

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
```diff
@@ -2,13 +2,24 @@ import * as YAML from 'js-yaml';
...
-  IacFileParsed,
-  IacFileData,
   IaCErrorCodes,
+  IacFileData,
+  IacFileParsed,
 } from '../types';
 
 const REQUIRED_K8S_FIELDS = ['apiVersion', 'kind', 'metadata'];
 
+export function validateYamlFile(fileData: IacFileData) {
```
rontalx, line 12: small sugesstion:
name slightly misleading, I would expect `validateYamlFile` to check the validity of a YAML file.
Maybe rename to something like: `isHelmFile():boolean`, and have it be called as part of the  `tryParsingKubernetesFile` method instead of before it?
e.g.
so `tryParsingKubernetesFile` encapsulates all logic to validate if it's a valid K8s file, and the `file-parser.ts` just calls it.

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
```diff
@@ -2,13 +2,24 @@ import * as YAML from 'js-yaml';
...
+  const lines: string[] = fileData.fileContent.split(/\r\n|\r|\n/);
+
+  lines.forEach((line) => {
+    const isHelmFile = line.includes('{{') && line.includes('}}');
```
rontalx, line 16: cc: @p0tr3c 
@ipapast 
are we 100% sure this won't mistakenly have some edge-case false-positives like we had with Golang templates for Terraform?

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -180,6 +180,7 @@ export enum IaCErrorCodes {
   // kubernetes-parser errors
   FailedToParseKubernetesYamlError = 1030,
   MissingRequiredFieldsInKubernetesYamlError = 1031,
+  FailedToParseHelmError = 1032,
```
rontalx, line 183: new error codes are fun fun fun!

test/jest/unit/iac-unit-tests/file-parser.spec.ts
```diff
@@ -64,4 +65,13 @@ describe('parseFiles', () => {
 
     await expect(parseFilesFn).rejects.toThrow(UnsupportedFileTypeError);
   });
+
+  it('will not parse if Helm file', async () => {
```
rontalx, line 69: nit (although suggested to move this logic inwards):
```suggestion
  it('does not call the kubernetes parser if detected a Helm file', async () => {
```

+can we add an assertion/another-test for the specific error throwing in case of a helm file?

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -54,6 +54,13 @@ Describe "Snyk iac test --experimental command"
       The output should include "We were unable to detect whether the YAML file"
     End
 
+    It "outputs an error for Helm files"
```
rontalx, line 57: so much coverageeeeee 😄 🎉

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
ipapast: I would not say I'm 100% sure this is the best way to detect them but this is based on my best knowledge of what a Helm config looks like and what I read in the docs. Do you have any scenarios for false-positives to think about?

```
[ ] {{- /*
[ ] This is a comment.
[ ] */}}
[ ] type: frobnitz
```

ipapast: I would not say I'm 100% sure this is the best way to detect them but this is based on my best knowledge of what a Helm config looks like and what I read in the docs. Do you have any scenarios for false-positives to think about?

test/jest/unit/iac-unit-tests/file-parser.spec.ts
ipapast: changed this to reflect the implementation

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
ipapast: what about `assertValidKubernetesFile`? 
I moved it inside the K8 function, but I did not make it a boolean, I tried to avoid extra nesting in there.
Now, if it's a valid type we scan, it will continue the parsing, if it is not, it will throw.

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
rontalx: In TF it was in cases wherever `{{}}` was present in string literals & heredocs, which is valid, but we identified it as invalid.
Not sure about Helm, although it does look very reasonable to do it this way.
Can you validate that from Helm's templating syntax perspective?

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
rontalx: sounds to me more like `assertHelmFile`/`assertHelmTemplate`, as the validations for it to be a valid K8s are later on (valid parsed yaml + required fields)

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
```diff
@@ -44,6 +57,15 @@ class FailedToParseKubernetesYamlError extends CustomError {
     this.userMessage = `We were unable to parse the YAML file "${filename}". Please ensure that it contains properly structured YAML`;
   }
 }
+
+export class HelmFileNotSupportedError extends CustomError {
+  constructor(filename: string) {
+    super('Failed to parse Helm file');
+    this.code = IaCErrorCodes.FailedToParseHelmError;
+    this.userMessage = `We were unable to parse the YAML file "${filename}" as we currently do not support scanning of Helm files.`;
```
ipapast, line 65: It just occured to me, should we add something like the TF validation user message here?

`This can be done by running it through the 'terraform validate' command` =>
 ```` Try generating a K8s file from your Helm configuration by running:
helm template ./k8s/helm > helm.yaml
snyk iac test helm.yaml
Also check 
https://support.snyk.io/hc/en-us/articles/360012429477-Test-your-Kubernetes-files-with-our-CLI-tool](https://support.snyk.io/hc/en-us/articles/360012429477-Test-your-Kubernetes-files-with-our-CLI-tool)This can be done by running it through the 'terraform validate' command`

rontalx (approved):

src/cli/commands/test/iac-local-execution/parsers/kubernetes-parser.ts
ipapast: Haha, sorry linguistic concerns.
I think assert means "guarantee", "verify", "ensure", etc. so I named this thinking that we 'guarantee' that it is a valid k8s file. Thinking again, we guarantee that any Helm file will fail.
Counter-suggestion "assertHelmAndThrow" ?

PR #1803, opened 2021-04-09
fix: JSON output for IaC shows line number [CC-680]

ipapast:

#### What does this PR do?

This PR adds the cloud-config-parser library which is a utility library for identifying the issue path in a YAML/JSON/HCL file and returning the relevant line number in order to highlight the relevant line to the users in their files.
This was an existing functionality in the old flow and it is now added in the `--experimental` flow. 

This way, we can add a line number for the specific issue path when running a test command with the `--json` output option.
If the parser library fails for some reason, we capture this in Debug and we set lineNumber as '-1' to indicate this.

- Package.json for the library
- Results-formatter.ts for adding the function where we call the parser functions to get the lineNumber.

- Run `npm run build` on this branch
- Run `node ~/snyk-repos/snyk/dist/cli iac test test/fixtures/iac/terraform/sg_open_ssh.tf --experimental --json`

[CC-680]

#### Screenshots
image

[CC-680]: https://snyksec.atlassian.net/browse/CC-680

### DIFF
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -9,12 +9,15 @@ import {
 import { SEVERITY } from '../../../../lib/snyk-test/common';
 import { IacProjectType } from '../../../../lib/iac/constants';
 import { CustomError } from '../../../../lib/errors';
+import {
+  issuesToLineNumbers,
+  CloudConfigFileTypes,
+} from '@snyk/cloud-config-parser';
+import { UnsupportedFileTypeError } from './file-parser';
+import * as analytics from '../../../../lib/analytics';
+import * as Debug from 'debug';
 
-// import {
-//   issuesToLineNumbers,
-//   CloudConfigFileTypes,
-// } from '@snyk/cloud-config-parser';
-
+const debug = Debug('iac-results-formatter');
 const SEVERITIES = [SEVERITY.LOW, SEVERITY.MEDIUM, SEVERITY.HIGH];
 
 export function formatScanResults(
@@ -32,20 +35,19 @@ export function formatScanResults(
   }
 }
 
-//
-// function getFileTypeForLineNumber(
-//   fileType: string,
-// ): CloudConfigFileTypes {
-//   switch (fileType) {
-//     case 'yaml':
-//     case 'yml':
-//       return CloudConfigFileTypes.YAML;
-//     case 'json':
-//       return CloudConfigFileTypes.JSON;
-//     default:
-//       return CloudConfigFileTypes.YAML;
-//   }
-// }
+function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
+  switch (fileType) {
+    case 'yaml':
+    case 'yml':
+      return CloudConfigFileTypes.YAML;
+    case 'json':
+      return CloudConfigFileTypes.JSON;
+    case 'tf':
+      return CloudConfigFileTypes.TF;
+    default:
+      throw new UnsupportedFileTypeError(fileType);
+  }
+}
 
 const engineTypeToProjectType = {
   [EngineType.Kubernetes]: IacProjectType.K8S,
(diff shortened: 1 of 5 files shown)


rontalx:
can you please double check the tests and as the fixtures are still using `-1`:
https://github.com/snyk/snyk/blob/master/test/jest/unit/iac-unit-tests/results-formatter.fixtures.ts#L74

ipapast:
> can you please double check the tests? as the fixtures are still using `-1`:
> https://github.com/snyk/snyk/blob/master/test/jest/unit/iac-unit-tests/results-formatter.fixtures.ts#L74

We still return -1 when the parser library fails, so that fixture would be a valid scenario. 
I am adding another with a "real " case

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -32,20 +34,17 @@ export function formatScanResults(
...
-// }
+function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
+  switch (fileType) {
+    case 'yaml':
+    case 'yml':
+      return CloudConfigFileTypes.YAML;
+    case 'json':
+      return CloudConfigFileTypes.JSON;
+    default:
+      return CloudConfigFileTypes.TF;
```
rontalx, line 45: super nit:
maybe let's be explicit with the `tf` case ?

test/jest/unit/iac-unit-tests/results-formatter.fixtures.ts
```diff
@@ -56,26 +56,33 @@ export const scanResults: Array<IacFileScanResult> = [
...
+  iacDescription: {
+    issue: anotherPolicyStub.issue,
+    impact: anotherPolicyStub.impact,
+    resolve: anotherPolicyStub.resolve,
+  },
+  severity: anotherPolicyStub.severity,
+  lineNumber: -1,
+};
+
 export const expectedFormattedResults = {
```
rontalx, line 74: nit:
shall we type it?

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -57,23 +56,22 @@ function formatScanResult(
...
+
+    let lineNumber: number;
+    try {
+      lineNumber = issuesToLineNumbers(
+        scanResult.fileContent,
+        getFileTypeForLineNumber(scanResult.fileType),
+        policy.msg.split('.'), // parser defaults to docId:0 and checks for the rest of the path
+      );
+    } catch (err) {
+      debug('Parser library failed. Could not assign lineNumber to issue');
```
rontalx, line 72: Let's add an error-code to analytics in that case like we did in the local cache when it fails to clean (but doesn't fail the process):
e.g.
```
    const err = new FailedToExtractLineNumberError();
    analytics.add('error-code', err.code);
    debug('Parser library failed. Could not assign lineNumber to issue');
```

rontalx (approved):
nice 👍 
left a few small comments

test/jest/unit/iac-unit-tests/results-formatter.spec.ts
```diff
@@ -1,12 +1,16 @@
+jest.mock('@snyk/cloud-config-parser');
```
jahed-snyk, line 1: Can this be placed below the `import` statements? Again, to keep imports at the top grouped together. Examples says it can: https://jestjs.io/docs/mock-functions#mocking-modules

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -9,11 +9,15 @@ import {
 import { SEVERITY } from '../../../../lib/snyk-test/common';
 import { IacProjectType } from '../../../../lib/iac/constants';
 import { CustomError } from '../../../../lib/errors';
+import * as analytics from '../../../../lib/analytics';
+import * as Debug from 'debug';
+const debug = Debug('iac-results-formatter');
```
jahed-snyk, line 14: Would be nice to place `const debug` below the `import` statements so imports are grouped together.

jahed-snyk (approved):

test/jest/unit/iac-unit-tests/results-formatter.spec.ts
ipapast: Yes, we can, good call  👍

PR #1804, opened 2021-04-09
feat: Add detection-depth flag to IaC for directory scan [CC-769]

teodora-sandu:

#### What does this PR do?
This PR adds the already defined `--detection-depth` flag to the `--experimental` flow for IaC. The flag was already defined and used in non-experimental mode, so this PR just passes the CLI options and `detectionDepth` value to the correct function. Unit tests and smoke tests were added to verify functionality, as well as tests to make sure that the flag gets initialised correctly when passed to the CLI.

The flow starts in `src/cli/commands/test/iac-local-execution/index.ts` by passing the `options` variable to the `loadFiles ` function. The rest of the code changes are in `src/cli/commands/test/iac-local-execution/file-loader.ts` and make use of the existing `makeDirectoryIterator` function which is used by the non-experimental flow.

1. Run `npm run build`
2. Run `node ./dist/cli iac test test/fixtures/iac/depth_detection --experimental --detection-depth=2` and notice that 2 projects were tested.
3. Run`node ./dist/cli iac test test/fixtures/iac/depth_detection --experimental --detection-depth=3`  and notice that 3 projects were tested.

N/A

[CC-769](https://snyksec.atlassian.net/browse/CC-769): https://snyksec.atlassian.net/browse/CC-769

#### Screenshots
Screenshot 2021-04-09 at 15 51 49

Screenshot 2021-04-12 at 09 46 00

#### Additional questions
N/A

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-loader.ts
+++ b/src/cli/commands/test/iac-local-execution/file-loader.ts
@@ -1,6 +1,11 @@
 import { makeDirectoryIterator } from '../../../../lib/iac/makeDirectoryIterator';
 import { promises as fs } from 'fs';
-import { IaCErrorCodes, IacFileData, IaCTestFlags, VALID_FILE_TYPES } from './types';
+import {
+  IaCErrorCodes,
+  IacFileData,
+  IaCTestFlags,
+  VALID_FILE_TYPES,
+} from './types';
 import { getFileType } from '../../../../lib/iac/iac-parser';
 import { IacFileTypes } from '../../../../lib/iac/constants';
 import { isLocalFolder } from '../../../../lib/detect';
--- a/src/cli/commands/test/iac-local-execution/file-loader.ts
+++ b/src/cli/commands/test/iac-local-execution/file-loader.ts
@@ -1,6 +1,6 @@
 import { makeDirectoryIterator } from '../../../../lib/iac/makeDirectoryIterator';
 import { promises as fs } from 'fs';
-import { IaCErrorCodes, IacFileData, VALID_FILE_TYPES } from './types';
+import { IaCErrorCodes, IacFileData, IaCTestFlags, VALID_FILE_TYPES } from './types';
 import { getFileType } from '../../../../lib/iac/iac-parser';
 import { IacFileTypes } from '../../../../lib/iac/constants';
 import { isLocalFolder } from '../../../../lib/detect';
@@ -8,11 +8,16 @@ import { CustomError } from '../../../../lib/errors';
 
 const DEFAULT_ENCODING = 'utf-8';
 
-export async function loadFiles(pathToScan: string): Promise<IacFileData[]> {
+export async function loadFiles(
+  pathToScan: string,
+  options: IaCTestFlags,
+): Promise<IacFileData[]> {
   let filePaths = [pathToScan];
 
   if (isLocalFolder(pathToScan)) {
-    filePaths = getFilePathsFromDirectory(pathToScan);
+    filePaths = getFilePathsFromDirectory(pathToScan, {
+      maxDepth: options.detectionDepth,
+    });
   }
 
   const filesToScan: IacFileData[] = [];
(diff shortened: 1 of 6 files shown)


test/jest/unit/iac-unit-tests/file-loader.spec.ts
```diff
@@ -78,21 +84,73 @@ describe('loadFiles', () => {
...
           anotherTerraformFileStub,
         ]);
       });
     });
 
+    describe('with nested files inside multiple-level directories', () => {
+      beforeEach(() => {
+        mockFs({
+          [level1Directory]: {
+            [path.basename(level2Directory)]: {
```
teodora-sandu, line 99: `mock-fs` does not like nesting files inside directories using the full file path, so we need to get the last portion of a path using `path.basename`

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -158,6 +158,23 @@ Describe "Snyk iac test --experimental command"
...
       The output should include "Failed to detect Kubernetes file, missing required fields"
     End
+
+    It "limits the depth of the directories"
+      When run snyk iac test ../fixtures/iac/depth_detection/ --experimental --detection-depth=2
+      The status should be success # no issues found
+      # Only File
+      The output should include "Testing one.tf..."
+      The output should include "Infrastructure as code issues:"
+      The output should include "Tested one.tf for known issues, found 0 issues"
```
ipapast, line 168:
```suggestion
      The output should include "Tested one.tf for known issues, found"
```
We are trying to make the smoke tests less explicit and not assert against a specific number of issues:
every time the experimental command runs, the latest policies are downloaded. If in the future we add more policies which detect an issue, but we still expect 0, this regression test will fail.

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -158,6 +158,23 @@ Describe "Snyk iac test --experimental command"
...
+
+      # Second File
+      The output should include "Testing root.tf..."
+      The output should include "Infrastructure as code issues:"
+      The output should include "Tested root.tf for known issues, found 0 issues"
```
ipapast, line 173:
```suggestion
      The output should include "Tested root.tf for known issues, found"
```
Same as above :)

src/cli/commands/test/iac-local-execution/file-loader.ts
```diff
@@ -1,18 +1,28 @@
...
 import { IacFileTypes } from '../../../../lib/iac/constants';
 import { isLocalFolder } from '../../../../lib/detect';
 import { CustomError } from '../../../../lib/errors';
 
 const DEFAULT_ENCODING = 'utf-8';
 
-export async function loadFiles(pathToScan: string): Promise<IacFileData[]> {
+export async function loadFiles(
+  pathToScan: string,
+  options: IaCTestFlags,
```
rontalx, line 18: nit:
maybe default to an empty object?
then in the tests you don't have to pass all those empty objects.

ipapast (approved):
Looks great, I also tested it locally with depths of 2,3,4. 🥳 
Only left a very small comment about the smoke tests, but approving ✅

test/jest/unit/iac-unit-tests/file-loader.spec.ts
```diff
@@ -86,13 +92,65 @@ describe('loadFiles', () => {
...
+          const loadedFiles = await loadFiles(level1Directory, {
+            detectionDepth: 2,
+          });
+          expect(loadedFiles).toEqual([level2FileStub]);
+        });
+      });
+
+      describe('with detectionDepth 3', () => {
+        it('returns the files at level 2 and level 3', async () => {
+          const loadedFiles = await loadFiles(level1Directory);
```
ekbsnyk, line 134: It doesn't look like detectionDepth is specified here or in the next test?

ekbsnyk (approved):

PR #1805, opened 2021-04-09
Support sarif output for container test without a `--file` parameter set

maxjeffos:

#### What does this PR do?
Allows you to run `snyk container test --sarif` without specifying `--file=`.
Now, when not specifying the `--file` we will not include the `locations` field in the results in the SARIF output. Note that `locations` is not a required field if `result`: 
- https://github.com/microsoft/sarif-tutorials/blob/main/docs/2-Basics.md#locations
- https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10540999

Try running `snyk container test --sarif` with and without a `--file=`.

We want to be able to not require the `--file` in the Snyk GitHub Action which uses Snyk's Docker images.
This other PR (https://github.com/snyk/snyk-images/pull/10) would become mergeable after change.
This makes `snyk container test --sarif` inline with `iac` and the default CLI mode (i.e. open source).

### DIFF
--- a/src/cli/commands/test/container-sarif-output.ts
+++ b/src/cli/commands/test/container-sarif-output.ts
@@ -73,24 +73,34 @@ export function getResults(testResult): sarif.Result[] {
     return results;
   }
   testResult.vulnerabilities.forEach((vuln) => {
-    results.push({
-      ruleId: vuln.id,
-      message: {
-        text: `This file introduces a vulnerable ${vuln.packageName} package with a ${vuln.severity} severity vulnerability.`,
-      },
-      locations: [
-        {
-          physicalLocation: {
-            artifactLocation: {
-              uri: testResult.displayTargetFile,
-            },
-            region: {
-              startLine: vuln.lineNumber || 1,
-            },
-          },
-        },
-      ],
-    });
+    results.push(getSarifResult(vuln, testResult.displayTargetFile));
   });
   return results;
+}
+
+export function getSarifResult(
+  vuln,
+  targetFile: string | undefined,
+): sarif.Result {
+  const result: sarif.Result = {
+    ruleId: vuln.id,
+    message: {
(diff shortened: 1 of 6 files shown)


test/acceptance/cli-test/cli-test.docker.spec.ts
```diff
@@ -735,7 +735,10 @@ export const DockerTests: AcceptanceTests = {
     '`test --docker --file=Dockerfile --sarif `': (params, utils) => async (
       t,
     ) => {
-      const testableObject = await testSarif(t, utils, params, { sarif: true });
+      const testableObject = await testSarif(t, utils, params, {
+        sarif: true,
+        file: 'Dockerfile',
```
maxjeffos, line 740: This is to make the what's being tested align with the description of this test - specifically that it should pass the `--file=Dockerfil`. The test wasn't quite correct before.

test/acceptance/fixtures/docker/sarif-container-result.json
```diff
@@ -33,7 +33,9 @@
           "locations": [
             {
               "physicalLocation": {
-                "artifactLocation": {},
+                "artifactLocation": {
+                  "uri": "Dockerfile"
```
maxjeffos, line 37: Again, this is to make the test align with its description.

src/cli/commands/test/container-sarif-output.ts
```diff
@@ -73,24 +73,34 @@ export function getResults(testResult): sarif.Result[] {
...
-        },
-      ],
-    });
+    results.push(getSarifResult(vuln, testResult.displayTargetFile));
   });
   return results;
 }
+
+export function getSarifResult(
+  vuln,
```
Avishagp, line 82: Can you add the type of `vuln`?

PR #1818, opened 2021-04-14
fix: snyk code respecting HTTPS_PROXY env vars

Spoor2709:
#### What does this PR do?
This introduces the Proxy env functionality in request.ts to deeproxy requests.


src/lib/plugins/sast/analysis.ts
```diff
@@ -21,15 +23,33 @@ export async function getCodeAnalysisAndParseResults(
 }
 
 async function getCodeAnalysis(root: string, options: Options): Promise<Log> {
-  const baseURL = config.CODE_CLIENT_PROXY_URL;
+  let baseURL = config.CODE_CLIENT_PROXY_URL;
```
miiila, line 26: [stepsize] this is the duplication of request.ts, we should find a way how to unify those

miiila (approved):

PR #1820, opened 2021-04-15
feat(@snyk/fix): issues breakdown by severity

lili2311:

#### What does this PR do?
- Show a summary of total issues + breakdown by severity
- Count how many are fixable via patch or upgrade from
remediation data for now as only projects with remediation data are supported.

#### Screenshots
CleanShot 2021-04-15 at 09 58 56@2x

#### Additional
Note `Fixed` issues count is coming in a next PR

### DIFF
--- a/packages/snyk-fix/src/lib/issues/fixable-issues.ts
+++ b/packages/snyk-fix/src/lib/issues/fixable-issues.ts
@@ -1,0 +1,45 @@
+import { DependencyPins, DependencyUpdates, TestResult } from '../../types';
+
+export function hasFixableIssues(
+  results: TestResult[],
+): {
+  hasFixes: boolean;
+  count: number;
+} {
+  let hasFixes = false;
+  let count = 0;
+  for (const result of Object.values(results)) {
+    const { remediation } = result;
+    if (remediation) {
+      const { upgrade, pin, patch } = remediation;
+      const upgrades = Object.keys(upgrade);
+      const pins = Object.keys(pin);
+      if (pins.length || upgrades.length) {
+        hasFixes = true;
+        // pins & upgrades are mutually exclusive
+        count += getUpgradableIssues(pins.length ? pin : upgrade);
+      }
+      const patches = Object.keys(patch);
+      if (patches.length) {
+        hasFixes = true;
+        count += patches.length;
+      }
+    }
+  }
+
+  return {
+    hasFixes,
+    count,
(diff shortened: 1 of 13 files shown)


packages/snyk-fix/src/lib/issues/fixable-issues.ts
```diff
@@ -0,0 +1,44 @@
...
+  results: TestResult[],
+): {
+  hasFixes: boolean;
+  count: number;
+} {
+  let hasFixes = false;
+  let count = 0;
+  for (const result of Object.values(results)) {
+    const { remediation } = result;
+    if (remediation) {
```
lili2311, line 13: using remediation data for now for the counts, when we can send new Issues in full we can migrate to use that.

packages/snyk-fix/src/lib/issues/fixable-issues.ts
```diff
@@ -0,0 +1,44 @@
...
+      const { upgrade, pin, patch } = remediation;
+      const upgrades = Object.keys(upgrade);
+      const pins = Object.keys(pin);
+      if (pins.length || upgrades.length) {
+        hasFixes = true;
+        // pins & upgrades are mutually exclusive
+        count += getUpgradableIssues(pins.length ? pin : upgrade);
+      }
+      if (Object.keys(patch).length) {
```
aarlaud, line 22: shouldn't we do the same as for pins and upgrades here? const patches = Object.keys(patch) etc so it's consistent?

packages/snyk-fix/src/types.ts
```diff
@@ -56,8 +56,8 @@ interface UpgradePath {
 }
 
 export interface FixInfo {
-  upgradePaths: UpgradePath[];
-  isPatchable: boolean;
+  upgradePaths?: UpgradePath[];
+  isPatchable?: boolean;
```
aarlaud, line 60: do we have isPinnable to?

aarlaud (approved):

packages/snyk-fix/src/types.ts
lili2311: it is not yet there, see slack

packages/snyk-fix/src/lib/issues/fixable-issues.ts
lili2311: good catch

PR #1821, opened 2021-04-15
fix: supports using --project-name flag with hex umbrella projects

JamesPatrickGill:

#### What does this PR do?

Elixir projects can be `umbrella` projects. These projects can have multiple manifests and dependency declarations so each manifest is scanned individually. 

However if the `--project-name` is provided when running `monitor` each of the sub projects get given the same name so overwrite each other in the ui. This adds a check to use a provided projectName from the scanned projects instead

### DIFF
--- a/test/jest/unit/lib/monitor/utils.spec.ts
+++ b/test/jest/unit/lib/monitor/utils.spec.ts
@@ -1,0 +1,29 @@
+import { getProjectName } from '../../../../../src/lib/monitor/utils';
+import { MonitorMeta } from '../../../../../src/lib/types';
+import { ScannedProject } from '@snyk/cli-interface/legacy/common';
+
+describe('utils', () => {
+  describe('getProjectName', () => {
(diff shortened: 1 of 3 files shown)


admons (approved):

test/jest/unit/lib/monitor/utils.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      const mockScannedProject = {
+        meta: { projectName: 'newPackageName' },
+      } as ScannedProject;
+
+      const mockMeta = { 'project-name': 'newPackageName' } as MonitorMeta;
+
+      expect(getProjectName(mockScannedProject, mockMeta)).toBe(
+        'newPackageName',
+      );
+    });
```
Avishagp, line 17: Can you please add a newline here for better visibility?

Avishagp (approved):

PR #1825, opened 2021-04-16
Feat: Support custom severities from Org IaC Settings

ipapast:

#### What does this PR do?
- This PR fetches organization settings data related to IaC, including customPolicies (custom severities) from an external endpoint. These are needed to be used for updating the local violatedPolicies with any custom severities before formatting the results to the user.
- We will also need the metadata for printing other useful info, such as organization name, etc  in a different PR.

There are 2 files of interest:
-  `src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts` to fetch the settings
- `src/cli/commands/test/iac-local-execution/apply-custom-severities.ts` to call the GET endpoint, update the existing severities with new ones
- Check that also a severity of `'none'` now results to the file being filtered out of the results (this happens in results-formatter.ts).

- Run npm run build
- Override the severity (via the UI or the local db in the Orgs table) of a rule. Here we are going to provide the example of `test/fixtures/iac/terraform/sg_open_ssh.tf`, the violated policy is 'SNYK-CC-TF-1'. The default severity was set to medium.
- Change it first to 'low' or 'high'
- Run `node ./dist/cli iac test test/fixtures/iac/terraform/sg_open_ssh.tf --experimental` and check that the issue is now shown as your updated severity
- Now change it to 'none'
- - Run `node ./dist/cli iac test test/fixtures/iac/terraform/sg_open_ssh.tf --experimental` and check that there are now 0 issues coming back as a result.

[CC-770]

#### Screenshots
Before, with the default customPolicy:
image

Updating it with 'high' severity:
image

After it with 'none' severity:
image

#### Additional questions

[CC-770]: https://snyksec.atlassian.net/browse/CC-770

### DIFF
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -13,6 +13,7 @@ import { cleanLocalCache, initLocalCache } from './local-cache';
 import { addIacAnalytics } from './analytics';
 import { TestResult } from '../../../../lib/snyk-test/legacy';
 import { IacFileInDirectory } from '../../../../lib/types';
+import { applyCustomSeverities } from './org-settings/apply-custom-severities';
 
 // this method executes the local processing engine and then formats the results to adapt with the CLI output.
 // the current version is dependent on files to be present locally which are not part of the source code.
@@ -30,7 +31,11 @@ export async function test(
   const filesToParse = await loadFiles(pathToScan, options);
   const { parsedFiles, failedFiles } = await parseFiles(filesToParse);
   const scannedFiles = await scanFiles(parsedFiles);
-  const formattedResults = formatScanResults(scannedFiles, options);
+  const resultsWithCustomSeverities = await applyCustomSeverities(scannedFiles);
+  const formattedResults = formatScanResults(
+    resultsWithCustomSeverities,
+    options,
+  );
   addIacAnalytics(formattedResults);
   cleanLocalCache();
 
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -104,21 +104,25 @@ function groupMultiDocResults(
   return Object.values(groupedData);
 }
 
-function filterPoliciesBySeverity(
+export function filterPoliciesBySeverity(
   violatedPolicies: PolicyMetadata[],
   severityThreshold?: SEVERITY,
 ): PolicyMetadata[] {
   if (!severityThreshold || severityThreshold === SEVERITY.LOW) {
-    return violatedPolicies;
+    return violatedPolicies.filter((violatedPolicy) => {
+      return violatedPolicy.severity !== 'none';
+    });
   }
 
   const severitiesToInclude = SEVERITIES.slice(
     SEVERITIES.indexOf(severityThreshold),
   );
-
-  return violatedPolicies.filter((policy) =>
-    severitiesToInclude.includes(policy.severity),
-  );
+  return violatedPolicies.filter((policy) => {
+    return (
+      policy.severity !== 'none' &&
+      severitiesToInclude.includes(policy.severity)
+    );
+  });
 }
 
 export class FailedToFormatResults extends CustomError {
--- a/src/cli/commands/test/iac-local-execution/types.ts
+++ b/src/cli/commands/test/iac-local-execution/types.ts
@@ -59,6 +59,17 @@ export type FormattedResult = {
   projectName: string;
 };
 
+export interface IacOrgSettings {
+  meta: TestMeta;
+  customPolicies: Record<string, { severity?: string }>;
+}
+export interface TestMeta {
+  isPrivate: boolean;
+  isLicensesEnabled: boolean;
+  org: string;
+  ignoreSettings?: IgnoreSettings | null;
+}
+
 export interface OpaWasmInstance {
   evaluate: (data: Record<string, any>) => { results: PolicyMetadata[] };
   setData: (data: Record<string, any>) => void;
@@ -81,7 +92,7 @@ export interface PolicyMetadata {
   title: string;
   // Legacy field, still included in WASM eval output, but not in use.
   description: string;
-  severity: SEVERITY;
+  severity: SEVERITY | 'none'; // the 'null' value can be provided by the backend
   msg: string;
   policyEngineType: 'opa';
   issue: string;
@@ -208,4 +219,7 @@ export enum IaCErrorCodes {
   // results-formatter errors
   FailedToFormatResults = 1070,
   FailedToExtractLineNumberError = 1071,
+
+  // get-iac-org-settings errors
+  FailedToGetIacOrgSettingsError = 1080,
 }
--- a/src/cli/commands/test/iac-local-execution/org-settings/apply-custom-severities.ts
+++ b/src/cli/commands/test/iac-local-execution/org-settings/apply-custom-severities.ts
@@ -1,0 +1,27 @@
+import { IacFileScanResult } from '../types';
+import { getIacOrgSettings } from './get-iac-org-settings';
+import _ = require('lodash');
+import { SEVERITY } from '../../../../../lib/snyk-test/common';
+
+export async function applyCustomSeverities(
+  scannedFiles: IacFileScanResult[],
+): Promise<IacFileScanResult[]> {
+  const iacOrgSettings = await getIacOrgSettings();
+  const customPolicies = iacOrgSettings.customPolicies;
+
+  if (Object.keys(customPolicies).length > 0) {
+    return scannedFiles.map((file) => {
+      const updatedScannedFiles = _.cloneDeep(file);
+      updatedScannedFiles.violatedPolicies.forEach((existingPolicy) => {
+        const customPolicyForPublicID = customPolicies[existingPolicy.publicId];
+        if (customPolicyForPublicID) {
+          existingPolicy.severity =
+            (customPolicyForPublicID.severity as SEVERITY) ??
+            existingPolicy.severity;
+        }
+      });
+      return updatedScannedFiles;
+    });
+  }
+  return scannedFiles;
+}
--- a/src/cli/commands/test/iac-local-execution/org-settings/get-iac-org-settings.ts
+++ b/src/cli/commands/test/iac-local-execution/org-settings/get-iac-org-settings.ts
@@ -1,0 +1,40 @@
+import { IaCErrorCodes, IacOrgSettings } from '../types';
+import { Payload } from '../../../../../lib/snyk-test/types';
+import * as config from '../../../../../lib/config';
+import { isCI } from '../../../../../lib/is-ci';
+import { api } from '../../../../../lib/api-token';
+import request = require('../../../../../lib/request');
+import { CustomError } from '../../../../../lib/errors';
+
+export function getIacOrgSettings(): Promise<IacOrgSettings> {
+  const payload: Payload = {
+    method: 'get',
+    url: config.API + '/iac-org-settings',
+    json: true,
+    headers: {
+      'x-is-ci': isCI(),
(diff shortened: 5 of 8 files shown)


src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -32,6 +34,36 @@ export function formatScanResults(
   }
 }
 
+export async function getCustomPoliciesAndUpdate(
+  scannedFiles: IacFileScanResult[],
+): Promise<IacFileScanResult[]> {
+  const filteredResults: IacFileScanResult[] = [...scannedFiles];
```
ipapast, line 40: I created a copy of the original array to be more "functional" (input/output)

src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return new Promise((resolve, reject) => {
+    request(payload, (error, res) => {
+      if (error) {
+        debug('Could not retrieve custom policies, an error occurred');
+        const err = new FailedToGetIacOrgSettings();
+        analytics.add('error-code', err.code);
+        return reject(error);
+      }
+      resolve(res.body);
```
aron, line 37: Feels like we should handle the `res.statusCode` here and only resolve if we have a 2xx response and reject otherwise? That will clean up the error handling in the caller where we check for the error in the body which shouldn't have an error property.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -30,7 +30,7 @@ export async function test(
   const filesToParse = await loadFiles(pathToScan, options);
   const { parsedFiles, failedFiles } = await parseFiles(filesToParse);
   const scannedFiles = await scanFiles(parsedFiles);
-  const formattedResults = formatScanResults(scannedFiles, options);
+  const formattedResults = await formatScanResults(scannedFiles, options);
```
aron, line 33: What made you decide to do the transformation inside formatScanResults vs. doing the transform here on a new line, then passing the transformed results into `formatScannedResults()`? This would mean that the format function remains only responsible for formatting the files and not applying overrides.

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -109,16 +144,20 @@ function filterPoliciesBySeverity(
   severityThreshold?: SEVERITY,
 ): PolicyMetadata[] {
   if (!severityThreshold || severityThreshold === SEVERITY.LOW) {
-    return violatedPolicies;
+    return violatedPolicies.filter((violatedPolicy) => {
+      return violatedPolicy.severity !== ('none' as SEVERITY);
```
aron, line 148: To fix this cast I suggest we update the `PolicyMetadata` interface to define `severity` as `SEVERITY | "none"` and add a comment that the "None" value can be provided by the backend.

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -59,6 +59,17 @@ export type FormattedResult = {
...
 };
 
+export interface IacOrgSettings {
+  meta: TestMeta;
+  customPolicies: Record<string, string>;
+}
+export interface TestMeta {
+  isPrivate: boolean;
+  isLicensesEnabled: boolean;
+  ignoreSettings: object;
```
aron, line 69: Can we fill this out completely based on our knowledge of the object from Registry.

aron (changes requested):
Looks great 🙌  some small changes requested inline to clean up the error handling and type casting. Also the loops in `getCustomPoliciesAndUpdate()` need a second look, as they're currently mutating the input and iterating over the files twice. 

Lastly, is the test file intended to be empty or was this pushed too early?

src/cli/commands/test/iac-local-execution/types.ts
ipapast: Sorry, I'm confused with this question(?). What do you mean? Registry would give us these fields only?

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -32,6 +39,34 @@ export function formatScanResults(
   }
 }
 
+export async function getCustomPoliciesAndUpdate(
+  scannedFiles: IacFileScanResult[],
+): Promise<IacFileScanResult[]> {
+  const updatedScannedFiles: IacFileScanResult[] = [...scannedFiles];
```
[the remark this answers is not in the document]
ipapast: 🤦

src/cli/commands/test/iac-local-execution/types.ts
aron: Sorry, the fields are defined here: https://github.com/snyk/registry/blob/c7a710b13447f1eb2909218a9983de50bb6c22f6/src/lib/types/index.ts#L134-L138

src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+  });
+}
+
+class FailedToGetIacOrgSettings extends CustomError {
+  constructor(message?: string) {
+    super(message || 'Failed to fetch IaC organization settings');
+    this.code = IaCErrorCodes.FailedToGetIacOrgSettings;
+    this.userMessage = ''; // Not a user facing error.
```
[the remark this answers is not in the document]
ipapast: Hm, yes debating it if we want to fail if registry fails. I'll double check with Ben 👍

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -32,6 +39,34 @@ export function formatScanResults(
...
+
+      if (customPolicyPublicId) {
+        updatedScannedFiles.forEach((fileToUpdate) => {
+          fileToUpdate.violatedPolicies.forEach((policy) => {
+            policy.severity = customPolicySeverity;
+          });
+        });
+      }
+    });
+  });
```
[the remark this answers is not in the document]
ipapast: 🤕  thank you!

src/cli/commands/test/iac-local-execution/apply-custom-severities.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getIacOrgSettings } from './get-iac-org-settings';
+import _ = require('lodash');
+
+export async function applyCustomSeverities(
+  scannedFiles: IacFileScanResult[],
+): Promise<IacFileScanResult[]> {
+  const iacOrgSettings: IacOrgSettings = await getIacOrgSettings();
+  const customPolicies: Record<string, object> = iacOrgSettings.customPolicies;
+
+  return scannedFiles.map((file) => {
```
rontalx, line 11: let's add a check to `customPolicies` object key count before that, so if it's empty we do not re-iterate on all the scanned files and clone them.

src/cli/commands/test/iac-local-execution/index.ts
ipapast: Yeah, I put it back to the index.ts now, I have split the modules according to what they do :)

ipapast:
Note: Some tests will be most probably converted to parameterized ones this morning, there is room to remove duplication.

src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts
```diff
@@ -0,0 +1,40 @@
+import { IaCErrorCodes, IacOrgSettings } from './types';
```
rontalx, line 1: minor nit:
shall we move `get-iac-org-settings.ts` and `apply-custom-severities.ts`
into a new folder? e.g. `org-settings`?

src/cli/commands/test/iac-local-execution/results-formatter.ts
ipapast: It's tricky, I can do that, but then there is another place it would break:
 https://github.com/snyk/snyk/blob/6aea05fe5b3df2adb029dfe02ffade1c249b767c/src/cli/commands/test/iac-local-execution/results-formatter.ts#L117-L125
On the `severitiesToInclude.includes(policy.severity)` we would get the warning:
 `TS2345: Argument of type 'SEVERITY | "none"' is not assignable to parameter of type 'SEVERITY'.   Type '"none"' is not assignable to type 'SEVERITY'.` so we would still need to cast it there.
 _Trying to see if there is a better way..._

src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+  });
+}
+
+export class FailedToGetIacOrgSettingsError extends CustomError {
+  constructor(message?: string) {
+    super(message || 'Failed to fetch IaC organization settings');
+    this.code = IaCErrorCodes.FailedToGetIacOrgSettingsError;
+    this.userMessage =
+      'We failed to fetch IaC organization settings, including custom severity overrides. Please run the command again with the `-d` flag and contact support@snyk.io with the contents of the output.';
```
rontalx, line 38: nit:
```suggestion
      'We failed to fetch your organization settings, including custom severity overrides for infrastructure-as-code policies. Please run the command again with the `-d` flag and contact support@snyk.io with the contents of the output.';
```

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -208,4 +219,7 @@ export enum IaCErrorCodes {
   // results-formatter errors
   FailedToFormatResults = 1070,
   FailedToExtractLineNumberError = 1071,
+
+  // get-iac-org-settings errors
+  FailedToGetIacOrgSettingsError = 1080,
```
ipapast, line 224: @aron 2 questions on this:
- do you think this is the best place to add this? Not sure if it actually needs a new "group" on its own

aron (approved):
👏 Very nice work. Left a couple of responses to your questions and a suggestion on how to better refine the typing, nothing blocking.

src/cli/commands/test/iac-local-execution/results-formatter.ts
ipapast: TIL!

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -59,6 +59,17 @@ export type FormattedResult = {
   projectName: string;
 };
 
+export interface IacOrgSettings {
+  meta: TestMeta;
+  customPolicies: Record<string, object>;
```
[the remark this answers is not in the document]
ipapast: Does the severity need to be optional? I think it's fine if it's mandatory, the backend would always return customPolicies with a severity key, if there are any. I'm asking on the optional because that would cause other TS 'undefined' type errors 🤦

src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts
ipapast: Can do

src/cli/commands/test/iac-local-execution/types.ts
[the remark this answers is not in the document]
aron: Yes, fine for now, I was thinking a little ahead where this `customPolicies` is really `Partial<Policy>` where the user could theoretically override any number of fields. In this case severity might not be present.

> I'm asking on the optional because that would cause other TS 'undefined' type errors

Yes, this would require a presence check before using the value.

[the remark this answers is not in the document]
ipapast: Right, I left it to optional then, I wasn't thinking that far ahead. 👍

src/cli/commands/test/iac-local-execution/get-iac-org-settings.ts
ipapast: Done done 🥳

PR #1834, opened 2021-04-16
docs: remove help note for tf-plan.json restriction [CC-793]

teodora-sandu:

#### What does this PR do?
This PR updates the `snyk --help` command to remove the note about the Terraform plan file name restriction.

Have a look at the markdown document for any spelling mistakes.

Run `npm run generate-help` and have a look at the sections that changed.

The code changes were delivered under https://github.com/snyk/snyk/pull/1822.

https://snyksec.atlassian.net/browse/CC-793

#### Screenshots
Screenshot 2021-04-16 at 16 22 15

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/file-parser.ts
@@ -57,10 +57,12 @@ function generateFailedParsedFile(
   };
 }
 
-function parseIacFileData(fileData: IacFileData): any[] {
+function parseYAMLOrJSONFileData(fileData: IacFileData): any[] {
   let yamlDocuments;
 
   try {
+    // the YAML library can parse both YAML and JSON content, as well as content with singe/multiple YAMLs
+    // by using this library we don't have to disambiguate between these different contents ourselves
     yamlDocuments = YAML.safeLoadAll(fileData.fileContent);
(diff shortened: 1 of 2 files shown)


ipapast (approved):

src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -61,6 +61,7 @@ function parseIacFileData(fileData: IacFileData): any[] {
   let yamlDocuments;
 
   try {
+    // the YAML library can parse both YAML and JSON content
```
aron, line 64: Nice! Two last points now I read this again.
1. Worth adding here that the reason we use YAML is it means we don't have to disambiguate between, JSON, single YAML doc and multiple YAML doc as the parser does it for us?
2. In the spirit of the names of the other functions and to be explicit perhaps this should be called `parseYAMLOrJSONFileData()` to be clear that this will not support a terraform file for example.

teodora-sandu: Is it a bit clearer now? I've tried to address both points, hopefully it makes sense 😃

aron (approved):
🙌  Looks amazing, thanks for iterating on the wording!

PR #1835, opened 2021-04-16
Feat: pipfile handler check & share common functionality

lili2311:

#### What does this PR do?
- drop micromatch it was under used and not needed, the check is much simpler here as CLI already had done the work of classifying the project type as python.
- add a new check with tests for a project that has a `Pipfile` manifest  + tests (nothing is using it yet)

### DIFF
--- a/packages/snyk-fix/src/plugins/python/get-handler-type.ts
+++ b/packages/snyk-fix/src/plugins/python/get-handler-type.ts
@@ -1,5 +1,3 @@
-import * as micromatch from 'micromatch';
-
 import { EntityToFix } from '../../types';
 import { SUPPORTED_HANDLER_TYPES } from './supported-handler-types';
 
@@ -18,11 +16,5 @@ export function getHandlerType(
 }
 
 export function isRequirementsTxtManifest(targetFile: string): boolean {
-  return micromatch.isMatch(
-    targetFile,
-    // micromatch needs **/* to match filenames that may include folders
-    ['*.txt'].map(
-      (f) => '**/' + f,
-    ),
-  );
+  return targetFile.endsWith('.txt');
 }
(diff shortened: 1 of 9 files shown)


packages/snyk-fix/src/plugins/python/get-handler-type.ts
```diff
@@ -18,11 +16,9 @@ export function getHandlerType(
...
 
 export function isRequirementsTxtManifest(targetFile: string): boolean {
-  return micromatch.isMatch(
-    targetFile,
-    // micromatch needs **/* to match filenames that may include folders
-    ['*.txt'].map(
-      (f) => '**/' + f,
-    ),
-  );
+  return targetFile.endsWith('.txt');
```
aarlaud, line 19: does that handle the folder situation mentioned on the left ?

packages/snyk-fix/src/plugins/python/handlers/valdidate-required-data.ts
```diff
@@ -0,0 +1,26 @@
+import { MissingRemediationDataError } from '../../../lib/errors/missing-remediation-data';
```
aarlaud, line 1: typo in filename

aarlaud (approved):

packages/snyk-fix/src/plugins/python/handlers/valdidate-required-data.ts
lili2311: which one is the typo? Not quite sure

packages/snyk-fix/src/plugins/python/get-handler-type.ts
lili2311: adding a test for that use-case, will be supported yep!

PR #1837, opened 2021-04-18
test: add dedup key to enable alert grouping

Avishagp:

#### What does this PR do?
https://developer.pagerduty.com/docs/events-api-v2/trigger-events/

> Submitting subsequent events for the same dedup_key will result in those events being applied to an open alert matching that dedup_key. Once the alert is resolved, any further events with the same dedup_key will create a new alert (for trigger events) or be dropped (for acknowledge and resolve events).

This change groups together repeating trigger events caused by smoke tests failing, meaning that PagerDuty won't alert us again and again if the first alert is still open.


aviadhahami (approved):
lgtm [boost]

PR #1845, opened 2021-04-20
Chore/upgrade sast testing

Spoor2709:

#### What does this PR do?
This introduces smoke tests + a fix for sarif output and additions to the original sast tests


src/lib/plugins/sast/index.ts
```diff
@@ -33,7 +33,14 @@ export const codePlugin: EcosystemPlugin = {
         path,
         options,
       );
+
+      const numOfIssues = sarifTypedResult.runs?.[0].results?.length || 0;
+      analytics.add('sast-issues-found', numOfIssues);
+
       if (options.sarif || options.json) {
+        if (numOfIssues > 0) {
+          hasIssues(jsonStringifyLargeObject(sarifTypedResult));
```
ArturSnyk, line 42: do we need to wrap the sarifTypedResult with  `jsonStringifyLargeObject` before sending to `hasIssues`.
I'm not sure if the "error" flow we're handling when sending an error (for exit code reasons) already handles it.
But if not, we should also handle the `sarifTypedResult` payload where we have the other `hasIssues` call here

Spoor2709: Yes I belive we do as this then throws the issue and does not get to the return statement outputting the actual sarif 😃

ArturSnyk (approved):

Avishagp (approved):

PR #1849, opened 2021-04-20
fix: hcl-to-json parsing with gopher js bundle

rontalx:
#### What does this PR do?
This PR adds a JS bundle which was created with [GopherJS](https://github.com/gopherjs/gopherjs) porting the following HCL-2-JSON Parser to javascript: https://github.com/tmccombs/hcl2json 
This is done in order to align our new experimental CLI flow with the same parsing mechanism we use in our SCM scanning.
It was tested with [a test-harness we wrote to validate that it's consistent with the SCM scanning](https://github.com/snyk/cloud-config-ops-runbooks/tree/main/test-harness) for all use-case fixture we have policies for.
Results showed that it is now 100% consistent with all use-cases we have rules against in https://github.com/snyk/cloud-config-opa-policies, besides two AZURE policies which we've aware of the issue and have opened a bug to fix these separately.
Test Harness results:
```
========= Issue detected while testing: /Users/rontal/dev/cloud-config-opa-policies/schemas/terraform/azure/SNYK_CC_AZURE_243/azurerm_sas_token_expire.tf ========= 

 CLI failed but CCPE did not ❌ 
 CLI failed with error: Failed to execute policy engine  

========= Issue detected while testing: /Users/rontal/dev/cloud-config-opa-policies/schemas/terraform/azure/SNYK_CC_AZURE_244/azurerm_sas_https_only.tf ========= 

 CLI failed but CCPE did not ❌ 
 CLI failed with error: Failed to execute policy engine  
```
Ticket opened to fix these:
https://snyksec.atlassian.net/browse/CC-804

This means that our CLI & SCM scanning results should now be 100% consistent and the same, which was a top priority for us in the new experimental flow.

Here are some comparisons we did for various approaches.
The GopherJS bundle is approximately 4MB, but after being GZipped it would be ~700K, therefore should be neglible in terms of installation time impact.
image

This is fully in-sync with Hammer team and we're collaborating to make the right decisions for the CLI:
https://snyk.slack.com/archives/C0127HWU0E7/p1618324426163000

The switch from the previous package used should be seamless and un-noticeable.

This is a long ticket we've worked on for a while and tested various approaches, more info can be found in:
https://www.notion.so/snyk/HCL-to-JSON-Parsing-IaC-CLI-ccc79634dd784287bad5472bce92a674
https://docs.google.com/document/d/1VX-ZQqBjDyrh-z6fCpVG1B0nAs1iN8WVAi--TYQaavo/edit#heading=h.ud812q5pdlu4

https://snyksec.atlassian.net/browse/CC-796


.prettierignore
```diff
@@ -1,5 +1,6 @@
 test/fixtures/**/*
 packages/**/test/fixtures/**/*
+src/cli/commands/test/iac-local-execution/parsers/hcl-to-json/parser.js
```
rontalx, line 3: avoiding from the minified gopherjs artifact to get automatically formatted by prettier as it expands it from 4MB's to 10MB's (and also obviously we don't need it formatted)

rontalx: disabled also linting to the artefact, in the first line of the artefact, otherwise lints will fail.
*we don't have an `eslintignore` file.

Avishagp (approved):

aron (approved):

PR #1850, opened 2021-04-20
feat: include org details in JSON file output [CC-772]

teodora-sandu:

#### What does this PR do?
This PR uses the org details we retrieve from registry to populate the results for the JSON file output flag, as well as for the organisation name in the CLI results.

The code flow can be read in the order of the changed files. This PR moves the `getIacOrgSettings` function call to the main index file, where we compute the org settings and then pass the contents of `customPolicies` field to the `applyCustomSeverities` call and then the contents of the `meta` field to `formatScanResults` in order to include the missing fields into the result.

2. Download `deployment_singleissue.yaml` from the Jira ticket
2. Run `node ./dist/cli iac test --experimental deployment_singleissue.yaml --json-file-output=experimental.json`
3. `node ./dist/cli iac test deployment_singleissue.yaml--json-file-output=original.json`
4. Compare using http://jsondiff.com/

https://snyksec.atlassian.net/browse/CC-772

#### Screenshots
Screenshot 2021-04-20 at 14 17 37

### DIFF
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -5,6 +5,7 @@ import {
   IacFileScanResult,
   IaCTestFlags,
   PolicyMetadata,
+  TestMeta,
 } from './types';
 import * as path from 'path';
 import { SEVERITY } from '../../../../lib/snyk-test/common';
@@ -20,12 +21,13 @@ const SEVERITIES = [SEVERITY.LOW, SEVERITY.MEDIUM, SEVERITY.HIGH];
 export function formatScanResults(
   scanResults: IacFileScanResult[],
   options: IaCTestFlags,
+  meta: TestMeta,
 ): FormattedResult[] {
   try {
     // Relevant only for multi-doc yaml files
     const scannedResultsGroupedByDocId = groupMultiDocResults(scanResults);
     return scannedResultsGroupedByDocId.map((iacScanResult) =>
-      formatScanResult(iacScanResult, options.severityThreshold),
+      formatScanResult(iacScanResult, meta, options.severityThreshold),
     );
   } catch (e) {
     throw new FailedToFormatResults();
@@ -39,6 +41,7 @@ const engineTypeToProjectType = {
 
 function formatScanResult(
   scanResult: IacFileScanResult,
+  meta: TestMeta,
   severityThreshold?: SEVERITY,
 ): FormattedResult {
   const formattedIssues = scanResult.violatedPolicies.map((policy) => {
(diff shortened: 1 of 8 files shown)


src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -74,15 +79,23 @@ function formatScanResult(
       ),
       projectType: projectTypeByFileType[scanResult.fileType],
     },
-    isPrivate: true,
-    packageManager: engineTypeToProjectType[scanResult.engineType],
-    targetFile: scanResult.filePath,
-    targetFilePath,
+    meta: {
```
teodora-sandu, line 82: I've only added `meta`, `filesystemPolicy`, `policy`, and `org`. I've just changed the order to match perfectly with the old flow. Same goes for the `types.ts` file below

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -74,15 +79,23 @@ function formatScanResult(
...
+      ...meta,
+      projectId: '', // we do not have a project at this stage
+      policy,
+    },
+    filesystemPolicy: !!policy,
```
rontalx, line 87: why is the const definition needed above instead of passing in `policy` & `filesystemPolicy` concrete explicit values?

rontalx (approved):
looking great!

src/cli/commands/test/iac-local-execution/results-formatter.ts
teodora-sandu: It was more for clarity in the future in case `policy` is ever not-empty, since `filesystemPolicy` changes based on its value. I can change this to the simpler explicit values if you think it's more confusing now?

rontalx: it was **slightly** confusing at first look, as also the term `policy` is over-used in Snyk.
I'd suggest we pass explicit values just for clarity.

teodora-sandu: Done 😃

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -74,15 +78,23 @@ function formatScanResult(
...
       projectType: projectTypeByFileType[scanResult.fileType],
     },
-    isPrivate: true,
-    packageManager: engineTypeToProjectType[scanResult.engineType],
-    targetFile: scanResult.filePath,
-    targetFilePath,
+    meta: {
+      ...meta,
+      projectId: '', // we do not have a project at this stage
+      policy: '',
```
rontalx, line 84: minor nit: 
maybe we should add a small comment on what this field is and why it is empty? (same as you did with the others?)
also here:
https://github.com/snyk/snyk/pull/1850/files#diff-2e4fd6421898cc5c5947d03530196bee66469cd25276fcdfcc0ba4381268b2c9R94

PR #1852, opened 2021-04-21
fix: IaC run with no vulnerabilities and --json flag should be successful [CC-805]

teodora-sandu:

#### What does this PR do?
This PR makes sure that when the `snyk iac test --experimental` command is run with the extra `--json` flag on a file with no vulnerabilities, the command returns successfully.

The first commit contains a new smoke test that covers this edge case. It modifies an existing test by adding the `--json` flag. As we can see from Circle, this new smoke test fails because the command does not return successfully. However, the original smoke test without the `--json` flag returns successfully.

The new IaC experimental flow returns the valid scannings (including vulnerabilities) separately from failed scannings (parse errors or any other errors to return to the user). So, the code needs to check first that there are no vulnerabilities (since there is an if statement a few lines below that deal with that case) and that there are no proper failures. If there aren't any of those, then we return successfully. 

2. Run the command on a file with no vulnerabilities (e.g. https://github.com/snyk/cloud-config-opa-policies/blob/master/schemas/terraform/gcp/SNYK_CC_GCP_91/provider.tf): `node ./dist/cli iac test --experimental path/to/cloud-config-opa-policies/schemas/terraform/gcp/SNYK_CC_GCP_91/provider.tf --json`
3. Notice that the command returns **and** says it has finished successfully
4. Run the existing CLI on the same file:`snyk iac test --experimental path/to/cloud-config-opa-policies/schemas/terraform/gcp/SNYK_CC_GCP_91/provider.tf --json`
5. Notice that the command returns **but** says it has finished with an error

This problem surfaced when looking at our analytics, where we noticed some errors without any error codes and with error messages that looked like valid scanning results. For more information read the ticket.

https://snyksec.atlassian.net/browse/CC-805

#### Screenshots
Before (notice red warning):
Screenshot 2021-04-21 at 10 56 41

After (notice green tick):
Screenshot 2021-04-21 at 10 56 26

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -171,8 +171,17 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
   } = extractDataToSendFromResults(results, jsonData, options);
 
   if (options.json || options.sarif) {
-    // if all results are ok (.ok == true) then return the json
-    if (errorMappedResults.every((res) => res.ok)) {
+    // the new experimental IaC flow does not have the ok field because it returns vulnerabilities and errors separately
+    // the legacy IaC flow may return errors that get mapped into errorMappedResults, but that flow will be removed soon
+    const successfulIacScanning =
+      options.iac &&
+      !foundVulnerabilities &&
+      (!iacScanFailures || iacScanFailures.length === 0);
+
+    // if running iac and it was successful, or
+    // if all results are ok (.ok == true)
+    // then return the json
+    if (successfulIacScanning || errorMappedResults.every((res) => res.ok)) {
       return TestCommandResult.createJsonTestCommandResult(
         stringifiedData,
         stringifiedJsonData,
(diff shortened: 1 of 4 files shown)


src/cli/commands/test/index.ts
```diff
@@ -171,8 +171,16 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
   } = extractDataToSendFromResults(results, jsonData, options);
 
   if (options.json || options.sarif) {
-    // if all results are ok (.ok == true) then return the json
-    if (errorMappedResults.every((res) => res.ok)) {
+    // the new experimental IaC flow does not have the ok field because it returns vulnerabilities and errors separately
```
teodora-sandu, line 174: @aron You mentioned adding a comment about how this can be removed in the future but after some more thinking I couldn't figure out why so I left that out for now 😅

Avishagp (approved):

src/cli/commands/test/index.ts
aron: Oh right, so I think this was because our results do not have an `ok` property so this will never be true for the new iac flow. And `errorMappedResults` is only ever going to (potentially) contain `Error` objects in the legacy `iac test` flow. So the comment was to say that when we remove the legacy flow we need to remove anything IaC related from `errorMappedResults` and it's generation.

aron (approved):

PR #1856, opened 2021-04-22
chore: fix issue type

lili2311:

#### What does this PR do?
Add missing raw issue data type. `IssuesData` is a combination of raw issue data & some computed properties.

### DIFF
--- a/src/cli/commands/fix/convert-legacy-test-result-to-new.ts
+++ b/src/cli/commands/fix/convert-legacy-test-result-to-new.ts
@@ -15,11 +15,12 @@ function convertVulnerabilities(
   const issues: Issue[] = [];
 
   vulns.forEach((vuln) => {
+    // TODO: map the rest as needed
     issuesData[vuln.id] = {
       id: vuln.id,
       severity: vuln.severity,
       title: vuln.title,
-    };
+    } as any;
     issues.push({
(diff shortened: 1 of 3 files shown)


src/cli/commands/fix/convert-legacy-test-result-to-new.ts
```diff
@@ -15,11 +15,12 @@ function convertVulnerabilities(
   const issues: Issue[] = [];
 
   vulns.forEach((vuln) => {
+    // TODO: map the rest as needed
     issuesData[vuln.id] = {
       id: vuln.id,
       severity: vuln.severity,
       title: vuln.title,
-    };
+    } as any;
```
gitphill, line 23: can you use Partial<IssueData> here? (or some other Typescript magic)

src/lib/snyk-test/legacy.ts
```diff
@@ -235,11 +236,8 @@ interface Issue {
   issueId: string;
   fixInfo: FixInfo;
 }
-
 interface TestDependenciesResult {
-  issuesData: {
-    [issueId: string]: IssueData;
-  };
+  issuesData: IssuesData;
```
gitphill, line 240: i think we should leave this legacy code alone

PR #1860, opened 2021-04-22
feat: Update snyk-docker-plugin to v4.20.1

snaftaly:

#### What does this PR do?

This PR updates the plugin version to add the algorithm prefix ('sha256:') to the imageId.

Run container test with this CLI version on a docker image will create a project where the imageId of the project test snapshot has a 'sha256:' prefix.

The full imageId for docker images should contain the algorithm prefix (as can be seen when running 'docker inspect' command). The new version of the plugin adds this prefix.

https://snyksec.atlassian.net/browse/DC-1195

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "semver": "^6.0.0",
     "snyk-config": "4.0.0",
     "snyk-cpp-plugin": "2.2.1",
-    "snyk-docker-plugin": "4.19.3",
+    "snyk-docker-plugin": "4.20.1",
     "snyk-go-plugin": "1.17.0",
     "snyk-gradle-plugin": "3.14.2",
     "snyk-module": "3.1.0",


package.json
```diff
@@ -118,7 +118,7 @@
     "semver": "^6.0.0",
     "snyk-config": "4.0.0",
     "snyk-cpp-plugin": "2.2.1",
-    "snyk-docker-plugin": "4.19.3",
+    "snyk-docker-plugin": "^4.20.1",
```
jan-stehlik, line 121: If we would like to be more strict about the internal dependencies and are not fine if users install latest minor version of a package we might want to remove `^` (atm with caret we can install e.g. versions 4.32.0 if that's the latest minor version which might not be what we want)
```suggestion
    "snyk-docker-plugin": "4.20.1",
```

jan-stehlik (commented):
Hey @snaftaly , I'm new here but I'm curious - what's the reason that we do not commit `package-lock.json` file to source control but have it [gitignored](https://github.com/snyk/snyk/blob/f73a9f6a6e3c50a8702c539df10fcd2929a5a451/.gitignore#L10) instead please? Are we okay with users potentionally getting quite different dependencies versions when installing snyk locally?

maxjeffos (approved):

package.json
snaftaly: Thanks! I removed the ^

snaftaly: Regarding package-lock.json, I have to admit that I'm not sure why it's not being used in this repo.
I'm sure there is a good explanation for that, and would love to know what it is..
Maybe @Avishagp could explain this?

jan-stehlik (approved):
LGTM 🚀

package.json
Avishagp: This is done in order to have the most similar dependency tree during test runs, as it would be after installing snyk as a library. Done in this PR:
https://github.com/snyk/snyk/pull/223
If you need any further explanation let me know :)

jan-stehlik: thanks @Avishagp for explaining, so I assume we expect users to use `npm install` as opposed to `npm ci` when installing snyk cli which is fair 👍

PR #1861, opened 2021-04-26
feat: use feature flag to enable local exec [CC-790]

teodora-sandu:

#### What does this PR do?
This PR rolls out the experimental CLI by using a feature flag. We want the experimental flow to be the default if the feature flag is enabled or if the `--experimental` flag is provided when running the CLI.

- Start in `src/cli/index.ts`, where if `iac` is running we now make a call to get the org name (i.e. `getIacOrgSettings`), required for checking if the experimental feature flag is enabled. 
- We only make the extra call to registry to see if the feature flag is enabled if the `experimental` is not passed, to minimise the number of network calls we need. 
- These checks are here instead of `src/cli/commands/test/iac-test-shim.ts` because the analytics report the `args` object and we use that object in our SLO dashboard.

The acceptance tests had to be modified with a new API call added to the fake server and some extra, temporary checks for testing the legacy flow. 

2. Start [registry](https://github.com/snyk/registry) in another terminal tab and in the tab you're running the CLI from run `export SNYK_API="http://localhost:8000/api"`
3. Authenticate now that you're pointing to a different Snyk API: `node ./dist/cli auth`
4. Run the command `node ./dist/cli iac test test/fixtures/iac/kubernetes/pod-privileged.yaml --debug`, which will run the original flow
5. Run the command `node ./dist/cli iac test test/fixtures/iac/terraform-plan/tf-plan.json --debug`, which will fail because the original flow does not support Terraform Plan scanning
6. Modify your org in the local database to have the `experimentalLocalExecIac` flag on.
7. Run the command `node ./dist/cli iac test test/fixtures/iac/kubernetes/pod-privileged.yaml --debug`, which will run the experimental flow
8. Run the command `node ./dist/cli iac test test/fixtures/iac/terraform-plan/tf-plan.json --debug`, which will work because experimental flow supports it

The flag has been added under https://github.com/snyk/registry/pull/20370 and this PR just starts checking if it's enabled to decide what execution flow to use.

https://snyksec.atlassian.net/browse/CC-790

#### Screenshots
- feature flag is enabled for org `teodora-sandu`
Screenshot 2021-04-26 at 08 24 19
Screenshot 2021-04-26 at 08 24 44

- feature flag is not enabled enabled for org `test`
Screenshot 2021-04-26 at 08 25 59
Screenshot 2021-04-26 at 08 26 09

- proof of analytics being sent correctly:
Screenshot 2021-04-26 at 12 48 18

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -244,9 +244,14 @@ async function main() {
       (args.options as unknown) as AllSupportedCliOptions,
     );
 
-    // modify args for IaC if experimental flag not provided based on feature flag
+    // IaC only: used for rolling out the experimental flow
+    // modify args if experimental flag not provided, based on feature flag
     // this can be removed once experimental becomes the default
-    if (args.options['iac'] && !args.options['experimental']) {
+    if (
+      args.options['iac'] &&
+      args.command === 'test' &&
+      !args.options['experimental']
+    ) {
       const iacOrgSettings = await getIacOrgSettings();
       const experimentalFlowEnabled = await isFeatureFlagSupportedForOrg(
         camelCase('experimental-local-exec-iac'),
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -3,6 +3,8 @@ import 'source-map-support/register';
 import * as Debug from 'debug';
 import * as pathLib from 'path';
 
+const camelCase = require('lodash.camelcase');
+
 // assert supported node runtime version
 import * as runtime from './runtime';
 // require analytics as soon as possible to start measuring execution time
@@ -38,6 +40,8 @@ import {
 } from '../lib/types';
 import { SarifFileOutputEmptyError } from '../lib/errors/empty-sarif-output-error';
 import { InvalidDetectionDepthValue } from '../lib/errors/invalid-detection-depth-value';
+import { getIacOrgSettings } from './commands/test/iac-local-execution/org-settings/get-iac-org-settings';
+import { isFeatureFlagSupportedForOrg } from '../lib/feature-flags';
 
 const debug = Debug('snyk');
 const EXIT_CODES = {
@@ -229,6 +233,7 @@ async function main() {
   checkRuntime();
 
   const args = argsLib(process.argv);
+
   let res;
   let failed = false;
   let exitCode = EXIT_CODES.ERROR;
@@ -238,6 +243,17 @@ async function main() {
     validateUnsupportedOptionCombinations(
       (args.options as unknown) as AllSupportedCliOptions,
     );
+
+    // modify args for IaC if experimental flag not provided based on feature flag
+    // this can be removed once experimental becomes the default
+    if (args.options['iac'] && !args.options['experimental']) {
+      const iacOrgSettings = await getIacOrgSettings();
+      const experimentalFlowEnabled = await isFeatureFlagSupportedForOrg(
+        camelCase('experimental-local-exec-iac'),
+        iacOrgSettings.meta.org,
+      );
+      args.options['experimental'] = !!experimentalFlowEnabled.ok;
+    }
 
     if (args.options['app-vulns'] && args.options['json']) {
       throw new UnsupportedOptionCombinationError([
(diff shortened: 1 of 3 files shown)


test/acceptance/fake-server.ts
```diff
@@ -55,7 +55,16 @@ export function fakeServer(root, apikey) {
   });
 
   server.use((req, res, next) => {
-    if (!server._nextResponse && !server._nextStatusCode) {
+    // these test don't run on the new experimental flow
+    // so once we deprecate legacy this check can be removed
+    const isExperimentalIac =
```
teodora-sandu, line 60: These checks are required because we use the `setNextResponse` function in the legacy acceptance tests, and now that we're making two extra calls (to get the org settings and to see if the feature flag is enabled), it has in-adverse side effects.

We could modify the `setNextResponse` function but it might affect the other acceptance tests. The legacy flow will be removed soon, and so will these acceptance tests. So this check is the cheapest and safest way to get these soon-to-be-deleted tests to pass.

test/acceptance/fake-server.ts
```diff
@@ -271,7 +280,10 @@ export function fakeServer(root, apikey) {
     root + '/cli-config/feature-flags/:featureFlag',
     (req, res, next) => {
       const flag = req.params.featureFlag;
-      if ((req as any).params.org === 'no-flag') {
+      if (
+        (req as any).params.org === 'no-flag' ||
+        flag === 'experimentalLocalExecIac'
```
teodora-sandu, line 285: We want this flag to be disabled for the acceptance tests, which only test the legacy flow

src/cli/index.ts
```diff
@@ -229,6 +233,18 @@ async function main() {
   checkRuntime();
 
   const args = argsLib(process.argv);
+
+  // modify args for IaC if experimental flag not provided based on feature flag
+  // this can be removed once experimental becomes the default
+  if (args.options['iac'] && !args.options['experimental']) {
```
teodora-sandu, line 239: We want to override the `experimental` flag if it's `false` and the feature flag is enabled. We do the check here because we want the analytics to pick up that this flag is now `true`

src/cli/index.ts
```diff
@@ -239,6 +244,17 @@ async function main() {
...
+    // this can be removed once experimental becomes the default
+    if (args.options['iac'] && !args.options['experimental']) {
+      const iacOrgSettings = await getIacOrgSettings();
+      const experimentalFlowEnabled = await isFeatureFlagSupportedForOrg(
+        camelCase('experimental-local-exec-iac'),
+        iacOrgSettings.meta.org,
+      );
+      args.options['experimental'] = !!experimentalFlowEnabled.ok;
+    }
+
```
rontalx, line 257: at first look I thought it isn't a good place for the change, and it's better of inside the `iac-test-shim.ts` file.
but on a 2nd look, because it's pure argument overriding logic, then this location seems correct.
I'd add another short comment at the beginning saying this is being used for rollout purposes.
+can you also add the `args.command === 'test'` to the if clause for extra explicitness?

rontalx (approved):
nice! 🎉
left one small comment.
did you try it locally with toggling the feature flag and verifying?

src/cli/index.ts
```diff
@@ -239,6 +244,22 @@ async function main() {
       (args.options as unknown) as AllSupportedCliOptions,
     );
 
+    // IaC only: used for rolling out the experimental flow
+    // modify args if experimental flag not provided, based on feature flag
+    // this can be removed once experimental becomes the default
+    if (
+      args.options['iac'] &&
+      args.options['test'] &&
```
rontalx, line 252: it's: `args.command['test']`

image

teodora-sandu:
Thanks @rontalx , I ran it locally before and after toggling the feature flag and ran it on a Kubernetes file and a Terraform Plan one (which would fail for the legacy flow). There are some screenshots of the test runs in the PR description 😄

src/cli/index.ts
teodora-sandu: 🤦  Yes .. I'll re-run the local tests before pushing the fix

Avishagp (approved):

PR #1870, opened 2021-04-29
fix: consistent command termination between --json and no --json flag

teodora-sandu:

#### What does this PR do?
This PR improves the consistency between running the experimental flow with the `--json` flag or without. We want the two commands to fail and succeed the same way, regardless of the flag's presence. This way we don't incorrectly report failed commands in our analytics.

Start in `src/lib/snyk-test/iac-test-result.ts`, where we add a new field to our IaC results, `ok`, which changes based on whether there are any vulnerabilities or not. This field fixes the original problem **and** removes the need for custom IaC checks in `src/cli/commands/test/index.ts` (added for https://snyksec.atlassian.net/browse/CC-805).

1. Build the CLI: `npm run build`
2. Run it on a directory containing a mix of valid and invalid files, with an without the `--json` flag:
    - contains one valid file with one vulnerability: `node ./dist/cli iac test test/fixtures/iac/kubernetes/ --severity-threshold=high --experimental`
    - contains one valid file with no vulnerabilities: `node ./dist/cli iac test test/fixtures/iac/no_vulnerabilities --severity-threshold=high --experimental`
3. Notice the same status code is returned.

We noticed via our [Grafana dashboard](https://grafana.dev.snyk-internal.net/d/zVy940uGz/iac-cli-experimental-dashboard?orgId=1) that we have a lot of  empty errors, more than we used to see before making the experimental flow the default in production a few days ago. 
Screenshot 2021-04-29 at 17 06 16

This was due to an edge case that we hadn't accounted for when testing the success/failure behaviour of the CLI and that some of our customers seem to run into a lot: scanning a directory with a mix of valid and invalid files, but the valid files having no vulnerabilities.

To be specific, this unexpected behaviour only happens when the `--json` flag is provided. This is the behaviour we were seeing when scanning directories with multiple files:
- without the `--json` flag:
    - everything parsed successfully, 0 issues found -> status code 0 (successful)
    - everything parsed successfully, >0 issues found -> status code 1 (failed)
    - some parsed successfully, 0 issues found -> status code 0 (successful)
    -  some parsed successfully, >0 issues found -> status code 1 (failed)

- with the `--json` flag:
    - everything parsed successfully, 0 issues found -> status code 0 (successful)
    - everything parsed successfully, >0 issues found -> status code 1 (failed)
    - some parsed successfully, 0 issues found -> status code 2 (failed)
    - some parsed successfully, >0 issues found -> status code 1 (failed)

As we can see, there was a difference in behaviour between running with and without and `--json` flag.
To highlight this, we implemented smoke tests covering those 8 cases above, and also modified our existing smoke tests for the experimental flow to specifically check the status code, since failure can look different depending on whether we find vulnerabilities or we encounter some other problem.

#### Screenshots
These correspond to the manual tests mentioned further above:
- contains one valid file with one vulnerability: 
Screenshot 2021-04-29 at 17 00 25
Screenshot 2021-04-29 at 17 00 05

- contains one valid file with no vulnerabilities: 
Screenshot 2021-04-29 at 17 00 45
Screenshot 2021-04-29 at 17 00 58

### DIFF
--- a/src/lib/snyk-test/iac-test-result.ts
+++ b/src/lib/snyk-test/iac-test-result.ts
@@ -49,6 +49,8 @@ export function mapIacTestResult(
     };
   }
 
+  const infrastructureAsCodeIssues =
+    iacTest?.result?.cloudConfigResults.map(mapIacIssue) || [];
   const {
     result: { projectType },
     ...filteredIacTest
@@ -56,8 +58,8 @@ export function mapIacTestResult(
   return {
     ...filteredIacTest,
     projectType,
-    [IAC_ISSUES_KEY]:
-      iacTest?.result?.cloudConfigResults.map(mapIacIssue) || [],
+    ok: infrastructureAsCodeIssues.length === 0,
+    [IAC_ISSUES_KEY]: infrastructureAsCodeIssues,
   };
 }
 
--- a/test/smoke/spec/iac/snyk_test_local_exec_spec.sh
+++ b/test/smoke/spec/iac/snyk_test_local_exec_spec.sh
@@ -9,7 +9,7 @@ Describe "Snyk iac test --experimental command"
   Describe "basic usage"
     It "outputs an error if the --experimental flag is mistyped"
       When run snyk iac test ../fixtures/iac/kubernetes/pod-invalid.yaml --experimentl
-      The status should be failure
+      The status should equal 2
       The output should include "Unsupported flag"
     End
   End
@@ -20,7 +20,7 @@ Describe "Snyk iac test --experimental command"
       When run snyk iac test ../fixtures/iac/file-logging -d --experimental
       # We expect the output, specifically the analytics block not to include
       # the following text from the file.
-      The status should be failure  # issues found
+      The status should equal 1 # issues found
       The output should not include "PRIVATE_FILE_CONTENT_CHECK"
       The error should not include "PRIVATE_FILE_CONTENT_CHECK"
     End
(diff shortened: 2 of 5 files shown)


src/lib/snyk-test/iac-test-result.ts
```diff
@@ -49,15 +49,17 @@ export function mapIacTestResult(
...
   const {
     result: { projectType },
     ...filteredIacTest
   } = iacTest;
   return {
     ...filteredIacTest,
     projectType,
-    [IAC_ISSUES_KEY]:
-      iacTest?.result?.cloudConfigResults.map(mapIacIssue) || [],
+    ok: infrastructureAsCodeIssues.length === 0,
```
teodora-sandu, line 61: The command should return with status code 1 if there are any vulnerabilities. So setting `ok` to false if there are any vulnerabilities takes care of that

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -177,14 +177,54 @@ Describe "Snyk iac test --experimental command"
...
+
+      Describe "Not using the --json flag"
+        It "returns 0 even if some files failed to parse"
+          When run snyk iac test ../fixtures/iac/no_vulnerabilities/ --experimental --severity-threshold=high
+          The status should equal 0
+          The output should not equal ""
+          The stderr should equal ""
+        End
+      End
+    End
```
jahed-snyk, line 219: Since the `--json` flag doesn't have IaC-specific logic now, do we need these IaC specific tests? We have specific `--json` flag tests in other test suites like: https://github.com/snyk/snyk/blob/master/test/acceptance/cli-json-file-output.test.ts

jahed-snyk (approved):
I don't think we need the `--json` smoke tests anymore, left a comment there. Not blocking on it though, I've approved.

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
teodora-sandu: We added these tests because we previously missed this inconsistency between IaC running with the `--json` flag and without. IaC has its own code path and so I'm thinking keeping these might ensure we don't accidentally break this behaviour over time

aron (approved):

PR #1874, opened 2021-05-02
fix: gradle scanning error due of mishandled root value

anthogez:
While processing Gradle snykGraphs from extractedJSON to depGraph,
projectName is generated using the basename of the rootDir,
we were not handling it if by any chance it was null or undefined.
Now that we are handling it properly, we are preventing possible scanning error.


mika-bar (approved):

Avishagp (approved):
lgtm!

PR #1878, opened 2021-05-04
fix: propagate argument to all modules

miiila:
Parse args as a first things, before other modules are required.
That will allow others to pick arguments.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -5,13 +5,19 @@ import * as pathLib from 'path';
 
 const camelCase = require('lodash.camelcase');
 
+// import args as a first internal module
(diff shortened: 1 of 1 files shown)


anthogez (commented):
Small question... why we are not propagating this globalArgs also in?
https://github.com/snyk/snyk/blob/b68c6882c7a1703428c58c191718a1d72bbe84ab/src/cli/index.ts#L427

miiila:
@anthogez We do: https://github.com/snyk/snyk/blob/b68c6882c7a1703428c58c191718a1d72bbe84ab/src/cli/index.ts#L313

anthogez (approved):
🤦 LGTM

maxjeffos (approved):

PR #1879, opened 2021-05-04
fix: fix invalid poetry detection

jan-stehlik:

#### What does this PR do?

Fix invalid detection of poetry project based solely on presence of `pyproject.toml`. Fixes https://github.com/snyk/snyk/issues/1802

Before we assumed that poetry package manager is present if `pyproject.toml` exists in a folder structure.

The solution is to parse  rely on `poetry.lock` alone to determine if project uses poetry or not.

Run the test cases below

2 testcases:

1. Run `snyk test` for project with `pyproject.toml` that doesn't use poetry
```bash
# in snyk repo
npm i
npm run build
cd test/jest/system/snyk-test/python/fixtures/pyproject-without-poetry/
pipenv install
cd -
node dist/cli/index.js test test/jest/system/snyk-test/python/fixtures/pyproject-without-poetry/ --all-projects
```

Expected result: only one file package manager is recognised pip for `Pipfile` manifest

2. Run `snyk test` for project with `pyproject.toml` that does use poetry
```bash
# in snyk repo
npm i
npm run build
cd test/jest/system/snyk-test/python/fixtures/pyproject-with-poetry/
pipenv install && poetry install
cd -
node dist/cli/index.js test test/jest/system/snyk-test/python/fixtures/pyproject-with-poetry/ --all-projects
```

Expected result: 2 package managers are recognised pip for `Pipfile` manifest and poetry for `pyproject.toml` manifest

https://github.com/snyk/snyk/issues/1802

### DIFF
--- a/src/lib/detect.ts
+++ b/src/lib/detect.ts
@@ -3,7 +3,10 @@ import * as pathLib from 'path';
 import * as debugLib from 'debug';
 const endsWith = require('lodash.endswith');
 import { NoSupportedManifestsFoundError } from './errors';
-import { SupportedPackageManagers } from './package-managers';
+import {
+  SupportedPackageManagers,
+  SUPPORTED_MANIFEST_FILES,
+} from './package-managers';
 
 const debug = debugLib('snyk-detect');
 
@@ -29,7 +32,6 @@ const DETECTABLE_FILES: string[] = [
   'composer.lock',
   'Podfile',
   'Podfile.lock',
-  'pyproject.toml',
   'poetry.lock',
   'mix.exs',
   'mix.lock',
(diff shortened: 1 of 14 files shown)


src/lib/detect.ts
```diff
@@ -99,6 +101,22 @@ const DETECTABLE_PACKAGE_MANAGERS: {
   'mix.exs': 'hex',
 };
 
+function inferPackageManagerFromPyProjectToml(file: string) {
+  const data = fs.readFileSync(file, { encoding: 'utf8' });
```
jan-stehlik, line 105: wrap in try .. catch + unit test

src/lib/detect.ts
```diff
@@ -177,10 +195,12 @@ export function detectPackageFile(root) {
...
 
-export function detectPackageManagerFromFile(
-  file: string,
-): SupportedPackageManagers {
-  let key = pathLib.basename(file);
+function isBasenameNonDeterministic(basename: string): boolean {
+  return basename in NON_DETERMINISTIC_BASENAMES;
+}
+
+export function detectPackageManagerFromFile(path: string) {
```
jan-stehlik, line 202: unit test for both usecases, when input param is
1. filename
2. absolute path

test/jest/system/snyk-test/python/snyk-test-pyproject.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+          foundProjectCount: undefined,
+          displayTargetFile: 'Pipfile',
+          platform: undefined,
+          path: fixturePath,
+        };
+        expect(result).toMatchObject({
+          result: JSON.stringify(expectedResultObject, null, 2),
+        });
+
+        done();
```
jan-stehlik, line 98: investigate if we really need this awful clause, check for any outstading callbacks that might not ever be called back in test cases

src/lib/detect.ts
```diff
@@ -99,6 +101,22 @@ const DETECTABLE_PACKAGE_MANAGERS: {
   'mix.exs': 'hex',
 };
 
+function inferPackageManagerFromPyProjectToml(file: string) {
+  const data = fs.readFileSync(file, { encoding: 'utf8' });
+  const parsedData = parseToml(data);
+
+  if (parsedData?.tool?.poetry) {
+    return 'poetry';
```
jan-stehlik, line 109: candidate for constant

src/lib/detect.ts
```diff
@@ -99,6 +101,22 @@ const DETECTABLE_PACKAGE_MANAGERS: {
...
+  }
+
+  return null;
+}
+
+// these filenames do not guarantee the usage of a package manager yet, and further checks are required to determine correct package manager
+const NON_DETERMINISTIC_BASENAMES = {
+  'pyproject.toml': inferPackageManagerFromPyProjectToml,
```
jan-stehlik, line 117: enum for keys

src/lib/detect.ts
jan-stehlik: addressed in https://github.com/snyk/snyk/pull/1879/commits/a32fe49e009cfe40e7263c3a64c2df9f7f460d1b

test/jest/system/snyk-test/python/snyk-test-pyproject.spec.ts
jan-stehlik: addressed in https://github.com/snyk/snyk/pull/1879/commits/a32fe49e009cfe40e7263c3a64c2df9f7f460d1b

src/lib/detect.ts
jan-stehlik: addressed in https://github.com/snyk/snyk/pull/1879/commits/a32fe49e009cfe40e7263c3a64c2df9f7f460d1b

src/lib/detect.ts
jan-stehlik: addressed in https://github.com/snyk/snyk/pull/1879/commits/a32fe49e009cfe40e7263c3a64c2df9f7f460d1b

orsagie:
Hey, so it definitely works, but I think a better solution that doesn't mess with detect.ts so much is to simply remove the `'pyproject.toml': 'poetry'` line. What this will do is to not allow us to detect a poetry project based on the toml file (which we require the poetry.lock anyway in order to scan a poetry project - see the python plugin code)

jan-stehlik:
> Hey, so it definitely works, but I think a better solution that doesn't mess with detect.ts so much is to simply remove the `'pyproject.toml': 'poetry'` line. What this will do is to not allow us to detect a poetry project based on the toml file (which we require the poetry.lock anyway in order to scan a poetry project - see the python plugin code)

hi @orsagie, thanks for feedback. Removing `pyproject.toml` -> `poetry` mapping is good idea, but it comes with a tradeoff - if project doesn't contain `project.lock` but it still uses `poetry` the app will fail silently and won't let users know that we detected poetry project but are unable to scan it because of missing `poetry.lock`. In current implementementation we do explicitely fail (see pr description) which I believe is a better DX.

To sum up options i think we have 2 options:
1. remove `pyproject.toml` -> `poetry` mapping
✅ pro - simple, easier to understand code
❌ con - won't detect poetry package manager at all if `poetry.lock` file is not present

2. keep current implementation
✅ pro - will detect poetry package manager even if `poetry.lock` file is not present
❌ con - more complicated, convoluted code

I'm happy with both options, but I lean towads option 1, for better DX reasons at a cost of more complexity. What do you think?

orsagie:
> con - won't detect poetry package manager at all if poetry.lock file is not present
The expectation for us to scan a project as a poetry project depends on a lockfile being present, if there is no lockfile, we will not create a project for it. This is why I think this option is the better way.

test/jest/system/snyk-test/python/snyk-test-pyproject.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+      });
+    });
+
+    describe('project does contain poetry.lock file', () => {
+      it('should scan poetry vulnerabilities', async (done) => {
```
orsagie, line 186:
```suggestion
      it('should scan python vulnerabilities', async (done) => {
```
If there's no lockfie, it should not detect a poetry project

orsagie (commented):
Looks good, love the new tests, will approve once https://github.com/snyk/snyk-python-plugin/pull/143 is merged and new version is released

orsagie (approved):

test/jest/system/snyk-test/python/snyk-test-pyproject.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+        mockedMakeRequest.mockImplementationOnce(() => {
+          return {
+            res: { statusCode: 200 },
+            body: {
+              result: { issuesData: {}, affectedPkgs: {} },
+              meta: { org: 'test-org', isPublic: false },
+              filesystemPolicy: false,
+            },
+          };
+        });
```
JamesPatrickGill, line 51: This could be nitpick so use your discretion, but this mocked implementation has a different type than the actual implementation, i.e. function returns a promise whereas this mock return the data that would be inside the promise. Think its managing it now but not sure if could cause trouble in the future. The `any` on line 11 is masking this.

jan-stehlik: nice catch @JamesPatrickGill , addressed in https://github.com/snyk/snyk/pull/1879/commits/ee3d0d2aa3f7fff10af8e30fd89a10b22437c8ee

PR #1888, opened 2021-05-05
fix: pick up .snyk file given absolute --file

jahed-snyk:
Work in progress.

--file (targetFile) can be a relative or absolute path so using `path.join` causes invalid paths. Instead we should `path.resolve`, using --path or project root as base and resolving --file to a valid path.

I've also run organise imports as matter of best practice to keep our import list... organised.

## To Do

- [ ] Repro test
- [ ] Check what `options.path` is and how it's populated.

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -5,7 +5,6 @@ import * as debugModule from 'debug';
 import * as fs from 'fs';
 import * as get from 'lodash.get';
 import * as path from 'path';
-import * as pathUtil from 'path';
 import * as Queue from 'promise-queue';
 import { parsePackageString as moduleToObject } from 'snyk-module';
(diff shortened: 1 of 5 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -697,7 +690,7 @@ async function assembleLocalPayloads(
         targetFile: project.plugin.targetFile,
 
         // TODO: Remove relativePath prop once we gather enough ruby related logs
-        targetFileRelativePath: `${targetFileRelativePath}`, // Forcing string
+        targetFileRelativePath,
```
jahed-snyk, line 693: I've checked Registry types and yes, `targetFileRelativePath` is optional. It doesn't make sense for it to be empty string as it's a path to the `targetFile` provided above.

PR #1909, opened 2021-05-12
feat: show only unique vulnerabilities in sarif format

admons:
Today in sarif output we have a result item per each "vulnerable path"
This will reduce the number of items to be the number of vulnerabilities


aviadhahami (approved):
lgtm

PR #1913, opened 2021-05-13
chore: Support iac --rules flag [CC-874]

aron:
Fixes CC-874

This PR implements the `--rules` flag for us to roll out an version of custom IaC rules for early feedback. The goal is to allow advanced users to generate their own policy bundle using our OPA framework and have the CLI evaluate it and provide results as with the built in policies.

A custom policy bundle is a tarball containing two files:

1. `custom_policy.wasm` - Which is the WASM module that is evaluated with the provided data.
2. `custom_data.json` - Which contains the metadata for each of the rules.

We keep the same control flow logic as with the standard test path with the exception of loading these custom rules via a new `EngineType.Custom` engine type. These policies are evaluated alongside our own rules.

1. We prep the policy cache using the provided tarball _as well as_ the one from S3.
2. We load and parse all matching files as usual across all file types YAML, JSON & HCL.
3. We duplicate the detected files and replace the `engineType` with our new CustomType.
4. We evaluate those files against the custom engine.
5. We return the results.

All in all I was so impressed with how easy this was to integrate, all I had to do was:

1. Add support for the flag.
1. Add a new EngineType and ProjectType for the "custom" rules.
1. Update the local-cache.ts to read the file from disk when a path is provided.
1. Pass the flag to the local cache.

**Testing**

* I've updated the unit tests for the cache flow to also verify the custom bundle is read + extracted.
* I've added a smoke test to verify the custom rules flag works as intended.

**Testing Locally**

The following is an example custom rule bundle to test out the flow. It contains a `custom_data.json` and `custom_policy.wasm` files which implement a CUSTOM-123 rule which will check a Terraform HCL file for a `tags` property in an AWS Security Group (see this [playground](https://play.openpolicyagent.org/p/396d1QokeC
) for the source).

[custom.tar.gz](https://github.com/snyk/snyk/files/6480163/custom.tar.gz)

    % node ./dist/cli/index.js iac test --rules=test/fixtures/iac/custom-rules/custom.tar.gz test/fixtures/iac/terraform/sg_open_ssh.tf

You should see the following output:

```
Testing sg_open_ssh.tf...


Infrastructure as code issues:
  ✗ Security Group allows open ingress [Medium Severity] [SNYK-CC-TF-1] in VPC
    introduced by resource > aws_security_group[allow_ssh] > ingress

  ✗ Missing tags [Low Severity] [CUSTOM-123]
    introduced by tags


Organization:      aron.carroll-ac3
Type:              Terraform
Target file:       test/fixtures/iac/terraform/sg_open_ssh.tf
Project name:      terraform
Open source:       no
Project path:      test/fixtures/iac/terraform/sg_open_ssh.tf

Tested sg_open_ssh.tf for known issues, found 2 issues
```


src/cli/commands/test/iac-local-execution/local-cache.ts
```diff
@@ -43,17 +57,43 @@ export function getLocalCachePath(engineType: EngineType) {
...
+    default:
+      assertNever(engineType);
   }
 }
 
-export async function initLocalCache(): Promise<void> {
-  const BUNDLE_URL = 'https://static.snyk.io/cli/wasm/bundle.tar.gz';
+export async function initLocalCache({
+  customRulesPath,
+}: { customRulesPath?: string } = {}): Promise<void> {
```
teodora-sandu, line 72: Why are we passing `customRulesPath` in an object? Are we expecting this function to take more arguments in the future? Just makes it slightly hard to read (for me)

aron: Ah, great question, I have a preference for using objects when the argument isn't particularly clear (booleans and numbers in particular) this makes the function definition more verbose but the call-site is more readable. e.g.

```ts
// What is the string? Is this the path to the cache?
initLocalCache("./some/path/to/bundle.tar.gz");

// Vs. Ah it's a path for custom rules bundle.
initLocalCache({customRulesPath: "./some/path/to/bundle.tar.gz"});
```

src/lib/iac/constants.ts
```diff
@@ -1,19 +1,22 @@
 export type IacProjectTypes =
   | 'k8sconfig'
   | 'terraformconfig'
+  | 'customconfig'
```
teodora-sandu, line 4: Are these changes going to show up anywhere? I thought these would be for the `--json` calls but we're still using the value returned by the respective scanner (see `projectType` in screenshot):
Screenshot 2021-05-17 at 10 51 59

teodora-sandu (approved):

src/lib/iac/constants.ts
aron: Hmm, I would expect this to show up, let me check.

ipapast (approved):
Looks good and it was easy to read, I guess the only thing left is to rename the file names?
Also there is an args.test.ts file where we sometimes test arguments, not sure if you can test anything on this feature except that it is true as an arguement when you provide it.
https://github.com/snyk/snyk/blob/master/test/args.test.ts#L5

aron:
> Also there is an args.test.ts file where we sometimes test arguments, not sure if you can test anything on this feature except that it is true as an argument when you provide it.

Thanks for this, I didn't know about it, I'll add that to my Hammer specific PR that adds the flag. #1920

src/lib/iac/constants.ts
aron: This is a "bug" or "feature" in our [`groupMultiDocResults()`](https://github.com/snyk/snyk/blob/a90abbce69472a4e0f104698318e804deeeabf77/src/cli/commands/test/iac-local-execution/results-formatter.ts#L104-L120) function that groups results by file path and uses the `packageType` of the first result it encounters. It was only written to handle multi-doc YAML results. I think this is fine for the moment but worth bearing in mind in future...

teodora-sandu: Yeah I think for the moment it's okay, it's not clear to me what the `customconfig` would add anyway since you can tell your custom rule ran from the results 🤔

PR #1914, opened 2021-05-13
chore: Use generics to simplify measurable-methods

aron:

#### What does this PR do?

Update the typings in the measurable-methods.ts  directory to use generics to determine the argument and return type of the wrapped function. This removes the need to do the casting to `AsyncMeasurableMethod` when passing the function into the wrapper. 

Start with `performanceAnalyticsDecorator()` and `asyncPerformanceAnalyticsDecorator()`. The main awkward bit of TypeScript here is the `Awaited<T>` type which is used to unwrap a `Promise`. We need to use this in the definition for `asyncPerformanceAnalyticsDecorator()` because TypeScript insists on the return value of an async method to be a `Promise<T>`.

```ts
// Should work as the return type:
async (...args: Parameters<T>) => ReturnType<T>

// TypeScript wants the async function to return a Promise so we have to unwrap and re-wrap the return value:
async (...args: Parameters<T>) => Promise<Awaited<ReturnType<T>>>
```

No behaviour changes.

### DIFF
--- a/src/cli/commands/test/iac-local-execution/measurable-methods.ts
+++ b/src/cli/commands/test/iac-local-execution/measurable-methods.ts
@@ -3,14 +3,6 @@ import { parseFiles } from './file-parser';
 import { scanFiles } from './file-scanner';
 import { formatScanResults } from './results-formatter';
 import { cleanLocalCache, initLocalCache } from './local-cache';
-import {
-  FormattedResult,
-  IacFileData,
-  IacFileScanResult,
-  IacOrgSettings,
-  ParsingResults,
-  TestReturnValue,
-} from './types';
 import { applyCustomSeverities } from './org-settings/apply-custom-severities';
 import { getIacOrgSettings } from './org-settings/get-iac-org-settings';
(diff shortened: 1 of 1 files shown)


rontalx (approved):
nice!

src/cli/commands/test/iac-local-execution/measurable-methods.ts
```diff
@@ -19,17 +11,19 @@ import {
   performanceAnalyticsObject,
 } from './analytics';
 
-export type AsyncMeasurableMethod<ReturnValue> = (
-  ...args: unknown[]
-) => Promise<ReturnValue>;
-
-export type MeasurableMethod<ReturnValue> = (...args: unknown[]) => ReturnValue;
+// Unwrap a promise: https://stackoverflow.com/questions/48011353/how-to-unwrap-type-of-a-promise
+type Awaited<T> = T extends PromiseLike<infer U> ? U : T;
```
ipapast, line 15: 🤯 wow.

PR #1917, opened 2021-05-13
chore: IaC --rules feature flag [CC-875]

aron:

Fixes CC-875
Depends on https://github.com/snyk/snyk/pull/1913

#### What does this PR do?

This PR adds support for a new `iacCustomRules` feature flag that will enable the `--rules` flag. For orgs that do not have this feature enabled the flag will be rejected with an error.

The implementation is all in test/local-exec/index.ts

I've also started migrating our Infrastructure as Code tests over to the `jest/acceptance` test folder. Most of our tests feel like they could live under here and talk to the fake backend. There are two files one containing the custom rules acceptance tests and the second a helper function that starts the fake server and returns a `run(cmd: string)` function for executing CLI commands and evaluating the output/exit code. The helper ensures that the CLI is configured (via environment variables) to talk to the correct backend with appropriate API token.

Test with feature flag disabled:

```
node ./dist/cli/index.js iac test --rules=test/fixtures/iac/custom-rules/custom.tar.gz test/fixtures/iac/terraform/sg_open_ssh.tf
```

<s>You should see an error. Enable the `iacCustomRules` feature flag for your org and run the command again.</s>
_EDIT: I've not implemented the Registry side of this yet_

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -19,6 +19,7 @@ test/smoke/.iac-data/ @snyk/cloudconfig
 test/fixtures/sast/ @snyk/sast-team
 test/snyk-code-test.spec.ts @snyk/sast-team
 test/jest/unit/iac-unit-tests/ @snyk/cloudconfig
+test/jest/acceptance/iac/ @snyk/cloudconfig
 src/lib/errors/invalid-iac-file.ts @snyk/cloudconfig
 src/lib/errors/unsupported-options-iac-error.ts @snyk/cloudconfig
 help/commands-docs/iac-examples.md @snyk/cloudconfig
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -34,7 +34,7 @@ function getFlagName(key: string) {
   return `${dashes}${flag}`;
 }
 
-class FlagError extends CustomError {
+export class FlagError extends CustomError {
   constructor(key: string) {
     const flag = getFlagName(key);
     const msg = `Unsupported flag "${flag}" provided. Run snyk iac test --help for supported flags.`;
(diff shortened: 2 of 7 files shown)


teodora-sandu (approved):
Approving to unblock you, but maybe we should make sure to include a test under https://github.com/snyk/snyk/blob/master/test/jest/acceptance/cli-args.spec.ts#L75 at least. It'd also be a good way to replace the smoke tests while we depend on the feature flag

aron:
@teodora-sandu thanks, the test for the flag is included here https://github.com/snyk/snyk/pull/1920 but I'll look at porting the smoke test over to the acceptance tests too.

.github/CODEOWNERS
```diff
@@ -19,6 +19,7 @@ test/smoke/.iac-data/ @snyk/cloudconfig
 test/fixtures/sast/ @snyk/sast-team
 test/snyk-code-test.spec.ts @snyk/sast-team
 test/jest/unit/iac-unit-tests/ @snyk/cloudconfig
+test/jest/acceptance/iac/ @snyk/cloudconfig
```
aron, line 22: Can anyone explain why this isn't overriding Hammer and Boost as codeowners of the iac specs?

teodora-sandu: Are they showing up as codeowners of the iac specs or `test/acceptance/fake-server.ts`?

aron: All three 🤔 
image

teodora-sandu: Oh I didn't know that's what that icon did. In VSCode they show up correctly as owned by`@snyk/cloudconfig`, that's very odd 🤔

Avishagp (approved):

.github/CODEOWNERS
aron: Turns out changes to CODEOWNERS are only applied after they are commited. Which makes sense.

PR #1921, opened 2021-05-17
feat: check for sast/snyk code setting on org before running code test

neonnoon:
#### What does this PR do?

Fixes [COD-123](https://snyksec.atlassian.net/browse/COD-123).

This PR prevents snyk code analysis unless enabled for the orgaization.

#### How to test?

Run `snyk code test` for an organization that does not have it enabled.

```
❯ snyk code test --org=testorg
Snyk Code is not supported for org testorg.
```

### DIFF
--- a/src/lib/plugins/sast/validate.ts
+++ b/src/lib/plugins/sast/validate.ts
@@ -1,24 +1,28 @@
 import { Options } from '../../types';
 import * as config from '../../config';
+
 import { AuthFailedError, FeatureNotSupportedForOrgError } from '../../errors';
 
 export async function validateCodeTest(options: Options) {
-  const featureFlag = 'snykCodeCli';
   const org = options.org || config.org;
 
+  // This is an unexpected path, code plugin executed for non-code command.
   if (!options.code) {
     throw new FeatureNotSupportedForOrgError(org);
   }
 
   // TODO: We would need to remove this once we fix circular import issue
-  const { isFeatureFlagSupportedForOrg } = require('../../feature-flags');
-  const isFeatureFlagRes = await isFeatureFlagSupportedForOrg(featureFlag, org);
+  const { getSastSettingsForOrg } = require('./settings');
+  const sastSettingsResponse = await getSastSettingsForOrg(org);
 
-  if (isFeatureFlagRes.code === 401 || isFeatureFlagRes.code === 403) {
-    throw AuthFailedError(isFeatureFlagRes.error, isFeatureFlagRes.code);
+  if (sastSettingsResponse.code === 401 || sastSettingsResponse.code === 403) {
+    throw AuthFailedError(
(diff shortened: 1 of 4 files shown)


neonnoon:
Smoke test fails because org `snyk-cli-smoke-test-with-snykcode` does not have snyk code enabled, looking into it...

src/lib/settings.ts
```diff
@@ -0,0 +1,25 @@
+import request = require('./request');
```
Spoor2709, line 1: Required: Since this file is SAST specific it needs to live in the src/lib/plugins/sast folder

src/lib/plugins/sast/validate.ts
```diff
@@ -1,24 +1,27 @@
...
   }
 
-  // TODO: We would need to remove this once we fix circular import issue
-  const { isFeatureFlagSupportedForOrg } = require('../../feature-flags');
-  const isFeatureFlagRes = await isFeatureFlagSupportedForOrg(featureFlag, org);
+  const sastSettingsResponse = await getSastSettingsForOrg(org);
 
-  if (isFeatureFlagRes.code === 401 || isFeatureFlagRes.code === 403) {
-    throw AuthFailedError(isFeatureFlagRes.error, isFeatureFlagRes.code);
+  if (sastSettingsResponse.code === 401 || sastSettingsResponse.code === 403) {
```
Spoor2709, line 17: Question: should we handle other error codes such as a 500?

j-sp4 (changes requested):
Great Job! 2 small comments to resolve 😃

src/lib/plugins/sast/validate.ts
neonnoon: Good question! From the previous (feature flag based) implementation I assumed that we do not need/want to show a specific error. Do you think it's necessary?

From a functionality point of view, any error will cause snyk code analysis to be skipped, which is correct IMHO.

src/lib/settings.ts
neonnoon: Depends whether we see this as a general place for (org/cli) setting checks to be implemented, or sast-specific, I guess?

neonnoon:
As smoke test runs towards app.snyk.io, this can only be merged & released once https://github.com/snyk/registry/pull/20845 is released to prod.

src/lib/settings.ts
Spoor2709: When I read this file I see it as very sast specific logic the cli should have as little notion as possible of sast

src/lib/plugins/sast/validate.ts
Spoor2709: This is a very good question. I think you are right here 😃

j-sp4 (approved):
Great Job! LGTM!

ghost (approved):

PR #1934, opened 2021-05-18
feat: add support for --org flag to snyk iac test [CC-858]

aron:

#### What does this PR do?

Fixes CC-858

This PR adds support for the `--org` flag (and `SNYK_CFG_ORG` environment variable) to the `snyk iac test` command. We use the same logic as elsewhere in the codebase and first check the `options.org` property and fall back to the `config.org` which looks in both the environment and then values set via `snyk config set org <org>`. If no org is provided it will use the default org associated with the account.

Start in the test/iac-local-execution/index.ts where the org is retrieved and passed to the feature flag helper as well as used when retrieving the org settings. I've added some additional acceptance tests that verify that the org is correctly passed to the backend and included in the JSON output.

This PR also re-enables the custom rules check that is used for testing the feature gating.

Run `npm run build` to generate the latest code and then test with various orgs.

```
node ./dist/cli/index.js iac test ./test/fixtures/iac/terraform/sg_open_ssh.tf --json --org=aron-123
```

You should see the relevant `meta.org` printed out in the JSON output. You can pipe the JSON into the following node command to print the org:

```
 node -pe 'JSON.parse(require("fs").readFileSync(0)).meta.org'
```

image

### DIFF
--- a/src/cli/commands/test/iac-local-execution/org-settings/get-iac-org-settings.ts
+++ b/src/cli/commands/test/iac-local-execution/org-settings/get-iac-org-settings.ts
@@ -7,11 +7,14 @@ import request = require('../../../../../lib/request');
 import { CustomError } from '../../../../../lib/errors';
 import { getErrorStringCode } from '../error-utils';
 
-export function getIacOrgSettings(): Promise<IacOrgSettings> {
+export function getIacOrgSettings(
+  publicOrgId?: string,
+): Promise<IacOrgSettings> {
   const payload: Payload = {
     method: 'get',
     url: config.API + '/iac-org-settings',
     json: true,
+    qs: { org: publicOrgId },
     headers: {
       'x-is-ci': isCI(),
       authorization: `token ${api()}`,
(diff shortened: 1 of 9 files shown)


src/cli/commands/test/iac-local-execution/org-settings/get-iac-org-settings.ts
```diff
@@ -7,11 +7,12 @@ import request = require('../../../../../lib/request');
 import { CustomError } from '../../../../../lib/errors';
 import { getErrorStringCode } from '../error-utils';
 
-export function getIacOrgSettings(): Promise<IacOrgSettings> {
+export function getIacOrgSettings(org?: string): Promise<IacOrgSettings> {
```
rontalx, line 10: nit:
can you rename the argument so it's clearer if this is a orgPublicId, orgId, orgName, etc'?

aron: Sure, I kept it as `org` because it matches the config and options naming.

aron: Updated to `orgPublicId`, are you reviewing the rest of the PR?

rontalx (approved):

JamesPatrickGill (approved):

PR #1936, opened 2021-05-19
chore: Introduce 'yaml' library to parse YAML IaC files [CC-854]

ipapast:
## TL;DR
This PR introduces the “yaml” library as a robust way of parsing YAML files for the IaC flow.
It also removes the `js-yaml` library which is a transitive dependency.

### The Problem
The past few months, we have identified a few issues with the npm library we are using for parsing YAML files (`js-yaml`).
It detects both false positives and false negatives when parsing YAML files, which has not been ideal for us. We used to do workarounds to avoid the real issue, but we are now introducing support for CloudFormation (another YAML type) and it's time to address this.

#### Examples
**Example 1:**
 Library gets this string: `{{"foo": "bar"}: {"qux": "baz"}}` and it should fail as this is invalid yaml.
It does not fail, but it parses it like this: `[{"[object Object]":{"qux":"baz"}}]` which is not correct.
Runkit: https://runkit.com/embed/9tf55g330u5h

**Example 2:**
Library gets a CloudFormation yaml file with AWS intrinsic functions such as: 
 `key: !Equals [ !Ref EnvironmentName, prod ]`
This will throw an exception for unknown tag, which is fair and will fail. However this is still valid YAML.

### Options/Solutions

- We use an outdated version of `js-yaml@3.14.1`. We can upgrade to a newer version, which provides us with a functionality to extend the default schema to support tags like !Ref and !Equals etc. Example: 
```
jsYaml.CORE_SCHEMA.extend({
implicit: [],      
explicit: [
'Fn::Base64',
'Fn::Cidr',
'Fn::FindInMap',
'Fn::GetAtt',
'Fn::GetAZs',
'Fn::ImportValue',
'Fn::Join',
'Fn::Select',
'Fn::Split',
'Fn::Sub',
'Fn::Transform',
'Ref',
'Condition',
'Fn::And',
'Fn::Equals',
'Fn::If',
'Fn::Not',
'Fn::Or',
]
```
However, if we print the dependency of the package (see below), we will see that it is not only used in our code, but also in `@yarn/core`. This means that `js-yaml@3.14.1` will always be a dependency until yarn updates.

```
% npm ls --production js-yaml
snyk@1.0.0-monorepo /Users/iliannapapastefanou/Snyk-repos/snyk
├─┬ @snyk/snyk-cocoapods-plugin@2.5.2
│ └─┬ @snyk/cocoapods-lockfile-parser@3.6.2
│   └── js-yaml@3.14.1  deduped
├─┬ snyk-nodejs-lockfile-parser@1.34.2
│ ├─┬ @yarnpkg/core@2.4.0
│ │ └─┬ @yarnpkg/parsers@2.3.0
│ │   └── js-yaml@3.14.1  deduped
│ └── js-yaml@4.1.0 
└─┬ snyk-policy@1.19.0
  └── js-yaml@3.14.1
 ```
- Replacing the `js-yaml` library with a new one in the IaC code. We spiked different libraries and we have found one that it currently supports everything that we want and it's also extensible. 

### The Proposal
This PR is a proposal to replace it with a different npm library, called `yaml`. This library parses YAML contents correctly and ensures our current functionality does not break. 

**The proposed package itself is small, so we think adding this package is worth it - 91.7kB for the Minified bundle and 27kB for the Minified + Gzipped. Also, there are 0 direct dependencies. More: 
https://bundlephobia.com/result?p=yaml@1.10.2**
image

[CC-854]

[CC-854]: https://snyksec.atlassian.net/browse/CC-854

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-loader.ts
+++ b/src/cli/commands/test/iac-local-execution/file-loader.ts
@@ -41,7 +41,6 @@ export async function loadFiles(
   if (filesToScan.length === 0) {
     throw new NoFilesToScanError();
   }
-
   return filesToScan;
 }
 
@@ -72,6 +71,10 @@ export async function tryLoadFileData(
     await fs.readFile(pathToScan, DEFAULT_ENCODING)
   ).toString();
 
+  if (fileContent === '') {
+    return null;
+  }
+
   return {
     filePath: pathToScan,
     fileType: fileType as IacFileTypes,
@@ -88,6 +91,7 @@ export class NoFilesToScanError extends CustomError {
       'Could not find any valid infrastructure as code files. Supported file extensions are tf, yml, yaml & json.\nMore information can be found by running `snyk iac test --help` or through our documentation:\nhttps://support.snyk.io/hc/en-us/articles/360012429477-Test-your-Kubernetes-files-with-our-CLI-tool\nhttps://support.snyk.io/hc/en-us/articles/360013723877-Test-your-Terraform-files-with-our-CLI-tool';
   }
 }
+
 export class FailedToLoadFileError extends CustomError {
   constructor(filename: string) {
     super('Failed to load file content');
--- a/src/cli/commands/test/iac-local-execution/file-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/file-parser.ts
@@ -1,4 +1,4 @@
-import * as YAML from 'js-yaml';
+import * as YAML from 'yaml';
 import {
   MissingRequiredFieldsInKubernetesYamlError,
   REQUIRED_K8S_FIELDS,
@@ -11,12 +11,12 @@ import {
 } from './parsers/terraform-plan-parser';
 
 import {
-  IacFileParsed,
-  IacFileData,
-  ParsingResults,
-  IacFileParseFailure,
   IaCErrorCodes,
+  IacFileData,
+  IacFileParsed,
+  IacFileParseFailure,
   IaCTestFlags,
+  ParsingResults,
   TerraformPlanScanMode,
 } from './types';
 import * as analytics from '../../../../lib/analytics';
@@ -67,7 +67,12 @@ function parseYAMLOrJSONFileData(fileData: IacFileData): any[] {
   try {
     // the YAML library can parse both YAML and JSON content, as well as content with singe/multiple YAMLs
     // by using this library we don't have to disambiguate between these different contents ourselves
-    yamlDocuments = YAML.safeLoadAll(fileData.fileContent);
+    yamlDocuments = YAML.parseAllDocuments(fileData.fileContent).map((doc) => {
+      if (doc.errors.length !== 0) {
+        throw doc.errors[0];
+      }
+      return doc.toJSON();
+    });
   } catch (e) {
     if (fileData.fileType === 'json') {
       throw new InvalidJsonFileError(fileData.filePath);
--- a/package.json
+++ b/package.json
@@ -137,7 +137,8 @@
     "tempfile": "^2.0.0",
     "update-notifier": "^5.1.0",
     "uuid": "^3.3.2",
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: src/cli/commands/test/iac-local-execution/file-parser.ts [hunk 1]
```diff
@@ -67,7 +67,12 @@ function parseYAMLOrJSONFileData(fileData: IacFileData): any[] {
   try {
     // the YAML library can parse both YAML and JSON content, as well as content with singe/multiple YAMLs
     // by using this library we don't have to disambiguate between these different contents ourselves
-    yamlDocuments = YAML.safeLoadAll(fileData.fileContent);
+    yamlDocuments = YAML.parseAllDocuments(fileData.fileContent).map((doc) => {
+      if (doc.errors.length !== 0) {
+        throw doc.errors[0];
```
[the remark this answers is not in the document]
ipapast: (This was to not break the existing jest tests for now)

src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -1,4 +1,4 @@
-import * as YAML from 'js-yaml';
```
gitphill, line 1: `js-yaml` was a transitive? that's a little naughty. Have we replaced it with `yaml` everywhere?

ipapast: It was, we also just recently realised this 😬 In the code, `js-yaml` is actually being imported only in this file.

gitphill: what about https://github.com/snyk/snyk/blob/master/src/lib/iac/iac-parser.ts#L2?

ipapast: 🤦 You're totally right. I just replaced it there too.

ghost (approved):

gitphill (approved):

src/cli/commands/test/iac-local-execution/file-loader.ts
```diff
@@ -27,10 +27,14 @@ export async function loadFiles(
...
 
   const filesToScan: IacFileData[] = [];
+  const emptyFiles: string[] = [];
   for (const filePath of filePaths) {
     try {
       const fileData = await tryLoadFileData(filePath);
       if (fileData) {
+        if (fileData.fileContent === '') {
+          emptyFiles.push(fileData.filePath);
+        }
```
ipapast, line 37: I'm pushing emptyFiles in a new array, so when we have a directory containing empty files, we can compare and skip them.

src/cli/commands/test/iac-local-execution/file-loader.ts
```diff
@@ -42,7 +46,12 @@ export async function loadFiles(
     throw new NoFilesToScanError();
   }
 
-  return filesToScan;
+  return filesToScan.filter(
+    (fileToScan) =>
+      !emptyFiles.some(
+        (emptyFilePath) => fileToScan.filePath === emptyFilePath,
+      ),
+  );
```
ipapast, line 54: @aron Can you have a look again? Now I compare the two arrays, so that we won't scan empty files at all.
I haven't added an Error for this case, but I think it can go in another PR?

aron: I don't think I understand why we even track the empty files. On line 76 of `tryLoadFileData()` we ignore files that don't have a supported extension, I think at this point, just after reading the file, we can also return null if the file is empty?

ipapast: We can do that and return null. The current flow returns an error: 
`NoFilesToScanError: Could not find any valid IaC files`
My idea was that we we skip the empty files and we let the user now which/how many files were skipped instead of an error.

aron (approved):

PR #1942, opened 2021-05-21
chore: add missing dependencies to package.json

ghost:
We have a bunch of imports that rely on transitive deps. This makes it difficult for us to reason about our dep tree (e.g. scanning) so I've added them to our package.json.

I'll create a separate PR to add a check for this to avoid future issues.

#1936 will remove the `js-yaml` imports which are also transitive.


aron (approved):
👏 Nice

gitphill (approved):

maxjeffos (approved):

PR #1950, opened 2021-05-25
Feat/adding cloudformation support [CC-850]

solgaezhova:
#### What does this PR do?

This PR enables users to scan CloudFormation files (both with `yaml` and `json` extenstions) via CLI. For this purpose it:

1. Adds EngineType for CloudFormation, packagemanager: cloudformationconfig.
2. Adds support for the new wasm bundle for CloudFormation files.
3. Extends the engineTypeToProjectType with the new EngineType=cloudformationconfig.
4. Implements the soultion for the detection of CloudFormation files in a new `k8s-or-cloudformation-parser.ts` by checking whether a file has a top-level `Resourses` field which is the only required field for ClodFormation templates. 
5. Introduces new unit tests and smoke tests for CloudFormation.

The reviewer should start with `src/cli/commands/test/iac-local-execution/file-parser.ts` file.

It can by manually tested by running `snyk-dev iac test test/fixtures/iac/cloudformation` command for testing a folder containing different CloudFormation templates for testing. Or by specifying CloudFormation file name in the same command for testing files individually.

https://snyksec.atlassian.net/jira/software/projects/CC/boards/93?selectedIssue=CC-850

#### Screenshots

1. Testing valid CloudFormation file with 1 issue.

Screen Shot 2021-05-25 at 11 18 23

2. Testing valid CloudFormation file with 1 issue, adding `--json` at the end of the command

Screen Shot 2021-05-25 at 11 25 16
Screen Shot 2021-05-25 at 10 59 55

3. Testing invalid CloudFormation file.

Screen Shot 2021-05-25 at 11 19 06

4. Testing valid CloudFormation file with no `Resources` field.

Screen Shot 2021-05-25 at 11 19 55

5. Testing CloudFormation file with `json` extension.

Screen Shot 2021-05-25 at 11 21 01

#### Additional questions

1. Before changes there was a specific error for k8s files that did not contain required fields - `MissingRequiredFieldsInKubernetesYamlError` saying `We were unable to detect whether the YAML file "${filename}" is a valid Kubernetes file, it is missing the following fields:...` Now CloudFormation and Kubernetes files which do not have required fields get the shared output error message in `FailedToDetectYamlFileError` saying `We were unable to detect whether the YAML file "${filename}" is a valid Kubernetes or CloudFormation file. For Kubernetes required fields are... For CloudFormation required fields are: ...` However, we did not remove `MissingRequiredFieldsInKubernetesYamlError` for now
to be able to use it in case we would like to have a specific error for k8s files. 
2. We renamed the `kubernetes-parser.ts` file to `k8s-or-cloudformation-parser.ts` to be consistent with the namings we had before. However, we think that a further renaming of the files is required (in a future PR) to be more consistent with the file content as in`k8s-or-cloudformation-parser.ts` we rather detect a config type after file parsing.

### DIFF
--- a/src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
@@ -9,6 +9,7 @@ import {
 } from '../types';
 
 export const REQUIRED_K8S_FIELDS = ['apiVersion', 'kind', 'metadata'];
+export const REQUIRED_CLOUDFORMATION_FIELDS = ['Resources'];
 
 export function assertHelmAndThrow(fileData: IacFileData) {
   const lines: string[] = fileData.fileContent.split(/\r\n|\r|\n/);
@@ -21,29 +22,44 @@ export function assertHelmAndThrow(fileData: IacFileData) {
   });
 }
 
-export function tryParsingKubernetesFile(
+export function detectConfigType(
   fileData: IacFileData,
-  yamlDocuments: any[],
+  parsedIacFiles: any[],
 ): IacFileParsed[] {
   assertHelmAndThrow(fileData);
 
-  return yamlDocuments.map((parsedYamlDocument, docId) => {
+  return parsedIacFiles.map((parsedIaCFile, docId) => {
     if (
-      REQUIRED_K8S_FIELDS.every((requiredField) =>
-        parsedYamlDocument.hasOwnProperty(requiredField),
-      )
+      checkRequiredFieldsMatch(parsedIaCFile, REQUIRED_CLOUDFORMATION_FIELDS)
     ) {
       return {
         ...fileData,
-        jsonContent: parsedYamlDocument,
+        jsonContent: parsedIaCFile,
+        projectType: IacProjectType.CLOUDFORMATION,
+        engineType: EngineType.CloudFormation,
+        docId,
+      };
+    } else if (checkRequiredFieldsMatch(parsedIaCFile, REQUIRED_K8S_FIELDS)) {
+      return {
+        ...fileData,
+        jsonContent: parsedIaCFile,
         projectType: IacProjectType.K8S,
         engineType: EngineType.Kubernetes,
         docId,
       };
     } else {
-      throw new MissingRequiredFieldsInKubernetesYamlError(fileData.filePath);
+      throw new FailedToDetectYamlConfigError(fileData.filePath);
     }
   });
+}
+
+export function checkRequiredFieldsMatch(
+  parsedDocument: any,
+  requiredFields: string[],
+) {
+  return requiredFields.every((requiredField) =>
+    parsedDocument.hasOwnProperty(requiredField),
+  );
 }
 
 export class HelmFileNotSupportedError extends CustomError {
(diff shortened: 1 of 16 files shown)


src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
```diff
@@ -21,31 +22,46 @@ export function assertHelmAndThrow(fileData: IacFileData) {
...
       };
     } else {
-      throw new MissingRequiredFieldsInKubernetesYamlError(fileData.filePath);
+      throw new FailedToDetectYamlFileError(fileData.filePath);
     }
   });
 }
 
+export function checkRequiredFieldsMatch(
+  parsedDocument: any[],
```
aron, line 57: I think this should be `any` as we're passing in a single document here. One TODO for a follow up PR we should look into replacing this and the other `any[]` instances with a `JSON` type e.g. https://github.com/microsoft/TypeScript/issues/1897#issuecomment-822032151

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
```diff
@@ -9,6 +9,7 @@ import {
 } from '../types';
```
aron, line 9: ^ Filename suggestion, `yaml-config-parser.ts`?

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
```diff
@@ -65,3 +81,18 @@ export class MissingRequiredFieldsInKubernetesYamlError extends CustomError {
     )}"`;
   }
 }
+
+export class FailedToDetectYamlFileError extends CustomError {
```
aron, line 85: Naming suggestion: `FailedToDetectYamlConfigError` mostly to dis-ambiguate because we know it's a parsed YAML file but it's not clear it's a _config file_.

aron (approved):
👏 fantastic work @solgaezhova this is a really clean PR, great description and implementation. I've left a couple of very minor comments around the naming of a file and error naming just to help us in future, other than that it's ready to go.

test/jest/unit/iac-unit-tests/file-parser.spec.ts
```diff
@@ -54,13 +62,15 @@ describe('parseFiles', () => {
       ...expectedMultipleKubernetesYamlsParsingResult,
       docId: 1,
```
aron, line 63: This line has broken in the test suite, I'm not sure from the changes why but it looks like we might be losing the docId... @ipapast @solgaezhova any thoughts here?

ipapast: Oops, that was me, I didn't update the fixture with a docId. I pushed it now 👍

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
solgaezhova: Actually `detectConfigType` function in `k8s-or-cloudformation-parser.ts` is used both for `yaml` and `json` cases. But we will think of other possible renaming of this file in next PR to make it more consistent to the content.

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
solgaezhova: Thank you, Aron, it makes sense, absolutely agree and like the suggested naming. @ipapast Ilianna, what is your opinion on this?

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
solgaezhova: Thank you @aron for reviewing and commenting!

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
solgaezhova: In case we rename this one, I think we should also rename `FailedToDetectJsonFileError` accordingly. @ipapast let's sync and decide when we do these renamings.

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
aron: Great point! Sounds good to me

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
solgaezhova: Renamed both `FailedToDetectYamlFileError` and `FailedToDetectJsonFileError`.

PR #1951, opened 2021-05-25
chore: upload test results to circleci

ghost:
We need better visibility on our tests and it looks like CircleCI has some features around it.

- jest-junit for jest
- tap-junit for tap. tap doesn't let you have both pretty output and junit output at the same time. But it does like you have TAP output, which I'm using to then convert to junit output for CircleCI.
  - However, the junit processers don't provide timing information. I'd send a feature request, but our tap version is so old that we'd first have to do due diligence and upgrade our tap version to make sure this is an issue with the processors and not our old tap version.
  - We could disable pretty output as tap's built-in xunit reporter provides timing, but that means we won't get pretty terminal output in CI. However, we will get failures in the CircleCI test tab.
- shellspec has built-in junit support

To see the results, open up this PR's workflow, choose a completed job, then click the "Tests" tab.

I've also refactored `check-dependencies` as I needed to add `jest-junit` to the ignore list. Now there's a root level `check-dependencies.config.ts` so people can easily add new ignore rules without needing to dig through the actual script. This is a separate commit for better readability.

I've added `--ci` to our jest tests as it's recommended. If we ever use jest snapshots, by default if it can't find an existing snapshot it will create one. With `--ci` it will fail instead as we'd expect an existing snapshot when running in CI.


.circleci/config.yml
```diff
@@ -252,47 +255,74 @@ jobs:
           steps:
             - run:
                 name: Run Package unit tests
-                command: npm run test:packages-unit
+                command: npm run test:packages-unit -- --ci --reporters="default" --reporters="jest-junit"
```
Avishagp, line 258: What does `--` without a flag mean?

jahed-snyk: It means "forward the rest of the arguments to the script (jest in this case) rather than to npm".

Avishagp (approved):

PR #1958, opened 2021-05-26
chore: add replay tests

admons:
Add a way to record tests


test/jest/replay/replay.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+describe('replay', () => {
+  nockBack.setMode('lockdown'); // switch to `record` for recording more tests
+
+  beforeAll(async () => {
+    await run('help'); // this will load all modules into memory
+  });
+
+  const testCases = ['test test/acceptance/workspaces/yarn-package'];
+
+  for (const command of testCases) {
```
anthogez, line 18: I would advice to go with it.each instead of `for (const command of testCases)`

PR #1959, opened 2021-05-26
bug: extract line numbers if sarif-file-output or json-file-output [CC-887]

teodora-sandu:

#### What does this PR do?
This PR enables the line number functionality for the `--sarif-file-output` and `--json-file-output` flags. Without this line number functionality, the GitHub action for IaC is broken because Sarif expects a non-negative `startLine`.

1. Run `npm run build`
2. Run `node ./dist/cli iac test test/fixtures/iac/terraform/sg_open_ssh.tf --sarif-file-output=snyk.sarif`
3. Run `cat snyk.sarif` and notice that the `startLine` is not -1 anymore.

In https://github.com/snyk/snyk/pull/1901 we wanted to make sure we didn't run the line number functionality if we weren't going to display the line number to a user. This line number is only rendered when we output JSON or Sarif. We forgot about the `--json-file-output` and `--sarif-file-output` flags however.

- [CC-887](https://snyksec.atlassian.net/browse/CC-887)

#### Screenshots
Before:
Screenshot 2021-05-26 at 11 39 19

After:
Screenshot 2021-05-25 at 19 06 39

### DIFF
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -41,7 +41,7 @@ const engineTypeToProjectType = {
 function formatScanResult(
   scanResult: IacFileScanResult,
   meta: TestMeta,
-  { severityThreshold, json, sarif }: IaCTestFlags,
+  options: IaCTestFlags,
 ): FormattedResult {
   const formattedIssues = scanResult.violatedPolicies.map((policy) => {
     const cloudConfigPath =
@@ -49,7 +49,15 @@ function formatScanResult(
         ? [`[DocId:${scanResult.docId}]`].concat(policy.msg.split('.'))
         : policy.msg.split('.');
 
-    const shouldExtractLineNumber = json || sarif;
+    const flagsRequiringLineNumber = [
+      'json',
+      'sarif',
+      'json-file-output',
+      'sarif-file-output',
+    ];
+    const shouldExtractLineNumber = flagsRequiringLineNumber.some(
+      (flag) => options[flag],
+    );
     const lineNumber: number = shouldExtractLineNumber
       ? extractLineNumber(scanResult, policy)
       : -1;
@@ -77,7 +85,7 @@ function formatScanResult(
     result: {
       cloudConfigResults: filterPoliciesBySeverity(
         formattedIssues,
-        severityThreshold,
+        options.severityThreshold,
       ),
       projectType: scanResult.projectType,
     },
--- a/test/jest/acceptance/iac/file-output.spec.ts
+++ b/test/jest/acceptance/iac/file-output.spec.ts
@@ -1,0 +1,65 @@
+import { readFileSync, unlinkSync } from 'fs';
+import * as path from 'path';
+import * as sarif from 'sarif';
+import { v4 as uuidv4 } from 'uuid';
+import { startMockServer } from './helpers';
+import { MappedIacTestResponse } from '../../../../src/lib/snyk-test/iac-test-result';
+jest.setTimeout(50000);
+
+describe('iac test --json-file-output', () => {
+  let run: (
+    cmd: string,
+  ) => Promise<{ stdout: string; stderr: string; exitCode: number }>;
+  let teardown: () => void;
+
+  beforeAll(async () => {
+    ({ run, teardown } = await startMockServer());
+  });
+
+  afterAll(async () => teardown());
+
+  it('can save JSON output to file while sending human readable output to stdout', async () => {
+    const jsonOutputFilename = path.join(__dirname, `${uuidv4()}.json`);
+    const { stdout } = await run(
+      `snyk iac test ./iac/terraform/sg_open_ssh.tf --json-file-output=${jsonOutputFilename}`,
+    );
+    expect(stdout).toMatch('Organization:');
+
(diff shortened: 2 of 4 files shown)


src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -41,15 +41,22 @@ const engineTypeToProjectType = {
...
 ): FormattedResult {
   const formattedIssues = scanResult.violatedPolicies.map((policy) => {
     const cloudConfigPath =
       scanResult.docId !== undefined
         ? [`[DocId:${scanResult.docId}]`].concat(policy.msg.split('.'))
         : policy.msg.split('.');
 
-    const shouldExtractLineNumber = json || sarif;
+    const shouldExtractLineNumber =
+      json || sarif || jsonFileOutput || sarifFileOutput;
```
aron, line 59: nit: Would it be better to have an array containing the flags and check each one?
```suggestion
    const flagsRequiringLineNumber = ['json', 'sarif', ...]
    const shouldExtractLineNumber = flagsRequiringLineNumber.some(flag => options[flag]);
```

test/jest/acceptance/iac/file-output.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  it('can save JSON output to file while sending human readable output to stdout', async () => {
+    const jsonOutputFilename = path.join(__dirname, `${uuidv4()}.json`);
+    const { stdout } = await run(
+      `snyk iac test ./iac/terraform/sg_open_ssh.tf --json-file-output=${jsonOutputFilename}`,
+    );
+    expect(stdout).toMatch('Organization:');
+
+    const outputFileContents = readFileSync(jsonOutputFilename, 'utf-8');
+    unlinkSync(jsonOutputFilename);
+    const jsonObj = JSON.parse(outputFileContents);
```
aron, line 31: Can we type the results here? Or at least the parts we're looking for so that we don't have all the casts in the test.
```suggestion
    const jsonObj: ??? = JSON.parse(outputFileContents);
```

test/jest/acceptance/iac/file-output.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  ) => Promise<{ stdout: string; stderr: string; exitCode: number }>;
+  let teardown: () => void;
+
+  beforeAll(async () => {
+    const result = await startMockServer();
+    run = result.run;
+    teardown = result.teardown;
+  });
+
+  afterAll(async () => teardown());
```
aron, line 52: Not for now, but we need a better setup boilerplate for this 🤔

aron (approved):
👏 amazing, thanks for knocking this out so quickly.

test/jest/acceptance/iac/file-output.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  it('can save Sarif output to file while sending human readable output to stdout', async () => {
+    const sarifOutputFilename = path.join(__dirname, `${uuidv4()}.sarif`);
+    const { stdout } = await run(
+      `snyk iac test ./iac/terraform/sg_open_ssh.tf --sarif-file-output=${sarifOutputFilename}`,
+    );
```
aron, line 58: Weird suggestion: If we wanted to avoid the tmp file write here (and these tests only run on unix) we can cheat here and have the sarif printed to stderr and read that directly:

```suggestion
    // stdout = regular output, stderr expects sarif output
    const { stdout, stderr } = await run(
      `snyk iac test ./iac/terraform/sg_open_ssh.tf --sarif-file-output=/dev/stderr`,
    );
```

teodora-sandu: I don't see why not! I didn't think of this, good thinking

test/jest/acceptance/iac/file-output.spec.ts
teodora-sandu: I copied these tests from `test/jest/acceptance/cli-json-file-output.spec.ts` but I was able to just remove the casts altogether, so we don't have to create a type for the tests

aron: I think this might be another todo for later. I think removing the casts for now is fine, but having a type definition for the JSON output would still be useful in maintaining the tests. I think `MappedIacTestResponse` or `IacTestResponse` have most of the fields that are used.

https://github.com/snyk/snyk/blob/2102fbc6c2669b612b98602cc0f584058dec8a17/src/lib/snyk-test/iac-test-result.ts#L79-L82

teodora-sandu: Actually turns out typing exposed that I wasn't doing the right check, so I'll include that in this PR!

test/jest/acceptance/iac/file-output.spec.ts
teodora-sandu: Do we raise tech debt issues for this sort of thing or just add a TODO?

aron: Great question, we decided to create tickets to track it, so that would be great.

test/jest/acceptance/iac/file-output.spec.ts
teodora-sandu: Tried this but the tests started failing in CircleCI and not locally. I couldn't recreate it locally at all so I just swapped back to using the tmp file

PR #1960, opened 2021-05-26
chore: enforce npm@7 for development

ghost:
We get a lot of contributors mentioning failed installs and builds as we use npm@7's workspaces and they're using an older npm version. This change will check the contributor's NodeJS and NPM version after every `npm install` via the `prepare` hook which only runs in development. This makes most sense as it's in the same hook as where we build the project.

This will fail our CI build as we have a npm@6 hack to fix (which we should fix anyway in the future). For now, I've setup a workaround.

Also we're using Node 10/12 for building, even though they're for testing. So I've avoided enforcing node@14 for now. Separating build and test setup will be a lot of work.

### DIFF
--- a/scripts/check-dev-environment.ts
+++ b/scripts/check-dev-environment.ts
@@ -1,0 +1,64 @@
+type Issue = {
+  target: string;
+  reason: string;
+  enforce: boolean;
+};
+
+/**
+ * We have a hack in .circleci/config.yml which uses npm@6 for regression
+ * tests. Once that's removed we can remove this check.
+ */
+const isCIHackInstall = (): boolean => {
+  return (
+    typeof process.env.CI === 'string' && process.env.NODE_ENV === 'production'
+  );
+};
+
+const checkDevEnvironment = async () => {
+  const issues: Issue[] = [];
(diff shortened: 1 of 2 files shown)


scripts/check-dev-environment.ts
```diff
@@ -0,0 +1,64 @@
...
+  enforce: boolean;
+};
+
+/**
+ * We have a hack in .circleci/config.yml which uses npm@6 for regression
+ * tests. Once that's removed we can remove this check.
+ */
+const isCIHackInstall = (): boolean => {
+  return (
+    typeof process.env.CI === 'string' && process.env.NODE_ENV === 'production'
```
Avishagp, line 13: What are we checking here?

jahed-snyk: This is checking for a production install (no dev deps) of our local repo in CI. It's a hack we use here:

https://github.com/snyk/snyk/blob/444a69f3494f81104a9bd922db83f09b1adc40ef/.circleci/config.yml#L188-L192

Avishagp (approved):

PR #1961, opened 2021-05-26
fix: code test json output should not print stack trace

ArturSnyk:

We had an issue with showing stack trace of an error when calling `snyk code test --json` because we didn't check if the "failure" is because we have issue. which means there are vulnerabilities. 

This pr respects the old current logic as well as when we have json output for a payload that has issues


orsagie (approved):
LGTM

PR #1962, opened 2021-05-26
test: improve interface for mock server

aron:
Consolidates all of the `startMockServer()` call behind a _slightly_ magical setup function. So now usage looks like:

```ts
    describe('some test', () => {
      const run = setupMockServer(beforeAll, afterAll);
 
      it('runs a command', () => {
         const { exitCode } = await run('echo "hello"');
         expect(exitCode).toEqual(1);
      });
    });
```

### DIFF
--- a/test/jest/acceptance/iac/helpers.ts
+++ b/test/jest/acceptance/iac/helpers.ts
@@ -1,6 +1,49 @@
+import * as assert from 'assert';
 import { exec } from 'child_process';
 import { join } from 'path';
 import { fakeServer } from '../../../acceptance/fake-server';
+
+type Lifecycle = (fn: () => Promise<void>, timeout?: number) => Promise<void>;
+
+/**
+ * Returns a `run()` function used to evaluate a shell command against the
+ * current environment. Will start a fake webserver in the background as part
+ * of the test beforeAll lifecycle and will ensure that it is torn down in the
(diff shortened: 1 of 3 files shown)


test/jest/acceptance/iac/helpers.ts
```diff
@@ -1,15 +1,61 @@
...
+ *     const run = setupMockServer(beforeAll, afterAll);
+ *
+ *     it('runs a command', () => {
+ *        const { exitCode } = await run('echo "hello"');
+ *        expect(exitCode).toEqual(1);
+ *     });
+ *   });
+ */
+export function setupMockServer(
+  beforeAll: Lifecycle,
```
teodora-sandu, line 26: Why do we need the `beforeAll` and `afterAll` if we're using our own `beforeEach` and `afterEach`?

aron: Good question, we want to pass in the Jest lifecycle methods so we can do the setup. Generally I feel the global versions should only be called directly in the test module itself. It feels odd to me that a sub dependency like this utils module would rely on the existence of the globals so I felt it better to just pass them in explicitly.

But this is a good example of where the whole setup is unclear and your suggestion to use a Jest environment might be on the whole a lot nicer.

PR #1968, opened 2021-05-27
chore: upgrade to jest@27

ghost:
Doesn't seem to be much in the way of breaking changes, mostly optimisations. https://jestjs.io/blog/2021/05/25/jest-27

Note: there was some issues with this PR but after upgrading our CircleCI instance to Large, it fixed itself.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,9 +1,5 @@
 module.exports = {
   preset: 'ts-jest',
-  testEnvironment: 'node',
-  collectCoverage: false, // not collecting coverage for now
-  collectCoverageFrom: ['src/**/*.ts'],
-  coverageReporters: ['text-summary', 'html'],
   testMatch: [
     '<rootDir>/test/*.spec.ts',
     '<rootDir>/test/iac-unit-tests/*.spec.ts',
--- a/package.json
+++ b/package.json
@@ -165,7 +165,7 @@
     "eslint": "6.8.0",
     "eslint-config-prettier": "^6.1.0",
     "fs-extra": "^9.1.0",
-    "jest": "^26.6.3",
+    "jest": "^27.0.6",
     "jest-junit": "^12.0.0",
     "lodash": "^4.17.20",
     "madge": "^3.4.4",
@@ -178,7 +178,7 @@
     "sinon": "^4.0.0",
     "tap": "^12.6.1",
     "tap-junit": "^4.2.0",
-    "ts-jest": "^26.5.2",
+    "ts-jest": "^27.0.4",
     "ts-node": "^8.0.0",
     "typescript": "^3.4.1"
   },
--- a/packages/snyk-fix/package.json
+++ b/packages/snyk-fix/package.json
@@ -48,13 +48,6 @@
     "strip-ansi": "6.0.0",
     "toml": "3.0.0"
   },
-  "devDependencies": {
-    "@types/jest": "26.0.20",
-    "@types/node": "14.14.25",
-    "jest": "26.6.3",
-    "ts-jest": "^26.5.3",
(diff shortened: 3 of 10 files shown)


RA80533:
Try adding a timeout for tests.

Looking at explore-node-modules.ts, it appears that file I/O is done synchronously which might result in horrendous performance when scanning the entire node_modules folder.

jest.config.js
```diff
@@ -1,9 +1,5 @@
 module.exports = {
   preset: 'ts-jest',
-  testEnvironment: 'node',
```
jahed-snyk, line 3: This is default now.

jest.config.js
```diff
@@ -1,9 +1,5 @@
...
-  collectCoverage: false, // not collecting coverage for now
-  collectCoverageFrom: ['src/**/*.ts'],
-  coverageReporters: ['text-summary', 'html'],
```
jahed-snyk, line 6: Coverage is `false` by default and we never use it so can just remove it.

packages/snyk-fix/package.json
```diff
@@ -48,13 +48,6 @@
     "strip-ansi": "6.0.0",
     "toml": "3.0.0"
   },
-  "devDependencies": {
-    "@types/jest": "26.0.20",
-    "@types/node": "14.14.25",
-    "jest": "26.6.3",
-    "ts-jest": "^26.5.3",
-    "typescript": "4.2.4"
-  },
```
jahed-snyk, line 57: @snyk/fix can use devDeps from parent since we're usng npm 7 workspaces. Avoids getting dev deps out of sync.

test/jest/acceptance/cli-json-file-output.spec.ts
```diff
@@ -91,7 +91,7 @@ describe('test --json-file-output ', () => {
     '`test --json-file-output produces same JSON output as normal JSON output to stdout`',
     (done) => {
       const jsonOutputFilename = `${uuidv4()}.json`;
-      return exec(
+      exec(
```
jahed-snyk, line 94: Jest 27 does not like tests using `done` and returning a promise.

package.json
```diff
@@ -165,7 +165,7 @@
     "eslint": "6.8.0",
     "eslint-config-prettier": "^6.1.0",
     "fs-extra": "^9.1.0",
-    "jest": "^26.6.3",
+    "jest": "^27.0.6",
```
maxjeffos, line 168: What about `@types/jest` looks like that can be updated to `^26.0.24`, though maybe it doesn't matter.
Also, it's weird that the major version of `@types/jest` types is out of sync with the `jest`.

test/jest/acceptance/cli-json-file-output.spec.ts
maxjeffos: why use `done` and not just make the test async?

maxjeffos (approved):

package.json
jahed-snyk: types/jest is community-maintained so it's a matter of waiting for someone to do the work. Though, jest does say it comes with typedefs so this shouldn't be needed. 🤔 I'll try removing types/jest and see what happens.

test/jest/acceptance/cli-json-file-output.spec.ts
jahed-snyk: More work unrelated to Jest 27's migration. I can make that a separate PR.

package.json
jahed-snyk: Doesn't work so rolling it back. Luckily 26's type defs work fine for now since 27 didn't change most of the API.

lili2311 (approved):

PR #1976, opened 2021-05-28
feat: yarn2 resolutions supported in workspaces

JamesPatrickGill:

#### What does this PR do?

This adds support to the workspaces parser to support resolutions (yarn2).

### DIFF
--- a/src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
+++ b/src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
@@ -66,12 +66,31 @@ export async function processYarnWorkspaces(
     },
     scannedProjects: [],
   };
+
+  let rootPkgJson = {} as lockFileParser.ManifestFile;
+  try {
+    rootPkgJson = JSON.parse(
+      getFileContents(
+        root,
+        pathUtil.join(Object.keys(yarnTargetFiles)[0], 'package.json'),
+      ).content,
+    );
+  } catch (e) {
+    throw new lockFileParser.InvalidUserInputError(
+      'root package.json parsing failed when parsing workspaces with error ' +
+        e.message,
+    );
+  }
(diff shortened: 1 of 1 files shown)


JamesPatrickGill:
Need to double check how this interacts with yarn1 project to make sure nothing breaks

src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
```diff
@@ -66,12 +66,23 @@ export async function processYarnWorkspaces(
     },
     scannedProjects: [],
   };
+
+  const rootPkgJson: lockFileParser.ManifestFile = JSON.parse(
```
aviadhahami, line 70: what if we throw here? who catches that?

src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
```diff
@@ -181,3 +192,27 @@ function packageJsonBelongsToWorkspace(
   );
   return match;
 }
+
+function createPackageJson(
+  root: string,
+  packageJsonFileName: string,
+  resolutions?: { [key: string]: string },
+) {
+  if (!resolutions) {
```
aviadhahami, line 201: can we assert the keys length instead of the object itself?

src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
JamesPatrickGill: This currently get thrown right out to the top, so you get a "This scan failed" type message. Right now the only place where we throw errors for parsing manifest file is in the parser, which makes sense, but atm this parsing is seperate so will just throw a more helpful error here for now, similar to the parser one and look to consolidate all the parsing logic into the plugin when it makes sense to

PR #1988, opened 2021-06-02
fix: paths in json and sarif output [CC-563]

teodora-sandu:

#### What does this PR do?
This PR fixes the logic for computing the `targetFile`, `targetFilePath`, and `projectName` for the outputs. The SARIF output for our GitHub action has been broken since making local exec GA and putting the legacy flow behind a flag. Comparing the legacy and GA output for SARIF, there was definitely a difference in the `physicalLocation` fields, **when** passing in the path to the folder as `.`.

The reason we are here today is we previously only tested the results returned by the CLI by scanning a single file, whereas now we've tested it for:
- a single file
- a folder which was passed in via the CLI by providing `.`
- a folder which was picked by default by the CLI by not providing any path to it
- a folder which is in a relative location to the current running directory

Build using `npm run build` and clone https://github.com/teodora-sandu/test. Change directory into `nested` and run the following commands one by one. The result of these commands can be found in the attached ZIP.
```
snyk iac test --json-file-output=output-ga.json
snyk iac test --json-file-output=output-legacy.json --legacy
snyk iac test --json-file-output=output-ga-path.json .
snyk iac test --json-file-output=output-legacy-path.json --legacy .

snyk iac test --sarif-file-output=output-ga.sarif
snyk iac test --sarif-file-output=output-legacy.sarif --legacy
snyk iac test --sarif-file-output=output-ga-path.sarif .
snyk iac test --sarif-file-output=output-legacy-path.sarif --legacy .

snyk-dev iac test --json-file-output=output-ga.json
snyk-dev iac test --json-file-output=output-legacy.json --legacy
snyk-dev iac test --json-file-output=output-ga-path.json .
snyk-dev iac test --json-file-output=output-legacy-path.json --legacy .

snyk-dev iac test --sarif-file-output=output-ga.sarif
snyk-dev iac test --sarif-file-output=output-legacy.sarif --legacy
snyk-dev iac test --sarif-file-output=output-ga-path.sarif .
snyk-dev iac test --sarif-file-output=output-legacy-path.sarif --legacy .
```

We discussed in https://snyk.slack.com/archives/CS61093QC/p1622121809384500 how the GitHub action for IaC is broken because the `physicalLocation` in the SARIF output is not relative to the `PROJECT_ROOT`, as it should be.

This PR fixes that but also improves the output so that we now include the paths correctly when we have nested directories:
e.g. I have a `nested` folder, with a `nested` folder inside it, both of which have their own `test1.tf` files. As we can see from the attached screenshots, now we can differentiate between the two better.

https://snyksec.atlassian.net/browse/CC-563 - fixes this
https://snyksec.atlassian.net/browse/CC-679 - maintains this

#### Screenshots
Can find inside these ZIP files a before and after comparison of:
- SARIF output
    - for GA code
    - for legacy code (unchanged)
- JSON output
    - for GA code
    - for legacy code (unchanged)
- terminal results
    - for GA code
    - for legacy code (unchanged)
[before.zip](https://github.com/snyk/snyk/files/6584876/before.zip)
[after.zip](https://github.com/snyk/snyk/files/6593379/after.zip)

Quick screenshot of running against a folder with path provided as `.`:
after-ga-path

Quick screenshot of running against a folder without a path provided:
after-ga

Quick screenshot of running against a folder which is located outside the current running directory:
after-relative-folder-path

Also quick screenshots of running against a single file (before vs after - no change):
before-ga-single-file

after-single-file

### DIFF
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -13,6 +13,7 @@ import { IacProjectType } from '../../../../lib/iac/constants';
 import { CustomError } from '../../../../lib/errors';
 import { extractLineNumber } from './extract-line-number';
 import { getErrorStringCode } from './error-utils';
+import { isLocalFolder } from '../../../../lib/detect';
 
 const SEVERITIES = [SEVERITY.LOW, SEVERITY.MEDIUM, SEVERITY.HIGH];
 
@@ -80,8 +81,10 @@ function formatScanResult(
     };
   });
 
-  const targetFilePath = path.resolve(scanResult.filePath, '.');
-
+  const { targetFilePath, projectName, targetFile } = computePaths(
+    scanResult.filePath,
+    options.path,
+  );
   return {
     result: {
       cloudConfigResults: filterPoliciesBySeverity(
@@ -100,13 +103,43 @@ function formatScanResult(
     dependencyCount: 0,
     licensesPolicy: null, // we do not have the concept of license policies
     ignoreSettings: null,
-    targetFile: scanResult.filePath,
-    projectName: path.basename(path.dirname(targetFilePath)),
+    targetFile,
+    projectName,
     org: meta.org,
     policy: '', // we do not have the concept of policy
     isPrivate: true,
     targetFilePath,
     packageManager: engineTypeToProjectType[scanResult.engineType],
+  };
+}
+
+function computePaths(
+  filePath: string,
+  pathArg = '.',
+): { targetFilePath: string; projectName: string; targetFile: string } {
+  const targetFilePath = path.resolve(filePath, '.');
+
+  // the absolute path is needed to compute the full project path
+  const cmdPath = path.resolve(pathArg);
+
+  let projectPath: string;
+  let targetFile: string;
+  if (!isLocalFolder(cmdPath)) {
+    // if the provided path points to a file, then the project starts at the parent folder of that file
+    // and the target file was provided as the path argument
+    projectPath = path.dirname(cmdPath);
+    targetFile = pathArg;
+  } else {
+    // otherwise, the project starts at the provided path
+    // and the target file must be the relative path from the project path to the path of the scanned file
+    projectPath = cmdPath;
+    targetFile = path.relative(projectPath, targetFilePath);
+  }
+
+  return {
+    targetFilePath,
+    projectName: path.basename(projectPath),
+    targetFile,
   };
 }
 
--- a/test/jest/acceptance/iac/file-output.spec.ts
+++ b/test/jest/acceptance/iac/file-output.spec.ts
@@ -18,7 +18,7 @@ describe('iac test --json-file-output', () => {
 
   afterAll(async () => teardown());
 
-  it('can save JSON output to file while sending human readable output to stdout', async () => {
+  it('contains a valid line number', async () => {
     const jsonOutputFilename = path.join(__dirname, `${uuidv4()}.json`);
     const { stdout } = await run(
       `snyk iac test ./iac/terraform/sg_open_ssh.tf --json-file-output=${jsonOutputFilename}`,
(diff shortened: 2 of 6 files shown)


teodora-sandu:
Also tested in GitHub by uploading the local SARIF file from running the command and it works for nested and non-nested files:
Screenshot 2021-06-03 at 16 04 07
Screenshot 2021-06-03 at 16 04 19

src/cli/commands/test/iac-output.ts
```diff
@@ -129,10 +129,15 @@ export function createSarifOutputForIac(
...
       // path, for directory tests only the filename is returned and we need
-      // too build the path manually.
-      const targetPath = res.targetFile.startsWith(res.path)
-        ? pathLib.join(res.targetFile)
-        : pathLib.join(res.path, res.targetFile);
+      // to build the path manually.
+      let targetPath: string;
+      if (res.targetFile.startsWith(res.path)) {
+        targetPath = pathLib.join(res.targetFile);
+      } else if (res.path.endsWith(res.targetFile)) {
```
teodora-sandu, line 136: This extra change was required otherwise a file would get appended to the end of a path that already contains the file:
Screenshot 2021-06-03 at 18 08 21

Before this PR this part of the code worked correctly though..

aron:
I can take a look at this, but might be later in the day.

test/jest/acceptance/iac/file-output.spec.ts
```diff
@@ -32,6 +32,75 @@ describe('iac test --json-file-output', () => {
     expect(lineNumber).not.toBeUndefined();
     expect(lineNumber).not.toEqual(-1);
   });
```
ipapast, line 34: Since the '-1' lineNumber is when we can not calculate it, is the name of the test correct or a bit misleading?

ipapast: or I might be reading it wrong? :D

ipapast (approved):
Thanks for sorting this out!

test/jest/acceptance/iac/file-output.spec.ts
teodora-sandu: This test validates that the line number is not undefined and not equal to -1 - I didn't want to use a real line number just in case we add any more rules and then these tests start failing.

Does the test make more sense now?

ipapast: It does, I think I read it wrong on Friday, and also I thought the last expect was equal and not "not equal". 😬

aron (approved):
Works great when I give it relative and absolute paths down from the current working directory. Nice job working through all of these!

I did find a couple of breaking corner cases...

Would suggest saving this script locally as sarif-path.js:

```js
const data = JSON.parse(require('fs').readFileSync(0, 'utf-8'));
const paths = new Set();
for (const run of data.runs) {
  const projectRoot = run.originalUriBaseIds.PROJECTROOT.uri;

  for (const result of run.results) {
    for (const loc of result.locations) {
      paths.add(projectRoot + loc.physicalLocation.artifactLocation.uri);
    }
  }
}

console.log(Array.from(paths).join('\n'));
```

Then run it locally with no path:

    % cd test/fixtures/iac/terraform-plan
    % snyk-dev iac test --sarif | node <path-to>/sarif-path.js
    
Output:

    file:///Users/Aron/Code/snyk/test/fixtures/iac/terraform-plan//Users/Aron/Code/snyk/test/fixtures/iac/terraform-plan/tf-plan-update.json

Also testing files outside of current subtree:

    % cd ~/Desktop
    % snyk-dev iac test <path-to-snyk>/test/fixtures/iac --sarif | node <path-to>/sarif-path.js 


```
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/cloudformation/aurora-valid.yml
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/cloudformation/fargate-valid.json
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/file-logging/file_content_logging.yaml
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/file-output/nested-folder/sg_open_ssh.tf
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/file-output/sg_open_ssh.tf
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/kubernetes/pod-privileged.yaml
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/kubernetes/pod-valid.json
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/no_vulnerabilities/pod-privileged.yaml
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/terraform/sg_open_ssh.tf
file:///Users/Aron/Desktop//Users/Aron/Code/snyk/test/fixtures/iac/terraform-plan/tf-plan-update.json
```

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -110,6 +113,36 @@ function formatScanResult(
...
+  if (!isLocalFolder(cmdPath)) {
+    // if the provided path points to a file, then the project starts at the parent folder of that file
+    // and the target file was provided as the path argument
+    projectPath = path.dirname(cmdPath);
+    targetFile = pathArg;
+  } else {
+    // otherwise, the project starts at the provided path
+    // and the target file must be the relative path from the project path to the path of the scanned file
+    projectPath = cmdPath;
+    targetFile = path.relative(projectPath, targetFilePath);
```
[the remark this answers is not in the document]
teodora-sandu: Can we ever end up in that scenario? `projectPath` is basically computed from the path that the scanned files would come from. And `targetFilePath` should be the path to one of those files. Is there a real scenario where this would happen?

teodora-sandu:
@aron The example you provided also doesn't work with the legacy code so I didn't try to fix it, I think it would require some changes around 
https://github.com/snyk/snyk/blob/f1c1b24d29265112ff95c4383d5f95b1cf5a3985/src/cli/commands/test/iac-output.ts#L133-L135

Something like the following might work, but it's getting uglier and uglier:
```
 const targetPath = res.targetFile.startsWith(res.path) || pathLib.resolve('.') === res.path
        ? pathLib.join(res.targetFile)
        : pathLib.join(res.path, res.targetFile);
```

aron:
Ah, sorry I missed the fact we were keeping the legacy behaviour. I think the core problem here is that the `path` provided to test is inconsistent depending on how you pass the paths to `snyk iac test`, in some cases it is relative, others absolute. If we have consistent file paths as input (absolute file paths) everywhere then we can remove all the workarounds.

This is a great start and fixes the most pressing issue so let's merge it. We can perhaps update the original ticket to note that there are still some edge cases that need fixing.

src/cli/commands/test/iac-local-execution/results-formatter.ts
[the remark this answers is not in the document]
aron: Hmm, I had assumed this was the problem we were seeing with providing a relative path and an absolute path with differing ancestors, but I think your later comment actually identifies the problem there.

PR #1995, opened 2021-06-03
test: migrate alerts tests to jest

ghost:


src/lib/alerts.ts
```diff
@@ -21,8 +21,8 @@ function registerAlerts(alerts: Alert[]) {
   });
 }
 
-function hasAlert(name: string) {
-  return registeredAlerts.find((a) => a.name === name);
+function hasAlert(name: string): boolean {
+  return registeredAlerts.some((a) => a.name === name);
```
jahed-snyk, line 25: `find` returns the value where as `some` returns a boolean. `hasAlert` is only used for boolean logic so changing it to `some` makes sense rather than using the result of `find` as a truthy/falsy.

gitphill (approved):

PR #2002, opened 2021-06-04
chore: improve npm version detection and don't enforce on errors

ghost:
`process.env.npm_config_user_agent` is not reliable. For example, Lerna changes it to its own thing:

https://github.com/lerna/lerna/blob/main/commands/publish/index.js#L93

So there's no reliable way to know given an `npm install` what version was used to run the install. We can only try our best. So I've improved version detection to be more strict about the user agent format and if it can't find it, it'll print out the error and not fail.


package.json
```diff
@@ -33,7 +33,8 @@
     "find-circular": "npm run build && madge --circular ./dist",
     "format": "prettier --write '{src,test,scripts}/**/*.{js,ts}'",
     "format:changes": "./scripts/format/prettier-changes.sh",
-    "prepare": "ts-node scripts/check-dev-environment.ts && npm run build",
+    "check-dev-environment": "ts-node ./scripts/check-dev-environment.ts",
+    "prepare": "npm run check-dev-environment && npm run build",
```
jahed-snyk, line 37: hopefully by using `npm run` we can increase the chance of grabbing an npm context rather than something else (whether it's lerna or some other tool).

JackuB (approved):

PR #2003, opened 2021-06-04
fix: Skip specific errors when parsing yaml IaC files

ipapast:

#### What does this PR do?

We identified a few cases that our services parse the yaml files in a different way. The node 'yaml' library is too 'strict', returning errors for validation of the YAML as well. 

In this PR we focus on two specific errors: bad indentation and duplicate keys. 
These two examples are not consistent with what Policy Engine returns so in this commit we intentionally skip errors for these cases, and return the parsed YAML document instead.

This will improve the consistency between the two parsers.

The changes are in `file-parser.ts` and `iac-parser.ts` where we check if the error is one of the two error strings.

- Create a file with a SemanticError, e.g. duplicate keys:
```
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
spec:
  containers:
  something: here
metadata:
  another: thing
```

Run `snyk-dev iac test duplicate_keys.yaml`
See that it is no longer failing but showing results (if any).

[CC-927]

#### Screenshots
Before:
image

After:
image

#### Additional questions

[CC-927]: https://snyksec.atlassian.net/browse/CC-927

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/file-parser.ts
@@ -23,6 +23,7 @@ import {
 import * as analytics from '../../../../lib/analytics';
 import { CustomError } from '../../../../lib/errors';
 import { getErrorStringCode } from './error-utils';
+import { shouldThrowErrorFor } from './file-utils';
 
 export async function parseFiles(
   filesData: IacFileData[],
@@ -69,7 +70,7 @@ function parseYAMLOrJSONFileData(fileData: IacFileData): any[] {
     // the YAML library can parse both YAML and JSON content, as well as content with singe/multiple YAMLs
     // by using this library we don't have to disambiguate between these different contents ourselves
     yamlDocuments = YAML.parseAllDocuments(fileData.fileContent).map((doc) => {
-      if (doc.errors.length !== 0) {
+      if (shouldThrowErrorFor(doc)) {
         throw doc.errors[0];
       }
       return doc.toJSON();
--- a/test/jest/unit/iac-unit-tests/file-parser.spec.ts
+++ b/test/jest/unit/iac-unit-tests/file-parser.spec.ts
@@ -27,6 +27,10 @@ import {
   expectedMultipleKubernetesYamlsParsingResult,
   invalidYamlFileDataStub,
   invalidJsonFileDataStub,
+  duplicateKeyYamlErrorFileDataStub,
+  expectedDuplicateKeyYamlErrorFileParsingResult,
+  expectedInsufficientIndentationYamlErrorFileParsingResult,
+  insufficientIndentationYamlErrorFileDataStub,
 } from './file-parser.fixtures';
 import { IacFileData } from '../../../../src/cli/commands/test/iac-local-execution/types';
 import { IacFileTypes } from '../../../../dist/lib/iac/constants';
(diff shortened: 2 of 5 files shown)


src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -69,7 +69,11 @@ function parseYAMLOrJSONFileData(fileData: IacFileData): any[] {
     // the YAML library can parse both YAML and JSON content, as well as content with singe/multiple YAMLs
     // by using this library we don't have to disambiguate between these different contents ourselves
     yamlDocuments = YAML.parseAllDocuments(fileData.fileContent).map((doc) => {
-      if (doc.errors.length !== 0) {
+      // a YAMLError can be any of these: 'YAMLReferenceError', 'YAMLSemanticError', 'YAMLSyntaxError', 'YAMLWarning'
+      // we decided to fail only on YAMLSyntaxError to be consistent with the Golang parser in Policy Engine.
+      if (doc.errors.length !== 0
+        && doc.errors[0].name === 'YAMLSyntaxError'
+      || !doc.errors[0].message.includes('expected flow map')) {
```
ipapast, line 76: I don't feel great about this line. I added it temporarily to discuss on Monday.
In this case: `{ "foo": "bar"` this would produce a FlowMap error, which is not a SyntaxError, but a SemanticError.

We need to think again how we approach these and compare them with Policy Engine.

src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -70,7 +70,15 @@ function parseYAMLOrJSONFileData(fileData: IacFileData): any[] {
     // by using this library we don't have to disambiguate between these different contents ourselves
     yamlDocuments = YAML.parseAllDocuments(fileData.fileContent).map((doc) => {
       if (doc.errors.length !== 0) {
-        throw doc.errors[0];
+        // the YAML Parser is more strict than the Golang one in Policy Engine,
+        // so we decided to skip specific errors in order to be consistent
+        const errorsToSkip = [
+          'Insufficient indentation in flow collection',
+          'Map keys must be unique',
+        ];
```
almog27, line 78: Can we extract this const outside of the map, maybe to a more shared place so it could be also used [here](https://github.com/snyk/snyk/pull/2003/files#diff-4e236e93db42d2bea509d66fbf89ab3bb0ac064b9ea08df6cf3e97a017f5ac5fR37-R40)

test/jest/unit/iac-unit-tests/file-parser.spec.ts
```diff
@@ -119,12 +121,19 @@ describe('parseFiles', () => {
...
-  it('throws an error for invalid YAML file types', async () => {
+  it('throws an error for invalid (syntax) YAML file types', async () => {
     await expect(parseFiles([invalidYamlFileDataStub])).rejects.toThrow(
       InvalidYamlFileError,
     );
   });
 
+  // the npm yaml parser by default fails on SemanticErrors like duplicate keys
+  // but we decided to skip this error in order to be consistent with the Policy Engine
+  it('parses files with SemanticYAMLError because of duplicate key correctly', async () => {
```
teodora-sandu, line 132: Can we also add a test case for  the`'Insufficient indentation in flow collection'` error? So that if the library somehow changes the error message, we are notified about it and know to update it

ipapast:
@almog27 @teodora-sandu  I extracted the check as a function and gave it a descriptive name, also added an extra test. If you could have a look again 🙇

almog27 (approved):

teodora-sandu (approved):

PR #2006, opened 2021-06-07
feat: support detour flow for c/c++ given ff

anthogez:
This pr will allows the CLI of pass to snyk-cpp-plugin a detour option to trigger a different logic if an
org's have snykSourceCpp ff enabled, this is an intentional and temporary approach to avoid breaking
changes

### DIFF
--- a/src/lib/ecosystems/enhance-options.ts
+++ b/src/lib/ecosystems/enhance-options.ts
@@ -1,0 +1,27 @@
+import { Options } from '../types';
+
+import { Ecosystem } from './types';
+import { AuthFailedError } from '../errors';
+import { isFeatureFlagSupportedForOrg } from '../feature-flags';
+
+/**
+ * @deprecated: do not use it, it's going to be deleted soon as the new snyk-cpp-plugin flow is completed
+ */
+export async function hasToDetourToLegacyFlow(
+  ecosystem: Ecosystem,
+  options: Options,
+): Promise<boolean> {
+  if (ecosystem !== 'cpp' || !options.org) {
+    return false;
+  }
+
+  const response = await isFeatureFlagSupportedForOrg(
+    'snykLegacyCpp',
+    options.org,
+  );
+  if (response.code === 401) {
+    throw AuthFailedError(response.error, response.code);
+  }
+
+  return response.ok ? true : false;
+}
--- a/src/lib/ecosystems/test.ts
+++ b/src/lib/ecosystems/test.ts
@@ -9,6 +9,7 @@ import { getPlugin } from './plugins';
 import { TestDependenciesResponse } from '../snyk-test/legacy';
 import { assembleQueryString } from '../snyk-test/common';
 import { getAuthHeader } from '../api-token';
+import { hasToDetourToLegacyFlow } from './enhance-options';
 
 export async function testEcosystem(
   ecosystem: Ecosystem,
@@ -16,26 +17,33 @@ export async function testEcosystem(
   options: Options,
 ): Promise<TestCommandResult> {
   const plugin = getPlugin(ecosystem);
+
+  const enhancedOptions = {
+    ...options,
+    detour: await hasToDetourToLegacyFlow(ecosystem, options),
+  };
+
   // TODO: this is an intermediate step before consolidating ecosystem plugins
   // to accept flows that act differently in the testDependencies step
   if (plugin.test) {
-    const { readableResult: res } = await plugin.test(paths, options);
+    const { readableResult: res } = await plugin.test(paths, enhancedOptions);
     return TestCommandResult.createHumanReadableTestCommandResult(res, '');
   }
+
   const scanResultsByPath: { [dir: string]: ScanResult[] } = {};
   for (const path of paths) {
     await spinner(`Scanning dependencies in ${path}`);
-    options.path = path;
-    const pluginResponse = await plugin.scan(options);
+    enhancedOptions.path = path;
+    const pluginResponse = await plugin.scan(enhancedOptions);
     scanResultsByPath[path] = pluginResponse.scanResults;
   }
   spinner.clearAll();
   const [testResults, errors] = await testDependencies(
     scanResultsByPath,
-    options,
+    enhancedOptions,
   );
   const stringifiedData = JSON.stringify(testResults, null, 2);
-  if (options.json) {
+  if (enhancedOptions.json) {
     return TestCommandResult.createJsonTestCommandResult(stringifiedData);
   }
   const emptyResults: ScanResult[] = [];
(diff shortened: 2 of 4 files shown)


gitphill (commented):
isn't a better pattern to pass a collection of Feature Flags? so instead of options.detour, you'd have options.flags - the plugin can make it's own decision on what that means. It's also more extensible by other ecosystems

anthogez:
> isn't a better pattern to pass a collection of Feature Flags? so instead of options.detour, you'd have options.flags - the plugin can make it's own decision on what that means. It's also more extensible by other ecosystems

Was thinking about that and went back - did not want to over design without consult @snyk/hammer as they own the cli, and yes I actually agree 100% that a list of ff would be beneficial for the plugins

test/jest/unit/lib/ecosystems/enhance-options.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { getEnhancedOptions } from '../../../../../src/lib/ecosystems/enhance-options';
+import { Options } from '../../../../../src/lib/types';
+import * as featureFlags from '../../../../../src/lib/feature-flags';
+
+describe('enhance-options', () => {
+  beforeEach(() => jest.restoreAllMocks());
```
jahed-snyk, line 6: Should be `afterEach` to avoid polluting future tests rather than just the current one.

src/lib/ecosystems/enhance-options.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+/**
+ * @deprecated: do not use it, it's going to be deleted soon as the new snyk-cpp-plugin flow is completed
+ */
+export async function getEnhancedOptions(
+  ecosystem: Ecosystem,
+  options: Options,
+): Promise<Options> {
+  return await EcosystemPlugins[ecosystem](options);
```
jahed-snyk, line 42: Since there's no compiler-level guarantee that every Ecosystem is mapped in EcosystemPlugins, we should have a fallback identity function. We can also remove the `docker` and `code` assignments as they can just fallback.

src/lib/ecosystems/enhance-options.ts
```diff
@@ -0,0 +1,43 @@
...
+  docker: async function(options: Options) {
+    return options;
+  },
+  code: async function(options: Options) {
+    return options;
+  },
+};
+
+/**
+ * @deprecated: do not use it, it's going to be deleted soon as the new snyk-cpp-plugin flow is completed
```
jahed-snyk, line 36: To ensure this is true, can we add a timer using `Date.now()` to the day we want this feature to stop working? Once the date is reached, we will always fallback to an identity function. A sort of contract enforcement for a likely future where we won't have time to come back to this.

src/lib/ecosystems/test.ts
```diff
@@ -22,6 +23,9 @@ export async function testEcosystem(
     const { readableResult: res } = await plugin.test(paths, options);
     return TestCommandResult.createHumanReadableTestCommandResult(res, '');
   }
+
+  await getEnhancedOptions(ecosystem, options);
```
jahed-snyk, line 27: I'd prefer we avoided mutation here and returned a new set of options. It would make it easier to understand what this does without having to dive into it.

However, I know mutating options is the norm currently so not blocking on this.

test/jest/unit/lib/ecosystems/enhance-options.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { getEnhancedOptions } from '../../../../../src/lib/ecosystems/enhance-options';
+import { Options } from '../../../../../src/lib/types';
+import * as featureFlags from '../../../../../src/lib/feature-flags';
+
+describe('enhance-options', () => {
```
jahed-snyk, line 5: There are no test cases to cover non-cpp ecosystems to make sure we do nothing. We should add one.

src/lib/ecosystems/test.ts
robcresswell: Its a bad norm, lets block on it. Avoid mutating please.

src/lib/ecosystems/enhance-options.ts
```diff
@@ -0,0 +1,43 @@
...
+  cpp: async function(options: Options) {
+    // temporary detour flow with ff till the new one is ready
+
+    if (options.org) {
+      const response = await isFeatureFlagSupportedForOrg(
+        'snykSourceCpp',
+        options.org,
+      );
+
+      if (response.code === 401) {
```
robcresswell, line 18: To me this is implying that all other 4xx / 5xx are okay? What happens in those cases?

anthogez: Being frankly honesty? As this code will be deleted as soon as the new flow will be introduced in `snyk-cpp-plugin` I temporarily reused this logic https://github.com/snyk/snyk/blob/master/src/lib/monitor/index.ts#L126

src/lib/ecosystems/enhance-options.ts
robcresswell: I don't really understand this. If its going away then why do we need all this logic for `code` and `docker`? This PR feels generally concerning to me with lots of talk of "temporary" code. Its a published version, which someone may well just pin in their CI system. There's no "temporary" here; its a contract.

Perhaps a safer route is to give every org in Snyk, including new ones a `snykLegacyCpp` flag, and if true we use the old behaviour, otherwise we use the new one. Then once that is deleted, everyone is just on the new flow.

This does really flag up that the CLI needs a proper deprecation / support policy.

anthogez: Well as you just flagged the CLI does not have a `proper deprecation / support policy.`.
And this `temporary` talk that you mentioned is the reason behind that. 

Happy to take in consideration the proposed solution even if I think that giving a ff for the new flow seems less problematic

anthogez: - ` 🚧 WIP label 9 hours ago ` and as I stated before https://github.com/snyk/snyk/pull/2006#issuecomment-856692457
was thinking about a feature flag flow (also raised internally) but I went back on my decision and I've just reduced that code.

- is that valid even for a closed beta? If yes I agree 100% with your point here

vikki: If flipping the feature flag around (as Rob suggested) allows us to safely remove the feature flag, and its just a case of inverting the logic and renaming it, that feels much safer, without too much work, right? 
We can (/need to) figure out more of proper deprecation policy for future work though (and we could also look at making isFeatureFlagSupportedForOrg a bit easier to work with)

anthogez: @vikki @robcresswell - feature flag was renamed to `snykLegacyCpp`
I will not merge this one till, it's spread across all the orgs.

Thanks for your feedback here - hope that we can have a proper policy soon.

JackuB (approved):

ghost (approved):

anthogez:
No longer necessary as we will swap out the legacy flow with new one.

PR #2009, opened 2021-06-08
test: migrate package-no-name test to jest

ghost:


test/jest/acceptance/snyk-test/package-no-name.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+test('packages with no name read dir', async () => {
+  const project = await createProject('package-sans-name');
+  const { code } = await runSnykCLI('test', {
+    cwd: project.path(),
+    env: {
+      ...process.env,
+      SNYK_TOKEN: requireSnykToken(),
+    },
+  });
+  expect(code).toEqual(1);
```
Avishagp, line 16: Why 1 if it should succeed?

jahed-snyk: Success in the old tap test wasn't based on no vulns, it was based on no errors (0 or 1). In the case of the fixtures being used, they have vulns so it's a 1.

Avishagp (approved):

PR #2019, opened 2021-06-11
[NEB-10] feat: track, check and enforce snyk code test limit on cli

neonnoon:
- [x] tests done

#### What does this PR do?

- validates if sast test limit has been reached
- adds error an alerts (if close to limit, or limits are not enforced)

#### Screenshots

```
❯ snyk code test
Test limit reached!

You have reached your monthly limit of 2 private tests for your org.
To learn more about our plans and increase your tests limit visit https://snyk.io/plans.
You have reached your monthly limit of 24 code tests for your org.
To learn more about our plans and increase your tests limit visit https://snyk.io/plans.
```

### DIFF
--- a/src/lib/plugins/sast/checks.ts
+++ b/src/lib/plugins/sast/checks.ts
@@ -9,6 +9,11 @@ interface SastSettings {
   error?: string;
 }
 
+interface TrackUsageResponse {
+  code?: number;
+  error?: string;
+}
+
 export async function getSastSettingsForOrg(org): Promise<SastSettings> {
   const response = await request({
     method: 'GET',
@@ -17,6 +22,21 @@ export async function getSastSettingsForOrg(org): Promise<SastSettings> {
     },
     qs: assembleQueryString({ org }),
     url: `${config.API}/cli-config/settings/sast`,
+    gzip: true,
+    json: true,
+  });
+
+  return (response as any).body;
+}
+
+export async function trackUsage(org): Promise<TrackUsageResponse> {
+  const response = await request({
+    method: 'POST',
+    headers: {
+      Authorization: `token ${snyk.api}`,
+    },
+    qs: assembleQueryString({ org }),
+    url: `${config.API}/track-sast-usage/cli`,
     gzip: true,
     json: true,
   });
--- a/src/lib/plugins/sast/checks.ts
+++ b/src/lib/plugins/sast/checks.ts
@@ -1,5 +1,5 @@
 import request = require('../../request');
-import snyk = require('../..');
+import { api as getApiToken } from '../../api-token';
 import * as config from '../../config';
 import { assembleQueryString } from '../../snyk-test/common';
 
@@ -18,7 +18,7 @@ export async function getSastSettingsForOrg(org): Promise<SastSettings> {
   const response = await request({
     method: 'GET',
     headers: {
-      Authorization: `token ${snyk.api}`,
+      Authorization: `token ${getApiToken()}`,
     },
     qs: assembleQueryString({ org }),
     url: `${config.API}/cli-config/settings/sast`,
@@ -33,7 +33,7 @@ export async function trackUsage(org): Promise<TrackUsageResponse> {
   const response = await request({
     method: 'POST',
     headers: {
-      Authorization: `token ${snyk.api}`,
+      Authorization: `token ${getApiToken()}`,
     },
     qs: assembleQueryString({ org }),
     url: `${config.API}/track-sast-usage/cli`,
--- a/src/lib/plugins/sast/checks.ts
+++ b/src/lib/plugins/sast/checks.ts
@@ -11,7 +11,7 @@ interface SastSettings {
 
 interface TrackUsageResponse {
   code?: number;
-  error?: string;
+  userMessage?: string;
 }
 
 export async function getSastSettingsForOrg(org): Promise<SastSettings> {
@@ -26,7 +26,7 @@ export async function getSastSettingsForOrg(org): Promise<SastSettings> {
     json: true,
   });
 
-  return (response as any).body;
+  return response.body as SastSettings;
 }
 
 export async function trackUsage(org): Promise<TrackUsageResponse> {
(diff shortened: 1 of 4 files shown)


src/lib/plugins/sast/checks.ts
```diff
@@ -23,3 +28,18 @@ export async function getSastSettingsForOrg(org): Promise<SastSettings> {
...
 }
+
+export async function trackUsage(org): Promise<TrackUsageResponse> {
+  const response = await request({
+    method: 'POST',
+    headers: {
+      Authorization: `token ${snyk.api}`,
+    },
+    qs: assembleQueryString({ org }),
+    url: `${config.API}/track-sast-usage/cli`,
```
miiila, line 39: `question`: is `track-sast-usage` route implemented in registry? Just curious if it is safe to start calling it :-)

miiila (approved):

src/lib/plugins/sast/checks.ts
neonnoon: Yes, see https://github.com/snyk/registry/pull/21469

test/jest/unit/snyk-code-test.spec.ts
```diff
@@ -202,6 +207,23 @@ describe('Test snyk code', () => {
...
+    isFeatureFlagSupportedForOrgSpy.mockResolvedValue({
+      ok: true,
+    });
+    trackUsageSpy.mockResolvedValueOnce({
+      code: 429,
+      userMessage: 'Test limit reached!',
+    });
+
+    try {
+      await cli.test('some/path', { code: true });
```
jahed-snyk, line 221: This will pass if `cli.test` does not throw. You should use something like:

```ts
await expect(cli.test('some/path', { code: true }))
  .rejects.toHaveProperty('userMessage', 'Test limit reached!')
```

src/lib/plugins/sast/checks.ts
```diff
@@ -23,3 +28,18 @@ export async function getSastSettingsForOrg(org): Promise<SastSettings> {
 
   return (response as any).body;
 }
+
+export async function trackUsage(org): Promise<TrackUsageResponse> {
+  const response = await request({
+    method: 'POST',
+    headers: {
+      Authorization: `token ${snyk.api}`,
```
JackuB, line 36: I think using a getter is safer here. It's defined in `api-token.ts` 

```suggestion
      Authorization: `token ${getApiToken()}`,
```

JackuB (approved):

JackuB:
Commit message could mention this affects only sast

src/lib/plugins/sast/checks.ts
```diff
@@ -23,3 +28,18 @@ export async function getSastSettingsForOrg(org): Promise<SastSettings> {
...
+    },
+    qs: assembleQueryString({ org }),
+    url: `${config.API}/track-sast-usage/cli`,
+    gzip: true,
+    json: true,
+  });
+
+  return (response as any).body;
```
jahed-snyk, line 44: Why do you need `as any` here? Can we make this type-safe? It looks like this response only has `userMessage` and `code` so it should not be difficult to parse.

Also it looks like you're not handling HTTP response codes (if the request fails) which will lead to an ambigious error. Maybe on non-ok status codes, there should be a fallback response with a user-friendly `userMessage`?

src/lib/plugins/sast/validate.ts
```diff
@@ -46,4 +50,12 @@ export async function validateCodeTest(options: Options) {
       'enable in Settings > Snyk Code',
     );
   }
+
+  const trackUsageResponse = await trackUsage(org);
+  if (trackUsageResponse.code === 429) {
+    throw new FailedToRunTestError(
+      trackUsageResponse.userMessage,
```
jahed-snyk, line 57: `trackUsageResponse` is a `TrackUsageResponse` type, but `TrackUsageResponse` does not have a `userMessage`. Should we define it? Where is it coming from?

src/lib/plugins/sast/checks.ts
```diff
@@ -9,6 +9,11 @@ interface SastSettings {
   error?: string;
 }
 
+interface TrackUsageResponse {
+  code?: number;
+  error?: string;
```
jahed-snyk, line 14: This `error` is not used anywhere by the looks of it. Did you mean `userMessage`?

ghost (changes requested):

src/lib/plugins/sast/checks.ts
JackuB: Wouldn't it be better to make this fail silently? It's only checking for the status code 429?

jahed-snyk: @JackuB I guess if the `body` is validated and falls back to a predefined body (whether that's an okay or a different failure, I'll leave that to SAST team to decide), it should handle any potential server issues. Something like:

```ts
try {
  return getTrackUsageResponse(response) // validate `code` and `userMessage` types
} catch (e) {
  debug(e)
  return okayTrackingUsageResponse
}
```

My main concern here is that `body` is not being validated so it can potentially be anything which defeats the point of using TypeScript.

test/jest/unit/snyk-code-test.spec.ts
neonnoon: Good point, fixed my new test and previous ones with the same issue.

src/lib/plugins/sast/checks.ts
neonnoon: There's (at least 😛 ) two problems here:
- we had a weird cyclic import for the api token which required! us to use require and thus we didn't have types -> fixed
- some of the types were incorrect (because not properly checked) -> fixed

What remains is because `request.ts` doesn't use generics we still need to assert the reponse type here, right?

src/lib/plugins/sast/validate.ts
neonnoon: Fixed

src/lib/plugins/sast/checks.ts
jahed-snyk: Yes, but you can make it typesafe instead of using `any`. This might need a larger discussion (we use `any` a lot in places) so I won't block on this.

ghost (approved):

Issue #1487, opened 2020-10-21
Pipfile: test should not update Pipfile.lock? 


damycra:
I run snyk manually\* from time to time in dev using the python-3 Docker image (lightly modified to allow psycopg2 to install, as I recall) but recently I was surprised to see my _Pipfile.lock_ being updated by the process. Therefore the test runs against an updated _Pipfile.lock_ and **not** the _Pipfile.lock_ that's committed (and is probably being built by Jenkins). 

Should it be
~~pipenv lock~~
`pipenv install --system --deploy`
https://github.com/snyk/snyk/blob/f35f39e96ef7aa69b22a846315dda015b12a4564/docker/docker-python-entrypoint.sh#L25

See also [Using pipenv for deployments](https://pipenv.pypa.io/en/latest/advanced/#using-pipenv-for-deployments).

Of course, my assumption is that _Pipfile.lock_ is always committed. Perhaps it should indeed run `pipenv lock` if one does not exist although this reduces the accuracy of the result. 

Happy to submit PR if desirable.

\* I appreciate that this is not how I should be doing it, so feel free to ignore me entirely! I `pipenv update` in dev then run tests locally before committing the _Pipfile.lock_ — then Jenkins takes over. If I run snyk at the beginning of this process, all is well; but if snyk was running in CI the problem recurs (with a small but definitely finite chance that libs will have updated between my local `pipenv update` and `snyk test`).

lili2311:
hi @damycra I can't think of any reason why it is doing that and I added it 🤔 If you are able to raise the PR to remove this, I can check with the relevant team to make sure they can review and are happy with this change.

damycra:
Thanks @lili2311 — 
the only reason I could see was if there was no Pipfile.lock, but I am conscious that I spend very little time on `pipenv` and only as an _end-user_ i.e. not a library maintainer or anything. So there may be something else hiding in there and I would hate to break everyone's scans! No worries if the risk/reward case for this chance can't be made, I am happy with my workflow for now.  

It seems like a `pipenv` bug that a harmless Mac only dependency is dropped by `pipenv lock`, so perhaps I should raise that with them, too.

lili2311:
@damycra your contribution has now been merged 🎉 Did that solve your issue? Can we close this?

damycra:
Thanks @lili2311 yes that fixes my issue. (While I'm here, are there any plans to push the Dockerfile.python-3.8 artefact to DockerHub?)

[...]

PR #2021, opened 2021-06-11
Fix: do not modify Pipfile.lock #1487

damycra:

#### What does this PR do?

Prevents _Pipfile.lock_ from being updated during a scan. 

Consider the behaviour of the different `pipenv` commands and flags against the expected input and output requirements (explicitly, I presume that a _Pipfile.lock_ must be present, based on external advice received)

Craft a _Pipfile_ with dependencies that would ordinarily be removed by `pipenv lock` (e.g. a Mac only dependency) and create the corresponding _Pipfile.lock_. Observe that the _Pipfile.lock_ is unchanged at the end of the scan.

Craft an outdated _Pipfile.lock_ and observe that the scan fails.

Check that the behaviour where a _Pipfile.lock_ is not present at the start is acceptable (i.e. presumably it will fail).  

I develop on Mac and run a snyk test in a lightly modified official Python 3 container, with my source mounted. This invariably modifies _Pipfile.lock_ (removing a Mac only dependency and adding a couple of others). I do not commit these changes and later Jenkins (on linux) builds containers for me apparently without any issues, from a fresh checkout.    

Without some change like this, there is presumably a chance that snyk could update dependencies thus fixing a security issue before a scan, while the committed _Pipfile.lock_ still contained the insecure library spec.    

#1487

### DIFF
--- a/docker/docker-python-entrypoint.sh
+++ b/docker/docker-python-entrypoint.sh
@@ -22,8 +22,7 @@ installRequirementsTxtDeps() {
 installPipfileDeps() {
     pushd "${PROJECT_PATH}/"
     echo "Found Pipfile"
-    pipenv lock
-    pipenv install --system
+    pipenv install --system --deploy
     popd
 }
 


aviadhahami:
Hi @damycra!
Thank you so much for the contribution :) 

This PR LGTM, so I'll continue with merging it via https://github.com/snyk/snyk/pull/2038 (in order to run ci jobs)
Will be closing this now in favor of the latter.

Thank you!

PR #2025, opened 2021-06-15
feat: resolveAndTestFacts with async polling

anthogez:
Introduces async request for C/C++ with polling, where its Facts are resolved/computed in our platform

### DIFF
--- a/src/lib/ecosystems/polling.ts
+++ b/src/lib/ecosystems/polling.ts
@@ -1,0 +1,120 @@
+import * as config from '../config';
+import { isCI } from '../is-ci';
+import { makeRequest } from '../request/promise';
+import { Options } from '../types';
+
+import { ResolveAndTestFactsResponse } from '../snyk-test/legacy';
+import { assembleQueryString } from '../snyk-test/common';
+import { getAuthHeader } from '../api-token';
+import { ScanResult } from './types';
+
+export async function requestPollingToken(
+  options: Options,
+  isAsync: boolean,
+  scanResult: ScanResult,
+): Promise<ResolveAndTestFactsResponse> {
+  const payload = {
+    method: 'POST',
+    url: `${config.API}/test-dependencies`,
+    json: true,
+    headers: {
+      'x-is-ci': isCI(),
+      authorization: getAuthHeader(),
+    },
+    body: {
+      isAsync,
+      scanResult,
+    },
+    qs: assembleQueryString(options),
+  };
+  return await makeRequest<ResolveAndTestFactsResponse>(payload);
+}
+
+export async function pollingWithTokenUntilDone(
+  token: string,
+  type: string,
+  options: Options,
+  pollInterval: number,
+  attemptsCount: number,
+  maxAttempts = Infinity,
+): Promise<ResolveAndTestFactsResponse> {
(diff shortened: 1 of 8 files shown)


src/lib/ecosystems/test.ts
```diff
@@ -30,10 +31,43 @@ export async function testEcosystem(
...
+      options,
+    );
+    return await getTestResultsOutput(
+      errors,
+      options,
+      testResults,
+      plugin,
+      scanResultsByPath,
+    );
+  }
```
anthogez, line 48: At the current stage I preferred to keep this logic as part of `testDependencies` instead of create a completely new flow for `resolveAndTestFacts`, have time to change it when required

src/lib/ecosystems/test.ts
```diff
@@ -53,6 +87,42 @@ export async function testEcosystem(
...
+        if (!token) {
+          throw 'Something went wrong, token n/a';
+        }
+        const response = await pollingWithTokenUntilDone(
+          token,
+          ecosystem,
+          options,
+          2000,
+        );
+        results.push(response);
```
anthogez, line 113: Keeping the current approach where we test/resolve one scanResult at the time.

Depending of resolution `AVG`  time for C/C++ unmanaged, we can change it to something like:

- Please generate a token (resolution process) for all the scanResults (if they are resolvable kind)
   then go over that list and keep polling each token in order to see if any is available, if yes then take out of the list till all the scanResults' token are processed either successfully or not

test/jest/unit/lib/ecosystems/test.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const [testResults, errors] = await resolveAndTestFacts(
+      'cpp',
+      scanResults,
+      {} as Options,
+    );
+
+    expect(testResults).toEqual([
+      {
+        components: [
```
anthogez, line 73: this will change to depGraph & issuesData

test/jest/unit/lib/ecosystems/test.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  it('resolveAndTestFacts', async () => {
+    /* eslint-disable @typescript-eslint/camelcase */
+    const scanResults = {
+      path: [
+        {
+          name: 'my-unmanaged-c-project',
+          facts: [
+            {
+              type: 'signatures',
```
robguinness-snyk, line 15: We called this `'cpp-signatures'` in feat/support-c-addon here.

src/lib/ecosystems/polling.ts
```diff
@@ -0,0 +1,84 @@
...
+        await pollingWithTokenUntilDone(
+          token,
+          type,
+          options,
+          pollInterval,
+          maxAttempts,
+          attemptsCount,
+        ),
+      );
+    }),
```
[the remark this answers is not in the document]
anthogez: 🤦 yes, thanks!

test/jest/unit/lib/ecosystems/test.spec.ts
anthogez: Good point, we are moving to `fileSignatures`

PR #2030, opened 2021-06-15
fix(protect): handle package paths in quotes

maxjeffos:

#### What does this PR do?
Makes `@snyk/protect` handle `.snyk` files with package paths which are wrapped in either single or double quotes.

The reason for this is because our Snyk Fix PR generates `.snyk` files which have single quotes. For example: 
```
# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
version: v1.19.0
ignore: {}
# patches apply the minimum changes required to fix a vulnerability
patch:
  SNYK-JS-LODASH-567746:
    - '@snyk/patchable-dep-fixture > lodash':
        patched: '2021-06-14T22:42:43.116Z'
```

### DIFF
--- a/packages/snyk-protect/src/lib/snyk-file.ts
+++ b/packages/snyk-protect/src/lib/snyk-file.ts
@@ -30,7 +30,8 @@ export function extractPatchMetadata(
               .substring(1)
               .split('>')
               .pop()
-              ?.trim();
+              ?.trim()
+              ?.replace(/['"]/g, '');
             if (!acc[writingTo].includes(destination)) {
               acc[writingTo].push(destination);
             }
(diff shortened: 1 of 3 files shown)


Avishagp (approved):

packages/snyk-protect/src/lib/snyk-file.ts
```diff
@@ -26,11 +26,17 @@ export function extractPatchMetadata(
           acc[key] = [];
         } else {
           if (key.startsWith('-')) {
-            const destination = key
+            let destination = key
               .substring(1)
               .split('>')
               .pop()
               ?.trim();
+            if (destination?.startsWith("'") || destination?.startsWith('"')) {
```
jahed-snyk, line 34: I think you can be a bit simpler here and just `replaceAll` `'` and `"` characters with empty string. Saves having to do the boolean logic and substrings.

maxjeffos: done, thanks @jahed-snyk

PR #2031, opened 2021-06-15
feat(protect): add successful message for protect

maxjeffos:

#### What does this PR do?
Adds the message `Successfully applied Snyk patches`, like the original protect implementation.

### DIFF
--- a/packages/snyk-protect/src/lib/index.ts
+++ b/packages/snyk-protect/src/lib/index.ts
@@ -84,6 +84,8 @@ async function protect(projectFolderPath: string) {
     });
   });
 
+  console.log('Successfully applied Snyk patches');
+
   sendAnalytics({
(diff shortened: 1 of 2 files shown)


packages/snyk-protect/src/lib/index.ts
```diff
@@ -84,6 +84,10 @@ async function protect(projectFolderPath: string) {
     });
   });
 
+  if (patchedModules.length > 0) {
+    console.log('Successfully applied Snyk patches');
```
jahed-snyk, line 88: We don't need this in another `if` as at this point we can assume we've patched something (otherwise we should've returned at the "Nothing to patch" branch above)

maxjeffos: Fixed, thanks @jahed-snyk

jahed-snyk (approved):

PR #2034, opened 2021-06-16
feat: adjust sast sarif output to github standard

ArturSnyk:

This Pr updates `code-client` library and adjust the sast sarif output, to support `message.text` the way github use it.
meaning that if we would have arguments for our message, it will appear only on `message.markdown`, where before this PR we had it on the `message.text` field.
This aligns with How github use sarif


aviadhahami (approved):
lgtm [loki]

pkey (approved):

PR #2042, opened 2021-06-18
fix: handle container no image name provided [CAP-228]

iuliams:
#### Missed Anything?
- [x] Tests updated/written
- [x] merge only after snyk-docker-plugin is merged

#### What does this PR do?
This PR does the following:
* checks if the `container test` and `container monitor` commands receive an argument (the container image name) that shall be analysed;
* handles the errors returned by the scanning functions of the test and monitor commands

Old:
```
$ snyk container test
{"type":"Buffer","data":[]}
```
New:
```
$ snyk container test
Could not detect an image. Specify an image name to scan and try running the command again.
```
Same for the `monitor` command.

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -34,7 +34,11 @@ import { convertMultiResultToMultiCustom } from '../../../lib/plugins/convert-mu
 import { convertSingleResultToMultiCustom } from '../../../lib/plugins/convert-single-splugin-res-to-multi-custom';
 import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
 import { getContributors } from '../../../lib/monitor/dev-count-analysis';
-import { FailedToRunTestError, MonitorError } from '../../../lib/errors';
+import {
+  FailedToRunTestError,
+  MonitorError,
+  MissingArgError,
+} from '../../../lib/errors';
 import { isMultiProjectScan } from '../../../lib/is-multi-project-scan';
 import { getEcosystem, monitorEcosystem } from '../../../lib/ecosystems';
 import { getFormattedMonitorOutput } from '../../../lib/ecosystems/monitor';
@@ -73,6 +77,12 @@ async function monitor(...args0: MethodArgs): Promise<any> {
 
   if (options.docker && options['remote-repo-url']) {
     throw new Error('`--remote-repo-url` is not supported for container scans');
+  }
+
+  // Handles no image arg provided to the container command until
+  // a validation interface is implemented in the docker plugin.
+  if (options.docker && paths.length === 0) {
+    throw new MissingArgError();
   }
 
   apiOrOAuthTokenExists();
--- a/src/cli/commands/process-command-args.ts
+++ b/src/cli/commands/process-command-args.ts
@@ -10,8 +10,8 @@ export function processCommandArgs<CommandOptions>(
   }
   args = args.filter(Boolean);
 
-  // populate with default path (cwd) if no path given
-  if (args.length === 0) {
+  // For repository scanning, populate with default path (cwd) if no path given
+  if (args.length === 0 && !options.docker) {
     args.unshift(process.cwd());
   }
 
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -5,6 +5,7 @@ import * as pathLib from 'path';
 const cloneDeep = require('lodash.clonedeep');
 const assign = require('lodash.assign');
 import chalk from 'chalk';
+import { MissingArgError } from '../../../lib/errors';
 
 import * as snyk from '../../../lib';
 import { isCI } from '../../../lib/is-ci';
@@ -67,6 +68,12 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
   const options = setDefaultTestOptions(originalOptions);
   validateTestOptions(options);
   validateCredentials(options);
+
+  // Handles no image arg provided to the container command until
+  // a validation interface is implemented in the docker plugin.
+  if (options.docker && paths.length === 0) {
+    throw new MissingArgError();
+  }
 
   const ecosystem = getEcosystemForTest(options);
   if (ecosystem) {
--- a/src/lib/ecosystems/monitor.ts
+++ b/src/lib/ecosystems/monitor.ts
@@ -52,6 +52,9 @@ export async function monitorEcosystem(
       ) {
         throw new DockerImageNotFoundError(path);
       }
+      if (ecosystem === 'docker' && error.message === 'invalid image format') {
+        throw new DockerImageNotFoundError(path);
+      }
 
       throw error;
     } finally {
--- a/src/lib/errors/missing-arg-error.ts
+++ b/src/lib/errors/missing-arg-error.ts
@@ -1,0 +1,11 @@
+import { CustomError } from './custom-error';
+
+export class MissingArgError extends CustomError {
+  constructor() {
+    const msg =
+      'Could not detect an image. Specify an image name to scan and try running the command again.';
+    super(msg);
+    this.code = 422;
+    this.userMessage = msg;
+  }
+}
--- a/src/lib/snyk-test/assemble-payloads.ts
+++ b/src/lib/snyk-test/assemble-payloads.ts
@@ -9,6 +9,7 @@ import { assembleQueryString } from './common';
 import spinner = require('../spinner');
 import { findAndLoadPolicyForScanResult } from '../ecosystems/policy';
 import { getAuthHeader } from '../../lib/api-token';
+import { DockerImageNotFoundError } from '../errors';
 
 export async function assembleEcosystemPayloads(
   ecosystem: Ecosystem,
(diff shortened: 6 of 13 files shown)


src/cli/commands/test/index.ts
```diff
@@ -265,6 +266,11 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
     error.code = errorResults[0].code;
     error.userMessage = errorResults[0].userMessage;
     error.strCode = errorResults[0].strCode;
+    if (options.docker) {
```
iuliams, line 269: Ok, this is meant to fix the following corner case:
$ snyk container test /image
Since 'container test' not handled as the other commands that use the `ecosystem` [library](https://github.com/snyk/snyk/tree/master/src/lib/ecosystems), I decided to approach it in this manner as I'm not sure how big the change is otherwise on the 'docker-plugin' side. @RotemS please let me know your thoughts and if you see a nicer way to do this.

src/cli/commands/test/index.ts
```diff
@@ -29,6 +29,7 @@ import {
   summariseErrorResults,
```
iuliams, line 29: Added this as a separate commit as it may have to be re-iterated.

src/lib/ecosystems/monitor.ts
```diff
@@ -52,8 +53,7 @@ export async function monitorEcosystem(
       ) {
         throw new DockerImageNotFoundError(path);
       }
-
-      throw error;
+      throw new InternalServerError("Scanning failed. Please make sure you are using the right parameters. If the issue still persists, please contact snyk support.");
```
iuliams: this is meant to fix e.g the following corner case (image name starting with '/'):
$ snyk container monitor /image

src/cli/commands/test/index.ts
iuliams: Hm actually, I just noticed that by simply using the ecosystem lib with the inner checks, this seems to scan the image correctly...:
```
--- a/src/lib/ecosystems/index.ts
+++ b/src/lib/ecosystems/index.ts
@@ -20,6 +20,9 @@ export function getEcosystemForTest(options: Options): Ecosystem | null {
   if (options.code) {
     return 'code';
   }
+  if (options.docker) {
+    return 'docker';
+  }
   return null;
 }
```

src/lib/errors/missing-arg-error.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomError } from './custom-error';
+
+export class MissingArgError extends CustomError {
+  constructor(command: string) {
+    const msg = `Could not detect an image to ${command}. Specify an image name to scan and try running the command again.`;
```
jahed-snyk, line 5: Not blocking on this one:

It looks like we only need `method`/`command` to generate this message. Do we really need `method`? I think given the context of what the user ran, we can just say:

```
$ snyk test --docker
Could not detect an image. Specify an image name to scan and try running the command again.
```

This also avoids having to change method signatures for `processCommandArgs` and expose the concept of a "method" to it.

ghost (dismissed):
Need to update tests but other than that looks good.

src/lib/ecosystems/monitor.ts
```diff
@@ -52,8 +53,11 @@ export async function monitorEcosystem(
       ) {
         throw new DockerImageNotFoundError(path);
       }
-
-      throw error;
+      if (ecosystem === 'docker') {
+        throw new InternalServerError(
+          'Scanning failed. Please make sure you are using the right parameters. If the issue still persists, please contact snyk support.',
+        );
+      }
```
jahed-snyk, line 60: Do we have test coverage for this? We did we stop throwing `error`?

src/cli/commands/process-command-args.ts
```diff
@@ -10,9 +10,11 @@ export function processCommandArgs<CommandOptions>(
...
   args = args.filter(Boolean);
 
-  // populate with default path (cwd) if no path given
   if (args.length === 0) {
-    args.unshift(process.cwd());
+    // For repository scanning, populate with default path (cwd) if no path given
+    if (!options.docker) {
+      args.unshift(process.cwd());
+    }
   }
```
jahed-snyk, line 18: Should we combine into a single `if` statement?

src/lib/snyk-test/assemble-payloads.ts
```diff
@@ -70,6 +71,17 @@ export async function assembleEcosystemPayloads(
     }
 
     return payloads;
+  } catch (error) {
+    if (ecosystem === 'docker' && error.message === 'authentication required') {
+      throw new DockerImageNotFoundError(options.path);
+    } else if (
```
jahed-snyk, line 77: Don't need the `else` here as previous statement is throwing. Might make it more readable as independant `if` statements.

src/cli/commands/monitor/index.ts
```diff
@@ -75,6 +79,10 @@ async function monitor(...args0: MethodArgs): Promise<any> {
     throw new Error('`--remote-repo-url` is not supported for container scans');
   }
```
iuliams, line 81: Add comment why we have to do this.

src/lib/snyk-test/assemble-payloads.ts
iuliams, line 77: -> if

test/process-command-args.spec.ts
```diff
@@ -9,6 +9,15 @@ describe('display help message', () => {
     expect(paths).toEqual([defaultCurrentPath]);
   });
 
+  it('should return expected options & paths when path is undefined and is a container run', () => {
```
jahed-snyk, line 12: Would be nice if these test cases were phrased with the expected result rather than just saying "expected" e.g. "should return no paths when running a container scan with no path"

src/cli/commands/test/index.ts
```diff
@@ -68,6 +69,10 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
   validateTestOptions(options);
   validateCredentials(options);
 
+  if (options.docker && paths.length === 0) {
+    throw new MissingArgError();
+  }
```
jahed-snyk, line 74: It looks like this `test` function is trying to be Ecosystem agnostic (other than IAC) as part of Project Envelope IIRC. Should this statement be in `snyk-docker-plugin` (which is executed via `testEcosystem`)?

ghost (commented):
I've left some comments. None of them are blocking this except the comment about Project Envelope. I don't have enough context on that so I'll see if someone else from the team can review my review.

src/cli/commands/test/index.ts
iuliams: Yes, totally agree and tried to make my way around it, but unfortunately for this specific check I cannot see another way because the docker plugin scan() is not even called, since there is no 'path' argument that it has to scan.
Let me know if you find the approach and added comment explanation ok for this scenario.

iuliams: See e.g for monitoring:
https://github.com/snyk/snyk/blob/master/src/lib/ecosystems/monitor.ts#L41
Path is empty -> no plugin scan() -> with the current implementation it cannot be done in snyk-docker-plugin.
One idea would be to implement an extra validation plugin handler, but this would probably be another issue.

ghost (approved):

src/cli/commands/test/index.ts
jahed-snyk: That makes sense. I've approved this PR 👍

I agree, seems we're missing an ecosystem validation step or we're making too many assumptions around the concept of "paths". Something to improve on in the future.

src/lib/ecosystems/monitor.ts
```diff
@@ -52,6 +52,9 @@ export async function monitorEcosystem(
       ) {
         throw new DockerImageNotFoundError(path);
       }
+      if (ecosystem === 'docker' && error.message === 'invalid image format') {
+        throw new DockerImageNotFoundError(path);
```
gitphill, line 56: looks same as exception thrown above - make one predicate that tests whether this exception should be thrown?

src/lib/snyk-test/assemble-payloads.ts
```diff
@@ -70,6 +71,15 @@ export async function assembleEcosystemPayloads(
     }
 
     return payloads;
+  } catch (error) {
+    if (ecosystem === 'docker' && error.message === 'authentication required') {
+      throw new DockerImageNotFoundError(options.path);
+    }
+    if (ecosystem === 'docker' && error.message === 'invalid image format') {
+      throw new DockerImageNotFoundError(options.path);
+    }
```
gitphill, line 80: same as https://github.com/snyk/snyk/pull/2042/files#r662954528

gitphill (approved):

agranado2k (commented):
I think it's missing the `snyk-docker-plugin` version being bumped.

src/lib/errors/missing-arg-error.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomError } from './custom-error';
+
+export class MissingArgError extends CustomError {
+  constructor() {
+    const msg = `Could not detect an image. Specify an image name to scan and try running the command again.`;
```
michael-c-snyk, line 5: Is the use of ``` deliberate here? I guess the linter is happy /shrug

src/lib/ecosystems/monitor.ts
iuliams: Thought about it, but then we discussed it's a bit more readable as such (without combining 4 conditions in one check). Wdyt?

gitphill (approved):

orsagie (approved):

PR #2051, opened 2021-06-22
feat: filter out non vulnerable test results

lili2311:

#### What does this PR do?
- Immediately filter out non-vulnerable test results in @snyk/fix (results with no issu
es at all) and do not try to fix any of them. 
- Add a line to the summary that states how many were not vulnerable.

https://github.com/snyk/snyk/compare/feat/hide-non-vulnerable-projects?expand=1#diff-efd77b5a15a1593cba78b737b731cbd8406edfbd0045f62c6226db58b1b20b45R34

#### Screenshots
**Before**
CleanShot 2021-06-22 at 18 21 26@2x

**After**
CleanShot 2021-06-22 at 18 18 18@2x

CleanShot 2021-06-23 at 12 54 30@2x

### DIFF
--- a/packages/snyk-fix/src/lib/output-formatters/show-results-summary.ts
+++ b/packages/snyk-fix/src/lib/output-formatters/show-results-summary.ts
@@ -1,7 +1,8 @@
 import * as chalk from 'chalk';
+import stripAnsi = require('strip-ansi');
 
 import { FixHandlerResultByPlugin } from '../../plugins/types';
-import { ErrorsByEcoSystem, Issue, TestResult } from '../../types';
+import { ErrorsByEcoSystem, FixOptions, Issue, TestResult } from '../../types';
 import { contactSupportMessage, reTryMessage } from '../errors/common';
 import { convertErrorToUserMessage } from '../errors/error-to-user-message';
 import { hasFixableIssues } from '../issues/fixable-issues';
@@ -14,6 +15,7 @@ export const PADDING_SPACE = '  '; // 2 spaces
 export async function showResultsSummary(
   resultsByPlugin: FixHandlerResultByPlugin,
   exceptionsByScanType: ErrorsByEcoSystem,
+  options: FixOptions,
 ): Promise<string> {
   const successfulFixesSummary = generateSuccessfulFixesSummary(
     resultsByPlugin,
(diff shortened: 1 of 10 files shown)


packages/snyk-fix/src/lib/output-formatters/show-results-summary.ts
```diff
@@ -12,6 +12,7 @@ import { formatUnresolved } from './format-unresolved-item';
 export const PADDING_SPACE = '  '; // 2 spaces
 
 export async function showResultsSummary(
+  notVulnerable: EntityToFix[],
```
aarlaud, line 15: Why is notVulnerable not going through the same groupEntitiesPerScanType function? Feels a bit strange to have this here vs the vulnerable one "encapsulated" in the resultsByPlugin? Like if we have notVulnerable, and vulnerable next to each other, that would make sense, but not just one without the other. Thoughts?

lili2311: `groupEntitiesPerScanType` is grouping items we want to fix per handler to send them to, since we don't want `@snyk/fix` to process non-vulnerable at all and only commend on how many there were they are filtered out first thing.

Design wise we wanted them to not appear in the results at all since the whole point of `snyk fix` is to fix vulneable items, but still show that we saw them and ignored them since there is nothing to do. Is it weird because it goes into the summary function so we count how many there were? Would just a number make more sense?

aarlaud (approved):

packages/snyk-fix/src/lib/output-formatters/show-results-summary.ts
aarlaud: it's more in the function signature. when I see notvulnerable, I immediately think vulnerable, and I would think of treating this generically, vulnerable or not, to display summary in a similar way for each. but nevermind me ;)

packages/snyk-fix/test/unit/lib/output-formatters/__snapshots__/show-results-summary.spec.ts.snap
```diff
@@ -53,6 +53,11 @@ exports[`generateUnresolvedSummary has failed upgrades & unsupported 1`] = `
   ✖ npm is not supported"
 `;
 
+exports[`showResultsSummary called with no vulnerable projects to fix 1`] = `
+"
+ ✖ No successful fixes"
```
mathild3r, line 58: why do we have 'no successful fixes' here and not 'no fixes needed'?

lili2311: Good point, we have not yet discussed a scenario where 100% of projects will be not vulnerable. Will raise with Andy

lili2311: Updated, checking a couple of other cases too

packages/snyk-fix/src/lib/output-formatters/show-results-summary.ts
lili2311: Updated the parameter names, better?

PR #2053, opened 2021-06-23
test: add new GitHub action test and fix the IaC SARIF output [CC-957]

teodora-sandu:
#### What does this PR do?
This implements the unit tests for the IaC GitHub action, with the aim of gating the CLI against breaking the SARIF output used by GitHub. Will start passing after https://github.com/snyk/snyk/pull/2052.

Run `npx jest test/jest/acceptance/iac/github-action/iac.spec.ts`

This PR came from an action to make sure we implement regression tests for our IaC GitHub action. The GitHub action would generally be used for scanning repos for security alerts, so may be run with the following file paths:
- `.`: the current way we run our action and which works
- nothing: equivalent of the above, but currently broken
- a relative directory nested inside the current directory
- a relative directory nested outside the current directory
- a file
The best way to test that the generated SARIF URIs are valid (without having to hard-code anything), is to build the paths and see if they match up to valid files in the local filesystem, as agreed in the [decision document](https://www.notion.so/snyk/Snyk-IaC-GitHub-action-Regression-testing-d3000b18cba44eaaae2f89d297b54ed7).

https://snyksec.atlassian.net/browse/CC-957

### DIFF
--- a/test/jest/acceptance/iac/github-action/runner.ts
+++ b/test/jest/acceptance/iac/github-action/runner.ts
@@ -1,0 +1,117 @@
+import {
+  readFileSync,
+  existsSync,
+  unlinkSync,
+  readdirSync,
+  statSync,
+} from 'fs';
+import * as path from 'path';
+import { v4 as uuidv4 } from 'uuid';
+import { pathToFileURL } from 'url';
+
+import { startMockServer } from '../helpers';
+
+const ROOT_DIR = './test/fixtures';
+
+export class GithubActionTestRunner {
+  constructor(
+    private command: string,
+    private workingDir: string,
+    private inputPath: string,
+    private run: (
+      cmd: string,
+      env: Record<string, string>,
+      cwd?: string,
+    ) => Promise<{ stdout: string; stderr: string; exitCode: number }>,
+    private teardown: () => void,
+  ) {}
+
+  static async build(
+    product: string,
+    relativeDir: string,
+    inputPath: string,
+  ): Promise<GithubActionTestRunner> {
+    const command = `snyk ${product} test ${path.join(inputPath)}`;
+    const workingDir = path.join(ROOT_DIR, relativeDir);
+
+    const { run, teardown } = await startMockServer();
+
+    return new GithubActionTestRunner(
+      command,
+      workingDir,
+      inputPath,
+      run,
+      teardown,
+    );
+  }
+
+  public async destroy() {
+    this.teardown();
+  }
+
+  public async test(flag: string) {
+    const sarif = await this.runAndGenerateSARIF(flag);
+    this.verifySARIFPaths(sarif);
+  }
+
+  private async runAndGenerateSARIF(flag: string): Promise<string> {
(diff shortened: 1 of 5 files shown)


test/jest/acceptance/iac/github-action/runner.ts
```diff
@@ -0,0 +1,123 @@
...
+
+function buildCommand(product: string, relativeDir: string, inputPath: string) {
+  // Before running the command, check if we need to change into a different directory
+  // And if it's windows we need to add /d
+  const isWindows =
+    osName()
+      .toLowerCase()
+      .indexOf('windows') === 0;
+  const changeDir =
+    relativeDir !== '' ? `cd ${isWindows ? '/d' : ''} ${relativeDir} &&` : '';
```
aron, line 115: I don't know if a more cross platform way to do this would be with [`process.chdir()`](https://nodejs.org/dist/latest-v14.x/docs/api/process.html#process_process_chdir_directory)

test/jest/acceptance/iac/github-action/runner.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+  }
+
+  private async verifySARIFPaths(sarif: string) {
+    const jsonObj = JSON.parse(sarif);
+
+    const actualPaths: Set<string> = new Set();
+    for await (const p of walk(this.scanPath)) {
+      actualPaths.add('file://' + p.replace(/\\/g, '/')); // URIs should use forward slash, not backward slash
```
aron, line 69: I think [`url.pathToFileURL()`](https://nodejs.org/dist/latest-v14.x/docs/api/url.html#url_url_pathtofileurl_path) will do what you want here on Windows... e.g.

```ts
actualPaths.add(url.pathToFileURL(p).href)
```

aron (approved):
👏 nice work, this looks great. A couple of cross platform suggestions to lean on the Node APIs vs doing workarounds with strings.

test/jest/acceptance/iac/github-action/runner.ts
teodora-sandu: This works and looks much cleaner, good find!

test/jest/acceptance/iac/github-action/runner.ts
teodora-sandu: I tried using `process.chdir()` but it changes the working directory of the Node.js process, rather than the working directory of the script. So it's a bit less clear what it's doing and requires some setup and cleanup before and after the tests run. I think as an alternative I'll split line 115 into multiple variables to make it easier to read

aron: Oh you're right, we could extend the `run` function to support providing a root directory... but for now I think this works. Perhaps leave a comment for the future?
https://github.com/snyk/snyk/blob/master/test/jest/acceptance/iac/helpers.ts#L54

teodora-sandu: Good suggestion, I've ended up extending the `run` function instead and it now works

aron: Nice!

PR #2057, opened 2021-06-24
test: do not fail when npm warnings are logged

ghost:
Following up from #2046, the reason tests were failing was because the assertions were strict on npm's stderr output. The change in npm@7.18 introduced warning output but the install itself was fine.

I've migrated the tests to acceptance tests written in Jest. Instead of importing specific modules, they're calling protect through Snyk CLI. They also copy the fixture before modifying it (via `createProject`) to avoid collisions with other tests using the same fixture. Similarly, they avoid using `process.chdir` which again introduces global state pollution.


test/jest/acceptance/snyk-protect/protect-patch-skip.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    await runCommand('npm', ['install'], {
+      cwd: project.path(),
+    }),
+  ).toEqual(
+    expect.objectContaining<RunCommandResult>({
+      code: 0,
+      stdout: expect.any(String),
+      stderr: expect.any(String),
+    }),
+  );
```
jahed-snyk, line 20: I know `objectContaining` looks a bit verbose and ugly but it's the best way to provide all command output (code, out, err) if any of them fail an assertion.

Once this becomes a common pattern, we can move it to its own custom matcher to reduce the verbosity. eg. 

```ts
expect(command).toExitWith({ code: 0 }) // out and err default to any string
```

test/jest/acceptance/snyk-protect/protect-patch-skip.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  expect(
+    await runCommand('npm', ['install'], {
+      cwd: project.path(),
+    }),
+  ).toEqual(
+    expect.objectContaining<RunCommandResult>({
+      code: 0,
+      stdout: expect.any(String),
+      stderr: expect.any(String),
+    }),
```
jahed-snyk, line 19: I know `objectContaining` looks a bit verbose and ugly but it's the best way to provide all command output (code, out, err) if any of them fail an assertion.

Once this becomes a common pattern, we can move it to its own custom matcher to reduce the verbosity. eg. 

```ts
expect(command).toExitWith({ code: 0 }) // out and err default to any string
```

test/jest/acceptance/snyk-protect/protect-patch.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect.objectContaining<RunCommandResult>({
+      code: 0,
+      stdout: 'Successfully applied Snyk patches',
+      stderr: expect.any(String),
+    }),
+  );
+
+  expect(await project.read('node_modules/semver/semver.js')).not.toEqual(
+    previousSemver,
+  );
```
jahed-snyk, line 34: In the Tap test we weren't asserting if the patch was being applied so I've added this assertion.

test/jest/util/runCommand.ts
```diff
@@ -15,23 +15,23 @@ const runCommand = (
   options?: RunCommandOptions,
 ): Promise<RunCommandResult> => {
   return new Promise((resolve, reject) => {
-    const cli = spawn(command, args, options);
+    const child = spawn(command, args, options);
```
jahed-snyk, line 18: Renamed as this has nothing to do with `cli` (it was migrated from a CLI-specific function). It's a generic `child` process.

test/jest/acceptance/snyk-protect/protect-patch-skip.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect.objectContaining<RunCommandResult>({
+      code: 0,
+      stdout: 'Successfully applied Snyk patches',
+      stderr: expect.any(String),
+    }),
+  );
+
+  expect(await project.read('node_modules/lodash/lodash.js')).toEqual(
+    previousLodash,
+  );
```
jahed-snyk, line 34: In the Tap test we weren't asserting if the patch was being *skipped* so I've added this assertion.

.circleci/config.yml
```diff
@@ -164,10 +164,7 @@ jobs:
           version: 19.03.13
           docker_layer_caching: true
       - install_shellspec
-      - run:
-          name: Install npm@7.17.0
-          command: |
-            sudo npm install -g npm@7.17.0
+      - run: sudo npm install -g npm@7
```
jahed-snyk, line 167: This PR fixed the tests on warnings so we don't need to lock ourselves to a specific version (which risks our npm version getting stale).

Avishagp (approved):

PR #2061, opened 2021-06-25
feat: ignore unrecognised IaC types [CC-947]

craigfurman:

#### What does this PR do?

ignore unrecognised IaC types

No longer return an error when scanning YAML or JSON files that do not
contain recognised infrastructure as code schemas, e.g. Kubernetes or
CloudFormation. Instead, ignore these files.

In YAML files that contain multiple documents, ignore documents that
contain unrecognised schemas. This includes empty documents.

I discovered while implementing this that mixed-recognised-config-type multi-doc yaml files were already supported: e.g. when scanning a 2-doc yaml file that contains a valid kubernetes manifest in 1 doc, and a valid cloudformation config in the other, issues are returned for both. This PR does not break this behaviour, although I suspect it was always undefined and I haven't added a test case for it.

In https://snyksec.atlassian.net/browse/CC-947 and https://docs.google.com/document/d/1IX6il8FN50Tv1bzEg9VvLfkvNTLz2jcyVbl4ceYscWo/edit#, we've had a fair amount of back and forth on what to do about unrecognised config types. This PR just moves such files from error cases to ignored, it doesn't print out a list/count of skipped json/yaml files.

https://snyksec.atlassian.net/browse/CC-947

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/file-parser.ts
@@ -1,5 +1,6 @@
 import { detectConfigType } from './parsers/k8s-or-cloudformation-parser';
 import { tryParsingTerraformFile } from './parsers/terraform-file-parser';
+import { NoFilesToScanError } from './file-loader';
 import {
   isTerraformPlan,
   tryParsingTerraformPlan,
@@ -34,6 +35,10 @@ export async function parseFiles(
       }
       failedFiles.push(generateFailedParsedFile(fileData, err));
     }
+  }
+
+  if (parsedFiles.length === 0) {
+    throw new NoFilesToScanError();
   }
 
   return {
--- a/src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
@@ -1,12 +1,5 @@
-import { CustomError } from '../../../../../lib/errors';
 import { IacProjectType } from '../../../../../lib/iac/constants';
-import { getErrorStringCode } from '../error-utils';
-import {
-  EngineType,
-  IaCErrorCodes,
-  IacFileData,
-  IacFileParsed,
-} from '../types';
+import { EngineType, IacFileData, IacFileParsed } from '../types';
 
 export const REQUIRED_K8S_FIELDS = ['apiVersion', 'kind', 'metadata'];
 export const REQUIRED_CLOUDFORMATION_FIELDS = ['Resources'];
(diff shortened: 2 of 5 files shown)


craigfurman:
Regarding the bot message above about smoke tests: a failing regression-smoke test is what led me to change the smoke tests in the first place. I don't really want to change my base branch, as that is a pain for PRs. Would it be ok to push a `smoke/` branch at an identical revision to this branch once I get the tests passing here, and link the action in a comment?

I can't find any guidance in https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md, but https://github.com/snyk/snyk/blob/feat/iac-skip-unrecognised/test/smoke/README.md sort-of-hints this might be ok practice?

craigfurman:
I pushed a smoke branch (https://github.com/snyk/snyk/runs/2916014071), and that seems to have added a bunch of checks here, presumably because github checks are commit-scoped and not branch-scoped.

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
```diff
@@ -15,79 +8,41 @@ export function detectConfigType(
...
 }
 
 export function checkRequiredFieldsMatch(
   parsedDocument: any,
   requiredFields: string[],
-) {
+): boolean {
+  if (!parsedDocument) {
+    return false;
+  }
```
aron, line 44: Under what conditions can we get a non-document here? Is this for the invalid YAML sub-document? I think it would be worth adding a comment or handling this in the layer above.

```
if (parsedFile && checkRequiredFieldsMatch(parsedFile, REQUIRED_K8S_FIELDS)) {
```

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
```diff
@@ -15,79 +8,41 @@ export function detectConfigType(
...
+        return {
+          ...fileData,
+          jsonContent: parsedFile,
+          projectType: IacProjectType.K8S,
+          engineType: EngineType.Kubernetes,
+          docId,
+        };
       } else {
-        throw new FailedToDetectYamlConfigError(fileData.filePath);
+        return null;
```
aron, line 32: I'm wondering if this is the correct layer to swallow this error. We currently have `IacFileParsed` and `IacFileParseFailure`, I'm wondering if here we should be returning an `IacFileParseFailure` or introducing a third type e.g. `IacFileParseSkipped`.

Then we would filter these out at the result rendering layer by default. This would also allow us to present the total number of skipped files to the user as part of the output and in future we could potentially introduce a `--strict` flag that shows the errors, as I think there is some value in the error messages provided here if you're unclear on why your file was skipped.

Thoughts?

aron (changes requested):
👏 Thanks for picking up this Tier 4 issue so quickly. Implementation looks good and thank you for extending the test suite. I've requested changes to get your thoughts on the question around surfacing the "skipped" files to the upper layers of the system.

craigfurman:
Note: this PR preserves the behaviour of printing an error and exiting non-zero when we encounter an unparseable/invalid JSON/yaml. I'm not sure if we want to change that as part of this PR.

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
craigfurman: Invalid YAML sub-docs, or invalid jsons, actually do still throw exceptions. This PR doesn't change that behaviour at all. Do we want to change this, and if so, now or later?

The branch you've highlighted covers the empty YAML sub-document case, which IIRC was one of the motivating cases for the ticket. Empty doc support is test-guarded in this fixture: https://github.com/snyk/snyk/pull/2061/files#diff-d87dcf90807ed0045cbb1e752f4ffbe6b4216d177be5abab4ebe57792b54f5c8R57, and the test that uses it.

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
craigfurman: I think we have a bunch of options here. I'll try to lay out my mental model of it:

1. Provide no way for the user to see what's skipped
   * This is what the PR does at the time of writing, and is what is motivating your comment. I can see why this might give us pause, in case users are confused.
1. Print a list of all files skipped
   * Optionally behind a flag, e.g. `--show-skipped`?
   * You mention possibly introducing a `--strict` that treats unrecognised yaml/json as errors, but personally I don't agree with this... it's semantically never an error to skip an unsupported schema. IMO a possible problem with this PR currently is that it's hard for users to figure out _why_ a file they did expect to be scanned is skipped, in case we have a bug in our detection logic.
   * We currently print failure summaries in bold red. Should this list be above or below the failure summary? Maybe non-bold blue?
   * I worry this could be very noisy. An IaC-oriented repository might contain 1000s of json and yaml files.
   * Arguably non json/yaml/tf files are also "skipped" by IaC, and no-one is suggesting erroring/listing these!
1. Print a count of skipped files
   * Possibly behind a feature flag
   * Less noisy than printing them all, but there is information loss here. It doesn't really help the user debug what is going on.

craigfurman: @aron and I had a zoom about this, summary:

- skipped file diagnostics are probably more noise than signal, so we won't implement them here
- we should ensure we return an error to the user when no recognised types are found, as this is probably a mistake
   - when we find no valid extensions, we already return an error for this reason
   - implicitly, we had an error when no recognised IaC type was found because we errored on every unrecognised yaml/json 😅.

craigfurman:
Thanks for the review and the conversation @aron 👍 

I've implemented what we agreed, would you mind taking another look?

src/cli/commands/test/iac-local-execution/file-parser.ts
```diff
@@ -36,6 +37,10 @@ export async function parseFiles(
     }
   }
 
+  if (parsedFiles.length == 0) {
+    throw new NoFilesToScanError();
```
craigfurman, line 41: wdyt about reusing this same error?

src/cli/commands/test/iac-local-execution/parsers/k8s-or-cloudformation-parser.ts
aron: That makes sense 👍

src/cli/commands/test/iac-local-execution/file-parser.ts
aron: I think it makes perfect sense to re-use the error. One thing to note is that this will mask any `failedFiles` so I wonder if this logic should be `if (parsedFiles.length === 0 && failedFiles.length !== 0) {` 

(also note the `===`, I'd have thought this would be picked up by the linter but generally strict equality `===` is preferred as a stylistic preference in JS vs. `==` which will perform coercion).

aron (approved):
Nice work, I tried this locally with a bunch of different combinations and it all felt like it was behaving as I would expect.

src/cli/commands/test/iac-local-execution/file-parser.ts
craigfurman: Do you actually mean `if (parsedFiles.length === 0 && failedFiles.length === 0)`? i.e, do propagate any exceptions from invalid files instead of masking them.

The way the code is currently written `if (parsedFiles.length == 0)`, when scanning a directory containing one invalid file and one random yaml (e.g. circleci.yaml), we'll get NoFilesToScanError.

If we use `if (parsedFiles.length === 0 && failedFiles.length === 0)`, we'll get "Testing invalid.json... Failed to parse JSON file".

I don't have a strong opinion on which error is more instructive in this case. My gut says NoFilesToScanError, since the user is clearly expecting some snyk iac functionality to be contextually relevant, but it isn't.

re: threequal vs `==`, d'oh 🤦. I keep forgetting this basic javascript mistake.

aron: Sorry you're right, my brain farted. Thanks for explaining your thinking, I'm sold, let's go with your logic and the triple equals.

PR #2065, opened 2021-06-29
feat: send `snyk fix` specific analytics

lili2311:

#### What does this PR do?
Add in custom `snyk fix` analytics
- track how many projects have failed, fixed & total
- track how many issues were fixable, fixed & total
- use latest version of @snyk/fix that sends these numbers


aarlaud (approved):

mathild3r (approved):

src/cli/commands/fix/index.ts
```diff
@@ -47,6 +48,12 @@ async function fix(...args: MethodArgs): Promise<string> {
   const { dryRun, quiet } = options;
   const { fixSummary, meta } = await snykFix.fix(results, { dryRun, quiet });
 
+  analytics.add('snykFixFailedProjects', meta.failed);
+  analytics.add('snykFixFixedProjects', meta.fixed);
+  analytics.add('snykFixTotalProjects', results.length);
+  analytics.add('snykFixVulnerableProjects', results.length);
+  analytics.add('snykFixSummary', fixSummary);
```
anthogez, line 55: naively question `results.length` could not contain also projects that had no vulnerabilities at all?

lili2311: good catch it is not meant to be results.length for `Vulnerable` 🦅 👁️

lili2311: updated

anthogez: chefkissemoji

anthogez: Perfect :chef-kiss: 🎉

anthogez (approved):

jan-stehlik (commented):
hey @lili2311 , looks good, can we please write test for this logic?

lili2311:
@jan-stehlik sure thing, do we have an example test where we capture analytics from a command? It will help me find one :)

anthogez:
> hey @lili2311 , looks good, can we please write test for this logic?

FYI even if we have test/jest/system/analytics.spec.ts, it's still is a poor sample of test impl/validation

anthogez:
> hey @lili2311 , looks good, can we please write test for this logic?

FYI @jan-stehlik  even if we have test/jest/system/analytics.spec.ts, it's still is a poor sample of CLI test impl/validation

lili2311:
@jan-stehlik updated, please 👀

src/cli/commands/fix/index.ts
```diff
@@ -47,6 +48,19 @@ async function fix(...args: MethodArgs): Promise<string> {
   const { dryRun, quiet } = options;
   const { fixSummary, meta } = await snykFix.fix(results, { dryRun, quiet });
 
+  // Analytics # of projects
+  analytics.add('snykFixFailedProjects', meta.failed);
```
JackuB, line 52: Is anyone using the snyk fix package as standalone? Are we concerned that we won't have usage data in that case?

lili2311: it is only part of the CLI at the moment.

JackuB (approved):

test/jest/unit/lib/commands/fix/fix.spec.ts
```diff
@@ -110,6 +114,36 @@ describe('snyk fix (functional tests)', () => {
         dryRun: true,
         quiet: true,
       });
+      expect(addAnalyticsSpy).toHaveBeenCalledTimes(8);
+      expect(addAnalyticsSpy.mock.calls[0]).toEqual([
+        'snykFixFailedProjects',
+        0,
+      ]);
```
jan-stehlik, line 121: nice work! this works, however, the more robust way to write a test if the order of execution doesn't matter might be
```suggestion
      expect(addAnalyticsSpy).toHaveBeenCalledWith('snykFixFailedProjects', 0);
```

the benefit here is that if the order of execution changes in the implementation, this test will still pass which I think is desired. This will prevent future test failures for wrong reason (implementation works it's just in different order of statements that's changed)

what do you think?

jan-stehlik (approved):
LGTM, well done @lili2311 🙌 I have only small suggestion

test/jest/unit/lib/commands/fix/fix.spec.ts
lili2311: i like it! forgot you can call it like that. Easier to read too. Will push an update

PR #2066, opened 2021-06-30
test(smoke): cover scenario when snyk is testing a specific library

JackuB:
We should cover the functionality of `snyk test LIBRARY`


JackuB:
Implemented in #2100

PR #2069, opened 2021-07-01
refactor: remove redundant file output tests for IaC [CC-957]

teodora-sandu:
#### What does this PR do?
These tests were added to verify the validity of the file output for our GitHub action.
We have implemented new regression tests under https://github.com/snyk/snyk/pull/2053, so
these tests don't add any value to us. They verify the output of the JSON file output, but
both the SARIF and JSON file output use the same logic so testing the GitHub action should
be enough.

- https://snyksec.atlassian.net/browse/CC-957


aron (approved):
👍 looks good

PR #2077, opened 2021-07-08
test: delete api-key-from-env and api-key tests

Avishagp:
Both of the scenarios tested in these tests are being tested numerous times in our Smoke Tests and CircleCI jobs, because we need to authenticate in order to run them.


test/jest/acceptance/api-key-from-env.spec.ts
```diff
@@ -0,0 +1,6 @@
+const key = (process.env.SNYK_TOKEN = '123456');
```
jahed-snyk, line 1: Modifying global variables like `process` causes noise. Can we improve it while we're here?

test/jest/acceptance/api-key-from-env.spec.ts
```diff
@@ -0,0 +1,6 @@
...
+import * as snyk from '../../../src/lib';
+
+test('api token from env', () => {
+  expect(key).toEqual(snyk.api);
```
jahed-snyk, line 5: `expect(snyk.api).toEqual('123456')` is more understandable here.

ghost (commented):
I think it's worth considering deleting this test. It's not doing much and is likely covered by other acceptance tests since it's testing auth which is a prerequisite for running most commands.

test/jest/acceptance/api-key-from-env.spec.ts
Avishagp: What do you mean by true acceptance test?

jahed-snyk: @Avishagp Testing by running the CLI as a whole.

Avishagp: Cool, thanks.
As for deleting this test, maybe the right thing would be to merge it with `api-key.test.ts`? Don't really see how separating these two scenarios to two different tests makes sense as `api-key-from-env` is so short.

jahed-snyk: Let's resolve this next time we pair. Should be a lot faster than doing it over PR comments 😬

ghost (approved):

PR #2079, opened 2021-07-08
fix: ensure tip to use --all-sub-projects is shown

lili2311:

#### What does this PR do?
Gradle plugin stopped sending root project as a subProject
snyk/snyk-gradle-plugin#180 and this resulted in the logic that was working around in in CLI to stop working correctly, this PR fixes it to display the `tip` to use `--all-sub-projects` when we found at least 1 gradle sub-project
#### Screenshots

CleanShot 2021-07-08 at 17 44 25@2x

### DIFF
--- a/src/lib/plugins/get-extra-project-count.ts
+++ b/src/lib/plugins/get-extra-project-count.ts
@@ -14,7 +14,7 @@ export async function getExtraProjectCount(
   if (
     inspectResult.plugin.meta &&
     inspectResult.plugin.meta.allSubProjectNames &&
-    inspectResult.plugin.meta.allSubProjectNames.length > 1
+    inspectResult.plugin.meta.allSubProjectNames.length > 0
   ) {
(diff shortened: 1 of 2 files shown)


anthogez:
> * [ ]  Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules
> 
> #### What does this PR do?
> Gradle plugin started sending less subProject names than expected
> [snyk/snyk-gradle-plugin#180](https://github.com/snyk/snyk-gradle-plugin/pull/180) and this broke
> the logic inside CLI that was expecting there to always be at least
> 1 project before considering to display the `tip` to use `--all-sub-projects` to scan all at once.
> 
> #### Screenshots
> CleanShot 2021-07-08 at 17 44 25@2x

Gradle has stopped adding root project as part of sub-project list names. Would be more accurate.

gitphill (approved):
thanks

PR #2080, opened 2021-07-09
refactor: move formatters to lib

lili2311:

#### What does this PR do?
Move all formatters into `lib/formatter` and away from `commands` which is for orchestrating the commands being called. Much of the formatting logic is inter used and keeping it together makes sense for now until we have better structure.
Added tests to all the formatters that did not have any, so we can start changing this output and capture any issues.

The formatters are lacking test coverage so will add some here for key flows.

See the comments left on the changes to explain any moves. Mostly the changes are adding tests

I want to add some extra information to the output but noticed I would have to create the tests in order to update them, so this is adding the tests to capture current functionality before I make further changes to the output.


src/cli/commands/test/index.ts
```diff
@@ -321,126 +298,3 @@ function shouldFail(vulnerableResults: any[], failOn: FailOn) {
   // should fail by default when there are vulnerable results
   return vulnerableResults.length > 0;
 }
-
-function displayResult(
```
lili2311, line 325: moved as is to it's own file inside `lib/formatters`

src/lib/formatters/remediation-based-format-issues.ts
```diff
@@ -153,10 +150,6 @@ export function formatIssuesWithRemediation(
   return results;
 }
 
-export function getSeverityValue(severity: SEVERITY): number {
```
lili2311, line 156: moved to it's own file inside `lib/formatters`

src/lib/formatters/test/format-test-results.ts
```diff
@@ -265,7 +263,11 @@ function getDockerSuggestionText(options, config): string {
   return dockerSuggestion;
 }
 
-function groupVulnerabilities(vulns): GroupedVuln[] {
+export function groupVulnerabilities(
+  vulns,
+): {
+  [vulnId: string]: GroupedVuln;
```
lili2311, line 269: original type was incorrect, this does not return an array

src/lib/snyk-test/iac-test-result.ts
```diff
@@ -4,13 +4,13 @@ export interface AnnotatedIacIssue {
   id: string;
   title: string;
   description: string;
-  severity: SEVERITY;
+  severity: SEVERITY | 'none';
```
lili2311, line 7: `none` is a valid severity expected by iac

src/lib/snyk-test/iac-test-result.ts
```diff
@@ -4,13 +4,13 @@ export interface AnnotatedIacIssue {
...
   isIgnored: boolean;
   cloudConfigPath: string[];
   type: string;
   subType: string;
-  path: string[];
-  documentation: string;
+  path?: string[];
+  documentation?: string;
```
lili2311, line 13: updating to match other types for `iac`

test/acceptance/workspaces/reachable-vulns/maven/test-dep-graph-response-reachable.json
```diff
@@ -177,8 +177,7 @@
     "upgrade": {
       "package-name@1.6.1": {
         "upgradeTo": "package-name@1.6.3",
-        "vulns": ["SNYK-JAVA-PACAKGE-12345"],
-        "isTransitive": false
```
lili2311, line 181: `isTransitive` is only set for Pip

test/jest/unit/format-test-results.spec.ts
```diff
@@ -1,157 +0,0 @@
-import { extractDataToSendFromResults } from '../../../src/cli/commands/test/formatters/format-test-results';
-import { Options } from '../../../src/lib/types';
-import * as fs from 'fs';
-
-describe('format-test-results', () => {
```
lili2311, line 5: moved the test as is

test/reachable-vulns.test.ts
```diff
@@ -21,151 +13,6 @@ import {
 import * as featureFlags from '../src/lib/feature-flags';
 import * as utils from './utils';
 
-test('output formatting', (t) => {
```
lili2311, line 24: moved to it's own test file as is

ipapast (commented):
Hey @lili2311, we undestand the intention behind moving the formatters into one place and that would be great in terms of structure. Our formatter, however, (iac) is not really reusable; it is quite specific to the cloud config files. 
Background: A few months ago, we moved all our files in a specific directory (`iac-local-execution`) to keep everything around iac in one place and the entire flow is in there.

Would you mind keeping the `results-formatter` in the same directory (iac-local-execution)? Except if you see any other value in that, happy to hear about it. ☺️

lili2311:
@ipapast sure can move it back, the idea was to move all the formatters into `lib/formatters` regardless of the project type because the CLI is not split up right now to have different libraries per type. 
Ideally the formatter comes from the plugin even and is not in the CLI. Was just trying to make some order of all the code we have everywhere.

lili2311:
@ipapast it's been moved back in the last commit

ipapast:
> @ipapast it's been moved back in the last commit

Thanks @lili2311, I understand the order we are trying to bring in. I discussed with some team members and we think for now it's best to be in its own folder, and we can consider again when we move to plugins in the future. The rest of the changes are still good, thanks!

ipapast (approved):

test/jest/unit/lib/formatters/test/format-test-results.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+  const jsonDataContainerFixture = JSON.parse(
+    fs.readFileSync('test/fixtures/basic-apk/jsonData.json', 'utf-8'),
+  );
+  const jsonDataGroupedContainerFixture = JSON.parse(
+    fs.readFileSync('test/fixtures/basic-apk/jsonDataGrouped.json', 'utf-8'),
+  );
+
+  it('should not create any JSON unless it is needed per options', () => {
+    const options = {} as Options;
+    const jsonStringifySpy = jest.spyOn(JSON, 'stringify');
```
jan-stehlik, line 28: why do we need to spy on javascript native functions? i think it's better to avoid testing implementation details (some function was called) and focus on testing public API (output equals expected output for given input). The test could look like this

```javascript
it('should not create any JSON unless it is needed per options', () => {
    const options = {} as Options;

    const res = extractDataToSendFromResults(
      resultsFixture,
      jsonDataFixture,
      options,
    );

    expect(res.stringifiedData).toBe('');
    expect(res.stringifiedJsonData).toBe('');
    expect(res.stringifiedSarifData).toBe('');
});

it('should create Snyk JSON and only Snyk JSON if `--json` is set in the options', () => {
    const options = {
      json: true,
    } as Options;

    const res = extractDataToSendFromResults(
      resultsFixture,
      jsonDataFixture,
      options,
    );

    expect(res.stringifiedData).toEqual(
        // ... actual response string
        // we can also do subset assertion of response is too long expect.stringContaining(...)
    );
    expect(res.stringifiedJsonData).toMatchObject({
        // ... actual response object
        // we can also do subset assertion of response is too long with expect.objectContaining(...)
    });
    expect(res.stringifiedSarifData).toBe('');
  });
```

The implications of testing implemenation details can be
- test will fail for wrong reason (e.g. when we change the serialisation mechanism but the output is still correct)
- test will not fail for right reason (e.g. when output changes a lot but we are still using the same serialisation mechanism)

The pros of testing public API over implementation details is
- our test will pass even if we change implementation details but the output stays the same (say we change the way to serialise data)
- our test will fail if the output is incorrect

I'm aware that sometimes testing against actual output might be very verbose and not desired. There are options to deal with that too:
- if you want to test against full output you can store expected value in standone fixtures files
- if you want to test against subset of return value you can leverage [expect.stringContaining](https://jestjs.io/docs/expect#expectstringcontainingstring) or [expect.objectContaining](https://jestjs.io/docs/expect#expectobjectcontainingobject)

Further reading on the matter https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

What do you think?

test/jest/unit/lib/formatters/format-reachability.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        reachableByFunctionVuln,
+        reachableByFunctionVuln,
+        reachableByPackageVuln,
+        noInfoVuln,
+      ]),
+    ).toEqual('In addition, found 2 vulnerabilities with a reachable path.');
+  });
+
+  test('formatReachablePaths', () => {
+    function reachablePathsTemplate(
```
jan-stehlik, line 88: i'm aware that this test was just copied over, but if you would like to improve it further i recommend to avoid putting extra logic in tests (what if logic in test is wrong - how do we ensure logic in test won't cause false positives). I think the best way from confidence and readibility point of view is to prefer as simple tests as possible without extra logic that can go wrong

Further reading https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

What do you think?

jan-stehlik (approved):
nice work @lili2311 , i especially like the new tests 🎉 i have only minor testing comments to consider

test/jest/unit/lib/formatters/test/format-test-results.spec.ts
lili2311: This was copied over as is, have not looked into this test 🤔 @maxjeffos might be more relevant to answer your qs. We can certainly improve this, but I want to make sure it is clear that I have not changed existing tests and therefore not broken any existing expectations. So perhaps as a separate PR afterwards so we don't have `move` and `changes` in the same PR as it will be confusing to see what changed.

test/jest/unit/lib/formatters/format-reachability.spec.ts
lili2311: same as above, want to make sure it is clear no existing tests have changed so assertions are the same. If it can be clear in the same PR then sure

lili2311: to simplify I will raise a separate PR with the proposed changes thanks!

test/jest/unit/lib/formatters/test/format-test-results.spec.ts
lili2311: to simplify I will raise a separate PR with the proposed changes thanks!

PR #2082, opened 2021-07-12
test: migrate cli-command-types test to jest

Avishagp:
See title


ghost (approved):
Needs to be in `unit/cli/commands/types.spec.ts`. Other than that, looks good.

PR #2087, opened 2021-07-14
chore: init support async test with polling

anthogez:
Initial support for async test of unmanaged ecosystems e.g. c/c++
adding polling to resolve and test fileSignatures fact.

The pr that will introduce the active usage of it will come later on.

These changes comes from https://github.com/snyk/snyk/pull/2025

Using chore as prefix, because at the moment it's not bringing any
effective feat yet.

### DIFF
--- a/src/lib/ecosystems/polling.ts
+++ b/src/lib/ecosystems/polling.ts
@@ -1,0 +1,129 @@
+import * as config from '../config';
+import { isCI } from '../is-ci';
+import { makeRequest } from '../request/promise';
+import { Options } from '../types';
+
+import { assembleQueryString } from '../snyk-test/common';
+import { getAuthHeader } from '../api-token';
+import { ScanResult } from './types';
+import { TestDependenciesResult, TestDepGraphMeta } from '../snyk-test/legacy';
+import { sleep } from '../common';
+
+type ResolveAndTestFactsStatus =
+  | 'CANCELLED'
+  | 'ERROR'
+  | 'PENDING'
+  | 'RUNNING'
+  | 'OK';
+
+interface PollingTask {
+  pollInterval: number;
+  maxAttempts: number;
+}
+
+interface ResolveAndTestFactsResponse {
+  token: string;
+  pollingTask: PollingTask;
+  result?: TestDependenciesResult;
+  meta?: TestDepGraphMeta;
+  status?: ResolveAndTestFactsStatus;
+  code?: number;
+  error?: string;
+  message?: string;
+  userMessage?: string;
+}
+
+export async function requestPollingToken(
+  options: Options,
+  isAsync: boolean,
+  scanResult: ScanResult,
+): Promise<ResolveAndTestFactsResponse> {
+  const payload = {
+    method: 'POST',
+    url: `${config.API}/test-dependencies`,
+    json: true,
+    headers: {
+      'x-is-ci': isCI(),
+      authorization: getAuthHeader(),
+    },
+    body: {
+      isAsync,
+      scanResult,
+    },
+    qs: assembleQueryString(options),
+  };
+  const response = await makeRequest<ResolveAndTestFactsResponse>(payload);
+  throwIfRequestPollingTokenFailed(response);
+  return response;
+}
+
+function throwIfRequestPollingTokenFailed(res: ResolveAndTestFactsResponse) {
+  const { token, status, pollingTask } = res;
+  const { maxAttempts, pollInterval } = pollingTask;
+  const isMissingPollingTask = !!maxAttempts && !!pollInterval;
+  if (!token && !status && isMissingPollingTask) {
+    throw 'Something went wrong, invalid response.';
+  }
+}
+
+export async function pollingWithTokenUntilDone(
+  token: string,
+  type: string,
+  options: Options,
+  pollInterval: number,
+  attemptsCount: number,
+  maxAttempts = Infinity,
+): Promise<ResolveAndTestFactsResponse> {
+  const payload = {
+    method: 'GET',
+    url: `${config.API}/test-dependencies/${token}`,
+    json: true,
+    headers: {
+      'x-is-ci': isCI(),
+      authorization: getAuthHeader(),
+    },
+    qs: { ...assembleQueryString(options), type },
+  };
+
+  const response = await makeRequest<ResolveAndTestFactsResponse>(payload);
+
+  if (pollingRequestHasFailed(response)) {
+    throw response;
+  }
+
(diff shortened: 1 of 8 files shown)


anthogez:
Next step will be "Refactoring" to make it work for both test and monitor flows

src/lib/ecosystems/types.ts
```diff
@@ -18,6 +22,31 @@ export interface ContainerTarget {
   image: string;
 }
 
+export type ResolveAndTestFactsStatus =
```
admons, line 25: I think those 3 types should sit in `polling.ts`

src/lib/snyk-test/legacy.ts
```diff
@@ -236,7 +236,7 @@ interface Issue {
   fixInfo: FixInfo;
 }
 
-interface TestDependenciesResult {
+export interface TestDependenciesResult {
```
admons, line 239: why?

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+        errors.push(`Could not test dependencies in ${path}`);
+      }
+    }
+  }
+  spinner.clearAll();
+  return [results, errors];
+}
+
+function throwIfRequestPollingTokenFailed(res: ResolveAndTestFactsResponse) {
```
admons, line 64: What do you think about adding this to the `requestPollingToken` function?

src/lib/snyk-test/legacy.ts
anthogez: `property` result of `interface ResolveAndTestFactsResponse` depends on testDependenciesResult, 
trying to keep the changes to minimum amount as possible as we still have to maintain legacy and new flow

src/lib/ecosystems/polling.ts
```diff
@@ -0,0 +1,118 @@
...
+  };
+
+  const response = await makeRequest<ResolveAndTestFactsResponse>(payload);
+
+  if (response.result && response.meta) {
+    return response;
+  }
+
+  if (pollingRequestHasFailed(response)) {
+    throw new Error('polling request has failed.');
```
admons, line 58: I think we need more context here for debugging (i.e. response, header.snyk-request-id)

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -0,0 +1,55 @@
...
+        const { maxAttempts, pollInterval } = res.pollingTask;
+        const response = await pollingWithTokenUntilDone(
+          res.token,
+          ecosystem,
+          options,
+          pollInterval,
+          maxAttempts,
+        );
+
+        if (response.meta && response.result) {
```
admons: you can remove this line as this is already. being verified in the `pollingWithTokenUntilDone` function

src/lib/ecosystems/polling.ts
```diff
@@ -0,0 +1,110 @@
...
+import { isCI } from '../is-ci';
+import { makeRequest } from '../request/promise';
+import { Options } from '../types';
+
+import { assembleQueryString } from '../snyk-test/common';
+import { getAuthHeader } from '../api-token';
+import { ScanResult } from './types';
+import { TestDependenciesResult, TestDepGraphMeta } from '../snyk-test/legacy';
+
+export type ResolveAndTestFactsStatus =
```
admons, line 11: do you need to export this?

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -0,0 +1,52 @@
...
+          results.push({
+            issues: response.result.issues,
+            issuesData: response.result.issuesData,
+            depGraphData: response.result.depGraphData,
+          });
+        }
+      } catch (error) {
+        const hasStatusCodeError = error.code >= 400 && error.code < 500;
+        if (hasStatusCodeError) {
+          throw new Error(error.message);
```
admons, line 39: 1. add logs for debugging everywhere you decide to throw an error
2. a possible scenario according to this logic is that you'll wait for an hour and the last scanResult will have an error (4xx status code) and you'll loose everything. I think you should add all errors to the errors and continue

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -0,0 +1,52 @@
...
+        const { maxAttempts, pollInterval } = res.pollingTask;
+        const response = await pollingWithTokenUntilDone(
+          res.token,
+          ecosystem,
+          options,
+          pollInterval,
+          maxAttempts,
+        );
+
+        if (response && response.meta && response.result) {
```
admons, line 29: remove this if statement as you're already verifing this in the `pollingWithTokenUntilDone` function

src/lib/ecosystems/polling.ts
```diff
@@ -0,0 +1,110 @@
...
+    const taskCompleted = response.result && response.meta;
+    if (taskCompleted) {
+      return response;
+    }
+    if (pollingRequestHasFailed(response)) {
+      throw response;
+    }
+    await sleep(pollInterval);
+  }
+}
```
admons, line 97: Don't you want to return/throw something here?

src/lib/ecosystems/polling.ts
```diff
@@ -0,0 +1,110 @@
...
+  result?: TestDependenciesResult;
+  meta?: TestDepGraphMeta;
+  status?: ResolveAndTestFactsStatus;
+  code?: number;
+  error?: string;
+  message?: string;
+  userMessage?: string;
+}
+
+export async function requestPollingToken(
```
admons, line 35: I would think about splitting this file to 2 different files 
i.e. `polling/token.ts` and `polling/get-status.ts` or something like that

admons (approved):
looking good, just please notice my comments

src/lib/ecosystems/polling.ts
anthogez:
```
if (pollingRequestHasFailed(response)) {
      throw response;
}
```

src/lib/ecosystems/resolve-test-facts.ts
anthogez: response could still be undefined by the way it was implemented in pollingWithToken.. to satisfy the type

src/lib/ecosystems/resolve-test-facts.ts
anthogez: `throw new Error(error.message);` good catch it should be pushed as the rest 😄 of errors

PR #2088, opened 2021-07-14
fix: make sure we suggest --all-sub-projects only when appropriate

lili2311:

#### What does this PR do?

Now that the https://github.com/snyk/snyk/pull/2080 is merged and contains some baseline tests:
- split up the tip generation into relevant functions
- test each function
- noticed a bug with the Gradle ` --all-sub-projects` so fixing that as well to only display a tip to try `--all-sub-projects` when all conditions are met:
    - the scan is not already using --all-sub-projects
    - project type is gradle
    - we detected some projects that could also be scanned

https://github.com/snyk/snyk/compare/refacfactor/display-tips?expand=1#diff-9b0a855c48a95a2016a99f89e5ab64cdb002c87b2e5271bce6e91c51adb543feR42

`snyk test` on a gradle project with options that match the test scenarios

### DIFF
--- a/src/lib/formatters/format-monitor-response.ts
+++ b/src/lib/formatters/format-monitor-response.ts
@@ -4,6 +4,8 @@ import * as url from 'url';
 
 import { MonitorResult } from '../types';
 import * as config from '../config';
+import { showAllProjectsTip } from './show-all-projects-tip';
+import { showGradleSubProjectsTip } from './show-all-sub-projects-tip';
 
 export function formatErrorMonitorOutput(
   packageManager,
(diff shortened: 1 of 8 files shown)


test/jest/unit/lib/formatters/__snapshots__/format-monitor-response.spec.ts.snap
```diff
@@ -38,8 +38,6 @@ Monitoring src/lib/jens...
 
 Explore this snapshot at https://example.com/project
 
-Tip: This project has multiple sub-projects (7), use --all-sub-projects flag to scan all sub-projects.
```
lili2311, line 41: We should not be suggesting to use `--all-sub-projects` when the user is already using that. So since this PR fixes that bug the snapshot can now be updated https://github.com/snyk/snyk/pull/2088/files#diff-53790338973f66dafe4a3c879afe8d04ffadf4855d9fbbbe0b86739305010be0L35

gitphill (approved):

src/lib/formatters/format-monitor-response.ts
```diff
@@ -52,18 +59,10 @@ export function formatMonitorOutput(
...
-          `Tip: This project has multiple sub-projects (${foundProjectCount}), ` +
-            'use --all-sub-projects flag to scan all sub-projects.\n\n',
-        )
-      : '') +
-    (advertiseAllProjectsCount && foundProjectCount
-      ? chalk.bold.white(
-          `Tip: Detected multiple supported manifests (${foundProjectCount}), ` +
-            'use --all-projects to scan all of them at once.\n\n',
-        )
+    (gradleSubProjectsTip
```
jan-stehlik, line 62: much more readable, nice one!

jan-stehlik (approved):
thanks @lili2311 , great cleanup 💯

PR #2089, opened 2021-07-14
feat: `snyk fix` suggestion tip for supported project types

lili2311:

#### What does this PR do?
- Suggest to users to try beta feature `snyk fix` when a vulnerable `pip` or `poetry` project is in the response.
- Refactor showing multi scan tips for `--all-projects` or `--all-sub-project` to be controlled from 1 function as it is one or the other and not both to be shown.
- Based on design feedback make sure we use `Tip:` and only bold the commands suggested and nothing else, so added changes here

`snyk test`

#### Screenshots

CleanShot 2021-07-15 at 15 37 54@2x

CleanShot 2021-07-15 at 15 42 29@2x


jan-stehlik (approved):
Love the new tests, thanks @lili2311 🚀

PR #2090, opened 2021-07-15
test: move cli-command-types test to unit tests folder

Avishagp:
See title


ghost (commented):
For unit tests, make sure to mirror the same name and directory structure as the module being tested. In this case:

```
test/jest/unit/cli/commands/types.spec.ts
```

This makes the tests easier to discover (just search for the module name and open the spec file with the same name).

PR #2092, opened 2021-07-15
feat: add --help to snyk protect

Avishagp:

#### What does this PR do?
Add `--help` flag to `@snyk/protect`.

### DIFF
--- a/packages/snyk-protect/help.txt
+++ b/packages/snyk-protect/help.txt
@@ -1,0 +1,18 @@
+$ snyk-protect --help
+
+NAME
+  @snyk/protect - Patch vulnerable code in your project's dependencies.
+
+DESCRIPTION
+  You don't typically need to add the @snyk/protect dependency manually.
+  For more information on how to setup and use @snyk/protect, visit:
+    https://github.com/snyk/snyk/tree/master/packages/snyk-protect
+
+SYNOPSIS
+  snyk-protect [<options>]
+
+OPTIONS
+  --help
+    Display this help text.
+  --version
+    Display the current version.
--- a/packages/snyk-protect/src/lib/index.ts
+++ b/packages/snyk-protect/src/lib/index.ts
@@ -13,6 +13,16 @@ import { getAllPatches } from './fetch-patches';
 import { sendAnalytics } from './analytics';
 
 async function protect(projectFolderPath: string) {
+  // Handle runs with flags
+  if (process.argv.includes('--help')) {
+    console.log(getHelp());
+    return;
+  }
+  if (process.argv.includes('--version')) {
+    console.log(getVersion());
+    return;
+  }
+
   const snykFilePath = path.resolve(projectFolderPath, '.snyk');
 
   if (!fs.existsSync(snykFilePath)) {
@@ -90,6 +100,19 @@ async function protect(projectFolderPath: string) {
     type: ProtectResultType.APPLIED_PATCHES,
     patchedModules,
   });
+}
+
+export function getHelp(): string {
+  const filePath = path.resolve(__dirname, '../../help.txt');
+  const helpText = fs.readFileSync(filePath, 'utf8');
+  return helpText;
+}
+
+export function getVersion() {
+  const filePath = path.resolve(__dirname, '../../package.json');
+  const rawData = fs.readFileSync(filePath, 'utf8');
+  const packageJSON = JSON.parse(rawData);
+  return packageJSON.version;
 }
 
 export default protect;
--- a/packages/snyk-protect/test/fixtures/version-flag/README.md
+++ b/packages/snyk-protect/test/fixtures/version-flag/README.md
@@ -1,0 +1,1 @@
+Test fixture for version flag
--- a/packages/snyk-protect/test/acceptance/protect.spec.ts
+++ b/packages/snyk-protect/test/acceptance/protect.spec.ts
@@ -1,6 +1,8 @@
 import protect from '../../src/lib';
+import { getVersion, getHelp } from '../../src/lib';
 import { createProject } from '../util/createProject';
 import { getPatchedLodash } from '../util/getPatchedLodash';
+import { runCommand } from '../util/runCommand';
 import * as http from '../../src/lib/http';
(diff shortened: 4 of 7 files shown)


aliscco (approved):

packages/snyk-protect/src/lib/help/help.md
```diff
@@ -0,0 +1,17 @@
+# @snyk/protect -- Patch vulnerable code in your project's dependencies.
```
aliscco, line 1: #

packages/snyk-protect/src/lib/help.ts
```diff
@@ -0,0 +1,12 @@
+import * as fs from 'fs';
+import stripAnsi from 'strip-ansi';
```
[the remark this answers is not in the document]
Avishagp: Added as dev-dependency

packages/snyk-protect/src/lib/help/help.txt
```diff
@@ -0,0 +1,26 @@
...
+
+       O   Under the "Fix Pull Request" section, ensure patches  are  enabled.
+           Snyk  will  now  include  patches  as  part of its Fix PRs for your
+           project.
+
+
+
+[1mF[0m[1mO[0m[1mR[0m [1mM[0m[1mO[0m[1mR[0m[1mE[0m [1mI[0m[1mN[0m[1mF[0m[1mO[0m[1mR[0m[1mM[0m[1mA[0m[1mT[0m[1mI[0m[1mO[0m[1mN[0m
+       Visit the Snyk Protect GitHub repository:
+       https://github.com/snyk/snyk/tree/master/packages/snyk-protect.
```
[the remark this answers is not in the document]
Avishagp: Gotta say I don't really agree with you on that point. I think that having more info for customers is more valuable, even if they can just click the link. If someone used this flag it means they need some assistance, so giving it is a must in my opinion.
Having a `--help` flag but not giving any information sounds weird to me.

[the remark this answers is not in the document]
jahed-snyk: That's true, but the help we're showing here is purely web-driven so IMO it should be presented on a web browser. Also, if the web flow / instructions change, they'll get the wrong information on older versions if it's baked into the CLI.

If it was something they can do on the terminal specifically on the version they're using, it would be helpful in the terminal.

Another note, the escape characters being used currently may not be supported in all terminals (esp. CI environments). We should avoid formatted text unless if we also want to maintain detecting terminals that support it and stripping ANSI manually (since we are using zero deps).

Taking all of that into account, how about something like:

```
$ snyk-protect --help

NAME
  @snyk/protect - Patch vulnerable code in your project's dependencies.

DESCRIPTION
  You don't typically need to add the @snyk/protect dependency manually.
  For more information on how to setup and use @snyk/protect, visit:
    https://github.com/snyk/snyk/tree/master/packages/snyk-protect

SYNOPSIS
  snyk-protect [<options>]

OPTIONS
  --help
    Display this help text.
  --version
    Display the current version.
```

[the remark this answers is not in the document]
Avishagp: Regarding stripping ANSI, I've already added it as a dev dependency. WDYT?
I like the output you proposed so I've changed `help.txt` accordingly. Thanks!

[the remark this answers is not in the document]
jahed-snyk: @Avishagp If we include `strip-ansi` and run it on users' machines, it will need to be a prod dependency (unless if we use a bundler to include it in the release build). IMO it's not worth the cost and maintenance for a sprinkle of bold text and colours.

Our output is pretty minimal so there's no need for additional formatting. The ALL CAPS and indentation is enough.

If we avoid formatting, we can also remove the markdown source (`help.md`) and move the text into a string template to avoid any complexity coming from the file system lookup (such as relative paths).

packages/snyk-protect/test/fixtures/version-flag/README.md
```diff
@@ -0,0 +1 @@
+Test fixture for version flag
```
jahed-snyk, line 1: We can re-use the same fixture for both help and version since they're copied to a temp directory for each test. Avoids duplication. e.g. name it "basic" or something.

packages/snyk-protect/help.txt
```diff
@@ -0,0 +1,18 @@
+$ snyk-protect --help
+
```
jahed-snyk, line 2: We don't need these lines (it's what the user types in)

packages/snyk-protect/src/lib/index.ts
```diff
@@ -92,4 +102,17 @@ async function protect(projectFolderPath: string) {
   });
 }
 
+export function getHelp(): string {
+  const filePath = path.resolve(__dirname, '../../help.txt');
+  const helpText = fs.readFileSync(filePath, 'utf8');
```
jahed-snyk, line 107: Suggestion but not blocking: Instead of using `fs`, the help text is simple enough now that we can use an inlined string templates. Avoids some complexity.

```ts
    const helpText = `
NAME
  ...
...
    `.trim()

File: packages/snyk-protect/src/lib/index.ts [hunk 7]
```diff
@@ -13,6 +13,16 @@ import { getAllPatches } from './fetch-patches';
 import { sendAnalytics } from './analytics';
 
 async function protect(projectFolderPath: string) {
+  // Handle runs with flags
```
jahed-snyk, line 16: Don't really need this comment, I think it's evident from the `if` statements.

jahed-snyk (approved):
Approved with some minor suggestions.

packages/snyk-protect/src/lib/index.ts
Avishagp: I think it'd make the file too long and less neat 🤔

jahed-snyk: We should move `getHelp` and `getVersion` to their own modules so the `index.ts` is only concerned with wiring the flow. That would also avoid the problem of being too long. But again not blocking, we can always do that later.

PR #2096, opened 2021-07-16
feat: track IaC local execution tests [CC-972]

craigfurman:

#### What does this PR do?

- Calls new IaC CLI tracker endpoint, so that tests performed via the local execution (default) flow can be accounted for.

https://snyksec.atlassian.net/browse/CC-972

### DIFF
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -8,6 +8,7 @@ import {
   EngineType,
 } from './types';
 import { addIacAnalytics } from './analytics';
+import { TestLimitReachedError } from './usage-tracking';
 import { TestResult } from '../../../../lib/snyk-test/legacy';
 import {
   initLocalCache,
@@ -17,6 +18,7 @@ import {
   getIacOrgSettings,
   applyCustomSeverities,
   formatScanResults,
+  trackUsage,
   cleanLocalCache,
 } from './measurable-methods';
 import { isFeatureFlagSupportedForOrg } from '../../../../lib/feature-flags';
@@ -62,6 +64,17 @@ export async function test(
       options,
       iacOrgSettings.meta,
     );
+
+    try {
+      await trackUsage(formattedResults);
+    } catch (e) {
+      if (e instanceof TestLimitReachedError) {
+        throw e;
+      }
+      // If something has gone wrong, err on the side of allowing the user to
+      // run their tests by squashing the error.
+    }
+
     addIacAnalytics(formattedResults);
 
     // TODO: add support for proper typing of old TestResult interface.
--- a/src/cli/commands/test/iac-local-execution/usage-tracking.ts
+++ b/src/cli/commands/test/iac-local-execution/usage-tracking.ts
@@ -1,0 +1,55 @@
+import request = require('../../../../lib/request');
+import config = require('../../../../lib/config');
+import { api as getApiToken } from '../../../../lib/api-token';
+import { CustomError } from '../../../../lib/errors';
+
+export async function trackUsage(
+  formattedResults: TrackableResult[],
+): Promise<void> {
+  const trackingData = formattedResults.map((res) => {
+    return {
+      isPrivate: res.meta.isPrivate,
+      issuesPrevented: res.result.cloudConfigResults.length,
+    };
+  });
+  const trackingResponse = await request({
+    method: 'POST',
+    headers: {
+      Authorization: `token ${getApiToken()}`,
+    },
+    url: `${config.API}/track-iac-usage/cli`,
+    body: { results: trackingData },
+    gzip: true,
+    json: true,
+  });
+  switch (trackingResponse.res.statusCode) {
+    case 200:
+      break;
+    case 429:
+      throw new TestLimitReachedError();
+    default:
+      throw new CustomError(
+        'An error occurred while attempting to track test usage: ' +
+          JSON.stringify(trackingResponse.res.body),
+      );
+  }
+}
+
+export class TestLimitReachedError extends CustomError {
+  constructor() {
+    super(
+      'Test limit reached! You have exceeded your infrastructure as code test allocation for this billing period.',
+    );
+  }
+}
+
+// Sub-interface of FormattedResult that we really only use to make test
+// fixtures easier to create.
+export interface TrackableResult {
+  meta: {
+    isPrivate: boolean;
+  };
+  result: {
+    cloudConfigResults: any[];
+  };
+}
--- a/test/acceptance/fake-server.ts
+++ b/test/acceptance/fake-server.ts
@@ -338,6 +338,12 @@ export function fakeServer(root, apikey) {
     return next();
   });
 
+  server.post(root + '/track-iac-usage/cli', (req, res, next) => {
+    res.status(200);
+    res.send({});
+    return next();
+  });
+
   server.setNextResponse = (response) => {
     server._nextResponse = response;
   };
--- a/src/cli/commands/test/iac-local-execution/analytics.ts
+++ b/src/cli/commands/test/iac-local-execution/analytics.ts
@@ -35,6 +35,7 @@ export enum PerformanceAnalyticsKey {
   OrgSettings = 'org-settings-ms',
   CustomSeverities = 'custom-severities-ms',
   ResultFormatting = 'results-formatting-ms',
+  UsageTracking = 'usage-tracking-ms',
   CacheCleanup = 'cache-cleanup-ms',
   Total = 'total-iac-ms',
 }
@@ -50,6 +51,7 @@ export const performanceAnalyticsObject: Record<
   [PerformanceAnalyticsKey.OrgSettings]: null,
   [PerformanceAnalyticsKey.CustomSeverities]: null,
   [PerformanceAnalyticsKey.ResultFormatting]: null,
+  [PerformanceAnalyticsKey.UsageTracking]: null,
   [PerformanceAnalyticsKey.CacheCleanup]: null,
   [PerformanceAnalyticsKey.Total]: null,
 };
--- a/src/cli/commands/test/iac-local-execution/measurable-methods.ts
+++ b/src/cli/commands/test/iac-local-execution/measurable-methods.ts
@@ -2,6 +2,7 @@ import { loadFiles } from './file-loader';
 import { parseFiles } from './file-parser';
 import { scanFiles } from './file-scanner';
 import { formatScanResults } from './results-formatter';
+import { trackUsage } from './usage-tracking';
 import { cleanLocalCache, initLocalCache } from './local-cache';
 import { applyCustomSeverities } from './org-settings/apply-custom-severities';
 import { getIacOrgSettings } from './org-settings/get-iac-org-settings';
@@ -87,6 +88,11 @@ const measurableFormatScanResults = performanceAnalyticsDecorator(
   PerformanceAnalyticsKey.ResultFormatting,
 );
 
+const measurableTrackUsage = asyncPerformanceAnalyticsDecorator(
+  trackUsage,
+  PerformanceAnalyticsKey.UsageTracking,
+);
+
 const measurableLocalTest = asyncPerformanceAnalyticsDecorator(
   test,
   PerformanceAnalyticsKey.Total,
@@ -100,6 +106,7 @@ export {
   measurableGetIacOrgSettings as getIacOrgSettings,
   measurableApplyCustomSeverities as applyCustomSeverities,
   measurableFormatScanResults as formatScanResults,
+  measurableTrackUsage as trackUsage,
   measurableCleanLocalCache as cleanLocalCache,
   measurableLocalTest as localTest,
 };
--- a/test/jest/unit/iac-unit-tests/usage-tracking.spec.ts
+++ b/test/jest/unit/iac-unit-tests/usage-tracking.spec.ts
@@ -1,0 +1,95 @@
+import {
+  trackUsage,
+  TestLimitReachedError,
+} from '../../../../src/cli/commands/test/iac-local-execution/usage-tracking';
+import { mocked } from 'ts-jest/utils';
+import { NeedleResponse } from 'needle';
+import makeRequest = require('../../../../src/lib/request/request');
+import { CustomError } from '../../../../src/lib/errors';
+
+jest.mock('../../../../src/lib/request/request');
+const mockedMakeRequest = mocked(makeRequest);
+
+const results = [
+  {
+    meta: {
+      isPrivate: true,
+    },
+    result: {
+      cloudConfigResults: ['an issue'],
+    },
+  },
+  {
+    meta: {
+      isPrivate: false,
+    },
+    result: {
+      cloudConfigResults: [],
+    },
+  },
+];
+
+describe('tracking IaC test usage', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('does not throw an error when backend returns HTTP 200', async () => {
+    mockedMakeRequest.mockImplementationOnce(() => {
+      return Promise.resolve({
+        res: { statusCode: 200 } as NeedleResponse,
+        body: {
+          foo: 'bar',
+        },
+      });
+    });
+
+    await trackUsage(results);
+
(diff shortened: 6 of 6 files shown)


src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -64,6 +63,63 @@ export async function test(
...
+      case 200:
+        // carry on
+        break;
+      case 430:
+        // Return only a failure, throw away the tests we've already done
+        return {
+          results: [],
+          failures: [
+            {
+              filePath: '',
```
craigfurman, line 94: Absolutely horrendous abuse of types here just to try out the new endpoint. We probably want to do this right instead.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -64,6 +63,63 @@ export async function test(
...
+                'ERROR: you have exhausted your limit of infrastructure-as-code tests in this billing period.',
+            },
+          ],
+        };
+      default:
+        // Something went wrong on our end. Return the results, but append a
+        // failure related to tracking the tests.
+        // TODO this does not display when scanning only one file, for some reason.
+        failure = {
+          filePath: '',
```
craigfurman, line 106: Similarly terrible abuse of types here. See other note about this.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -64,6 +63,63 @@ export async function test(
...
+      json: true,
+    });
+
+    let failure;
+    switch (trackingResponse.res.statusCode) {
+      case 200:
+        // carry on
+        break;
+      case 430:
+        // Return only a failure, throw away the tests we've already done
```
craigfurman, line 89: ToDo: validate this is the behaviour we want.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -64,6 +63,63 @@ export async function test(
...
+            {
+              filePath: '',
+              fileType: 'json',
+              failureReason:
+                'ERROR: you have exhausted your limit of infrastructure-as-code tests in this billing period.',
+            },
+          ],
+        };
+      default:
+        // Something went wrong on our end. Return the results, but append a
```
craigfurman, line 102: ToDo: validate that this is the behaviour we want.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -64,6 +63,63 @@ export async function test(
...
+        // TODO this does not display when scanning only one file, for some reason.
+        failure = {
+          filePath: '',
+          fileType: 'json',
+          jsonContent: null,
+          engineType: null,
+          err: new Error(),
+          fileContent: '',
+          failureReason:
+            'TODO CHANGEME: print response body with error message',
```
craigfurman, line 113: let's not forget this.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -62,6 +64,42 @@ export async function test(
...
+        // carry on
+        break;
+      case 429:
+        throw new Error(
+          'Test limit reached! You have exceeded your infrastructure as code test allocation for this billing period.',
+        );
+      default:
+        // If something has gone wrong, err on the side of allowing the user to
+        // run their tests.
+        console.log(
```
craigfurman, line 96: I'm not at all sure that this is the best way to present non-fatal errors to CLI users.

craigfurman:
The IaC smoke tests are failing, and I think this is expected, because they run against prod which does not contain the new endpoint introduced in https://github.com/snyk/registry/pull/22111.

src/cli/commands/test/iac-local-execution/index.ts
ipapast: No problem! For errors, we follow this pattern:
1) We create a new error code under IaCErrorCodes here https://github.com/snyk/snyk/blob/93624dcbc620dd966c391de2d1ecef1781b8b6c4/src/cli/commands/test/iac-local-execution/types.ts#L211-L250
2) Then we create a Constructor, usually close to where the error is used (same file). For this you can have a look at any of our Error constructors, for example: https://github.com/snyk/snyk/blob/77336a02c75432da563402091f94653dc1e3d585/src/cli/commands/test/iac-local-execution/parsers/terraform-file-parser.ts#L29-L36 

These errors extend a CustomError which is used across the CLI.
The `userMessage` field will hold the user facing error for the CLI.
The string code will be generated automatically according to the name, and it will also be sent to the CLI analytics.

ipapast (commented):
I tried it with different files and the registry branch and I think it does what it needs to 😉 

Suggestions to take it out of the draft:
1) Export it on its own function
2) Move it to its own file or even in the analytics.ts - in there we also have the issuesCount so we do not need to do `res.result.cloudConfigResults.length` twice. 

3) Ron introduced some timers at some point, which can count how long a call or a flow took. This might be a good thing to add so that we can see how much time our post does. These are located in analytics.ts too, in the `iac-metrics` object.

Edit: I saw the comment about smoke tests, I am not sure it's related, it's all the tests with `--json` flag. 🤔 We can have a look together tomorrow if you like.

src/cli/commands/test/iac-local-execution/index.ts
craigfurman: We ended up chatting on zoom about this. Summary: throwing an error isn't what we want to do. We want to present a non-fatal error to the user that is not in the context of any tests. This might be a new domain concept.

src/cli/commands/test/iac-local-execution/usage-tracking.ts
```diff
@@ -0,0 +1,47 @@
+import { FormattedResult } from './types';
+import request = require('../../../../lib/request');
+import config = require('../../../../lib/config');
+import { api as getApiToken } from '../../../../lib/api-token';
+import { CustomError } from '../../../../lib/errors';
+
+//  TODO analytics shows 0ms
```
craigfurman, line 7: When testing this against my local registry (the companion registry PR has not yet hit production) 
 with `SNYK_API=http://localhost:8000/api/v1 snyk-dev iac test testdata -d`, I noticed that the new analytics I've introduced here shows 0ms. That doesn't look right!

As far as I can tell this is because the measurable-methods timer wrapper only works on sync functions:

https://github.com/snyk/snyk/blob/2aecd40ceda2b18bfe715e09f5e5e9fbdf723cf1/src/cli/commands/test/iac-local-execution/measurable-methods.ts#L42-L44

If the wrapped function returns a promise, we only measure the time taken to yield that promise - not to wait on it. I think we can see this manifest itself in `custom-severities-ms` too - cc @ipapast. It's possible I've missed the point, please do set me straight 😁 

So, I tried to convert this function to a sync one, with the following code:

```diff
diff --git a/src/cli/commands/test/iac-local-execution/index.ts b/src/cli/commands/test/iac-local-execution/index.ts
index 428dc007..201c228f 100644
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -66,7 +66,10 @@ export async function test(
     );
 
     try {
-      await trackUsage(formattedResults);
+      trackUsage(formattedResults);
+      console.log(
+        'should be unreachable if an error is thrown, but is reached',
+      );
     } catch (e) {
       if (e instanceof TestLimitReachedError) {
         throw e;
diff --git a/src/cli/commands/test/iac-local-execution/usage-tracking.ts b/src/cli/commands/test/iac-local-execution/usage-tracking.ts
index f6633fea..cd7462e9 100644
--- a/src/cli/commands/test/iac-local-execution/usage-tracking.ts
+++ b/src/cli/commands/test/iac-local-execution/usage-tracking.ts
@@ -5,16 +5,14 @@ import { api as getApiToken } from '../../../../lib/api-token';
 import { CustomError } from '../../../../lib/errors';
 
 //  TODO analytics shows 0ms
-export async function trackUsage(
-  formattedResults: FormattedResult[],
-): Promise<void> {
+export function trackUsage(formattedResults: FormattedResult[]): void {
   const trackingData = formattedResults.map((res) => {
     return {
       isPrivate: res.meta.isPrivate,
       issuesPrevented: res.result.cloudConfigResults.length,
     };
   });
-  const trackingResponse = await request({
+  request({
     method: 'POST',
     headers: {
       Authorization: `token ${getApiToken()}`,
@@ -23,19 +21,20 @@ export async function trackUsage(
     body: { results: trackingData },
     gzip: true,
     json: true,
+  }).then((trackingResponse) => {
+    switch (trackingResponse.res.statusCode) {
+      case 200:
+        // carry on
+        break;
+      case 429:
+        throw new TestLimitReachedError();
+      default:
+        throw new CustomError(
+          'An error occurred while attempting to track test usage: ' +
+            JSON.stringify(trackingResponse.res.body),
+        );
+    }
   });
-  switch (trackingResponse.res.statusCode) {
-    case 200:
-      // carry on
-      break;
-    case 429:
-      throw new TestLimitReachedError();
-    default:
-      throw new CustomError(
-        'An error occurred while attempting to track test usage: ' +
-          JSON.stringify(trackingResponse.res.body),
-      );
-  }
 }
 
 export class TestLimitReachedError extends CustomError {
```

In the happy path, when no exception is thrown from this function, everything seems to work well, and I get non-zero analytics timers for this function.

However, when I test a sad path, e.g. by dropping the SNYK_API to hit production, which doesn't have the new endpoint yet and returns HTTP 404, I hit my "shouldn't be reached" `console.log`, and the exception is dumped at the end of the CLI's output, presumably by some higher-level code that can wait on the unresolved promise. This is probably a basic javascript error, but everyday's a school day 🤦. @rontalx would you mind taking a look?

craigfurman: With fresh eyes I think I see the problem straight away: `then` doesn't block on the callback finishing, otherwise it wouldn't be very useful 😅

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -62,6 +64,21 @@ export async function test(
...
+      await trackUsage(formattedResults);
+    } catch (e) {
+      if (e instanceof TestLimitReachedError) {
+        throw e;
+      }
+
+      // If something has gone wrong, err on the side of allowing the user to
+      // run their tests.
+      // TODO propagate error
+      console.log(e);
```
craigfurman, line 78: @rontalx @ipapast what do you reckon about just logging these errors, and if so, where and how? From Ben on https://snyksec.atlassian.net/browse/CC-972:

> We shouldn’t block the local exec from running if we can’t report the test data. The test data helps us understand usage and if we chose to enforce test limits would help with that, but i’d optimise for the developer experience of being able to run a test

This produces output that looks like this:

```
CustomError: An error occurred while attempting to track test usage: {"code":429,"message":"Test limit reached!","userMessage":"Test limit reached!","error":"Test limit reached!","errorRef":"c38824c4-539e-412a-98c7-7bfc994831e2"}
    at trackUsage (/Users/craig/workspace/github.com/snyk/snyk/src/cli/commands/test/iac-local-execution/usage-tracking.ts:34:13)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at Object.trackUsage (/Users/craig/workspace/github.com/snyk/snyk/src/cli/commands/test/iac-local-execution/measurable-methods.ts:29:25)
    at test (/Users/craig/workspace/github.com/snyk/snyk/src/cli/commands/test/iac-local-execution/index.ts:69:7)
    at Object.localTest (/Users/craig/workspace/github.com/snyk/snyk/src/cli/commands/test/iac-local-execution/measurable-methods.ts:29:25)
    at test (/Users/craig/workspace/github.com/snyk/snyk/src/cli/commands/test/index.ts:89:39)
    at runCommand (/Users/craig/workspace/github.com/snyk/snyk/src/cli/index.ts:57:25)
    at main (/Users/craig/workspace/github.com/snyk/snyk/src/cli/index.ts:313:11) {
  code: undefined,
  strCode: undefined,
  innerError: undefined,
  userMessage: undefined
}

Testing k8s.yaml...


Infrastructure as code issues:
  ✗ Container is running in privileged mode [High Severity] [SNYK-CC-K8S-1] in Deployment
...
```

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -62,6 +64,21 @@ export async function test(
...
+    }
+
+    // TODO can't collect analytics if we throw an exception above
     addIacAnalytics(formattedResults);
```
craigfurman, line 82: placeholder for a conversation. I think this could possibly be OK, or at least no worse than today: there were already many code paths in which an exception causes `addIacAnalytics` to never be called.

src/cli/commands/test/iac-local-execution/index.ts
ipapast: Hmm I like the idea of not failing the run, but how do we beautify this? :D

ipapast: Should we just console log the message? 
```suggestion
      console.log(e.userMessage);
```

craigfurman: @rontalx and I paired a bit, and he suggested that the user shouldn't really care about these non-fatal errors. I agreed, so we removed the log. I think we can rely on server-side monitoring and alerting on this endpoint.

test/acceptance/fake-server.ts
```diff
@@ -338,6 +338,12 @@ export function fakeServer(root, apikey) {
     return next();
   });
 
+  server.post(root + '/track-iac-usage/cli', (req, res, next) => {
+    res.status(200);
```
craigfurman, line 342: I wanted to add a sad-path test too, but could not figure out how to grab the org from the request context here and do logic based on it, e.g. to return 429 for one org.

craigfurman:
ready for review, and if approved I'll squash before merge

craigfurman:
I'm guessing a lot of these additional code owners are for fake-server.ts?

ipapast:
> ready for review, and if approved I'll squash before merge.
> 
> @ipapast re:
> 
> > Move it to its own file or even in the analytics.ts - in there we also have the issuesCount so we do not need to do res.result.cloudConfigResults.length twice.
> 
> I did end up re-doing the math. Performance-wise, it's a drop in the ocean compared to the network calls and filesystem usage, and duplicating it allows us to decouple unrelated concerns: performance analytics, and usage tracking.

sounds good to me!

>I'm guessing a lot of these additional code owners are for fake-server.ts?
Yes, it belongs to "all" 😄

src/cli/commands/test/iac-local-execution/usage-tracking.ts
```diff
@@ -0,0 +1,46 @@
...
+      Authorization: `token ${getApiToken()}`,
+    },
+    url: `${config.API}/track-iac-usage/cli`,
+    body: { results: trackingData },
+    gzip: true,
+    json: true,
+  });
+  switch (trackingResponse.res.statusCode) {
+    case 200:
+      // carry on
```
ipapast, line 28: nit: do we need this comment? it does not give me any valuable info, I think :)

test/acceptance/fake-server.ts
ipapast: hmm.. we only have the org name after we get it back from `/iac-org-settings` so not sure if you can do it in this context. except if you do both calls and get it from the response? 😬

ipapast (approved):
Looks great, I also tested it locally and checked that the tests I'm running are getting tracked.

After this is merged, I guess we can check if the usage goes through to BQ and to Reporting.

 Nice one!👌🤩

src/cli/commands/test/iac-local-execution/usage-tracking.ts
craigfurman: I was thinking commenting an empty branch makes it clear it's not a mistake, but this doesn't add any real context. I'll just remove it.

ipapast: you don't have to, just wanted to check. :)
I personally don't like comments if they don't offer anything or you can not read it in code, but as I said, personal opinion. 😆

craigfurman: no worries, I think you're right 🙂

anthogez (approved):

PR #2097, opened 2021-07-16
test(protect): update to match fix pr setup

jahed-snyk:
Originally the Fix PR test (used for acceptance tests and production monitoring) use `prepublish` and `npx`, but after recent discoveries while working on migrating Fix PRs to use `@snyk/protect`, we changed that approach. Now we add it to `dependencies` as `latest` and execute it with `npm run snyk-protect`. This test has been update to reflect that.


packages/snyk-protect/test/util/createProject.ts
```diff
@@ -22,6 +25,8 @@ const createProject = async (fixtureName: string): Promise<TestProject> => {
     await useLocalPackage(projectPath);
   }
 
+  debug('createProject: %s', projectPath);
```
jahed-snyk, line 28: Added some debug logging as it's convenient to know where a test project is created for debugging. I used `__filename` as it's useful to know exactly which module is doing the logging.

```
NODE_DEBUG='@snyk/*' npx jest ....
```

packages/snyk-protect/test/util/createProject.ts
```diff
@@ -2,13 +2,16 @@ import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
 import { useLocalPackage } from './useLocalPackage';
+import { debuglog } from 'util';
```
maxjeffos, line 5: did you want camel case on?

packages/snyk-protect/test/util/useLocalPackage.ts
```diff
@@ -1,36 +1,35 @@
...
+import { debuglog } from 'util';
 
 type PackageJSON = {
-  scripts?: Record<string, string>;
+  dependencies?: Record<string, string>;
 };
 
+const debug = debuglog('@snyk' + __filename);
+
 const useLocalPackage = async (projectPath: string) => {
```
maxjeffos, line 12: How about some comments on what this function is for both here and in createProject.ts where you use it? It's not super easy to follow.

maxjeffos (approved):

packages/snyk-protect/test/util/createProject.ts
jahed-snyk: Do you mean for `debuglog`? It's a NodeJS module so don't have a choice here. https://nodejs.org/api/util.html#util_util_debuglog_section_callback

packages/snyk-protect/test/util/useLocalPackage.ts
jahed-snyk: I'll add some comments for it in a separate PR.

PR #2105, opened 2021-07-21
The Webpack One

JackuB:
This PR changes CLI build process to bundle all the code with Webpack before shipping.
This will bring us safer and predictable builds and tests.

Also results in much faster and smaller installation for the users.

I love this show...


src/lib/version.ts
```diff
@@ -4,8 +4,8 @@ import { executeCommand } from './exec';
 
 export async function getVersion(): Promise<string> {
   const root = path.resolve(__dirname, '../..');
-  const filename = path.resolve(root, 'package.json');
-  const version = require(filename).version;
+
+  const { version } = require('../../package.json');
```
jahed-snyk, line 8: If we use `require('package.json')` in our source code, then the bundle will contain the older version before the version bump. So whenever someone runs `snyk --version` they'll get `1.0.0-monorepo`.

We need to build after we perform the minor version bump.

Otherwise we need to use `fs.readFile` so we use the outer package.json rather than the bundled one.

I prefer the latter approach is it align with the package.json that NPM sees (rather than the bundled one).

ghost (approved):

Avishagp (approved):

PR #2107, opened 2021-07-23
Use await syntax in analytics

maxjeffos:

#### What does this PR do?
Switches from callback syntax to await syntax in analytics module.

### DIFF
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -71,11 +71,10 @@ export async function postAnalytics(
     return Promise.resolve();
   }
 
-  const isStandalone = version.isStandaloneBuild();
-
-  const snykVersion = await version.getVersion();
-
   try {
+    const isStandalone = version.isStandaloneBuild();
+    const snykVersion = await version.getVersion();
+
     data.version = snykVersion;
     data.os = osName(os.platform(), os.release());
     data.nodeVersion = process.version;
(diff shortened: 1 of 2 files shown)


src/lib/analytics.ts
```diff
@@ -73,74 +73,70 @@ export function postAnalytics(
...
-        json: true,
-        method: 'post',
-        headers: headers,
-      });
-    })
-    .catch((error) => {
-      debug('analytics', error); // this swallows the analytics error
+  const snykVersion = await version.getVersion();
+
+  try {
```
JackuB, line 78: Why is the getVersion outside of try/catch?

src/lib/protect/apply-patch.js
```diff
@@ -5,7 +5,7 @@ const exec = require('child_process').exec;
 const path = require('path');
 const fs = require('fs');
 const semver = require('semver');
-const errorAnalytics = require('../analytics').postAnalytics;
+const analytics = require('../analytics');
```
JackuB, line 8: Could we use modules?

```suggestion
import { addDataAndSend }  from '../analytics';
```

src/lib/analytics.ts
maxjeffos: because it was before, but that's not a very good reason... fixed

src/lib/protect/apply-patch.js
maxjeffos: 👍

JackuB (approved):

PR #2119, opened 2021-07-27
fix: IaC multi-doc yaml indexing

craigfurman:

#### What does this PR do?

In the IaC local execution data flow, the stage in results-formatter.ts
takes a list of IacFileScanResult as input, and returns a list of
FormattedResult. One key difference between these two objects is that
the former list contains a distinct object per document in a multi-doc
yaml file, but for the latter we are grouping these so that there is a
distinct object per file.

FormattedResults each contain a list of vulnerabilities, so we end up
with vulnerabilities from different YAML documents (in the same file)
sharing a list. Vulnerabilities with the same "resource path"
(Kubernetes example: `spec.containers[web].securityContext.privileged`),
but in distinct YAML documents in the same file, should be
differentiated by the "cloudConfigPath" (analogous to the "path" of
libraries used to address dependecies in the open-source flows), and
also by the line number. Both of these distinguishing fields were
actually being clobbered by their values from the first document in a
file: doc ID was always 0, and the line numbers referred to identical
resources in the first doc rather than subsequent docs.

This needs to be resolved in order to deliver IaC CLI issue-ignore
support in a sane way, and so is tangentially related to
https://snyksec.atlassian.net/browse/CC-990.

This will also help any tooling that infers information from line
numbers, which could help us build things like editor plugins and
language servers in the future.

- https://snyksec.atlassian.net/browse/CC-990

### DIFF
--- a/src/cli/commands/test/iac-local-execution/extract-line-number.ts
+++ b/src/cli/commands/test/iac-local-execution/extract-line-number.ts
@@ -1,4 +1,4 @@
-import { IaCErrorCodes, IacFileScanResult, PolicyMetadata } from './types';
+import { IaCErrorCodes } from './types';
 import { CustomError } from '../../../../lib/errors';
 import {
   CloudConfigFileTypes,
@@ -25,14 +25,15 @@ function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
 }
 
 export function extractLineNumber(
-  scanResult: IacFileScanResult,
-  policy: PolicyMetadata,
+  fileContent: string,
+  fileType: string,
+  cloudConfigPath: string[],
 ): number {
   try {
     return issuesToLineNumbers(
-      scanResult.fileContent,
-      getFileTypeForLineNumber(scanResult.fileType),
-      policy.msg.split('.'), // parser defaults to docId:0 and checks for the rest of the path
+      fileContent,
+      getFileTypeForLineNumber(fileType),
+      cloudConfigPath,
     );
   } catch {
     const err = new FailedToExtractLineNumberError();
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -23,11 +23,18 @@ export function formatScanResults(
   meta: TestMeta,
 ): FormattedResult[] {
   try {
-    // Relevant only for multi-doc yaml files
-    const scannedResultsGroupedByDocId = groupMultiDocResults(scanResults);
-    return scannedResultsGroupedByDocId.map((iacScanResult) =>
-      formatScanResult(iacScanResult, meta, options),
-    );
+    const groupedByFile = scanResults.reduce((memo, scanResult) => {
+      const res = formatScanResult(scanResult, meta, options);
+      if (memo[scanResult.filePath]) {
+        memo[scanResult.filePath].result.cloudConfigResults.push(
+          ...res.result.cloudConfigResults,
+        );
+      } else {
+        memo[scanResult.filePath] = res;
+      }
+      return memo;
+    }, {} as { [key: string]: FormattedResult });
+    return Object.values(groupedByFile);
   } catch (e) {
     throw new FailedToFormatResults();
   }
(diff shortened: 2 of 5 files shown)


src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -61,7 +68,11 @@ function formatScanResult(
       (flag) => options[flag],
     );
     const lineNumber: number = shouldExtractLineNumber
-      ? extractLineNumber(scanResult, policy)
+      ? extractLineNumber(
```
craigfurman, line 71: this fix to line number parsing is not actually tested, which is a big shame. I didn't find a test suite that already exercises this code path, and I didn't take the time to add one. Perhaps this can stay as tech debt for now 😞

test/jest/unit/iac-unit-tests/results-formatter.fixtures.ts
```diff
@@ -89,7 +104,24 @@ export function generateCloudConfigResults(
...
       severity: anotherPolicyStub.severity,
       lineNumber: withLineNumber ? 3 : -1,
-      documentation: anotherPolicyStub.documentation,
+      documentation: 'https://snyk.io/security-rules/SNYK-CC-K8S-2',
+    },
+    {
+      ...yetAnotherPolicyStub,
+      id: yetAnotherPolicyStub.publicId,
+      name: yetAnotherPolicyStub.title,
+      cloudConfigPath: ['[DocId: 1]'].concat(
```
craigfurman, line 113: Having added a fixture input with docID > 0, this part of the expected-result fixture would have failed the assertion due to the clobbering by the first doc that the commit message discusses.

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -23,11 +23,18 @@ export function formatScanResults(
   meta: TestMeta,
 ): FormattedResult[] {
   try {
-    // Relevant only for multi-doc yaml files
-    const scannedResultsGroupedByDocId = groupMultiDocResults(scanResults);
-    return scannedResultsGroupedByDocId.map((iacScanResult) =>
-      formatScanResult(iacScanResult, meta, options),
-    );
+    const groupedByFile = scanResults.reduce((memo, scanResult) => {
```
craigfurman, line 26: the heart of the fix is that we group by file after plucking the docID onto each vuln, rather than reducing that information away and clobbering everything with zero.

src/cli/commands/test/iac-local-execution/extract-line-number.ts
```diff
@@ -25,14 +25,15 @@ function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
...
+  cloudConfigPath: string[],
 ): number {
   try {
     return issuesToLineNumbers(
-      scanResult.fileContent,
-      getFileTypeForLineNumber(scanResult.fileType),
-      policy.msg.split('.'), // parser defaults to docId:0 and checks for the rest of the path
+      fileContent,
+      getFileTypeForLineNumber(fileType),
+      cloudConfigPath,
```
craigfurman, line 36: the path already has a docID prefix, allowing the line parser to extract correct line numbers. We weren't passing it in before!

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -48,7 +55,7 @@ function formatScanResult(
   const formattedIssues = scanResult.violatedPolicies.map((policy) => {
     const cloudConfigPath =
       scanResult.docId !== undefined
-        ? [`[DocId:${scanResult.docId}]`].concat(policy.msg.split('.'))
+        ? [`[DocId: ${scanResult.docId}]`].concat(policy.msg.split('.'))
```
craigfurman, line 58: Introducing a space here achieves 2 things:

- bends this into a format recognised by the line-parsing library (cloud-config-parser), so that line parsing works when docId > 0
- aligns CLI with registry when it comes to cloudConfigPath formatting. Alignment here will be useful when we come to the issue to respect `.snyk` ignores in SCM scans, and we'd rather not change registry's format if we can avoid it, since that is saved in the database. We'd either have to tackle a fiddly data migration, or break existing ignores / reporting data. Maybe both!

src/cli/commands/test/iac-local-execution/results-formatter.ts
aron: I pushed a quick and dirty acceptance test for this functionality

aron (approved):
Nice sleuthing!

craigfurman:
Nice, thanks for pushing that test. LGTM.

PR #2122, opened 2021-07-28
fix: IaC path parsing

craigfurman:

#### What does this PR do?

Fix a bug in which IaC resource paths (more context below) was naively
split by dot, without distinguishing between separator dots and dots
inside string literals.

For example: "foo.bar.baz" was split into 3 words correctly, but
"metadata.annotations['container.apparmor.security.beta.kubernetes.io/web']"
was not.

The "path" / "cloud config path" is a sort of address used to identify
locations in files where vulnerabilities reside. It is very similar to a
jsonpath. The user receives it as an array of components, which in the
future can be useful for building tooling around, e.g. the upcoming CLI
ignores.

The code that evaluates policies against source files is also
responsible for returning a representation of this path, which it does
as a dot-separated string.

Implementation note: with PEGjs, if we want to we can pre-compile the grammar into a javascript file, which could offer some small performance improvements. I didn't do this to preserve simplicity of iteration, because realistically the other operations of the CLI, including network calls, take orders of magnitude longer than compiling this parser.

`snyk iac test deployment.yaml` against https://github.com/snyk/infrastructure-as-code-goof/blob/master/k8s/PrivilegedPod/web/deployment.yaml returns vulnerabilities including:

```
  ✗ Container is running without AppArmor profile [Low Severity] [SNYK-CC-K8S-32] in Deployment
    introduced by metadata > annotations['container > apparmor > security > beta > kubernetes > io/web']
```

You can see the mangled path there above. With this code, the same output is:

```
  ✗ Container is running without AppArmor profile [Low Severity] [SNYK-CC-K8S-32] in Deployment
    introduced by metadata > annotations['container.apparmor.security.beta.kubernetes.io/web']
```

The `--json` output shows the fix similarly.

IaC CLI scans are "local execution" by default, and do no processing on registry. For IaC SCM imported projects, registry contains its own logic for parsing paths: https://github.com/snyk/registry/blob/96aebaf2e16f14bac9c939f9b0717caa09fcce4f/src/lib/snapshots/cloud-config-issues.ts#L155-L164

As you can see, it achieves the same aims as this PR but only for "annotations", a Kubernetes special case.

Ideally, we'll align on the same code in here and registry (maybe in https://github.com/snyk/cloud-config-parser) in order to minimise differences in path handling, which will help us avoid bugs in an upcoming feature to have registry respect `.snyk` ignores for IaC imported projects.

Tangentially related to https://snyksec.atlassian.net/browse/CC-990. In order to deliver CLI ignores, we want the user to copy-paste resource paths from the CLI's output into the `.snyk` file. If the resource paths are mangled this will be confusing!

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "open": "^7.0.3",
     "ora": "5.4.0",
     "os-name": "^3.0.0",
+    "pegjs": "^0.10.0",
     "promise-queue": "^2.2.5",
     "proxy-from-env": "^1.0.0",
     "rimraf": "^2.6.3",
--- a/src/cli/commands/test/iac-local-execution/parsers/path.ts
+++ b/src/cli/commands/test/iac-local-execution/parsers/path.ts
@@ -1,0 +1,35 @@
+import * as peg from 'pegjs';
+
+const grammar = `
+start
+  = element+
+
+element
+  = component:component "."? { return component; }
+
+component
+  = $(identifier index?)
+
+identifier
+  = $([^'"\\[\\]\\.]+)
+
+index
+  = $("[" ['"]? [^'"\\]]+ ['"]? "]")
+`;
+
+export const parsePath = createPathParser();
+
+function createPathParser(): (expr: string) => string[] {
+  const parser = peg.generate(grammar);
+  return (expr: string) => {
+    try {
+      return parser.parse(expr);
+    } catch (e) {
+      // I haven't actually been able to write a testcase that triggers this
+      // code path, but I've included it anyway as a fallback to allow users to
+      // keep using the CLI even if this does occur. Their paths might look
+      // strange, but that's better than nothing.
+      return expr.split('.');
+    }
+  };
+}
--- a/src/cli/commands/test/iac-local-execution/results-formatter.ts
+++ b/src/cli/commands/test/iac-local-execution/results-formatter.ts
@@ -7,6 +7,7 @@ import {
   PolicyMetadata,
   TestMeta,
 } from './types';
+import { parsePath } from './parsers/path';
 import * as path from 'path';
 import { SEVERITY } from '../../../../lib/snyk-test/common';
 import { IacProjectType } from '../../../../lib/iac/constants';
@@ -55,7 +56,7 @@ function formatScanResult(
   const formattedIssues = scanResult.violatedPolicies.map((policy) => {
     const cloudConfigPath =
       scanResult.docId !== undefined
-        ? [`[DocId: ${scanResult.docId}]`].concat(policy.msg.split('.'))
+        ? [`[DocId: ${scanResult.docId}]`].concat(parsePath(policy.msg))
         : policy.msg.split('.');
 
     const flagsRequiringLineNumber = [
--- a/test/jest/unit/iac-unit-tests/path-parser.spec.ts
+++ b/test/jest/unit/iac-unit-tests/path-parser.spec.ts
@@ -1,0 +1,19 @@
+import { parsePath } from '../../../../src/cli/commands/test/iac-local-execution/parsers/path';
+
+describe('parsing cloudConfigPath', () => {
+  it.each([
+    ['foo', ['foo']],
+    ['foo.bar.baz', ['foo', 'bar', 'baz']],
+    ['foo_1._bar2.baz3_', ['foo_1', '_bar2', 'baz3_']],
+    ['foo.bar[abc].baz', ['foo', 'bar[abc]', 'baz']],
+    ['foo.bar[abc.def].baz', ['foo', 'bar[abc.def]', 'baz']],
+    ["foo.bar['abc.def'].baz", ['foo', "bar['abc.def']", 'baz']],
+    ['foo.bar["abc.def"].baz', ['foo', 'bar["abc.def"]', 'baz']],
+    ["foo.bar['abc/def'].baz", ['foo', "bar['abc/def']", 'baz']],
+    ["foo.bar['abcdef'].baz", ['foo', "bar['abcdef']", 'baz']],
+    ["bar['abc.def']", ["bar['abc.def']"]],
+    ["fo%o.bar['ab$c/def'].baz", ['fo%o', "bar['ab$c/def']", 'baz']],
+  ])('%s', (input, expected) => {
+    expect(parsePath(input)).toEqual(expected);
+  });
+});


src/cli/commands/test/iac-local-execution/parsers/path.ts
```diff
@@ -0,0 +1,27 @@
...
+identifier
+  = $([a-zA-Z0-9_]+)
+
+index
+  = $("[" ['"]? [^'"\\]]+ ['"]? "]")
+`;
+
+export const parsePath = createPathParser();
+
+function createPathParser(): (expr: string) => string[] {
```
teodora-sandu, line 22: Just curious, was there no other way to implement this? You basically want to split by dots, apart from when they're inside square brackets, right?

Would  a JavaScript implementation of this not work? Just a bit worried about having to come back to this and re-understand the grammar.
Screenshot 2021-07-28 at 16 39 58

craigfurman: Good question! There are definitely other ways to implement this, as your screenshot suggests! 😁 

I can understand why the choice of a parser generator might be controversial. The generic argument in favour of them is that parser _code_ is a solved problem, and often infamously tangled and hard to follow, so split policy and protocol by passing a grammar to a thing that generates code. On the other hand, 

I made a subjective judgement call that it would be overall simpler to read and extend. The grammar and its client code is very small, and if the problem required regular expressions and a lot of `if` statements I'd argue more strongly in favour of keeping it. However, now that I look again, I can see that the problem definition as it currently stands doesn't even really require regular expressions, so it's a close call. Personally, if we ever had to change this code again to go beyond splitting by dot, I'd prefer the parser generator implementation, but I'm not sure the problem ever will be extended in that way.

Anyway, this was probably too many words to say that I'm not really sure anymore which style of implementation is better. Please feel free to PR this branch, or post a git patch here. If it passes the tests in this PR, it's a workable solution!

teodora-sandu: I'll add a commit patch just so it's all documented in one (should have done that to begin with but to be honest, didn't know it was a thing 😅 )

src/cli/commands/test/iac-local-execution/parsers/path.ts
```diff
@@ -0,0 +1,27 @@
...
+  const parser = peg.generate(grammar);
+  return (expr: string) => {
+    return parser.parse(expr);
+  };
+}
```
teodora-sandu, line 27:
```suggestion
export function parsePath(s: string): string[] {
  const paths: string[] = [];

  let i = s.indexOf('[');
  while (i >= 0) {
    const startPath = s.substring(0, i).lastIndexOf('.');
    const endPath = i + s.substring(i).indexOf(']');

    if (startPath > 0) {
      paths.push(...s.substring(0, endPath).split('.'));
    }
    paths.push(s.substring(startPath + 1, endPath + 1));

    s = s.substring(endPath + 2);
    i = s.indexOf('[');
  }
  if (s.length > 0) {
    paths.push(...s.split('.'));
  }
  return paths;
}

```

craigfurman: That doesn't quite work - we appear to be getting some extra array elements, e.g. in the `foo.bar['abc.def'].baz` test case. I haven't debugged why yet (it's too late in the day for that!), but it does kind of illustrate that this sort of code is always going to be relatively fiddly.

teodora-sandu: Sorry I did some refactoring before copying it over and didn't re-run the tests! I've updated the suggestion and actually checked the tests again this time

src/cli/commands/test/iac-local-execution/parsers/path.ts
craigfurman: oh yeah, multiline suggestions - I meant `git diff | pbcopy` -> paste into comment, but your way is better 😉

src/cli/commands/test/iac-local-execution/parsers/path.ts
craigfurman: @teodora-sandu I pushed another commit to be a little less restrictive about user input, and provide a fallback path in case there is some input we haven't thought of that would trigger a parse exception.

Your suggestion passes all the tests 🎊 

After sleeping on this, I do still think the PEG implementation is easier to read and change than the code in the suggestion. Not that I think the code in the suggestion is "bad" - it appears to be the most reasonable function that passes the tests. Also my opinion here isn't very strong tbh.

I'm not sure how to resolve this, apart from getting more opinions.

teodora-sandu: Yes please ask for more opinions, I haven't properly reviewed this PR since this is a biased suggestion and I didn't want to block you!

craigfurman:
@p0tr3c this fixes the consumer end, but I'd sleep better if we had guards on the producer end too 😅 

Ideally, we'd have tests in opa-policies that assert that the `msg` field on returned results can be parsed in the way we parse it here. In our rego test harness, do you know if it's possible for us to add it in a way that runs automatically for every policy, and doesn't require policy authors to copy-paste the same test every time we write a new policy?

I could look into this myself but figured I'd take a shortcut by asking someone who is very familiar with that repo first 😁

p0tr3c:
> @p0tr3c this fixes the consumer end, but I'd sleep better if we had guards on the producer end too 😅
> 
> Ideally, we'd have tests in opa-policies that assert that the `msg` field on returned results can be parsed in the way we parse it here. In our rego test harness, do you know if it's possible for us to add it in a way that runs automatically for every policy, and doesn't require policy authors to copy-paste the same test every time we write a new policy?
> 
> I could look into this myself but figured I'd take a shortcut by asking someone who is very familiar with that repo first 😁

Great timing of yesterday rego session :-D. We can add test [here](https://github.com/snyk/cloud-config-opa-policies/blob/master/lib/testing/main.rego#L29) to implement the check. Have a go at it and I am happy to review

src/cli/commands/test/iac-local-execution/parsers/path.ts
craigfurman: Actually, just to be difficult, I think my opinion has hardened a bit already - not unshakeably so, but still a bit 😛 

Even though this looks easy, it is a parsing problem. Probably the simplest possible "parsing problem" that can't be solved with a few lines and a regex, but a parsing problem nonetheless. Writing parser code is fiddly, and the resultant code is inevitably hard to follow. Parser generators solve this problem in a well-established way, and create a neat split between policy (the grammar) and protocol (the parser generation code itself).

src/cli/commands/test/iac-local-execution/parsers/path.ts
craigfurman: thread continues on the suggestion below 👇

src/cli/commands/test/iac-local-execution/parsers/path.ts
```diff
@@ -0,0 +1,35 @@
+import peg = require('pegjs');
```
gitphill, line 1: import from? (not require)

gitphill (approved):

package.json
```diff
@@ -114,6 +114,7 @@
     "open": "^7.0.3",
     "ora": "5.4.0",
     "os-name": "^3.0.0",
+    "pegjs": "^0.10.0",
```
jahed-snyk, line 117: > PEG.js is still very much work in progress. There are no compatibility guarantees until version 1.0

Not so sure about this. Also hasn't had a stable release for 5 years.

https://github.com/pegjs/pegjs/tree/master/packages/pegjs
https://snyk.io/advisor/npm-package/pegjs

craigfurman: @jahed-snyk I think the decision tree here is:

First: Parser generator vs handwritten parser. The problem I’m solving in the PR is admittedly right on the boundary between “a few lines and a regex” and “a proper parser is actually the most optimal solution”. Please see threads above for a discussion between me and @teodora-sandu on this.

If we do keep the parser generator approach, that does basically necessitate the use of an open source library. The next decision is: which one? Please see https://snyk.slack.com/archives/CFTAUCYQ7/p1627569110241000?thread_ts=1627479562.212600&cid=CFTAUCYQ7 for some discussion on that topic too.

I'm not trying to shut down the discussion, but I am trying to keep it to as few fronts as possible.

craigfurman: or, let's move it here from slack to be async-friendly 🙂 cc @gitphill

craigfurman: if we do want to keep a parser generator library (and I have to admit this much discussion is making me rethink that a bit), do you think https://snyk.io/advisor/npm-package/peggy would be less controversial? It's touted as a successor to pegjs, but is much newer.

craigfurman: Another alternative we could move pegjs/peggy to dev dependencies, and generate and check in the parser rather than dynamically generating it.

ghost (approved):

teodora-sandu (approved):

package.json
craigfurman: discussion continued here: https://snyk.slack.com/archives/C0127HWU0E7/p1627569982037100. we decided to keep pegjs 🤞

jahed-snyk: To summarise, an alternative solution will be explored if and when pegjs causes a notable problem for us.

PR #2125, opened 2021-07-29
chore: execute snyk and snyk-protect using bin file

Avishagp:

#### What does this PR do?
Execute `bin/snyk` and `packages/snyk-protect/bin/snyk-protect` instead of the respective `index.js` files.
This solves this bug: https://github.com/npm/cli/issues/2632 and makes running snyk more predictable and clear.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "files": [
     "help",
     "dist",
+    "bin",
     "config.default.json",
     "SECURITY.md",
     "LICENSE",
@@ -18,7 +19,7 @@
     "doc": "help/commands-help"
   },
   "bin": {
-    "snyk": "dist/cli/index.js"
+    "snyk": "bin/snyk"
   },
   "engines": {
     "node": ">=10"
@@ -186,7 +187,8 @@
   },
   "pkg": {
     "scripts": [
-      "dist/**/*.js"
+      "dist/**/*.js",
+      "bin/snyk"
     ],
     "assets": [
       "config.default.json",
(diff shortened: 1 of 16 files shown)


muscar (approved):

ipapast (approved):

orsagie (approved):

package.json
```diff
@@ -186,14 +187,16 @@
...
+      "bin"
     ],
     "assets": [
       "config.default.json",
       "test-unpublished.json",
       "help/**/*.txt",
       "dist/STANDALONE",
-      "node_modules/@deepcode/dcignore"
+      "node_modules/@deepcode/dcignore",
+      "bin"
```
jahed-snyk, line 199: Why do we need `bin` in `assets` if it's also in `scripts`?

packages/cli-alert/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "dist/index.js",
   "private": true,
   "scripts": {
-    "start": "npm run build && node dist/index.js",
+    "start": "npm run build && node bin/snyk",
```
jahed-snyk, line 8: This package.json is for `cli-alert`, not Snyk CLI. We can leave it as it is.

ghost (commented):
Left some comments.

It's difficult to tell from GitHub's interface but make sure to `chmod +x` any files under `./bin` so they can be executed directly using the shebangs.

packages/cli-alert/package.json
Avishagp: Nice catch, thanks!

ghost (approved):

JackuB:
Merging, even without Tech Services approval, as it will greatly advance #2105 

Screen Shot 2021-07-29 at 18 43 18

Snyk Fix is not using the bin field that we removed

PR #2131, opened 2021-07-29
Add acceptance test for analytics

maxjeffos:

#### What does this PR do?

Adds an acceptance test for analytics which verifies the shape of data being sent using fake-server.

### DIFF
--- a/test/jest/acceptance/analytics.spec.ts
+++ b/test/jest/acceptance/analytics.spec.ts
@@ -1,0 +1,90 @@
+import { fakeServer } from '../../acceptance/fake-server';
+import { runSnykCLI } from '../util/runSnykCLI';
+
+describe('analytics module', () => {
+  let server;
+  let env: Record<string, string>;
+
+  beforeAll((done) => {
+    const port = process.env.PORT || process.env.SNYK_PORT || '12345';
+    const baseApi = '/api/v1';
+    env = {
+      ...process.env,
+      SNYK_API: 'http://localhost:' + port + baseApi,
+      SNYK_HOST: 'http://localhost:' + port,
+      SNYK_TOKEN: '123456789',
+    };
+
+    server = fakeServer(baseApi, env.SNYK_TOKEN);
+    server.listen(port, () => {
+      done();
+    });
+  });
+
+  afterAll((done) => {
+    server.close(() => {
+      done();
+    });
+  });
+
+  test('validate actual analytics call', async () => {
+    server.setNextResponse({});
+    const { code } = await runSnykCLI(`version --org=fooOrg --all-projects`, {
+      env,
+    });
+    expect(code).toBe(0);
+
+    const lastRequest = server.popRequest();
+    expect(lastRequest).toMatchObject({
+      headers: {
+        host: 'localhost:12345',
+        accept: 'application/json',
+        authorization: 'token 123456789',
+        'content-type': 'application/json; charset=utf-8',
+        'x-snyk-cli-version': '1.0.0-monorepo',
+      },
+      query: {
+        org: 'fooOrg',
+      },
+      body: {
+        data: {
+          args: [
+            {
+              org: 'fooOrg',
+              allProjects: true,
+            },
+          ],
+          ci: expect.any(Boolean),
+          command: 'version',
+          durationMs: expect.any(Number),
+          environment: {
+            npmVersion: expect.any(String),
+          },
+          id: expect.any(String),
+          integrationEnvironment: '',
+          integrationEnvironmentVersion: '',
+          integrationName: '',
+          integrationVersion: '',
+          // prettier-ignore
+          metrics: {
+            'network_time': {
+              type: 'timer',
+              values: [],
+              total: expect.any(Number),
+            },
+            'cpu_time': {
+              type: 'synthetic',
+              values: [expect.any(Number)],
+              total: expect.any(Number),
+            },
+          },
+          nodeVersion: process.version,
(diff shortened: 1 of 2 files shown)


test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  expect(lastRequest.headers.host).toBe('localhost:12345');
+  expect(lastRequest.headers.accept).toBe('application/json');
+  expect(lastRequest.headers.authorization).toBe('token 123456789');
+  expect(lastRequest.headers['content-type']).toBe(
+    'application/json; charset=utf-8',
+  );
+  expect(lastRequest.headers['x-snyk-cli-version']).toBe('1.0.0-monorepo');
+  expect(lastRequest.query).toEqual({
+    org: 'fooOrg',
+  });
```
jahed-snyk, line 56: This seems really verbose. Did you try using `toEqual` instead of matching individual properties? Alternatively `toMatchObject` should work for partial matching.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,80 @@
+import { fakeServer } from '../../acceptance/fake-server';
+import { runSnykCLI } from '../util/runSnykCLI';
+
+let server;
+let env: Record<string, string>;
```
jahed-snyk, line 5: Having these on module scope might cause a memory leak (when run alongside lots of tests). I don't remember if that's true or not.

Might be worth moving this into a `describe` (with everything else) just in case.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  const port = process.env.PORT || process.env.SNYK_PORT || '12345';
+  const BASE_API = '/api/v1';
+  env = {
+    ...process.env,
+    SNYK_API: 'http://localhost:' + port + BASE_API,
+    SNYK_HOST: 'http://localhost:' + port,
+    SNYK_TOKEN: '123456789',
+  };
+
+  const apiKey = '123456789';
```
jahed-snyk, line 17: Can remove this and use `env.SNYK_TOKEN`.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+afterAll((done) => {
+  server.close(() => {
+    done();
+  });
+});
+
+afterAll(async () => {
+  await server.close();
+});
```
jahed-snyk, line 32: You've got 2 `afterAll`s here.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,80 @@
+import { fakeServer } from '../../acceptance/fake-server';
+import { runSnykCLI } from '../util/runSnykCLI';
+
+let server;
+let env: Record<string, string>;
+
+beforeAll((done) => {
+  const port = process.env.PORT || process.env.SNYK_PORT || '12345';
+  const BASE_API = '/api/v1';
```
jahed-snyk, line 9: This isn't a global constant so can be in camelCase.

test/jest/acceptance/analytics.spec.ts
maxjeffos: Great suggestion - done

test/jest/acceptance/analytics.spec.ts
maxjeffos: done

test/jest/acceptance/analytics.spec.ts
maxjeffos: thanks!

test/jest/acceptance/analytics.spec.ts
maxjeffos: thanks

test/jest/acceptance/analytics.spec.ts
maxjeffos: done

test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+          durationMs: expect.any(Number),
+          environment: {
+            npmVersion: expect.any(String),
+          },
+          id: expect.any(String),
+          integrationEnvironment: '',
+          integrationEnvironmentVersion: '',
+          integrationName: '',
+          integrationVersion: '',
+          // prettier-ignore
```
jahed-snyk, line 73: What's this ignoring?

test/jest/acceptance/analytics.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    server.setNextResponse({});
+
+    const { code, stdout } = await runSnykCLI(
+      `version --org=fooOrg --all-projects`,
+      {
+        env,
+      },
+    );
+    expect(code).toBe(0);
+    expect(stdout).toBe('1.0.0-monorepo');
```
jahed-snyk, line 40: Since this test is for analytics we should avoid asserting on `--version` so that any changes to that flag's behaviour doesn't weirdly fail a test about analytics.

ghost (approved):

test/jest/acceptance/analytics.spec.ts
maxjeffos: The `'network_time'` and `'cpu_time'`: without prettier ignored, it will reformat to take away the single quotes, but then we get an ESLint error for not using camel case.

PR #2132, opened 2021-07-29
chore: remove broken release notes action

ghost:
The "Release Notes Confirmation" check keeps hanging since yesterday. See #2127 and #2123

We decided it's not worth keeping around. It's more or less just previewing release commits. DangerJS already enforces various formatting rules and prompts if a PR will trigger a release so "Release Notes Preview" doesn't do much.


JackuB (approved):
2f0

PR #2134, opened 2021-07-30
feat: IaC ignore support

craigfurman:

#### What does this PR do?

- Non-JSON output includes full targetFile pathname, not just the basename. These are paths relative to the root of the scan, which might be needed by users in order to write path-scoped ignore rules (more below).
- Support for IaC ignores with 3 levels of specificity: all instances of an issue, all instances in a file, and one specific instance by resource path.
- Add some scaffolding for an ignored-issues analytics counter. We can't turn this on until we ensure there is backend support for it (https://snyksec.atlassian.net/browse/CC-995).

All the usual policy flags from the CLI are supported, including `--policy-path` to pass in an out-of-tree `.snyk`. The logic used to load the policy is the same as used by other CLI subsystems (open-source, container).

Without `--policy-path`, `.snyk` files will be searched for in each directory passed to `snyk iac test` (default is the CWD). In this way you can use multiple, directory-scoped `.snyk`s. File paths in file-scoped ignore rules must be relative to the directory under test. Note that there is a 1:1 correspondence between .snyk policy files and root directories under test. If you only test 1 directory, recursively, you must use only one `.snyk` file. This is a limitation of the current policy file framework, not a new limitation introduced by this PR.

When using `--policy-path`, you can pass it either a directory in which it will look for a file named `.snyk`, or a file path - but the file must be named `.snyk`. This is a bit unexpected, but is a limitation of the current policy file framework, not a new limitation introduced by this PR. If we want to fix it, we can do that separarately.

All of this affects only `snyk iac test`, not `snyk test`. Ignore support has not been added to the `snyk iac test test --legacy` flow either (but the output formatting does affect this flow).

I'd recommend going commit-by-commit. The middle commit that actually adds the ignore support is a little chunky admittedly, but I couldn't really break it down much.

First, a developer can run snyk ignore --id=SNYK-CC-K8S-1 , which will generate a .snyk policy file with the following content if it doesn’t already exist. If it does exist, that command will append the relevant policy stanza:

```yaml
version: v1.19.0
ignore:
  SNYK-CC-K8S-1:
    - '*':
        reason: None Given
        expires: 2021-08-26T08:40:35.249Z
        created: 2021-07-27T08:40:35.251Z
```

You can also handwrite this file, or copy-paste a template from our docs.

Now, snyk iac test (with an optional --policy-path=/path/to/policy-file if you don’t want to keep .snyk in the root of a repository) will ignore all instances of that particular issue.

If you want to scope the ignore down to a single file, you can edit that * to instead be the relative path to the file in question. For example:

```yaml
version: v1.19.0
ignore:
  SNYK-CC-K8S-1:
    - 'staging/deployment.yaml':
        reason: None Given
        expires: 2021-08-26T08:40:35.249Z
        created: 2021-07-27T08:40:35.251Z
```

Finally, in order to scope the ignore down to one particular instance, you can append the `>`-separated “resource path” to the filename, which you can copy-paste from the CLI’s output. For example:

```yaml
version: v1.19.0
ignore:
  SNYK-CC-K8S-1:
    - 'k8s.yaml > [DocId:1] > spec > template > spec > containers[web] > securityContext > privileged':
        reason: None Given
        expires: 2021-08-26T08:40:35.249Z
        created: 2021-07-27T08:40:35.251Z
```

- I haven't written documentation for this new feature yet, but that's next on my list.
- formatted-results.json was captured by console.logging a `FormattedResult[]` from an actual scan. It has some metadata about my org and computer in there, but I don't see any security risks. Perhaps double check that.]
- Because I changed the smoke tests, I pushed a `smoke/feat/iac-ignores` branch to the same HEAD. You can see the resultant github checks in the regular checks area below.

https://snyksec.atlassian.net/browse/CC-990

### DIFF
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -22,6 +22,10 @@ const keys: (keyof IaCTestFlags)[] = [
   'quiet',
   'scan',
   'legacy',
+
+  // PolicyOptions
+  'ignore-policy',
+  'policy-path',
 ];
 const allowed = new Set<string>(keys);
 
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -69,10 +69,13 @@ export async function test(
       iacOrgSettings.meta,
     );
 
-    const filteredResults = filterIgnoredIssues(policy, formattedResults);
+    const { filteredIssues, ignoreCount } = filterIgnoredIssues(
+      policy,
+      formattedResults,
+    );
 
     try {
-      await trackUsage(filteredResults);
+      await trackUsage(filteredIssues);
     } catch (e) {
       if (e instanceof TestLimitReachedError) {
         throw e;
@@ -81,11 +84,11 @@ export async function test(
       // run their tests by squashing the error.
     }
 
-    addIacAnalytics(filteredResults);
+    addIacAnalytics(filteredIssues, ignoreCount);
 
     // TODO: add support for proper typing of old TestResult interface.
     return {
-      results: (filteredResults as unknown) as TestResult[],
+      results: (filteredIssues as unknown) as TestResult[],
       // NOTE: No file or parsed file data should leave this function.
       failures: isLocalFolder(pathToScan)
         ? failedFiles.map(removeFileContent)
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -9,6 +9,7 @@ import {
 } from './types';
 import { addIacAnalytics } from './analytics';
 import { TestLimitReachedError } from './usage-tracking';
+import { filterIgnoredIssues } from './policy';
 import { TestResult } from '../../../../lib/snyk-test/legacy';
 import {
   initLocalCache,
@@ -24,6 +25,7 @@ import {
 import { isFeatureFlagSupportedForOrg } from '../../../../lib/feature-flags';
 import { FlagError } from './assert-iac-options-flag';
 import config = require('../../../../lib/config');
+import { findAndLoadPolicy } from '../../../../lib/policy/find-and-load-policy';
 
 // this method executes the local processing engine and then formats the results to adapt with the CLI output.
 // this flow is the default GA flow for IAC scanning.
@@ -37,6 +39,8 @@ export async function test(
     const customRulesPath = await customRulesPathForOrg(options.rules, org);
 
     await initLocalCache({ customRulesPath });
+
+    const policy = await findAndLoadPolicy(pathToScan, 'iac', options);
 
     const filesToParse = await loadFiles(pathToScan, options);
     const { parsedFiles, failedFiles } = await parseFiles(
@@ -65,8 +69,10 @@ export async function test(
       iacOrgSettings.meta,
     );
 
+    const filteredResults = filterIgnoredIssues(policy, formattedResults);
+
     try {
-      await trackUsage(formattedResults);
+      await trackUsage(filteredResults);
     } catch (e) {
       if (e instanceof TestLimitReachedError) {
         throw e;
@@ -75,11 +81,11 @@ export async function test(
       // run their tests by squashing the error.
     }
 
-    addIacAnalytics(formattedResults);
+    addIacAnalytics(filteredResults);
 
     // TODO: add support for proper typing of old TestResult interface.
     return {
-      results: (formattedResults as unknown) as TestResult[],
+      results: (filteredResults as unknown) as TestResult[],
       // NOTE: No file or parsed file data should leave this function.
       failures: isLocalFolder(pathToScan)
         ? failedFiles.map(removeFileContent)
--- a/src/cli/commands/test/iac-local-execution/policy.ts
+++ b/src/cli/commands/test/iac-local-execution/policy.ts
@@ -8,7 +8,9 @@ export function filterIgnoredIssues(
   if (!policy) {
     return { filteredIssues: results, ignoreCount: 0 };
   }
-  const vulns = results.map((res) => policy.filter(toIaCVulnAdapter(res)));
+  const vulns = results.map((res) =>
+    policy.filter(toIaCVulnAdapter(res), undefined, 'exact'),
+  );
   const ignoreCount: number = vulns.reduce(
     (totalIgnored, vuln) => totalIgnored + vuln.filtered.ignore.length,
     0,
@@ -44,10 +46,8 @@ function toIaCVulnAdapter(result: FormattedResult): IacVulnAdapter {
         // splice.
         // Insert the targetFile into the path so that it is taken into account
         // when determining whether an ignore rule should be applied.
-        // Insert garbage into the first element because the policy library
-        // ignores it.
         const path = [...annotatedResult.cloudConfigPath];
-        path.splice(0, 0, 'GARBAGE', result.targetFile);
+        path.splice(0, 0, result.targetFile);
 
         return {
           id: cloudConfigResult.id,
@@ -73,7 +73,7 @@ function toFormattedResult(adapter: IacVulnAdapter): FormattedResult {
         // including target file context. As that logic changes, so must this.
         const annotatedResult = res as AnnotatedResult;
         const significantPath = [...annotatedResult.cloudConfigPath];
-        significantPath.splice(0, 0, 'GARBAGE', original.targetFile);
+        significantPath.splice(0, 0, original.targetFile);
 
         if (vuln.from.length !== significantPath.length) {
           return false;
--- a/src/cli/commands/test/iac-local-execution/policy.ts
+++ b/src/cli/commands/test/iac-local-execution/policy.ts
@@ -4,13 +4,17 @@ import { Policy } from '../../../../lib/policy/find-and-load-policy';
 export function filterIgnoredIssues(
   policy: Policy | undefined,
   results: FormattedResult[],
-): FormattedResult[] {
+) {
   if (!policy) {
-    return results;
+    return { filteredIssues: results, ignoreCount: 0 };
   }
-  return results
-    .map((res) => policy.filter(toIaCVulnAdapter(res)))
-    .map((vuln) => toFormattedResult(vuln));
+  const vulns = results.map((res) => policy.filter(toIaCVulnAdapter(res)));
+  const ignoreCount: number = vulns.reduce(
+    (totalIgnored, vuln) => totalIgnored + vuln.filtered.ignore.length,
+    0,
+  );
+  const filteredIssues = vulns.map((vuln) => toFormattedResult(vuln));
+  return { filteredIssues, ignoreCount };
 }
 
 type IacVulnAdapter = {
@@ -19,6 +23,7 @@ type IacVulnAdapter = {
     from: string[];
   }[];
   originalResult: FormattedResult;
+  filtered?: { ignore: any[] };
 };
 
 // This is a total cop-out. The type I really want is AnnotatedIacIssue from
--- a/src/cli/commands/test/iac-local-execution/policy.ts
+++ b/src/cli/commands/test/iac-local-execution/policy.ts
@@ -1,0 +1,87 @@
+import { FormattedResult, PolicyMetadata } from './types';
+import { Policy } from '../../../../lib/policy/find-and-load-policy';
+
+export function filterIgnoredIssues(
+  policy: Policy | undefined,
+  results: FormattedResult[],
+): FormattedResult[] {
+  if (!policy) {
+    return results;
+  }
+  return results
+    .map((res) => policy.filter(toIaCVulnAdapter(res)))
+    .map((vuln) => toFormattedResult(vuln));
+}
+
+type IacVulnAdapter = {
+  vulnerabilities: {
+    id: string;
+    from: string[];
+  }[];
+  originalResult: FormattedResult;
+};
+
+// This is a total cop-out. The type I really want is AnnotatedIacIssue from
+// src/lib/snyk-test/iac-test-result.ts, but that appears to only be used in the
+// legacy flow and I gave up on adapting it to work in both flows. By the time
+// this code is called, cloudConfigPath is present on the result object.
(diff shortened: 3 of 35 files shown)


src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
```diff
@@ -22,6 +22,12 @@ const keys: (keyof IaCTestFlags)[] = [
   'quiet',
   'scan',
   'legacy',
+
+  // PolicyOptions
+  'ignore-policy',
+  'trust-policies',
+  'policy-path',
+  'loose',
```
craigfurman, line 30: These flags from PolicyOptions had to be duplicated in a couple of places to enable them for IaC flows.

src/cli/commands/test/iac-local-execution/policy.ts
```diff
@@ -0,0 +1,81 @@
+import { FormattedResult } from './types';
+import { Policy } from '../../../../lib/policy/find-and-load-policy';
+
+export function filterIgnoredIssues(
```
craigfurman, line 4: This is the core of the ignore implementation. I basically map the IaC issues into a data type that the snyk-policy library expects, and then map the result back.

src/lib/policy/find-and-load-policy.ts
```diff
@@ -11,11 +11,11 @@ const debug = debugModule('snyk');
 
 export async function findAndLoadPolicy(
   root: string,
-  scanType: SupportedPackageManagers | 'docker',
+  scanType: SupportedPackageManagers | 'docker' | 'iac',
```
craigfurman, line 14: Add an allowed value, but don't do any special logic with it.

src/lib/policy/find-and-load-policy.ts
```diff
@@ -11,11 +11,11 @@ const debug = debugModule('snyk');
...
   options: PolicyOptions,
   pkg?: PackageExpanded,
   scannedProjectFolder?: string,
-): Promise<string | undefined> {
+): Promise<Policy | undefined> {
```
craigfurman, line 18: This function never returned a string. See new Policy interface below, which takes baby steps towards proper typing. I stopped short of actually fully speccing out a Vulnerability type though, because of IaC's fiddly "adapter" needs. See policy.ts.

src/lib/policy/find-and-load-policy.ts
```diff
@@ -43,9 +43,13 @@ export async function findAndLoadPolicy(
   } catch (err) {
     // note: inline catch, to handle error from .load
     // if the .snyk file wasn't found, it is fine
-    if (err.code !== 'ENOENT') {
+    if (err.code !== 'ENOENT' && err.code !== 'ENOTDIR') {
```
craigfurman, line 46: Unlike snyk open source scans, it's not uncommon in IaC to pass file (non-directory) paths to `snyk iac test`. In these cases there is no root directory to check for a `.snyk`, so we consider ENOTDIR as well as ENOENT as business as usual.

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -149,7 +149,7 @@ Describe "Snyk iac local test command"
     It "finds issues in terraform file"
       When run snyk iac test ../fixtures/iac/terraform/sg_open_ssh.tf
       The status should equal 1 # issues found
-      The output should include "Testing sg_open_ssh.tf..."
+      The output should include "Testing ../fixtures/iac/terraform/sg_open_ssh.tf..."
```
craigfurman, line 152: In cases where we pass a file (as opposed to a directory) to `snyk iac test`, the targetFile is the relative path from where the test is run, not relative to the directory under test. We now print the whole targetFile.

test/smoke/spec/iac/snyk_test_local_exec_spec.sh
```diff
@@ -236,9 +236,9 @@ Describe "Snyk iac local test command"
       When run snyk iac test ../fixtures/iac/depth_detection/  --detection-depth=2
       The status should equal 1 #  issues found
       # Only File
-      The output should include "Testing one.tf..."
+      The output should include "Testing one/one.tf..."
```
craigfurman, line 239: In cases where we pass a directory to `snyk iac test`, the targetFile is the relative path to the directory under test. We now print the whole targetFile.

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -126,6 +127,12 @@ export type IaCTestFlags = Pick<
   | 'severityThreshold'
   | 'json'
   | 'sarif'
+
+  // PolicyOptions
+  | 'ignore-policy'
+  | 'trust-policies'
+  | 'policy-path'
+  | 'loose'
```
craigfurman, line 135: @ipapast  Now that I look at these flags again, only policy-path (override path to policy file) and ignore-policy (ignore any policy files that are found) have semantic meaning in our context. 

`loose` appears to be a snyk-policy library option that ignores ENOENT, but we just do that ourselves in findAndLoadPolicy. There is no user-facing behaviour.

`trust-policies` is an open-source-specific flag that looks for `.snyk` files in dependencies themselves.

Maybe I'll remove these 2 from IaC's allowed flags list?

ipapast:
We can see the new path output in both `introducedBy` and `tested ...` fields  in this screenshot 👌(adding visuals for the future readers)

image

src/cli/commands/test/iac-local-execution/types.ts
ipapast: yes, let's do that if we don't use these flags in IaC

craigfurman: done in another commit that I'll probably squash when all the review is done

src/cli/commands/test/iac-local-execution/policy.ts
```diff
@@ -0,0 +1,92 @@
...
+  policy: Policy | undefined,
+  results: FormattedResult[],
+) {
+  if (!policy) {
+    return { filteredIssues: results, ignoreCount: 0 };
+  }
+  const vulns = results.map((res) => policy.filter(toIaCVulnAdapter(res)));
```
ipapast, line 11: nit: Should we rename all references of "vulns" as issues to keep the language consistent to configuration issues?

e.g. issues, toIaCIssueAdapter, etc.

craigfurman: I'm a bit torn here. I used `vuln` for these objects to minic the naming inside the policy library itself. The IaCVulnAdapter adapts IaC issues to vulns!

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -38,6 +40,8 @@ export async function test(
 
     await initLocalCache({ customRulesPath });
 
+    const policy = await findAndLoadPolicy(pathToScan, 'iac', options);
```
craigfurman, line 43: @ipapast I spotted something a bit weird: https://github.com/snyk/snyk/blob/b7a190b4098d87cd885c73dc76ec8492388c52b4/src/lib/policy/find-and-load-policy.ts#L34

We increment the policies analytics counter even when no policies are actually found. This is the case for open source and container flows too, so maybe we just roll with it to keep everything consistent. I have no idea whether or not we're inferring a lot of meaning from that number - I'd guess no, because we're only just introducing policies to IaC. WDYT?

ipapast: > I have no idea whether or not we're inferring a lot of meaning from that number - I'd guess no, because we're only just introducing policies to IaC. WDYT?

I have no idea either, but as you said, we haven't used it yet, so it can't be harmful.

src/cli/commands/test/iac-local-execution/policy.ts
ipapast: We had a chat with Craig on zoom about this. We got on the same page about it. 👍

test/jest/unit/iac-unit-tests/policy.spec.ts
```diff
@@ -115,6 +115,14 @@ describe('filtering ignored issues', () => {
...
   });
 
+  it('filters no issues when path is file path in the wrong directory', async () => {
+    const { fixture, filtered, ignoreCount } = await filterFixture(
+      'policy-ignore-file-path-wrong-dir.yml',
+    );
+    expect(filtered).toEqual(fixture);
+    expect(ignoreCount).toEqual(0);
+  });
+
```
ipapast, line 125: Yaaaaas! 🙌

craigfurman: don't get too excited - this is a failing test that depends on https://github.com/snyk/policy/pull/58 to pass.

craigfurman: now it should pass, since that PR is merged and I've bumped the dependency here

src/cli/commands/test/iac-local-execution/policy.ts
```diff
@@ -0,0 +1,92 @@
...
+    vulnerabilities: result.result.cloudConfigResults.map(
+      (cloudConfigResult) => {
+        const annotatedResult = cloudConfigResult as AnnotatedResult;
+
+        // Copy the cloudConfigPath array to avoid modifying the original with
+        // splice.
+        // Insert the targetFile into the path so that it is taken into account
+        // when determining whether an ignore rule should be applied.
+        const path = [...annotatedResult.cloudConfigPath];
+        path.splice(0, 0, result.targetFile);
```
ipapast, line 50: ah, we can do without 'GARBAGE'. nice

ipapast (approved):
Nice! 🙌 The tests passes as well so the wrond-dir scenario is working too.

Just a last question: do we need to update the version in the `.snyk` file at the root of this repo too?

craigfurman:
@ipapast good question... I don't think we _need_ to update the version. The reason I updated it in the test fixtures is because the version was being generated freshly, and we were asserting it matched after updating the policy lib.

src/cli/commands/test/iac-local-execution/policy.ts
craigfurman: yes, less garbage is always nice 😅

src/lib/policy/find-and-load-policy.ts
jahed-snyk: If a file is passed to `snyk iac test`, shouldn't we look at the file's directory for a policy file (and pass that in as `scannedProjectFolder` or `root`?

And if that's not possible (I'm not 100% on IAC's flow), shouldn't we avoid passing in invalid values (non-directory paths) to `policyLocations` in the first place?

src/cli/commands/test/index.ts
```diff
@@ -198,12 +198,14 @@ async function test(...args: MethodArgs): Promise<TestCommandResult> {
...
   }
 
+  const noopPathTransformation = (path: string) => path;
   let response = results
     .map((result, i) => {
       return displayResult(
         results[i] as LegacyVulnApiResult,
         resultOptions[i],
         result.foundProjectCount,
+        options.iac ? noopPathTransformation : undefined, // Use default transform unless IaC
```
jahed-snyk, line 208: Is there a point in this workaround (using noop for iac) if in `displayResult`,  `targetFileTransform` is only used in iac contexts anyway (i.e. if (options.iac ...)? We can just remove that logic, right?

craigfurman: I don't think it is only used in an IaC context. I started off doing this unconditionally, then some failing smoke tests pointed showed me that the targetFile is being printed in some non-IaC contexts, so I think we need this logic in order to preserve that behaviour.

re: "If a file is passed to snyk iac test" (for some reason github isn't offering me a reply prompt there), this is a really long story 😅 

1. Non-IaC-specific core/framework parts of the CLI determine what the targetFile is differently for a given file under test depending on whether that file is a member of a directory argument passed to `snyk test` or `snyk iac test`, or whether a single-file arg is passed.
2. When the argument is a directory, `targetFile` is the relative path to that file from the directory
3. When the argument is a file, `targetFile` is the relative path from the current working directory
4. Many IaC ignores in the wild are going to be filepath-scoped. To write a meaningful `.snyk`, these must be paths relative to that policy file.
5. To properly support automatic policy file discovery for file arguments, we could easily do what you're suggesting and search the dirname of the file - but it would not be that useful, because the relative paths in that file would then be broken due to the different treatment of targetFile.
6. I started writing some special-case logic to support automatic .snyk discovery for file arguments, but we decided it wasn't worth it. We can always support that edge-case if it becomes an issue.
7. This doesn't really become a problem in non-IaC contexts because the argument to snyk-test is usually a directory, and file-path-scoped ignores aren't a thing outside of IaC yet as far as I know.

jahed-snyk: RE: reply prompt, it's because it's a comment on the same line as your initial comment so you need to use the first thread (higher up) to reply. It's not great UX... Might be worth editing/moving it there to avoid confusion.

jahed-snyk: I double-checked and it definitely looks like it's only used in `options.iac` contexts.

To clarify, I'm suggesting instead of adding a new `targetFileTransform` parameter to `displayResult`, it's possible to change all occurences of `pathLib.basename(res.targetFile)` to just `res.targetFile`.

I won't block on this (I can try refactoring another time to verify) but it's worth checking again.

ghost (approved):

src/lib/policy/find-and-load-policy.ts
craigfurman: I replied here, because github confused me 😅 : https://github.com/snyk/snyk/pull/2134#discussion_r684240252

src/cli/commands/test/index.ts
craigfurman: Thanks, I'll take another look at that then and might change it before merging. I'm going to assume that because you approved you're happy for me to make a judgement without re-review 🙏. Holler if not!

craigfurman: huh, you're absolutely right and it's really obvious why: https://github.com/snyk/snyk/blob/c9f28ebbc6c263d7f732e68f6da83e0c128c7a97/src/lib/formatters/test/display-result.ts#L41-L43

I don't understand why I wrote that transform code, and have no memory of doing it. Weird. Will simplify this before merge.

PR #2139, opened 2021-08-01
Migrate Tap tests for Analytics to Jest

maxjeffos:

#### What does this PR do?

Migrates the last Tap tests for analytics over to Jest. 

- Most of the tests from the now-deleted `test/analytics.test.ts` are replaced by acceptance tests in `test/jest/acceptance/analytics.spec.ts`
- The rest of the tests from the now-deleted `test/analytics.test.ts` are replaced by `test/jest/unit/lib/analytics/index.spec.ts`
- The now-deleted `test/jest/system/analytics.spec.ts` tests were made redundant by a combination of the acceptance tests in `test/jest/acceptance/analytics.spec.ts` and by extracting `test/jest/unit/lib/analytics/getStandardData.spec.ts` and having some unit tests on that: `test/jest/unit/lib/analytics/getStandardData.spec.ts`


src/lib/analytics/types.ts
```diff
@@ -0,0 +1,21 @@
+export type Environment =
+  | {
+      npmVersion: string | undefined;
+    }
+  | {};
```
jahed-snyk, line 5: `{}` as a type means "any object" since TypeScript uses duck typing. We can change this to be:

```ts
export type Environment = {
  npmVersion?: string | undefined;
}
```

test/jest/acceptance/analytics.spec.ts
```diff
@@ -27,11 +29,13 @@ describe('analytics module', () => {
...
   });
 
-  test('validate actual analytics call', async () => {
-    server.setNextResponse({});
-    const { code } = await runSnykCLI(`version --org=fooOrg --all-projects`, {
-      env,
-    });
+  it('sends correct analytics data for simple command (`snyk version`)', async () => {
+    const { code } = await runSnykCLI(
+      `version --org=fooOrg --all-projects --integrationName=JENKINS --integrationVersion=1.2.3`,
```
jahed-snyk, line 34: The flags should be in kebab case right? e.g. `--integration-name=JENKINS`

The integration flags might be better off as environment variables as that's how they're typically used.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -27,11 +29,13 @@ describe('analytics module', () => {
     });
   });
 
-  test('validate actual analytics call', async () => {
-    server.setNextResponse({});
-    const { code } = await runSnykCLI(`version --org=fooOrg --all-projects`, {
-      env,
-    });
+  it('sends correct analytics data for simple command (`snyk version`)', async () => {
```
jahed-snyk, line 32: We can remove this test as the other ones are more thorough and make the same / similar assertions.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -87,4 +93,293 @@ describe('analytics module', () => {
       },
     });
   });
+
+  // test for `snyk test` with a project that has no vulns
+  // improves upon the `snyk version` test because the `snyk test` path will include hitting `analytics.add`
+  it('sends correct analytics data for `snyk test` command', async () => {
```
jahed-snyk, line 99: Instead of a comment, we can name the test more clearly. `sends analytics when no vulns are found`. Same for the other tests.

test/jest/unit/lib/analytics/getStandardData.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    expect(standardData).toMatchObject({
+      os: expect.any(String),
+      version: '1.0.0-monorepo',
+      id: expect.any(String),
+      ci: expect.any(Boolean),
+      environment: {
+        npmVersion: await getCommandVersion('npm'),
+      },
+      // prettier-ignore
```
jahed-snyk, line 18: What is this ignoring?

test/jest/unit/lib/analytics/index.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as snyk from '../../../../../src/lib';
+import * as request from '../../../../../src/lib/request';
+import { argsFrom } from './utils';
+
+describe('analytics module', () => {
+  let old: string | undefined;
+
+  beforeEach(() => {
+    old = snyk.config.get('disable-analytics');
+    snyk.config.delete('disable-analytics');
```
jahed-snyk, line 11: Instead of modifying global state (snyk config), we can have another acceptance test with analytics disabled using environment variables (`SNYK_CONF_...`, or whatever it was) and a simple `snyk version` so that it's fast.

test/jest/unit/lib/analytics/utils.ts
```diff
@@ -0,0 +1,15 @@
+import { ArgsOptions } from '../../../../../src/cli/args';
+
+/**
+ * Test helper to make a proper ArgsOptions[] out of a simpler input.
+ * @param args basic key value pairs object
+ * @returns a ArgsOptions[] with just the stuff we need for the tests.
+ */
+export function argsFrom(args: { [key: string]: string }): ArgsOptions[] {
```
jahed-snyk, line 8: This function doesn't do much. It wraps a shallow copy of an object in an array. It's only used 3 times. Do we need it?

maxjeffos: Not really, but creating an ArgsOptions[] for test is excessively verbose.

test/jest/acceptance/analytics.spec.ts
maxjeffos: done and switched to using the env var

src/lib/analytics/types.ts
maxjeffos: thanks - done

test/jest/acceptance/analytics.spec.ts
maxjeffos: done

test/jest/acceptance/analytics.spec.ts
maxjeffos: done

test/jest/unit/lib/analytics/getStandardData.spec.ts
maxjeffos: `'network_time'` and `'cpu_time'`
If I don't ignore it, prettier will drop the single quotes but then eslint will complain that it isn't camel case.

test/jest/unit/lib/analytics/index.spec.ts
maxjeffos: done - made an acceptance test to cover these.

ghost (approved):

PR #2146, opened 2021-08-09
chore: add prebuild step

teodora-sandu:
The `./dist` folder doesn't seem to get deleted when we run `npm run build`, so if we have run it before and have any old files left behind, they will stay in the `./dist` folder.

We noticed this by running `snyk-dev iac test <IaC file> --debug` and getting an error from a file that doesn't exist in our latest pull from `master`:
Screenshot 2021-08-09 at 17 37 30

If we add the `prebuild` step, the `./dist` folder gets re-created from scratch, without leftover files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   ],
   "scripts": {
     "clean": "rimraf dist tsconfig.tsbuildinfo",
+    "prebuild": "npm run clean",
     "build": "tsc --build",
     "build-watch": "tsc -w",
     "find-circular": "npm run build && madge --circular ./dist",


ghost (changes requested):
This will make every re-build and install time really slow (since `build` is run in the `prepare` hook which is executed after `install`). `tsbuildinfo` holds a cache of `dist`. Worse once we enable `incremental` builds and Webpack (#2105).

This may be best done on a need-to basis rather than on every `npm run build`. Which is why we added the `clean` script. Kind of like `rm -rf node_modules`.

To avoid this, maybe we can run `npm run build` and on failure, show a message saying "try running `npm run clean` first"?

For reference, there's no official solution to this problem yet: https://github.com/Microsoft/TypeScript/issues/16057

Edit: thinking about it, once we have #2105 merged, this will solve itself as there won't be leftover files, just the bundle file.

teodora-sandu:
This was a runtime failure so I don't know if we can add an error, but I agree that we did notice a longer build time than usual. I'll close this, thanks for the review! 🙂

PR #2147, opened 2021-08-09
Update tsconfig settings

maxjeffos:

#### What does this PR do?

Update the tsconfig.settings.json

This is a pre-PR for https://github.com/snyk/snyk/pull/2105

### DIFF
--- a/tsconfig.settings.json
+++ b/tsconfig.settings.json
@@ -3,11 +3,13 @@
     "outDir": "./dist",
     "rootDir": ".",
     "pretty": true,
-    "target": "es2017",
-    "lib": ["es2017", "dom"],
+    "moduleResolution": "node",
+    "target": "es2018",
+    "lib": ["ES2018", "dom"],
     "module": "commonjs",
     "allowJs": true,
     "sourceMap": true,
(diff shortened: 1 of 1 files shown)


tsconfig.settings.json
```diff
@@ -3,11 +3,13 @@
     "outDir": "./dist",
     "rootDir": ".",
     "pretty": true,
-    "target": "es2017",
-    "lib": ["es2017", "dom"],
+    "moduleResolution": "node",
```
jahed-snyk, line 6: This is default right? Since we use "module: commonjs" which defaults this to node. So do we need this?

https://www.typescriptlang.org/tsconfig#moduleResolution

Since we're using an old TS, maybe the docs don't line up with our usage?

ghost (approved):

tsconfig.settings.json
maxjeffos: It does appear to be the default, so I'll take it out.

jahed-snyk: I checked a few OSS repos can seems common to make it explicit. So let's keep it in.

PR #2148, opened 2021-08-09
Chore/ci optimization

maxjeffos:

#### What does this PR do?

A few optimizations to the CI pipeline ahead of https://github.com/snyk/snyk/pull/2105.


ghost (approved):
Looks good. What kind of performance improvements do we see with using Large in these cases? Just wondering in terms of cost/benefit. (We had to do it for the other jobs because of memory issues, whereas here we don't have that problem AFAIK)

maxjeffos:
> Looks good. What kind of performance improvements do we see with using Large in these cases? Just wondering in terms of cost/benefit. (We had to do it for the other jobs because of memory issues, whereas here we don't have that problem AFAIK)

I don't think this is just for performance. @JackuB knows as this is basically from his big webpack PR, but I think he needed to bump these up to get it to finish, not just for performance.
@JackuB can you confirm?

PR #2152, opened 2021-08-10
chore: remove update notifier

maxjeffos:

#### What does this PR do?

Removes the update notification mechanism along with the `update-notifier` dependency.

This is a pre-PR to #2105


tsconfig.settings.json
```diff
@@ -5,7 +5,7 @@
     "pretty": true,
     "moduleResolution": "node",
     "target": "es2018",
-    "lib": ["ES2018", "dom"],
+    "lib": ["ES2018", "ES2019", "dom"],
```
maxjeffos, line 8: @Avishagp @JackuB this seems to be the trick to getting that `Object.fromEntries` is not a function error to go away. Apparently, that came about in the ES2019.

Avishagp: @maxjeffos Yeah, seen that trick. But it still doesn't make sense that the test is broken only for this PR right?

maxjeffos: @Avishagp, I have theory... I was thinking that since `update-notifier` has many deps/transitive deps, along with `@types/update-notifier`, probably something in those deps was defining `Object.fromEntries`. I think it's quite likely the case.

Avishagp (approved):

PR #2156, opened 2021-08-11
test: use async/await instead of callbacks

ghost:
As promised in #1968. I've migrated this test to use async/await for better readability.

### DIFF
--- a/test/jest/acceptance/cli-json-file-output.spec.ts
+++ b/test/jest/acceptance/cli-json-file-output.spec.ts
@@ -1,120 +1,62 @@
-import { exec } from 'child_process';
-import { sep, join } from 'path';
-import { readFileSync, unlinkSync } from 'fs';
-import { v4 as uuidv4 } from 'uuid';
 import { fakeServer } from '../../acceptance/fake-server';
-import cli = require('../../../src/cli/commands');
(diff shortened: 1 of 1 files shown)


test/jest/acceptance/cli-json-file-output.spec.ts
```diff
@@ -1,120 +1,62 @@
...
 
-  const BASE_API = '/api/v1';
-  const SNYK_API = 'http://localhost:' + port + BASE_API;
-  const SNYK_HOST = 'http://localhost:' + port;
+  afterAll((done) => {
+    server.close(() => done());
+  });
 
-  const server = fakeServer(BASE_API, apiKey);
+  it('`can save JSON output to file while sending human readable output to stdout`', async () => {
```
maxjeffos, line 28: Any idea why some tests seem to backtick the entire test description?

maxjeffos (approved):
Great @jahed-snyk

test/jest/acceptance/cli-json-file-output.spec.ts
jahed-snyk: Must be that some tests said things like `"runs `snyk test` blah blah"` and over time after many copy-and-pastes it evolved into what we're seeing... Life finds a way.

PR #2157, opened 2021-08-11
test: increase analytics test timeout

ghost:
Launching process alone takes 5 seconds which is the default timeout. Causes unreliable tests.

https://app.circleci.com/pipelines/github/snyk/snyk/6706/workflows/1ae30a6e-a106-4d71-a337-dffd58512a09/jobs/44733

Also made some small fixes (see commits).

### DIFF
--- a/test/jest/acceptance/analytics.spec.ts
+++ b/test/jest/acceptance/analytics.spec.ts
@@ -3,6 +3,8 @@ import { createProject } from '../util/createProject';
 import { runSnykCLI } from '../util/runSnykCLI';
 import * as fs from 'fs';
 
+jest.setTimeout(1000 * 30);
+
 describe('analytics module', () => {
   let server;
   let env: Record<string, string>;
--- a/test/jest/acceptance/analytics.spec.ts
+++ b/test/jest/acceptance/analytics.spec.ts
@@ -1,7 +1,6 @@
 import { fakeServer } from '../../acceptance/fake-server';
 import { createProject } from '../util/createProject';
 import { runSnykCLI } from '../util/runSnykCLI';
-import * as request from '../../../src/lib/request';
 import * as fs from 'fs';
 
 describe('analytics module', () => {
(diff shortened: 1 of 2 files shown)


test/jest/acceptance/analytics.spec.ts
```diff
@@ -27,7 +30,7 @@ describe('analytics module', () => {
   });
 
   afterEach(() => {
-    jest.restoreAllMocks();
+    server.clearRequests();
```
jahed-snyk, line 33: Added this as we should typically clear state between tests so we aren't asserting on invalid state.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -1,8 +1,11 @@
...
-import { createProject } from '../util/createProject';
+import {
+  createProjectFromFixture,
+  createProjectFromWorkspace,
+} from '../util/createProject';
 import { runSnykCLI } from '../util/runSnykCLI';
-import * as request from '../../../src/lib/request';
-import * as fs from 'fs';
+
+jest.setTimeout(1000 * 30);
```
jahed-snyk, line 8: This is the main thing to avoid test timeouts.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -301,14 +301,15 @@ describe('analytics module', () => {
...
       env: {
         ...env,
         SNYK_DISABLE_ANALYTICS: '1',
       },
     });
     expect(code).toBe(0);
-    expect(requestSpy).not.toBeCalled();
+
+    const lastRequest = server.popRequest();
+    expect(lastRequest).toBeUndefined();
```
jahed-snyk, line 313: Spying won't work as spies are bound to the current Node context; which is Jest. We are launching Snyk CLI in its own process where the spy does not exist.

To test this correctly, we can ensure fakeServer didn't receive a request.

test/jest/acceptance/analytics.spec.ts
maxjeffos: 👍

maxjeffos (approved):
Nice

PR #2159, opened 2021-08-11
Use OAUTH token if set for analytics

maxjeffos:

#### What does this PR do?

Update Analytics to use the OAUTH token if it is set.

### DIFF
--- a/src/lib/api-token.ts
+++ b/src/lib/api-token.ts
@@ -44,3 +44,7 @@ export function getAuthHeader(): string {
   }
   return `token ${api()}`;
 }
+
+export function someTokenExists(): boolean {
+  return Boolean(getOAuthToken() || getDockerToken() || api());
+}
--- a/test/jest/acceptance/analytics.spec.ts
+++ b/test/jest/acceptance/analytics.spec.ts
@@ -22,7 +22,6 @@ describe('analytics module', () => {
       SNYK_INTEGRATION_NAME: 'JENKINS',
       SNYK_INTEGRATION_VERSION: '1.2.3',
     };
-
     server = fakeServer(baseApi, env.SNYK_TOKEN);
     server.listen(port, () => {
       done();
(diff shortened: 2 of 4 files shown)


test/jest/acceptance/analytics.spec.ts
```diff
@@ -300,6 +300,120 @@ describe('analytics module', () => {
...
+    });
+
+    expect(code).toBe(0);
+
+    const lastRequest = server.popRequest();
+    expect(lastRequest).toMatchObject({
+      headers: {
+        host: 'localhost:12345',
+        accept: 'application/json',
+        authorization: 'Bearer oauth-jwt-token',
```
jahed-snyk, line 376: We can be less verbose in this test case and just match on the authorization header since that's all we care about.

test/jest/acceptance/analytics.spec.ts
```diff
@@ -300,6 +300,120 @@ describe('analytics module', () => {
...
+    delete envWithoutToken.SNYK_TOKEN;
+    const { code } = await runSnykCLI('version', {
+      cwd: project.path(),
+      env: envWithoutToken,
+    });
+
+    expect(code).toBe(0);
+
+    const lastRequest = server.popRequest();
+    expect(lastRequest).toMatchObject({
```
jahed-snyk, line 315: Could do something like `not.toHaveProperty('headers.authorization')` to be less verbose.

src/lib/api-token.ts
```diff
@@ -44,3 +44,10 @@ export function getAuthHeader(): string {
   }
   return `token ${api()}`;
 }
+
+export function someTokenExists(): boolean {
+  const oauthToken: string | undefined = getOAuthToken();
+  const dockerToken: string | undefined = getDockerToken();
+  const apiToken: string | undefined = api();
+  return Boolean(oauthToken || dockerToken || apiToken);
```
jahed-snyk, line 52: We can inline these variables. I don't think naming them adds much, makes it more noisy (esp. with type annotations). And maybe check for `api()` first as it's the most common.

ghost (approved):

test/jest/acceptance/analytics.spec.ts
maxjeffos: 👍

src/lib/api-token.ts
maxjeffos: 👍

PR #2165, opened 2021-08-13
Unblock pipeline with additional parallelism

maxjeffos:

#### What does this PR do?

- do not wait for regression tests to complete before launching all the other test suites
- break the Tap acceptance tests out into their own job to get more parallelism

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -505,8 +505,6 @@ workflows:
       - dev-release:
           name: Dev Release
           node_version: '14'
-          requires:
-            - Regression Test
       - test-windows:
           name: Windows, Node 14 - Packages, Jest, Acceptance, System tests
           context: nodejs-install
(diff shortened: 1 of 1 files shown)


ghost (approved):

maxjeffos:
> @maxjeffos Looks like this is causing all tests to run on merge commits.
> 
> https://app.circleci.com/pipelines/github/snyk/snyk/6780/workflows/bfde2afd-e0cd-4b44-824c-5154322908bb
> 
> We'll need to add this to every test job:
> 
> https://github.com/snyk/snyk/blob/875b9fab73733c7586d6c7cca3e406e43a7fcc0c/.circleci/config.yml#L501-L504
> 
> Previously since all test jobs needed Regression Tests (which included that filter), they didn't need it. Now they do.
> 
> Though, could argue that running tests on merges is a good idea. If so, we need to restructure it so the "Release" job runs after tests pass.

@jahed-snyk https://github.com/snyk/snyk/pull/2172

PR #2167, opened 2021-08-16
refactor: replace cc parser with split functions [CC-1021]

almog27:

#### What does this PR do?
This PR updates the version of the `cloud-config-parser` dependency to use a newer version that takes care of a performance issue.
Instead of calculating the treeMap for each line in the file, we are calculating the treeMap once for each file, and sending it to the parser to get the correct line number.
Therefore, for a file that has multiple issues, we only calculate the treeMap once.

1. Download the new branch and install dependencies
2. Run `snyk iac test` with `--json` or `--sarif` flags.
3. See the line number returns in the response

Jira Ticket [CC-1021](https://snyksec.atlassian.net/browse/CC-1021)

#### Snapshots
Screen Shot 2021-08-16 at 11 37 49

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
   "dependencies": {
     "@open-policy-agent/opa-wasm": "^1.2.0",
     "@snyk/cli-interface": "2.11.0",
-    "@snyk/cloud-config-parser": "^1.9.2",
+    "@snyk/cloud-config-parser": "^1.10.2",
     "@snyk/code-client": "4.0.0",
     "@snyk/dep-graph": "^1.27.1",
     "@snyk/fix": "1.650.0",
@@ -147,7 +147,7 @@
   "devDependencies": {
     "@types/cross-spawn": "^6.0.2",
     "@types/fs-extra": "^9.0.11",
-    "@types/jest": "^26.0.20",
+    "@types/jest": "^27.0.1",
     "@types/lodash": "^4.14.161",
     "@types/needle": "^2.0.4",
     "@types/node": "^14.14.31",
--- a/src/cli/commands/test/iac-local-execution/extract-line-number.ts
+++ b/src/cli/commands/test/iac-local-execution/extract-line-number.ts
@@ -2,7 +2,8 @@ import { IaCErrorCodes } from './types';
 import { CustomError } from '../../../../lib/errors';
 import {
   CloudConfigFileTypes,
-  issuesToLineNumbers,
+  MapsDocIdToTree,
+  getLineNumber,
 } from '@snyk/cloud-config-parser';
 import { UnsupportedFileTypeError } from './file-parser';
 import * as analytics from '../../../../lib/analytics';
@@ -10,7 +11,7 @@ import * as Debug from 'debug';
 import { getErrorStringCode } from './error-utils';
 const debug = Debug('iac-extract-line-number');
 
-function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
+export function getFileTypeForParser(fileType: string): CloudConfigFileTypes {
   switch (fileType) {
     case 'yaml':
     case 'yml':
@@ -25,16 +26,12 @@ function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
 }
 
 export function extractLineNumber(
-  fileContent: string,
-  fileType: string,
   cloudConfigPath: string[],
+  fileType: CloudConfigFileTypes,
+  treeByDocId: MapsDocIdToTree,
 ): number {
   try {
-    return issuesToLineNumbers(
-      fileContent,
-      getFileTypeForLineNumber(fileType),
-      cloudConfigPath,
-    );
+    return getLineNumber(cloudConfigPath, fileType, treeByDocId);
   } catch {
     const err = new FailedToExtractLineNumberError();
     analytics.add('error-code', err.code);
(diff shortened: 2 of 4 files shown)


src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -53,28 +65,32 @@ function formatScanResult(
   meta: TestMeta,
   options: IaCTestFlags,
 ): FormattedResult {
+  const shouldExtractLineNumber = flagsRequiringLineNumber.some(
+    (flag) => options[flag],
+  );
```
aron, line 70: Apologies for the drive by comment, unless performance is significantly affected, we should aim to remove this entirely now. This will avoid confusion should we want to add the line numbers to the other output types in future.

ipapast (commented):
Looks good. Tests are breaking though, you will need to update the mocking

src/cli/commands/test/iac-local-execution/extract-line-number.ts
```diff
@@ -1,16 +1,16 @@
...
+import { CloudConfigFileTypes, getLineNumber } from '@snyk/cloud-config-parser';
 import { UnsupportedFileTypeError } from './file-parser';
 import * as analytics from '../../../../lib/analytics';
 import * as Debug from 'debug';
 import { getErrorStringCode } from './error-utils';
+import { MapsDocIdToTree } from '@snyk/cloud-config-parser/dist/types';
 const debug = Debug('iac-extract-line-number');
 
-function getFileTypeForLineNumber(fileType: string): CloudConfigFileTypes {
+export function getFileTypeForLineNumber(
```
teodora-sandu, line 11: nit: This is the first time I've looked at this code and this function name confused me a bit. if you're up for it, could you please rename it to something like `getFileTypeForParser`?

src/cli/commands/test/iac-local-execution/results-formatter.ts
```diff
@@ -53,27 +58,24 @@ function formatScanResult(
...
+    // we still pass an undefined tree and not calculated line number for those
+  }
+
   const formattedIssues = scanResult.violatedPolicies.map((policy) => {
     const cloudConfigPath =
       scanResult.docId !== undefined
         ? [`[DocId: ${scanResult.docId}]`].concat(parsePath(policy.msg))
         : policy.msg.split('.');
 
-    const flagsRequiringLineNumber = [
```
teodora-sandu, line 62: Why did we remove these checks? I think we added these checks here so we don't unnecessarily make line number calls and increase the running time

test/jest/unit/iac-unit-tests/results-formatter.spec.ts
```diff
@@ -51,6 +52,39 @@ describe('formatScanResults', () => {
...
+    );
+
+    expect(formattedResults.length).toEqual(1);
+    expect(formattedResults[0]).toEqual(
+      expectedFormattedResultsWithoutLineNumber,
+    );
+  });
+});
+
+describe('tests', () => {
```
teodora-sandu, line 72: This test name is not very helpful - can we move the inner `it` to the `formatScanResults` `describe` and also move the `jest.clearAllMocks();` line from all tests to an `afterEach`?

src/cli/commands/test/iac-local-execution/extract-line-number.ts
```diff
@@ -1,16 +1,16 @@
...
-import {
-  CloudConfigFileTypes,
-  issuesToLineNumbers,
-} from '@snyk/cloud-config-parser';
+import { CloudConfigFileTypes, getLineNumber } from '@snyk/cloud-config-parser';
 import { UnsupportedFileTypeError } from './file-parser';
 import * as analytics from '../../../../lib/analytics';
 import * as Debug from 'debug';
 import { getErrorStringCode } from './error-utils';
+import { MapsDocIdToTree } from '@snyk/cloud-config-parser/dist/types';
```
teodora-sandu, line 8: I see this import is used in `registry` too, but can we export `MapsDocIdToTree` from https://github.com/snyk/cloud-config-parser/blob/master/lib/index.ts and change these imports in a future PR?

teodora-sandu (changes requested):
Looks good, I asked for a few small changes

test/jest/unit/iac-unit-tests/results-formatter.spec.ts
ipapast: hey @teodora-sandu, we paired a bit with Almog to check the tests and this was a commit for me to see when Almog had to disconnect for a bit. Apologies :D The state of the tests now should be good.

src/cli/commands/test/iac-local-execution/results-formatter.ts
almog27: @teodora-sandu please see @aron's comment - https://github.com/snyk/snyk/pull/2167#discussion_r689351195

almog27: And there was some performance issue due to the fact that we were doing the whole calculation every time - this PR change this behaviour - so the big calculation happens only once, and the small calculations happen for every issue detected.
And that way, we can also support line number for the remain cases if wanted in the future to present it, and not only for JSON/SARIF

teodora-sandu (approved):

src/cli/commands/test/iac-local-execution/extract-line-number.ts
almog27: Fixed - will later create fix for this at Registry too

package.json
```diff
@@ -147,7 +147,7 @@
   "devDependencies": {
     "@types/cross-spawn": "^6.0.2",
     "@types/fs-extra": "^9.0.11",
-    "@types/jest": "^26.0.20",
+    "@types/jest": "^27.0.1",
```
almog27, line 150: This dependency was updated due to failure in regression tests due to the fact that the `jest` dependency is a newer version than the types

JackuB (approved):

PR #2175, opened 2021-08-19
fix(protect): skip previously patched files

jahed-snyk:
In the old `snyk protect` we used to add placeholder files to patched modules so indicate we've already patched them. The new snyk/protect doesn't do this so it attempts to patch modules that have already been patched, causing a failure.

This PR fixes that by creating the same placeholder file. The difference is that it's simpler and should work with any existing ignore rules a project might use to exclude them in their builds.

`snyk protect`:

`[filename].snyk-[vuln-id].flag` containing a timestamp.

Now:

`[filename].snyk-protect.flag` containing nothing.

So a project using `*.snyk-*.flag` will still match the same files. Like:

https://github.com/signalapp/Signal-Desktop/blob/f96246fecf4855db88d571e5825132ee4d5605c8/package.json#L435

<details>
<summary>See error logs</summary>
<pre>
node ➜ .../snyk-protect/test/fixtures/fix-pr (master ✗) $ npm install
> fix-pr@1.0.0 prepare
> npm run snyk-protect
> fix-pr@1.0.0 snyk-protect
> snyk-protect
patched /workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/lodash/lodash.js
Successfully applied Snyk patches
added 2 packages, and audited 3 packages in 9s
1 high severity vulnerability
To address all issues, run:
  npm audit fix --force
Run `npm audit` for details.
node ➜ .../snyk-protect/test/fixtures/fix-pr (master ✗) $ npm install
> fix-pr@1.0.0 prepare
> npm run snyk-protect
> fix-pr@1.0.0 snyk-protect
> snyk-protect
Expected
  line from local file
 "        if ((key === '__proto__' || key === 'constructor' || key === 'prototype')) {" 
  to match patch line
 "        if (index != lastIndex) {" 
(node:7862) UnhandledPromiseRejectionWarning: Error: File to be patched does not match, not patching
    at patchString (/workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/@snyk/protect/dist/lib/patch.js:59:27)
    at Object.applyPatchToFile (/workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/@snyk/protect/dist/lib/patch.js:9:29)
    at /workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/@snyk/protect/dist/lib/index.js:68:29
    at Array.forEach (<anonymous>)
    at /workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/@snyk/protect/dist/lib/index.js:67:39
    at Array.forEach (<anonymous>)
    at /workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/@snyk/protect/dist/lib/index.js:66:24
    at Array.forEach (<anonymous>)
    at /workspaces/snyk/packages/snyk-protect/test/fixtures/fix-pr/node_modules/@snyk/protect/dist/lib/index.js:65:94
    at Array.forEach (<anonymous>)
(node:7862) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:7862) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
up to date, audited 3 packages in 2s
1 high severity vulnerability
To address all issues, run:
  npm audit fix --force
Run `npm audit` for details.
</pre>
<details>

### DIFF
--- a/packages/snyk-protect/bin/snyk-protect
+++ b/packages/snyk-protect/bin/snyk-protect
@@ -1,2 +1,7 @@
 #!/usr/bin/env node
-require('../dist/index.js').protect();
+require('../dist/index.js')
+  .protect()
+  .catch((error) => {
+    // don't block pipelines on unexpected errors
+    console.error(error);
+  });
--- a/packages/snyk-protect/src/lib/index.ts
+++ b/packages/snyk-protect/src/lib/index.ts
@@ -26,7 +26,7 @@ async function protect(projectFolderPath: string) {
   const snykFilePath = path.resolve(projectFolderPath, '.snyk');
 
   if (!fs.existsSync(snykFilePath)) {
-    console.log('No .snyk file found');
+    console.log('No .snyk file found.');
     sendAnalytics({
       type: ProtectResultType.NO_SNYK_FILE,
     });
@@ -67,7 +67,7 @@ async function protect(projectFolderPath: string) {
   > = await getAllPatches(vulnIdAndPackageNames, packageNameToVersionsMap);
 
   if (packageAtVersionsToPatches.size === 0) {
-    console.log('Nothing to patch');
+    console.log('Nothing to patch.');
     sendAnalytics({
       type: ProtectResultType.NOTHING_TO_PATCH,
     });
@@ -83,7 +83,8 @@ async function protect(projectFolderPath: string) {
     vuldIdAndPatches?.forEach((vp) => {
       vp.patches.forEach((patchDiffs) => {
         patchDiffs.patchDiffs.forEach((diff) => {
-          applyPatchToFile(diff, fpp.path);
+          const patchedPath = applyPatchToFile(diff, fpp.path);
+          console.log(`Patched: ${patchedPath}`);
         });
       });
       patchedModules.push({
@@ -94,7 +95,7 @@ async function protect(projectFolderPath: string) {
     });
   });
 
-  console.log('Successfully applied Snyk patches');
+  console.log('Applied Snyk patches.');
 
   sendAnalytics({
     type: ProtectResultType.APPLIED_PATCHES,
--- a/packages/snyk-protect/src/lib/patch.ts
+++ b/packages/snyk-protect/src/lib/patch.ts
@@ -9,7 +9,7 @@ export function applyPatchToFile(patchContents: string, baseFolder: string) {
   const contentsToPatch = fs.readFileSync(targetFilePath, 'utf-8');
   const patchedContents = patchString(patchContents, contentsToPatch);
   fs.writeFileSync(targetFilePath, patchedContents);
-  console.log(`patched ${targetFilePath}`);
+  return targetFilePath;
 }
 
 export function extractTargetFilePathFromPatch(patchContents: string): string {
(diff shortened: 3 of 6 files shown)


packages/snyk-protect/bin/snyk-protect
```diff
@@ -1,2 +1,7 @@
 #!/usr/bin/env node
-require('../dist/index.js').protect();
+require('../dist/index.js')
+  .protect()
+  .catch((error) => {
+    // don't block pipelines on unexpected errors
+    console.error(error);
+  });
```
jahed-snyk, line 7: This is existing behaviour. Previously we let NodeJS handle "unhandled rejections" which it does by printing a warning and exiting with zero. However, the warnings state they'll change it to a non-zero in the future so it's best that we handle it ourselves. We still exit with zero as errors on our end should not block pipelines.

packages/snyk-protect/src/lib/index.ts
```diff
@@ -83,7 +83,8 @@ async function protect(projectFolderPath: string) {
     vuldIdAndPatches?.forEach((vp) => {
       vp.patches.forEach((patchDiffs) => {
         patchDiffs.patchDiffs.forEach((diff) => {
-          applyPatchToFile(diff, fpp.path);
+          const patchedPath = applyPatchToFile(diff, fpp.path);
+          console.log(`Patched: ${patchedPath}`);
```
jahed-snyk, line 87: I moved this log line here so it's clear exactly what we're printing to stdout for the user to see in a single top-level file.

packages/snyk-protect/src/lib/patch.ts
```diff
@@ -66,16 +71,13 @@ export function patchString(
       }
       case ' ': {
         if (currentLine !== nextLine) {
-          console.log(
-            'Expected\n  line from local file\n',
-            JSON.stringify(currentLine),
-            '\n  to match patch line\n',
-            JSON.stringify(nextLine),
-            '\n',
-          );
```
jahed-snyk, line 75: I moved this into the error message, otherwise it prints out back-to-front which is confusing.

packages/snyk-protect/src/lib/index.ts
```diff
@@ -94,7 +95,7 @@ async function protect(projectFolderPath: string) {
     });
   });
 
-  console.log('Successfully applied Snyk patches');
+  console.log('Applied Snyk patches.');
```
jahed-snyk, line 98: "Applied" already implies success. Less words, better.

packages/snyk-protect/test/acceptance/fix-pr.smoke.spec.ts
```diff
@@ -1,13 +1,34 @@
...
 describe('Fix PR', () => {
+  let project: TestProject;
+  let patchedLodash: string;
+
+  beforeAll(async () => {
+    project = await createProject('fix-pr');
+    patchedLodash = await getPatchedLodash();
+  });
+
   test('patches vulnerable dependencies on install', async () => {
```
jahed-snyk, line 16: I tried spliting this test up so that one runs `npm install` then another runs it again so it's clear what the second run is looking for. However Jest doesn't fail fast, so if the first test fails, it will still execute the second test which wouldn't make sense.

JackuB (approved):

PR #2180, opened 2021-08-22
test: migrate is-docker to jest

Avishagp:
Migrate is-docker to jest, including sinon usage


test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as path from 'path';
+
+import { isDocker } from '../../../src/lib/is-docker';
+
+describe('isDocker', () => {
+  it('inside a Docker container (.dockerenv test)', async () => {
+    delete require.cache[path.join(__dirname, 'index.js')];
+    const statSyncStub = sinon.stub(fs, 'statSync').returns({} as any);
+    expect(isDocker()).toBeTruthy();
+    statSyncStub.restore();
```
anthogez, line 12: Hey nice work here @Avishagp! I would suggest to also replace sinon with jest as it already provide built-in capabilities e.g. jest.spyOn

 ```
  it('inside a Docker container (.dockerenv test)', async () => {
    delete require.cache[path.join(__dirname, 'index.js')];
    const statSyncSpy = jest.spyOn(fs, 'statSync').mockReturnValue({} as never);
    expect(isDocker()).toBeTruthy();
    expect(statSyncSpy).toHaveBeenCalledTimes(1);
    expect(statSyncSpy).toHaveBeenLastCalledWith('/.dockerenv');
  });
  ```
  
  What do you think?

Avishagp: Sounds good, thanks for your input :)
Could you please explain why you used `never` in `mockReturnValue({} as never);` instead of `any`?

anthogez: never as the statSync Stats class do not contains value, but `any` type can work as well as it's extended by a class containing a number

test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    delete require.cache[path.join(__dirname, 'index.js')];
+
+    const statSyncStub = sinon.stub(fs, 'statSync');
+    statSyncStub
+      .withArgs('/.dockerenv')
+      .throws("ENOENT, no such file or directory '/.dockerinit'");
+    const readFileSyncStub = sinon.stub(fs, 'readFileSync');
+    readFileSyncStub
+      .withArgs('/proc/self/cgroup', 'utf8')
+      .returns('xxx docker yyyy');
```
anthogez, line 25: for here I would suggest something like:

```

    const statSyncSpy = jest.spyOn(fs, 'statSync');
    const readFileSyncSpy = jest.spyOn(fs, 'readFileSync');

    statSyncSpy.mockImplementationOnce((path, options): any => {
      if (path === '/.dockerenv') {
        throw new Error("ENOENT, no such file or directory '/.dockerinit'");
      }
    });

    readFileSyncSpy.mockImplementationOnce((path, options): any => {
      if (path === '/proc/self/cgroup' && options === 'utf8') {
        return 'xxx docker yyyy';
      }
    });
```

test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,54 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { isDocker } from '../../../src/lib/is-docker';
+
+describe('isDocker', () => {
+  it('inside a Docker container (.dockerenv test)', async () => {
```
anthogez, line 7: I would suggest also an `afterEach(() => jest.restoreAllMocks())`

test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    const readFileSyncStub = jest.spyOn(fs, 'readFileSync');
+
+    statSyncSpy.mockImplementationOnce((path): any => {
+      if (path === '/.dockerenv') {
+        throw new Error("ENOENT, no such file or directory '/.dockerinit'");
+      }
+    });
+
+    readFileSyncStub.mockImplementationOnce((): any => {
+      throw new Error("ENOENT, no such file or directory '/.dockerinit'");
```
anthogez, line 53: readFileSync was not supposed to contain

```
 if (path === '/proc/self/cgroup' && options === 'utf8') {
        return 'xxx docker yyyy';
 }
```

?

Avishagp: No, that was the `inside a Docker container (cgroup test)` test.
Renaming that now, oops 👍

anthogez: @Avishagp 🤦 I thought you updated https://github.com/snyk/snyk/pull/2180/commits/e4ee8eabc74be8b32256cef790c77ea504e2af5a#diff-df937d847f2ff47cfd06de3ad2ae63a7a769976252c63dc8a482b829dec17610R31

Apologies for the first part of the comment 😄

test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { isDocker } from '../../../src/lib/is-docker';
+
+describe('isDocker', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('inside a Docker container (.dockerenv test)', async () => {
+    delete require.cache[path.join(__dirname, 'index.js')];
```
jahed-snyk, line 12: AFAIK this isn't doing anything 🤔 Can we remove it (same with the other test)? Looks dodgy.

test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+    });
+
+    readFileSyncSpy.mockImplementationOnce((path, options): any => {
+      if (path === '/proc/self/cgroup' && options === 'utf8') {
+        return 'xxx docker yyyy';
+      }
+    });
+
+    expect(isDocker()).toBeTruthy();
```
jahed-snyk, line 37: Should use `toEqual(true)` instead to be more exact.

test/jest/unit/is-docker.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+      if (path === '/.dockerenv') {
+        throw new Error("ENOENT, no such file or directory '/.dockerinit'");
+      }
+    });
+
+    readFileSync.mockImplementationOnce((): any => {
+      throw new Error("ENOENT, no such file or directory '/.dockerinit'");
+    });
+
+    expect(isDocker()).toBeFalsy();
```
jahed-snyk, line 56: Should use `toEqual(false)` instead to be more exact.

ghost (approved):

test/jest/unit/is-docker.spec.ts
anthogez: isDocker  is a function that returns only a boolean value

```
export function isDocker(): boolean {
  return hasDockerEnv() || hasDockerCGroup();
}
```

isEqual is used to validate properties of an object are the same,

`.toBeFalsy()` IMO is more adequate than `isEqual`

jahed-snyk: Yes, it's minor improvement but IMO worth pointing out for readability (now and the future). If we know it's a boolean, so it's either false or true, it's more clear to keep within that framework and say it's false rather than false-like (which introduces other potential values into the reader's mind for no benefit).

anthogez: Thanks for sharing your point about readability here, if that's the case then `toBe` would be enough to both validate the primitive value either to false/true and maintain the readability

jahed-snyk: Yeah, in the context of primitives it doesn't matter much.

I use `toEqual` by default because it's a safer default that covers more use cases (like deep equality). In general, there's not much need for `toBe` unless if there's a need to care about mutations or something related to references.

I do wish `toEqual` was more like `toStrictEqual` though. One of Jest's many problems...

PR #2181, opened 2021-08-22
test: migrate old-snyk-format to jest

Avishagp:


test/jest/unit/old-snyk-format.spec.ts
```diff
@@ -0,0 +1,18 @@
+import * as policy from 'snyk-policy';
+import { loadJson } from '../../utils';
+import path = require('path');
+
+it('test sensibly bails if gets an old .snyk format', async () => {
+  try {
```
jahed-snyk, line 6: The try/catch should be only around the main function under test (`config.filter`). Otherwise it will catch and incorrectly assert on setup errors.

test/jest/unit/old-snyk-format.spec.ts
```diff
@@ -0,0 +1,18 @@
+import * as policy from 'snyk-policy';
+import { loadJson } from '../../utils';
+import path = require('path');
```
jahed-snyk, line 3: We should use `import * as path from 'path'` (progress towards standard imports)

ghost (approved):

test/jest/unit/old-snyk-format.spec.ts
Avishagp: The original test fails at `policy.load`, will change try/catch accordingly

PR #2182, opened 2021-08-23
fix: missing no supported sast files strategy

ArturSnyk:

#### What does this PR do?

This Pr adds an appropriate message when we do not support files for `snyk code` scanning

#### Screenshots
image

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -30,6 +30,7 @@ import {
   UnsupportedOptionCombinationError,
   ExcludeFlagBadInputError,
   CustomError,
+  NoSupportedSastFiles,
 } from '../lib/errors';
 import stripAnsi from 'strip-ansi';
 import { ExcludeFlagInvalidInputError } from '../lib/errors/exclude-flag-invalid-input';
@@ -50,7 +51,7 @@ const debug = Debug('snyk');
 const EXIT_CODES = {
   VULNS_FOUND: 1,
   ERROR: 2,
-  NO_SUPPORTED_MANIFESTS_FOUND: 3,
+  NO_SUPPORTED_PROJECTS_DETECTED: 3,
 };
 
 async function runCommand(args: Args) {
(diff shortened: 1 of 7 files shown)


src/cli/index.ts
```diff
@@ -92,12 +92,16 @@ async function handleError(args, error) {
   spinner.clearAll();
   let command = 'bad-command';
   let exitCode = EXIT_CODES.ERROR;
-  const noSupportedManifestsFound = error.message?.includes(
+  const noSupportedProjectsDetectedPossibleMessage = [
     'Could not detect supported target files in',
+    'We found 0 supported files',
+  ];
+  const noSupportedProjectsDetected = noSupportedProjectsDetectedPossibleMessage.some(
```
pkey, line 99: optional: this is quite shaky in case the message gets changed :/ maybe it's better to check based on Error type? You could create an error which inherits CustomError and then check based on that, for example.

pkey: It seems like you do have `NoSupportedSastFiles`

ArturSnyk: I agree i don't love it either. and I considered refactoring the whole error messaging part. But I stopped because it turned up to be more work then expected.

I liked the point you raised, regarding the error type. I issue another commit to check by error type
https://github.com/snyk/snyk/pull/2182/commits/bf361511e01a2e6f2b7c66a9dcef750afbe513ca

pkey (approved):

src/cli/index.ts
```diff
@@ -92,12 +93,16 @@ async function handleError(args, error) {
   spinner.clearAll();
   let command = 'bad-command';
   let exitCode = EXIT_CODES.ERROR;
+
   const noSupportedManifestsFound = error.message?.includes(
     'Could not detect supported target files in',
   );
+  const noSupportedSastFiles = error instanceof NoSupportedSastFiles;
+  const noSupportedProjectsDetected =
+    noSupportedManifestsFound || noSupportedSastFiles;
```
novalex, line 102: Shouldn't this be an `and` condition?
```suggestion
  const noSupportedProjectsDetected =
    noSupportedManifestsFound && noSupportedSastFiles;
```

ArturSnyk: Nope, we would like to exit with code `3` (NO_SUPPORTED_PROJECTS_DETECTED) when we don't have supported files detected. and we can reach this state with one of those checks

orsagie (approved):

PR #2184, opened 2021-08-23
fix: don't render two newlines at the end

ghost:
When trying to use command substitution (`$(snyk ...)`), because we render 2 new lines, commands like `snyk config get api` will return `value\n\n`. Command substitution can handle 1 `\n` but not 2. Generally, it also just looks wrong from a CLI perspective.

Before

```
node ➜ /workspaces/snyk (feat/webpack) $ ./bin/snyk config get api
123456789

node ➜ /workspaces/snyk (feat/webpack ✗) $
```

After

```
node ➜ /workspaces/snyk (master) $ ./bin/snyk config get api
123456789
node ➜ /workspaces/snyk (master) $
```

The fix is to only use `console.log` if we have alerts to show. Otherwise don't.

I did not look into what "alerts" are. I assume they are extra messages from the API server. I've moved them to stderr (using `console.warn`) so if there are alerts, then `$(snyk config get api)` will not include alerts and break things since it'll only provide stdout. stderr will be printed to console as usual (unless if the developer chooses to redirect it).

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -317,7 +317,10 @@ async function main() {
   }
 
   if (!globalArgs.options.json) {
-    console.log(alerts.displayAlerts());
+    const alertsMessage = alerts.displayAlerts();
+    if (alertsMessage) {
+      console.log(alertsMessage);
+    }
   }
 
   if (!process.env.TAP && failed) {
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -319,7 +319,7 @@ async function main() {
   if (!globalArgs.options.json) {
     const alertsMessage = alerts.displayAlerts();
     if (alertsMessage) {
-      console.log(alertsMessage);
+      console.warn(alertsMessage);
     }
   }
 
--- a/test/jest/acceptance/cli-args.spec.ts
+++ b/test/jest/acceptance/cli-args.spec.ts
@@ -108,7 +108,7 @@ test('test --file=blah --scan-all-unmanaged', async () => {
     const { code, stdout } = await runSnykCLI(
       `test --${arg} --yarn-workspaces`,
     );
-    expect(stdout).toEqual(
+    expect(stdout).toMatch(
       `The following option combination is not currently supported: ${arg} + yarn-workspaces`,
     );
     expect(code).toEqual(2);
@@ -118,7 +118,7 @@ test('test --file=blah --scan-all-unmanaged', async () => {
     const { code, stdout } = await runSnykCLI(
       `monitor --${arg} --yarn-workspaces`,
     );
-    expect(stdout).toEqual(
+    expect(stdout).toMatch(
       `The following option combination is not currently supported: ${arg} + yarn-workspaces`,
     );
     expect(code).toEqual(2);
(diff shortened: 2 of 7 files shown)


src/cli/index.ts
```diff
@@ -319,7 +319,10 @@ async function main() {
   }
 
   if (!globalArgs.options.json) {
-    console.log(alerts.displayAlerts());
+    const alertsMessage = alerts.displayAlerts();
+    if (alertsMessage) {
+      console.log(alertsMessage);
```
JackuB, line 324: Pushing alerts to stderr could also improve the situation

jahed-snyk: Agreed.

jahed-snyk: Done.

src/cli/index.ts
```diff
@@ -317,7 +317,10 @@ async function main() {
   }
 
   if (!globalArgs.options.json) {
-    console.log(alerts.displayAlerts());
+    const alertsMessage = alerts.displayAlerts();
+    if (alertsMessage) {
+      console.warn(alertsMessage);
+    }
```
jahed-snyk, line 323: This is the main change.

test/jest/acceptance/cli-args.spec.ts
```diff
@@ -108,7 +108,7 @@ test('test --file=blah --scan-all-unmanaged', async () => {
     const { code, stdout } = await runSnykCLI(
       `test --${arg} --yarn-workspaces`,
     );
-    expect(stdout).toEqual(
+    expect(stdout).toMatch(
```
jahed-snyk, line 111: These tests cared about new lines in stdout which is a bit too strict. I could add `\n` to the expected string, but thought as long that expected message is in the stdout somewhere, this test should be okay with it. Formatting is none of its concern.

test/jest/acceptance/snyk-config/snyk-config.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  const expectedToken = 'my-test-token';
+
+  const { code, stdout } = await runSnykCLI('config get api', {
+    env: {
+      ...process.env,
+      SNYK_CFG_API: expectedToken,
+    },
+  });
+  expect(code).toEqual(0);
+  expect(stdout).toEqual(expectedToken + '\n');
```
jahed-snyk, line 15: We care about formatting here so it's using `toEqual` instead of `toMatch`.

test/jest/util/runCommand.ts
```diff
@@ -34,12 +34,8 @@ const runCommand = (
...
       resolve({
         code: code || 0,
-        stdout: Buffer.concat(stdout)
-          .toString('utf-8')
-          .trim(),
-        stderr: Buffer.concat(stderr)
-          .toString('utf-8')
-          .trim(),
+        stdout: Buffer.concat(stdout).toString('utf-8'),
+        stderr: Buffer.concat(stderr).toString('utf-8'),
```
jahed-snyk, line 38: New lines matter so I've removed the `trim` from these as it's a bad assumption to make and can break features (like command substitution in bash).

maxjeffos (approved):

PR #2187, opened 2021-08-24
chore: organise package scripts based on workflows

ghost:
The scripts in package.json are in a bit of a mess so I've organised them based on common workflow.

- Clean
- Install
- Format
- Lint
- Build
- Test


package.json
```diff
@@ -48,20 +57,7 @@
...
     "test:test": "tap test/*.test.* -Rspec --timeout=300 --node-arg=-r --node-arg=ts-node/register",
-    "test:smoke": "./scripts/run-smoke-tests-locally.sh",
-    "test": "npm run test:acceptance && npm run test:system && npm run test:test && npm run test:jest && test:packages-unit && test:packages-acceptance",
-    "test-windows": "npm run test:acceptance-windows && npm run test:system && npm run test:test && npm run test:jest",
-    "lint": "run-p --aggregate-output lint:*",
-    "lint:js": "eslint --color --cache '**/*.{js,ts}'",
-    "lint:formatting": "prettier --check '**/*.{js,ts,json,yaml,yml,md}'",
-    "format": "prettier --write '**/*.{js,ts,json,yaml,yml,md}'",
-    "format:changes": "./scripts/format/prettier-changes.sh",
-    "check-tests": "! grep 'test\\.only' test/*.test.js -n",
```
jahed-snyk, line 59: This `check-tests` is very loose and misleading. It only tests old tap tests at the top-level and only for test.only. We're better without it. We can improve how we detect accidental `.only` commits in a separate PR (as it needs to cover tap and jest).

.circleci/config.yml
```diff
@@ -173,9 +173,7 @@ jobs:
       - run: sudo apt-get install python-pip
       - attach_workspace:
           at: ~/snyk
-      - run:
-          name: Run tests
-          command: npm run test:common
```
jahed-snyk, line 178: These aren't really "tests", it's the linter step.

.circleci/config.yml
```diff
@@ -221,9 +219,8 @@ jobs:
           name: Update local .npmrc file (Windows)
           command: echo "//registry.npmjs.org/:_authToken=$env:NPM_TOKEN" >> .npmrc
       - run:
-          name: Run auth
-          shell: bash.exe
-          command: npm run snyk-auth-windows
+          name: Add Snyk Token to Snyk Config
+          command: node ./bin/snyk config set "api=$env:SNYK_API_KEY"
```
jahed-snyk, line 223: This step tended to fail now and then during contact with Registry. So I've changed it to just set the token and not validate it. Our tests will validate it anyway.

package.json
```diff
@@ -30,14 +30,23 @@
     "packages/*"
   ],
   "scripts": {
-    "clean": "rimraf dist tsconfig.tsbuildinfo pysrc",
-    "build": "npm run build:dev",
-    "build-watch": "tsc -w",
-    "find-circular": "npm run build && madge --circular ./dist",
```
jahed-snyk, line 36: We don't use this anymore. Also removed "madge" dependency.

maxjeffos (approved):

Avishagp (approved):

PR #2189, opened 2021-08-24
fix: support oauth tokens for feature flags

love-bhardwaj:

#### What does this PR do?

Utilized the newer `getAuthHeader` function. This function returns the OAuth token when set. Typical use cases include Snyk Plugin in AWS CodePipelines.

Set/export environmental variable `SNYK_OAUTH_TOKEN` and the run `snyk monitor` for a package manager that calls the experimental dep graph API endpoint which is behind a feature flag.

We at Team Moose are implementing the monitor functionality in AWS CodePipeline. We found that for package managers such as `yarn` or `npm` it calls experimental dep graph API endpoint. Since the dep graph API endpoint in the `dev` environment is behind a feature flag. The call to check if the org had a feature flag was failing because of `Authorization: token undefined` header.

### DIFF
--- a/test/jest/acceptance/oauth-token.spec.ts
+++ b/test/jest/acceptance/oauth-token.spec.ts
@@ -12,14 +12,17 @@ describe('test using OAuth token', () => {
     const apiPath = '/api/v1';
     const apiPort = process.env.PORT || process.env.SNYK_PORT || '12345';
     env = {
-      ...process.env,
+      PATH: process.env.PATH || '',
       SNYK_API: 'http://localhost:' + apiPort + apiPath,
-      SNYK_TOKEN: '123456789',
       SNYK_OAUTH_TOKEN: 'oauth-jwt-token',
     };
 
     server = fakeServer(apiPath, env.SNYK_TOKEN);
     server.listen(apiPort, () => done());
+  });
+
+  afterEach(() => {
+    server.clearRequests();
   });
 
   afterAll((done) => {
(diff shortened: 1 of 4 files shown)


test/jest/acceptance/oauth-token.spec.ts
```diff
@@ -57,4 +58,16 @@ describe('test using OAuth token', () => {
     expect(requests[0].headers.authorization).toBe('Bearer oauth-jwt-token');
     expect(requests[0].method).toBe('PUT');
   });
+
+  it('should successfully check if feature flag in enabled with OAuth token', async () => {
+    await featureFlags.isFeatureFlagSupportedForOrg(
+      'experimentalDepGraph',
+      'test',
+    );
```
jahed-snyk, line 66: These are acceptance tests so you'll need to use the same structure as the other tests in this file. Use `runSnykCLI` instead of a method call.

test/jest/acceptance/oauth-token.spec.ts
```diff
@@ -57,4 +58,16 @@ describe('test using OAuth token', () => {
...
+    const request = server.popRequest();
+    expect(request.headers.authorization).toBe('Bearer oauth-jwt-token');
+    expect(request.url).toContain(
+      '/api/v1/cli-config/feature-flags/experimentalDepGraph',
+    );
```
jahed-snyk, line 71: `toContain` is for arrays. Use `toMatch` for partial string matching.

https://jestjs.io/docs/expect#tomatchregexp--string

test/jest/acceptance/oauth-token.spec.ts
```diff
@@ -57,4 +58,16 @@ describe('test using OAuth token', () => {
     expect(requests[0].headers.authorization).toBe('Bearer oauth-jwt-token');
     expect(requests[0].method).toBe('PUT');
   });
+
+  it('should successfully check if feature flag in enabled with OAuth token', async () => {
```
jahed-snyk, line 62: To highlight the specific thing this test cares about, might be more clear as:

```
uses oauth token for feature flag requests
```

test/jest/acceptance/oauth-token.spec.ts
```diff
@@ -57,4 +58,16 @@ describe('test using OAuth token', () => {
...
+    await featureFlags.isFeatureFlagSupportedForOrg(
+      'experimentalDepGraph',
+      'test',
+    );
+    const request = server.popRequest();
```
jahed-snyk, line 67: This will only give you the last request which is probably monitor or analytics. You might need to use `popRequests` instead and pass a number to manually find the correct entry that makes the feature flag request.

It isn't great, we're looking to improve it in the near future.

test/jest/acceptance/oauth-token.spec.ts
lovebhardwajsnyk: The reason for writing the test this way is because we are not sure which commands are behind a FF or how to trigger them. Please let us know what CLI command would trigger the `isFeatureFlagSupportedForOrg` and we would be happy to refractor. Or if unit test for `isFeatureFlagSupportedForOrg` is a better choice.

ghost (dismissed):

ghost (approved):

PR #2192, opened 2021-08-25
chore: don't warn about changes to test fixtures in dangerjs

Avishagp:
`dangerfile.js` contained some logic regarding adding new Tap tests instead of Jest test. This logic resulted in a redundant warning when adding new test fixtures. This should stop these warnings.


jahed-snyk (approved):
Approved.

Why do we have test files in our fixtures? We should look into cleaning that up.

PR #2202, opened 2021-08-27
refactor: remove experimental-dep-graph flag

ghost:
The deprecation plan for the `experimentalDepGraph` feature flag is a bit vague.
Looking at: https://github.com/snyk/snyk/blob/63975cecee24ffae37e0738016ddef1c53bb41f6/src/lib/monitor/index.ts

We have a bunch of options.

1. Do we want to remove `experimentalMonitorDepGraphFromDepTree`? That is, always use `monitorDepTree`. 
2. Do we want to just remove the `monitorGraphSupportedRes` logic (always call `experimentalMonitorDepGraphFromDepTree`, provided `GRAPH_SUPPORTED_PACKAGE_MANAGERS` is true.
3. Do we want to remove `monitorDepTree` and always use `experimentalMonitorDepGraphFromDepTree`. Renaming it to `monitorDepGraphFromDepTree`  

TL;DR: This PR is going with Option 2. See self-review for details on changes.

## Option 1 ❌

I originally thought 1 going by the `TODO`s and `@deprecation` but I think that was a trap.

https://github.com/snyk/snyk/blob/63975cecee24ffae37e0738016ddef1c53bb41f6/src/lib/monitor/index.ts#L119-L124

https://github.com/snyk/snyk/blob/63975cecee24ffae37e0738016ddef1c53bb41f6/src/lib/monitor/index.ts#L505-L510

I ended up trying it but it's a mess. Doesn't seem like the right track: https://github.com/snyk/snyk/pull/2196

But it did help me understand what this was all about. Looking at how:

-  `snyk test` always converts trees to graphs:

    https://github.com/snyk/snyk/blob/b7a190b4098d87cd885c73dc76ec8492388c52b4/src/lib/snyk-test/run-test.ts#L732-L735

-  `experimentalDepGraph` defaults to `true` in the Snyk API:

    https://github.com/snyk/registry/blob/98fd3ab36a5aadb6d04213e77e6aa28b5af1f265/src/lib/features.ts#L114-L117

I’m guessing what was actually meant was 2 or 3.

## Option 2 ✅ 

I’m going with 2 as it’s the safest and gives us what we want. That is, stop making the feature flag request and just assume it’s `true`. This is what this PR does.

## Option 3 ❓ 
It would be nice to do 3, so that `snyk monitor` matches `snyk test`, but I have no idea if we need `monitorDepTree` so I’ll just leave it for now. Maybe some package managers (those not in `GRAPH_SUPPORTED_PACKAGE_MANAGERS`) don't support `experimentalMonitorDepGraphFromDepTree` or something.

### DIFF
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -1,6 +1,5 @@
 import * as Debug from 'debug';
 import * as path from 'path';
-
 import * as depGraphLib from '@snyk/dep-graph';
 import * as snyk from '..';
 import { apiOrOAuthTokenExists, getAuthHeader } from '../api-token';
@@ -20,7 +19,6 @@ import {
   Contributor,
 } from '../types';
 import * as projectMetadata from '../project-metadata';
-
 import {
   MonitorError,
   ConnectionTimeoutError,
@@ -54,8 +52,6 @@ import { abridgeErrorMessage } from '../error-format';
 const debug = Debug('snyk');
 
 const ANALYTICS_PAYLOAD_MAX_LENGTH = 1024;
-
-// TODO(kyegupov): clean up the type, move to snyk-cli-interface repository
 
 interface MonitorBody {
   meta: Meta;
@@ -187,8 +183,6 @@ async function monitorDepTree(
     root,
     meta.isDocker ? 'docker' : packageManager!,
     options,
-    // TODO: fix this and send only send when we used resolve-deps for node
-    // it should be a ExpandedPkgTree type instead
     depTree,
     targetFileDir,
   );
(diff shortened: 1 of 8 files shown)


anthogez:
@jahed-snyk thanks a lot for raising this pull request and add the different details/options.

I agree, let's go with `Option 2 ` and incrementally we can go beyond and apply further 
refactors till we arrive to the point where we will have just a single "monitorDepGraph" call

Small note: 

The ecosystems that are under this "feature flag" 
flow they are still creating dependencyTree from scratch and converting to graph.
Once all of those have been migrated to a graph from scratch we could easily get rid of it as well
and finally have a single `monitorDepGraph func` which IMO would be the best scenario.

Regarding monitorDepTree, it might still be used by other products, worth verify before getting rid of it

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -85,8 +83,16 @@ if (!isWindows) {
     t.end();
   });
 
+  beforeEach(async () => {
+    server.clearRequests();
+  });
```
jahed-snyk, line 88: Making sure tests aren't polluting each other with requests from a previous test.

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -185,28 +191,6 @@ if (!isWindows) {
     t.pass('succeed');
   });
 
-  test('`monitor npm-package with experimental-dep-graph enabled, but bad auth token`', async (t) => {
```
jahed-snyk, line 188: Test not needed.

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -331,45 +314,7 @@ if (!isWindows) {
     t.notOk(packageC2.deps.length, 'a.d.c has no dependencies');
   });
 
-  test('`monitor npm-package-pruneable --prune-repeated-subdependencies --experimental-dep-graph`', async (t) => {
```
jahed-snyk, line 334: Test not needed.

test/acceptance/cli-monitor/cli-monitor.acceptance.test.ts
```diff
@@ -388,10 +333,7 @@ if (!isWindows) {
       loadPlugin.restore();
     });
 
-    await cli.monitor('sbt-simple-struts', {
-      'experimental-dep-graph': true,
-      'sbt-graph': true,
```
jahed-snyk, line 393: `sbt-graph` isn't used by sbt-plugin. The feature it was gating is always enabled.

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -23,10 +23,9 @@ export const AllProjectsTests: AcceptanceTests = {
         allProjects: true,
         detectionDepth: 2,
       });
-      // Pop all calls to server and filter out calls to `featureFlag` endpoint
-      const requests = params.server
-        .popRequests(4)
-        .filter((req) => req.url.includes('/monitor/'));
+      const requests = params.server.requests.filter((req) =>
+        req.url.includes('/monitor/'),
+      );
```
jahed-snyk, line 28: No need to pop specific number of requests. Just filter for the ones we care about. That way we don't need to worry about feature flag requests at all.

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -263,39 +249,43 @@ export const AllProjectsTests: AcceptanceTests = {
       const mavenAll = requests.find((req) => req.url.indexOf('maven') > -1);
       const sbtAll = requests.find((req) => req.url.indexOf('sbt') > -1);
 
+      params.server.clearRequests();
```
jahed-snyk, line 252: This test runs `cli.monitor` multiple times so just making sure requests aren't been polluted from previous runs.

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -263,39 +249,43 @@ export const AllProjectsTests: AcceptanceTests = {
...
       await params.cli.monitor('mono-repo-project', {
         file: 'package-lock.json',
       });
-      // Pop all calls to server and filter out calls to `featureFlag` endpoint
-      const [npmFile] = params.server
-        .popRequests(2)
-        .filter((req) => req.url.includes('/monitor/'));
+      const npmFile = params.server.popRequest();
```
jahed-snyk, line 268: No more feature flag to filter out.

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
```diff
@@ -399,21 +389,6 @@ export const AllProjectsTests: AcceptanceTests = {
...
-          ffRequests.every((req) => req.url.includes('experimentalDepGraph')),
-          true,
-          'all left requests are feature flag requests',
-        );
-        t.equal(requests.length, 9, 'sends expected # requests'); // extra feature-flags request
-        t.equal(
-          params.server.requests.length,
-          0,
-          `${params.server.requests.length} pending requests`,
-        );
```
jahed-snyk, line 415: It seems the only reason this test cares so much about feature flags is because it wanted to clean up `server.requests` for the next test. This made it really brittle. We can just use `server.clearRequests` to reset the requests array and have the same effect.

test/cli-server.js
```diff
@@ -55,7 +55,7 @@ module.exports = function(root, apikey, notAuthorizedApiKey) {
     return next();
   });
 
-  server.put(root + '/monitor/npm', function(req, res) {
+  server.put(root + '/monitor/npm/graph', function(req, res) {
```
jahed-snyk, line 58: All npm requests go through the `graph` endpoint. Previously, this wasn't happening because of bad mocks. While `fakeServer` defaults all feature flags to `true`, some tests used this old `cli-server` mock which doesn't. Hopefully we can migrate away from this `cli-server` to avoid such divergence in the future.

test/jest/acceptance/oauth-token.spec.ts
```diff
@@ -61,36 +61,12 @@ describe('OAuth Token', () => {
...
+    server.requests.forEach((request) => {
+      expect(request).toMatchObject({
         headers: {
           authorization: 'Bearer oauth-jwt-token',
         },
       });
     });
   });
-
-  it('uses oauth token when fetching feature flags', async () => {
```
jahed-snyk, line 73: Don't need this test as all requests are now captured and enforced by the tests above.

src/lib/monitor/index.ts
```diff
@@ -116,33 +109,16 @@ export async function monitor(
...
-      camelCase('experimental-dep-graph'),
-      options.org || config.org,
+    return await monitorDepGraphFromDepTree(
+      root,
+      meta,
+      scannedProject,
+      pluginMeta,
+      options,
+      targetFileRelativePath,
+      contributors,
```
jahed-snyk, line 120: This is the main change. We no longer request the feature flag (which defaults to true) and instead assume its always true. Not that it also never checked for `--experimental-dep-graph` so the feature flag was always requested.

src/lib/monitor/index.ts
```diff
@@ -258,91 +232,74 @@ async function monitorDepTree(
     callGraphPayload = callGraph;
   }
 
-  // TODO(kyegupov): async/await
```
jahed-snyk, line 261: The changes below are because I made this change for all three functions (as a separate commit). Migrating `new Promise` to async/await.

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
maxjeffos: Makes me think we should add a `.requestsMatching(url: string): Request[]` method to fake server (because we're using this pattern quite a bit now in recent PRs. Not suggesting for this PR, just putting it out.

maxjeffos (approved):

test/acceptance/cli-monitor/cli-monitor.all-projects.spec.ts
jahed-snyk: Yeah, definitely.

robcresswell:
I _think_ this has broken the CLI for onprem customers, see https://snyk.zendesk.com/agent/tickets/16148

PR #2205, opened 2021-08-29
chore: enforce usage of ES6 modules

Avishagp:
Enforce usage of ES6 modules in `dangerfile.js` by reading the modified files and warning if any of them contains old-style `import` or `export`.

Screen Shot 2021-09-01 at 15 49 21

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,4 +1,5 @@
 const { danger, warn, fail, message } = require('danger');
+const fs = require('fs');
 
 const MAX_COMMIT_MESSAGE_LENGTH = 72;
 const SMOKE_TEST_BRANCH = 'smoke/';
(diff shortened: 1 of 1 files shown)


ghost (approved):

maxjeffos (approved):

dangerfile.js
```diff
@@ -99,4 +100,19 @@ if (danger.github && danger.github.pr) {
...
+    const fileContent = fs.readFileSync(f, 'utf8');
+    return (
+      fileContent.includes('module.exports') ||
+      fileContent.includes('= require(')
+    );
+  });
+
+  if (filesUsingOldModules) {
+    warn(
+      "Looks like a file you've modified is using an old-style import or export, either `module.exports` or `require()`. Please update them to ES6 [import syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [export syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export).",
```
JackuB, line 115: You should also say which file this is affecting with (not sure if this can handle multiple files?)

```
// Highlights low-priority issues, but does not fail the build. Message is shown inside a HTML table.
warn(message: MarkdownString, file?: string, line?: number) => void
```
https://danger.systems/js/reference.html

And maybe changing to messaging as to "why" and whether it's a required thing or a plea?

JackuB: Similar to this for example:
https://github.com/snyk/snyk/blob/a52ddeb3a247e46bd39e8bf1c3b9fd1ee2773a22/dangerfile.js#L65-L69

PR #2210, opened 2021-08-31
chore(ci): use env for npm token

ghost:
This is a follow-up from previous attempts. The goal is to avoid touching `.npmrc` which has a risk for credential leakage and also complicates our CI pipeline.

The NPM docs suggest in CI two approaches to providing the NPM_TOKEN:

https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow

1. `NPM_TOKEN` environment variable
2. `.npmrc` with a reference to an environment various.

Option 2 was attempted in #1974. Due to yarn smoke test failures (yarn fails if the env var isn't defined), it was reverted in #2010.

According to docs, since we already have `NPM_TOKEN` in our environment, Option 1 should work. We don't need any additional steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,12 +34,6 @@ commands:
       - run:
           name: Install dependencies
           command: npm ci
-  update_local_npmrc_linux:
-    description: Update local .npmrc file (Linux)
-    steps:
-      - run:
-          name: Update local .npmrc file (Linux)
-          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
   build_ts:
     description: Build js files from ts
     steps:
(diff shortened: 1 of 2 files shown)


.npmrc
```diff
@@ -1,4 +1,3 @@
-@snyk:registry=https://registry.npmjs.org
```
jahed-snyk, line 1: This is the default registry anyway so we don't need this.

.circleci/config.yml
```diff
@@ -368,12 +357,11 @@ jobs:
           name: Update package versions
           command: ./release-scripts/update-versions.sh
       - run:
-          name: Make git ignore changes relating to version updates and to .npmrc
+          name: Make git ignore changes relating to version updates
           command: |
             git update-index --skip-worktree -- ./lerna.json
             git update-index --skip-worktree -- ./package.json
             git update-index --skip-worktree -- ./packages/snyk-protect/package.json
-            git update-index --skip-worktree -- ./.npmrc
```
jahed-snyk, line 376: `.npmrc` is no longer modified.

maxjeffos (approved):

PR #2211, opened 2021-08-31
feat: support yarn workspaces with transitives that are also workspaces

lili2311:

#### What does this PR do?
Yarn workspaces introduces complexity where local packages
are mentioned as required dependency but are not present on the lockfile
this means that the lockfile is considered deeply out of sync by the plugin.

Allow to skip the check if user is calling with StrictOutOfSync=false not just
for top level dependencies but also the transitives as a package can rely on another local workspace which lists a workspace as a dep.

https://github.com/snyk/nodejs-lockfile-parser/pull/119

https://github.com/snyk/nodejs-lockfile-parser/pull/119

Scan https://github.com/backstage/backstage

https://github.com/snyk/snyk/issues/2126

https://github.com/snyk/snyk/issues/2126

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "snyk",
       "version": "1.0.0-monorepo",
       "license": "Apache-2.0",
       "workspaces": [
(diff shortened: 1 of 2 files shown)


jan-stehlik (approved):
LGTM 🎉

Avishagp (approved):

PR #2214, opened 2021-09-01
chore(ci): fix npm authentication

ghost:
Follow up from #1974  and #2210.

npm doesn't actually support `NPM_TOKEN`. At all. It's just that certain services like GitHub Actions and Heroku have their own thing that happens to use it to log into npm.

https://github.com/npm/npm/issues/8356

So this PR is sort of making the most out of a bad situation. Instead of using `echo >>` like we did before, it's using `npm config set` to modify `~/.npmrc` (home directory, instead of the `.npmrc` in our git repo). It also uses a literal `${NPM_TOKEN}` rather than the value itself so that there's less risk in exposing the token.

```bash
$ npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
$ cat ~/.npmrc
//registry.npmjs.org/:_authToken=${NPM_TOKEN}
$ export NPM_TOKEN='<token>'
$ npm whoami
snyk-user
```

https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,6 +345,7 @@ jobs:
       - aws-cli/install:
           version: 2.2.32
       - run: sudo npm install -g npm@7.21.0
+      - run: npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
       - run:
           name: Install osslsigncode
           command: sudo apt-get install -y osslsigncode


.circleci/config.yml
```diff
@@ -345,6 +345,7 @@ jobs:
       - aws-cli/install:
           version: 2.2.32
       - run: sudo npm install -g npm@7.21.0
+      - run: npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
```
JackuB, line 348: Surprised about the quotes around npm token

JackuB (approved):

maxjeffos (approved):

PR #2215, opened 2021-09-01
chore(ci): cache downloads

ghost:
Constantly downloading remote resources makes our CI runs slow and unreliable. The servers we download from can slow down or fall down. In such scenarios, it's best we minimise how much it affects us and how much we contribute to it.

This PR introduces caching steps for downloading via:
- SDKMAN
- Chocolatey (limited, see self-review)
- NPM

While speed is a minor benefit, the main benefit is reliability.

I've also added a few more changes:

- CircleCI tends to use words like "Installing software" rather than "Install software" for steps. Not consistently, but most of the time. So I've changed our wording to fit that for consistency.
- Windows is now only tested with Node 14 (default). See self-review for why.
- As setting up caching needed me to reorganise the steps, I took the opportunity to reduce some duplication between `dev-release` and `prod-release` (formerly known as `release`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ commands:
         default: '7.21.1'
     steps:
       - restore_cache:
-          name: Restoring NPM cache
+          name: Restoring npm cache
           keys:
             - npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
       - when:
@@ -48,7 +48,7 @@ commands:
             equal: ['win/default', << parameters.os >>]
           steps:
             run:
-              name: Installing NPM
+              name: Installing npm
               command: npm install -g npm@<< parameters.npm_version >>
       - when:
           condition:
@@ -56,10 +56,10 @@ commands:
               equal: ['win/default', << parameters.os >>]
           steps:
             run:
-              name: Installing NPM
+              name: Installing npm
               command: sudo npm install -g npm@<< parameters.npm_version >>
       - run:
-          name: Configuring NPM
+          name: Configuring npm
           command: npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
       - run:
           name: Installing project dependencies
@@ -69,7 +69,7 @@ commands:
             equal: ['win/default', << parameters.os >>]
           steps:
             - save_cache:
-                name: Saving NPM cache
+                name: Saving npm cache
                 key: npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
                 paths:
                   - ~\AppData\Local\npm-cache
@@ -79,7 +79,7 @@ commands:
               equal: ['win/default', << parameters.os >>]
           steps:
             - save_cache:
-                name: Saving NPM cache
+                name: Saving npm cache
                 key: npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
                 paths:
                   - ~/.npm
@@ -141,12 +141,29 @@ commands:
       - run:
           name: Installing NodeJS
           command: choco install nodejs --version=<< parameters.node_version >> --no-progress
-  install_shellspec:
-    description: Install Shellspec
+  install_shellspec_dependencies:
     steps:
-      - run: curl -fsSL https://git.io/shellspec | sh -s -- -y
-      - run: sudo ln -s ${HOME}/.local/lib/shellspec/shellspec /usr/local/bin/shellspec
-      - run: sudo apt-get install jq
+      - run:
+          name: Installing test dependencies
+          command: sudo apt-get install jq python-pip
+      - run:
+          name: Installing ShellSpec
+          command: |
+            curl -fsSL https://git.io/shellspec | sh -s -- -y
+            sudo ln -s ${HOME}/.local/lib/shellspec/shellspec /usr/local/bin/shellspec
+  pack_snyk_cli:
+    steps:
+      - run:
+          name: Packing Snyk CLI
+          command: |
+            mkdir ./dist-pack
+            cd ./dist-pack
+            npm pack ..
+  install_release_dependencies:
+    steps:
+      - run:
+          name: Installing release dependencies
+          command: sudo apt-get install -y osslsigncode
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ defaults: &defaults
   parameters:
     node_version:
       type: string
-      default: ''
+      default: '14.17.5'
```
jahed-snyk, line 12: Made this default so we don't need to specify it in places like builds and releases. It's only relevent when testing.

.circleci/config.yml
```diff
@@ -28,88 +28,144 @@ defaults: &defaults
...
+              name: Installing NPM
+              command: npm install -g npm@<< parameters.npm_version >>
+      - when:
+          condition:
+            not:
+              equal: ['win/default', << parameters.os >>]
+          steps:
+            run:
+              name: Installing NPM
+              command: sudo npm install -g npm@<< parameters.npm_version >>
```
jahed-snyk, line 61: Had to use a `when` condition because Linux needs `sudo` for global installs.

.circleci/config.yml
```diff
@@ -28,88 +28,144 @@ defaults: &defaults
...
+      - when:
+          condition:
+            not:
+              equal: ['win/default', << parameters.os >>]
+          steps:
+            - save_cache:
+                name: Saving NPM cache
+                key: npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
+                paths:
+                  - ~/.npm
```
jahed-snyk, line 86: Had to use a `when` because Linux and Windows use different paths for NPM cache.

.circleci/config.yml
```diff
@@ -28,88 +28,144 @@ defaults: &defaults
...
-      - run:
-          name: Installing sbt
-          command: sdk install sbt 1.5.5
+      - restore_cache:
+          name: Restoring SDKMAN cache
+          keys:
+            - sdkman-cache-v2-{{ arch }}-{{ checksum ".circleci/install-sdks-linux.sh" }}
+      - run:
+          name: Installing SDKs
+          command: ./.circleci/install-sdks-linux.sh
```
jahed-snyk, line 126: I moved the `sdk install` steps to a script so that we can use its checksum as a cache key. Kind of like a package-lock. Also makes it more similar to Chocolatey's steps.

.circleci/config.yml
```diff
@@ -28,88 +28,144 @@ defaults: &defaults
...
+          paths:
+            - ~\AppData\Local\Temp\chocolatey
+  install_sdks_linux:
     steps:
       - run:
           name: Installing SDKMAN
           command: |
             curl -s "https://get.sdkman.io?rcupdate=false" | bash
             echo -e '\nsource "/home/circleci/.sdkman/bin/sdkman-init.sh"' >> $BASH_ENV
             source $BASH_ENV
```
jahed-snyk, line 119: I didn't bother to cache the SDKMAN installer as it's quite complicated. Ideally we'd cache the ZIP file that it downloads (containing all the various scripts), but the installer script they provide doesn't have a cache feature. I might send them a PR to support it.

.circleci/config.yml
```diff
@@ -28,88 +28,144 @@ defaults: &defaults
...
+          keys:
+            - chocolatey-cache-v2-{{ arch }}-{{ checksum ".circleci/chocolatey.config" }}
+      - run:
+          name: Installing SDKs
+          command: choco install .circleci/chocolatey.config --no-progress
+      - save_cache:
+          name: Saving Chocolatey cache
+          key: chocolatey-cache-v2-{{ arch }}-{{ checksum ".circleci/chocolatey.config" }}
+          paths:
+            - ~\AppData\Local\Temp\chocolatey
```
jahed-snyk, line 111: The caching here is limited. Chocolately doesn't provide a consistent cache folder and this one only contains `sbt` and `gradle`. For proper caching, we'll need to buy a pro license.

This doesn't save much time. Most time is spent installaling software rather than downloading. We can't cache installed software because Windows has various steps outside of the filesystem (like setting up PATHs, registry, etc.).

What it does do is make our builds more reliable since we don't need to rely on gradle and sbt's servers to always be up.

.circleci/config.yml
```diff
@@ -28,88 +28,144 @@ defaults: &defaults
...
+          name: Removing pre-installed NodeJS
           command: |
             $current_node_version = node --version
             nvm uninstall $current_node_version
-            choco install nodejs --version=<< parameters.node_version >>
-  install_shellspec:
-    description: Install Shellspec
+      - run:
+          name: Installing NodeJS
+          command: choco install nodejs --version=<< parameters.node_version >> --no-progress
```
jahed-snyk, line 145: We still install NodeJS separately from the other Chocolatey downloads because the version is dynamic (can't be placed in the chocolatey.config XML file).

I didn't bother with caching as there's nothing to cache (see the other comment on how Chocolatey's caching works).

.circleci/config.yml
```diff
@@ -422,49 +444,40 @@ workflows:
...
           context: nodejs-install
           requires:
             - Build
           filters:
             branches:
               ignore:
                 - master
-          matrix:
-            parameters:
-              node_version: ['10.24.1', '12.22.5', '14.17.5']
```
jahed-snyk, line 435: Windows is only run against the default version now (14.17.5). We just want to make sure OS-specific stuff like path separators work. We can rely on the Linux tests for checking compatibility with different Node versions.

.circleci/config.yml
```diff
@@ -505,10 +518,18 @@ workflows:
...
-      - release:
-          name: Release
+      - dev-release:
+          name: Development Release
+          requires:
+            - Build
+          filters:
+            branches:
+              ignore:
+                - master
```
jahed-snyk, line 528: Moved this down to be next to `prod-release`.

.circleci/config.yml
```diff
@@ -505,10 +518,18 @@ workflows:
...
+
+      - prod-release:
```
jahed-snyk, line 530: Renamed `release` to `prod-release` to help differentiate it from `dev-release`.

.circleci/config.yml
```diff
@@ -28,88 +28,142 @@ defaults: &defaults
...
+            run:
+              name: Installing npm
+              command: sudo npm install -g npm@<< parameters.npm_version >>
       - run:
-          name: Install dependencies
+          name: Configuring npm
+          command: npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
+      - run:
+          name: Installing project dependencies
           command: npm ci
```
JackuB, line 66: Screen Shot 2021-09-03 at 11 04 16

Can't we use cached npm deps before running `npm ci`? Currently it installs them every time in the Build step

JackuB (approved):
Tons of great changes 🎉

.circleci/config.yml
jahed-snyk: Do you mean caching `node_modules`? `npm ci` removes `node_modules` and does a clean install so it's faster without `node_modules` than with.

https://docs.npmjs.com/cli/v7/commands/npm-ci

Right now we're only caching `~/.npm`, the actual npm cache that contains downloaded tarballs. It's safer and the recommended directory to cache. `node_modules` can get really large and contain runtime artifacts like binaries and symlinks (which is why `npm ci` clears it, it's faster to remove than to validate).

PR #2219, opened 2021-09-02
fix: exclude-base-image-vulns to work with autodetected base image

RotemS:

#### What does this PR do?

Fixes an issue where `--exclude-base-image-vulns` does not work properly when the base image is auto detected, without using `--file`

`snyk container test alpine:3.10` with/without `--exclude-base-image-vulns`

### DIFF
--- a/src/lib/snyk-test/legacy.ts
+++ b/src/lib/snyk-test/legacy.ts
@@ -430,7 +430,11 @@ function convertTestDepGraphResultToLegacy(
       ? undefined
       : options.severityThreshold;
 
-  if (options.docker && options.file && options['exclude-base-image-vulns']) {
+  if (
+    options.docker &&
+    dockerRes?.baseImage &&
+    options['exclude-base-image-vulns']
+  ) {
     vulns = vulns.filter((vuln) => (vuln as DockerIssue).dockerfileInstruction);
   }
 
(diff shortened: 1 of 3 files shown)


agatakrajewska (dismissed):

shaninja:
pushed an amendment to the fake server, that returns the "docker" property on the result, which we now rely on. also changed `dockerRes.baseImage` to `dockerRes?.baseImage`.
intentionally didn't add it as a separate commit, as I believe that no single commit that would go to master should break tests. 

checked that after the mock addition, when reverting back to the original code, the test `has no vulns with flag without a dockerfile` fails.

shaninja:
my updates:
1. stop tests from blowing up by adding optional chaining to `dockerRes.baseImage`
2. added vuln info to fake-server in a manner that would enable us to test excluding base image vulns (PS, I am not very prod in the manner that it's done. welcoming suggestions!)
3. made sure that `has no vulns with flag without a dockerfile` tests failed if code is reverted to the way it was before

note: I added the tests under the windows exclusion. the current --exclude-base-image-vulns test was already there, and those tests do in fact don't pass the windows suite. not entirely sure why it's needed. [dropped a question in the commit that introduced it.](https://github.com/snyk/snyk/commit/2c5b41d6313bba4dc068f60674d3114c8f000d00#r57460386)

thisislawatts:
🧹 Closing as stale.

PR #2220, opened 2021-09-03
chore(ci): use ramdisk

ghost:
CircleCI suggests using RAM disk to make builds more reliable. Physical disks are shared between Docker containers so they can be unreliable when CircleCI is under heavy load.

https://support.circleci.com/hc/en-us/articles/360054908812-Speed-up-steps-using-a-RAM-disk

https://circleci.com/blog/the-issue-of-speed-and-determinism-in-ci/

This doesn't speed up our CI runs but like #2215, it's more about improving reliability.

## Windows

RAM disk is only pre-configured for Docker-based jobs. CircleCI mentions VMs won't encounter disk contention as much as Docker containers so we may not need it for reliability.

However, our Windows runs are really slow so RAM disk might help there. We may be able to use ImDisk. Though, we'll need to see how much time it adds to download, install and configure.

https://community.chocolatey.org/packages/imdisk

## To Do

- [x] Move test outputs to use RAM disk path.
- [x] Try setting up RAM disk on Windows.
  - Tried it. Because CircleCI needs a working directory before we can create a RAM Disk, it causes a mess.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ defaults: &defaults
     package_tests:
       type: boolean
       default: false
-  working_directory: ~/snyk
+  working_directory: /mnt/ramdisk/snyk
 
 commands:
   install_project_dependencies:
@@ -38,11 +38,14 @@ commands:
       npm_version:
         type: string
         default: '7.21.1'
+      npm_cache_directory:
+        type: string
+        default: /mnt/ramdisk/.npm
     steps:
       - restore_cache:
           name: Restoring npm cache
           keys:
-            - npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
+            - npm-cache-v2-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
       - when:
           condition:
             equal: ['win/default', << parameters.os >>]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -25,64 +25,54 @@ defaults: &defaults
...
     steps:
       - restore_cache:
           name: Restoring npm cache
           keys:
-            - npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
+            - npm-cache-v2-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
       - when:
           condition:
-            equal: ['win/default', << parameters.os >>]
+            not: << parameters.npm_global_sudo >>
```
jahed-snyk, line 51: Easier to see intent using a named variable instead of `os`.

.circleci/config.yml
```diff
@@ -25,64 +25,54 @@ defaults: &defaults
     package_tests:
       type: boolean
       default: false
-  working_directory: ~/snyk
+  working_directory: /mnt/ramdisk/snyk
```
jahed-snyk, line 28: Using RAM disk for repo.

.circleci/config.yml
```diff
@@ -25,64 +25,54 @@ defaults: &defaults
...
             run:
               name: Installing npm
               command: sudo npm install -g npm@<< parameters.npm_version >>
       - run:
           name: Configuring npm
-          command: npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
+          command: |
+            npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
+            npm config set cache << parameters.npm_cache_directory >>
+            npm config set prefer-offline true
```
jahed-snyk, line 67: Moving npm cache directory to RAM disk.

.circleci/config.yml
```diff
@@ -25,64 +25,54 @@ defaults: &defaults
...
-            - save_cache:
-                name: Saving npm cache
-                key: npm-cache-v1-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
-                paths:
-                  - ~/.npm
+      - save_cache:
+          name: Saving npm cache
+          key: npm-cache-v2-{{ arch }}-node<< parameters.node_version >>-npm<< parameters.npm_version >>-{{ checksum "package-lock.json" }}
+          paths:
+            - << parameters.npm_cache_directory >>
```
jahed-snyk, line 75: Since this is just a parameter now, we don't need the `when` logic anymore.

.circleci/config.yml
```diff
@@ -280,13 +272,18 @@ jobs:
     docker:
       - image: circleci/node:<< parameters.node_version >>
     resource_class: large
+    environment:
+      TEMP: /mnt/ramdisk/tmp
```
jahed-snyk, line 276: Using RAM disk for temporary fixtures created by tests.

.circleci/config.yml
```diff
@@ -187,13 +177,13 @@ jobs:
       - install_project_dependencies:
           node_version: << parameters.node_version >>
       - attach_workspace:
-          at: ~/snyk
+          at: .
```
jahed-snyk, line 180: Changed various paths to use relative paths from working directory. Makes changing working directory more easier in the future.

.circleci/config.yml
```diff
@@ -372,7 +369,7 @@ jobs:
     <<: *defaults
     docker:
       - image: circleci/node:<< parameters.node_version >>
-    resource_class: small
+    resource_class: large
```
jahed-snyk, line 372: Changed this to large for consistency with every other job. Don't see why not.

maxjeffos (approved):

PR #2223, opened 2021-09-03
fix: skip patching when old-style flag file exists

maxjeffos:
#### What does this PR do?

When `@snyk/protect` runs, we already skip patching a file if there exists a `<filename>.snyk-protect.flag` file adjacent to the target file. This PR enhances this behaviour to also check for the old-style (pre-`@snyk/protect`) flag file, i.e. `<filename>.orig`.


packages/snyk-protect/src/lib/patch.ts
```diff
@@ -8,7 +8,8 @@ export function applyPatchToFile(patchContents: string, baseFolder: string) {
   );
 
   const flagPath = `${targetFilePath}.snyk-protect.flag`;
-  if (fs.existsSync(flagPath)) {
+  const origPatchFlagPath = `${targetFilePath}.orig`;
```
jahed-snyk, line 11: I thought the old flag is in the form of `${targetFilePath}.snyk-<vuln-id>.flag`?

https://github.com/snyk/snyk/blob/75c167addf2795d1e78e7f0c72eda3c8b388ba84/src/lib/protect/write-patch-flag.js#L17

See #2175 for more details.

maxjeffos: That file you mention is also being created.

However, `<filename>.orig` is being created and is a copy of the original, un-patched file: https://github.com/snyk/snyk/blob/master/src/lib/protect/apply-patch.js#L97

I think this would be a more accurate and simpler way of doing things, since it specifies the actual file(s) that were patched, rather than just the intended paths to patch per the .snyk file which, do not match well to the physical files on disk in later npm versions.

jahed-snyk (approved):

packages/snyk-protect/test/acceptance/fix-pr.smoke.spec.ts
```diff
@@ -41,4 +47,29 @@ describe('Fix PR', () => {
...
   });
+
+  it('does not apply patches if `.orig` file from old protect exists', async () => {
+    const pathToProtectBin = path.resolve(__dirname, '../../bin/snyk-protect');
+    const project = await createProject(
+      'single-patchable-module-patched-with-old-protect',
+    );
+    // sleep 2 ms to guarentee that before/after modified times cannot be the
+    // same because the file was written and then modified within a single millisecond.
+    sleep(2);
```
jahed-snyk, line 58: This needs an `await`

PR #2224, opened 2021-09-03
Cleanup some Jest tests

maxjeffos:
#### What does this PR do?

- Uses fake-server for the `test/jest/acceptance/snyk-test/trust-policies.spec.ts` acceptance tests to improve performance.
- use `runSnykCLI` in `test/jest/acceptance/proxy-behavior.spec.ts`
 - pass all env vars into CLI process in acceptance tests to so that we can debug through the process (if we only pass in the PATH (plus whatever other env vars might be required by the tests), it seems like we can't debug locally through the CLI process.
    - `test/jest/system/lib/commands/fix/fix.spec.ts`
    - `test/jest/acceptance/oauth-token.spec.ts`
- move wrongly classified test `test/jest/acceptance/cli-command-types.spec.ts` to a proper home at `test/jest/unit/cli/commands/types.spec.ts`

### DIFF
--- a/test/jest/acceptance/oauth-token.spec.ts
+++ b/test/jest/acceptance/oauth-token.spec.ts
@@ -12,7 +12,7 @@ describe('OAuth Token', () => {
     const apiPath = '/api/v1';
     const apiPort = process.env.PORT || process.env.SNYK_PORT || '12345';
     env = {
-      PATH: process.env.PATH || '',
+      ...process.env,
       SNYK_API: 'http://localhost:' + apiPort + apiPath,
       SNYK_OAUTH_TOKEN: 'oauth-jwt-token',
       SNYK_DISABLE_ANALYTICS: '1',
--- a/test/jest/acceptance/oauth-token.spec.ts
+++ b/test/jest/acceptance/oauth-token.spec.ts
@@ -32,8 +32,7 @@ describe('OAuth Token', () => {
 
   it('uses oauth token for authorised requests when testing projects', async () => {
     const project = await createProjectFromWorkspace('fail-on/no-vulns');
-    const jsonObj = JSON.parse(await project.read('vulns-result.json'));
-    server.setNextResponse(jsonObj);
+    server.depGraphResponse = await project.readJSON('vulns-result.json');
 
     const { code } = await runSnykCLI(`test --json`, {
       cwd: project.path(),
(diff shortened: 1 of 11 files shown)


test/jest/acceptance/oauth-token.spec.ts
```diff
@@ -12,7 +12,7 @@ describe('OAuth Token', () => {
     const apiPath = '/api/v1';
     const apiPort = process.env.PORT || process.env.SNYK_PORT || '12345';
     env = {
-      PATH: process.env.PATH || '',
+      ...process.env,
```
jahed-snyk, line 15: Only assigning PATH (and other specific vars as-needed) might be better to avoid accidental pollution from parent process' env vars. It'll also help to understand exactly which env vars this feature is working with.

test/jest/acceptance/snyk-test/trust-policies.spec.ts
```diff
@@ -1,18 +1,44 @@
...
+  afterAll((done) => {
+    server.close(() => {
+      done();
+    });
+  });
 
   test('`snyk test` detects suggested ignore policies', async () => {
     const project = await createProject('qs-package');
+    const depGraphResStr = await project.read('test-dep-graph-result.json');
+    server.depGraphResponse = JSON.parse(depGraphResStr);
```
jahed-snyk, line 40: I think it's time we introduced `project.readJSON`. We do it in a lot of tests.

test/jest/acceptance/proxy-behavior.spec.ts
```diff
@@ -1,123 +1,86 @@
-import { exec } from 'child_process';
-import { sep } from 'path';
-const main = './bin/snyk'.replace(/\//g, sep);
+import { runSnykCLI } from '../util/runSnykCLI';
 
 const SNYK_API_HTTPS = 'https://snyk.io/api/v1';
 const SNYK_API_HTTP = 'http://snyk.io/api/v1';
 const FAKE_HTTP_PROXY = 'http://localhost:12345';
```
jahed-snyk, line 5: If we change this to `127.0.0.1` instead of `localhost`, it'll make tests easier to read. We won't need the `.split` in our assertions.

test/jest/acceptance/proxy-behavior.spec.ts
```diff
@@ -1,123 +1,86 @@
...
-              'TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"',
-            );
-            done();
-          },
-        );
-      },
-      testTimeout,
-    );
+      // TODO: incorrect behavior when Needle tries to upgrade connection after 301 http->https and the Agent option is set to a strict http/s protocol
+      // See lines with `keepAlive` in request.ts for more details
```
jahed-snyk, line 80: This isn't really a `TODO`, it's more documentation. (It'd be good to avoid `TODO`s anyway as they lack the bigger picture and can mislead people)

The `request.ts` reference is bound to get out of date/sync as test isn't testing it directly. Can we document the necessary details here so people don't need to look around?

ghost (approved):
Approved with some suggestions.

test/jest/acceptance/snyk-test/trust-policies.spec.ts
maxjeffos: I'll do that as another commit in this PR

PR #2229, opened 2021-09-06
chore: introduce jest eslint plugin

ghost:
Installing Jest's eslint plugin to catch typos and usage issues. I also needed to update our TypeScript eslint plugin as it's a peer dependency.

There are over 1000 issues to solve, so for now, I moved the failing rules from error level to warn level. We should gradually fix them as we see them. Ideally we should be in a position where we use `recommended` rules and don't need to discuss and maintain custom overrides.

Follow up from #2222


.eslintrc.json
```diff
@@ -42,13 +41,20 @@
     {
       "files": ["*.ts"],
       "rules": {
+        "@typescript-eslint/ban-types": "warn",
```
jahed-snyk, line 44: This is a new recommended default. I switched it to `warn` as there's too many to fix in one PR.

.eslintrc.json
```diff
@@ -6,8 +6,7 @@
   // },
   "env": {
     "node": true,
-    "es6": true,
-    "jest": true
```
jahed-snyk, line 10: This is enabled under `jest/recommended`'s defaults.

package.json
```diff
@@ -150,15 +150,16 @@
     "@types/restify": "^8.4.2",
     "@types/sarif": "^2.1.2",
     "@types/sinon": "^7.5.0",
-    "@typescript-eslint/eslint-plugin": "2.18.0",
-    "@typescript-eslint/parser": "^2.0.0",
+    "@typescript-eslint/eslint-plugin": "^4.30.0",
+    "@typescript-eslint/parser": "^4.30.0",
```
jahed-snyk, line 154: Had to upgrade these to use the latest ESLint Jest plugin.

packages/cli-alert/src/index.ts
```diff
@@ -32,7 +32,6 @@ async function discoverConsecutiveFailures(
     await octokit.actions.listJobsForWorkflowRun({
       owner: 'snyk',
       repo: 'snyk',
-      // eslint-disable-next-line @typescript-eslint/camelcase
```
jahed-snyk, line 35: This rule no longer exists and its replacement (`naming-convention`) has new defaults that don't complain about these lines.

packages/snyk-fix/test/unit/fix.spec.ts
```diff
@@ -184,7 +184,6 @@ describe('Snyk fix', () => {
       'prod.txt',
       'django===1.6.1',
     );
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
jahed-snyk, line 187: This rule is now just `@typescript-eslint/ban-ts`. Also, these explicit disables aren't needed as we don't fail on this. Guessing they're leftovers from when we did fail on them.

src/cli/commands/protect/prompts.ts
```diff
@@ -487,10 +487,6 @@ function canBeUpgraded(vuln) {
   });
 }
 
-interface IgnoreMeta {
-  review: unknown;
-}
```
jahed-snyk, line 492: Unused type.

src/lib/plugins/sast/analysis.ts
```diff
@@ -56,7 +59,7 @@ async function getCodeAnalysis(root: string, options: Options): Promise<Log> {
     fileOptions: { paths: [root] },
   });
 
-  return result?.analysisResults.sarif!;
+  return result?.analysisResults.sarif || null;
```
jahed-snyk, line 62: Everything that uses this function expects it can be null so no point in hiding that here. (new rules rightly complain about using `?` with `!`).

test/jest/acceptance/cli-args.spec.ts
```diff
@@ -112,7 +112,7 @@ describe('cli args', () => {
     expect(code).toEqual(2);
   });
 
-  test('test multiple paths with --project-name=NAME', async () => {
+  test('snyk test multiple paths with --project-name=NAME', async () => {
```
jahed-snyk, line 115: Jest rules don't like a test starting with `test`. The `test` here is referring to `snyk test` so worth being explicit.

test/jest/unit/lib/formatters/sarif-output.spec.ts
```diff
@@ -6,13 +6,6 @@ describe('createSarifOutputForContainers', () => {
   it('general', () => {
     const testFile = getTestResult();
     const sarif = createSarifOutputForContainers([testFile]);
-    expect(sarif.runs).toHaveLength(1);
-
-    const [run] = sarif.runs;
-    expect(run.tool.driver.name).toEqual('Snyk Container');
-    expect(run.tool.driver.rules).toHaveLength(1);
-    expect(run.results).toHaveLength(1);
-    expect(run.results?.[0].level === 'error');
```
jahed-snyk, line 15: This test is snapshot testing anyway so we don't need these individual assertions.

Jest rules rightly caught the `expect(run.results?.[0].level === 'error')` which isn't actually making an assertion and isn't even correct ('error' level is under `defaultConfiguration`).

.eslintrc.json
```diff
@@ -42,13 +41,20 @@
...
     {
       "files": "*.spec.*",
+      "extends": ["plugin:jest/recommended"],
       "rules": {
-        "@typescript-eslint/ban-ts-ignore": "off"
+        "jest/no-done-callback": "warn",
+        "jest/valid-title": "warn",
+        "jest/no-conditional-expect": "warn",
+        "jest/no-try-expect": "warn",
+        "jest/no-identical-title": "warn",
```
jahed-snyk, line 56: Too many to fix in one PR so I've switched them to warn so we can fix them as we go.

.eslintrc.json
```diff
@@ -42,13 +41,20 @@
...
+        "@typescript-eslint/ban-ts-comment": "warn"
```
jahed-snyk, line 57: This used to be "off" but it should be "warn" at least since ignoring type errors makes our codebase less safe.

.eslintrc.json
```diff
@@ -42,13 +41,20 @@
...
       "files": ["*.ts"],
       "rules": {
+        "@typescript-eslint/ban-types": "warn",
         "id-blacklist": ["error", "exports"] // in TS, use "export" instead of Node's "module.exports"
       }
     },
     {
       "files": "*.spec.*",
+      "extends": ["plugin:jest/recommended"],
       "rules": {
```
JackuB, line 51: We might want to add `jest/expect-expect` to the list. It got some false positives

Screen Shot 2021-09-07 at 13 24 43

lili2311 (approved):

.eslintrc.json
jahed-snyk: I'll take a look. The error message here isn't great but the flagging of such lines, I kind of agree with.

I think for tests, we should always mention `snyk test` with a `snyk` prefix to give context. In this case, it should be aliased to `snykTest`. Using the word `test` inside tests to test snyk test is really confusing. Also there's a risk of conflicting with the global `test` introduced by Jest.

teodora-sandu (approved):

src/lib/plugins/sast/analysis.ts
pkey: confirmed that `null` is already handled elsewhere. Thank for the improvements @jahed-snyk!

pkey (approved):

maxjeffos (approved):

PR #2231, opened 2021-09-07
feat(@snyk/fix): send more detailed error analytics

lili2311:

#### What does this PR do?
When projects fail we do not surface the full error in the summary,
it is shown to the user in the warning output.
make sure we send the error in analytics to help debug & improve the
plugin.

https://github.com/snyk/snyk/compare/feat/snyk-fix-send-error-analytics?expand=1#diff-496e8bb7be43118118a857255a018b53e916cd7611050049ec55a6fba8a39d7aR70

run `snyk fix` on a project that fails & observe analytics

### DIFF
--- a/src/cli/commands/fix/index.ts
+++ b/src/cli/commands/fix/index.ts
@@ -45,21 +45,21 @@ export default async function fix(...args: MethodArgs): Promise<string> {
     (res) => Object.keys(res.testResult.issues).length,
   );
   const { dryRun, quiet } = options;
-  const { fixSummary, meta } = await snykFix.fix(results, { dryRun, quiet });
+  const { fixSummary, meta, results: resultsByPlugin } = await snykFix.fix(
+    results,
+    {
+      dryRun,
+      quiet,
+    },
+  );
 
-  // Analytics # of projects
-  analytics.add('snykFixFailedProjects', meta.failed);
-  analytics.add('snykFixFixedProjects', meta.fixed);
-  analytics.add('snykFixTotalProjects', results.length);
-  analytics.add('snykFixVulnerableProjects', vulnerableResults.length);
-
-  // Analytics # of issues
-  analytics.add('snykFixFixableIssues', meta.fixableIssues);
-  analytics.add('snykFixFixedIssues', meta.fixedIssues);
-  analytics.add('snykFixTotalIssues', meta.totalIssues);
-
-  analytics.add('snykFixSummary', fixSummary);
-
+  setSnykFixAnalytics(
+    fixSummary,
+    meta,
+    results,
+    resultsByPlugin,
+    vulnerableResults,
+  );
   // `snyk test` did not return any test results
   if (results.length === 0) {
     throw new Error(fixSummary);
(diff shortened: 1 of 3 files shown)


ghost (approved):

src/cli/commands/fix/index.ts
```diff
@@ -60,6 +66,16 @@ export default async function fix(...args: MethodArgs): Promise<string> {
 
   analytics.add('snykFixSummary', fixSummary);
 
+  // Analytics for errors
+  const errors: string[] = [];
+  for (const plugin of Object.keys(resultsByPlugin)) {
+    const failedToFix = resultsByPlugin[plugin].failed;
```
gitphill, line 72:
```typescript
for (const { failed } of resultsByPlugin) {
  if (failed.length > 0) {
    errors.push(...failed.map((f) => f.error?.message));
  }
}
```

is more concise, but do we need some indication of what plugin each error relates to? instead of a flat array of strings

gitphill (approved):

src/cli/commands/fix/index.ts
lili2311: hmm yeah that will probably be useful, can we send objects in analytics?

lili2311: that might not be very readable, perhaps `${plugin}: ${Error}` as a string array

src/cli/commands/fix/index.ts
```diff
@@ -60,6 +66,16 @@ export default async function fix(...args: MethodArgs): Promise<string> {
 
   analytics.add('snykFixSummary', fixSummary);
 
+  // Analytics for errors
+  const errors: string[] = [];
```
jan-stehlik, line 70: nitpick: i wonder if this could be extracted into a separate function, this one has grown quite a lot

test/jest/unit/lib/commands/fix/fix.spec.ts
```diff
@@ -294,6 +298,93 @@ describe('snyk fix (functional tests)', () => {
...
+          _doubleDashArgs: [],
+          _: [],
+        });
+      } catch (e) {
+        res = e.message;
+      }
+      expect(snykFixSpy.mock.calls[0][1]).toEqual({
+        dryRun: true,
+        quiet: true,
+      });
```
jan-stehlik, line 359: i find this a little more readable

```suggestion
      expect(snykFixSpy).toHaveBeenCalledWith({
        dryRun: true,
        quiet: true,
      });
```

jan-stehlik (approved):
nice one @lili2311 💯

src/cli/commands/fix/index.ts
jahed-snyk: `[{ plugin, message }]` maybe? It's useful to split plugin into its own field so we can query by plugin if needed. otherwise we'll need to use string functions which can get messy.

AFAIK yes we support arbritrary JSON objects (since it used in a few places).

test/jest/unit/lib/commands/fix/fix.spec.ts
lili2311: unfortunately if I do that it expects me to pass in the entirety of the options sent and it's a large array of object from CLI

src/cli/commands/fix/index.ts
lili2311: Ive gone with 
```
snykFixErrors', {
        python: ['SolverProblemError Because x (2.6) depends on y (>=1.19)'],
        npm: []
        ....
      }
```

aarlaud (approved):

PR #2233, opened 2021-09-07
feat: exclude files/folders

Arvi3d:

#### What does this PR do?

Bumped @snyk/code-client dependency version to 4.2.1, that provides support for new feature of excluding folders and files

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "@open-policy-agent/opa-wasm": "^1.2.0",
     "@snyk/cli-interface": "2.11.0",
     "@snyk/cloud-config-parser": "^1.10.2",
-    "@snyk/code-client": "4.0.0",
+    "@snyk/code-client": "^4.2.2",
     "@snyk/dep-graph": "^1.27.1",
     "@snyk/fix": "file:packages/snyk-fix",
     "@snyk/gemfile": "1.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@open-policy-agent/opa-wasm": "^1.2.0",
         "@snyk/cli-interface": "2.11.0",
         "@snyk/cloud-config-parser": "^1.10.2",
-        "@snyk/code-client": "4.0.0",
+        "@snyk/code-client": "^4.2.2",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "@open-policy-agent/opa-wasm": "^1.2.0",
     "@snyk/cli-interface": "2.11.0",
     "@snyk/cloud-config-parser": "^1.10.2",
-    "@snyk/code-client": "4.0.0",
+    "@snyk/code-client": "4.2.1",
```
JackuB, line 76: This needs to be updated in the package-lock as well

I’m surprised the build worked 🤔

JackuB (changes requested):

package.json
Arvi3d: done, updated lock file as well

JackuB:
Screen Shot 2021-09-07 at 20 10 42

Somehow all "resolved" and "integrity" fields were removed by this change? Could you try it again? I think npm version could be the cause here - could you use npm@7?

package.json
jahed-snyk: If you rollback and just run (on npm 7) `npm install --save @snyk/code-client@4.2.1` you should get a proper minimal package-lock change. Should be similar to:

```
$ git diff --stat package-lock.json
 package-lock.json | 108 ++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------------
 1 file changed, 48 insertions(+), 60 deletions(-)
```

Arvi3d: done

JackuB:
Screen Shot 2021-09-08 at 10 59 06

Seems like this caused some regression in the CLI. Maybe we should improve the coverage in this repo as well

Arvi3d:
all seems fine now. please review

Arvi3d:
> Screen Shot 2021-09-08 at 10 59 06
> 
> Seems like this caused some regression in the CLI. Maybe we should improve the coverage in this repo as well

fixed, please review

package.json
Arvi3d: please review. all fixed now

JackuB (approved):

PR #2236, opened 2021-09-08
refactor: npm/standalone are not integration names

ghost:
We never use `NPM` and `STANDALONE` as integration names so it's misleading. Also they are more like distribution methods as we can still have either of them depending on how the integration installs Snyk CLI. We have the `standalone` key in analytics for this, though right now we're not forwarding that data to storage (to be fixed in another PR).

This PR also fixes a bunch of ESLint and DangerJS warnings for related files.

### DIFF
--- a/src/lib/analytics/sources.ts
+++ b/src/lib/analytics/sources.ts
@@ -8,12 +8,14 @@
   Integration name is validated with a list
 */
 
-const debug = require('debug')('snyk');
+import { exec } from 'child_process';
+import * as createDebug from 'debug';
 import * as fs from 'fs';
+import { join } from 'path';
 import * as semver from 'semver';
 import { ArgsOptions } from '../../cli/args';
-import { join } from 'path';
-const { exec } = require('child_process');
+
+const debug = createDebug('snyk');
 
 export const INTEGRATION_NAME_ENVVAR = 'SNYK_INTEGRATION_NAME';
 export const INTEGRATION_VERSION_ENVVAR = 'SNYK_INTEGRATION_VERSION';
--- a/src/lib/analytics/sources.ts
+++ b/src/lib/analytics/sources.ts
@@ -22,10 +22,6 @@ export const INTEGRATION_ENVIRONMENT_VERSION_ENVVAR =
   'SNYK_INTEGRATION_ENVIRONMENT_VERSION';
 
 enum TrackedIntegration {
-  // Distribution builds/packages
-  NPM = 'NPM', // registry - how can we detect this one? Not standalone?
-  STANDALONE = 'STANDALONE', // Tracked by detecting "isStandalone" flag
-
   // tracked by passing envvar on CLI invocation
   HOMEBREW = 'HOMEBREW',
   SCOOP = 'SCOOP',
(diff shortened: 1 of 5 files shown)


test/jest/unit/query-strings.spec.ts
```diff
@@ -1,45 +1,47 @@
...
+      ['docker', expect.any(String)],
+    ]);
   });
 
   it('uses integration name and version', () => {
-    process.env.SNYK_INTEGRATION_NAME = 'NPM';
-    process.env.SNYK_INTEGRATION_VERSION = '1.2.3';
-
-    // Object.fromEntries is not available for Node 8 and 10. Testing this in latest Node is enough
-    if (Object.fromEntries) {
```
jahed-snyk, line 29: Jest doesn't like conditional assertions and we can avoid the conditional here anyway.

Query parameters can have multiple entries for the same key, so while converting an object to query params is valid (one to many), doing it the other way around loses information and makes our tests wrong.

```js
Object.fromEntries([['a', 1], ['a', 2]])
// Object { a: 2 }
```

Ideally, these tests should be broken down more so we don't need to do this weird conversion in the first place but I didn't want to get too deep into it. Small steps...

test/jest/unit/analytics-sources.spec.ts
```diff
@@ -188,7 +187,7 @@ describe('sources - getIntegrationEnvironment', () => {
   });
 });
 
-describe('sources - getIntegrationEnvironment', () => {
```
jahed-snyk, line 191: Duplicate and wrong test name. I changed the other test names to be similar and less repetitive.

src/lib/analytics/sources.ts
```diff
@@ -22,10 +24,6 @@ export const INTEGRATION_ENVIRONMENT_VERSION_ENVVAR =
   'SNYK_INTEGRATION_ENVIRONMENT_VERSION';
 
 enum TrackedIntegration {
-  // Distribution builds/packages
-  NPM = 'NPM', // registry - how can we detect this one? Not standalone?
-  STANDALONE = 'STANDALONE', // Tracked by detecting "isStandalone" flag
```
jahed-snyk, line 27: This is the main change. (See PR opening post)

test/jest/unit/query-strings.spec.ts
```diff
@@ -1,45 +1,47 @@
...
-        utm_medium: 'cli',
-        utm_campaign: 'NPM',
-        utm_campaign_content: '1.2.3',
-        os: expect.any(String),
-        docker: expect.any(String),
-      });
-    }
-
-    delete process.env.SNYK_INTEGRATION_NAME;
-    delete process.env.SNYK_INTEGRATION_VERSION;
```
jahed-snyk, line 43: Avoiding environment pollution. These are now passed in as args which makes more sense as we want to make sure args are forwarded. The handling of global env var fallbacks is tested in other modules.

tsconfig.settings.json
```diff
@@ -5,7 +5,7 @@
     "pretty": true,
     "moduleResolution": "node",
     "target": "es2018",
-    "lib": ["ES2018", "ES2019", "dom"],
+    "lib": ["ES2018", "ES2019", "DOM", "DOM.Iterable"],
```
jahed-snyk, line 8: `DOM.Iterable` types are needed for various iterable methods like `URLSearchParams.entries`.

Avishagp (approved):

PR #2240, opened 2021-09-09
refactor: extract re-usable code from poetry fix

lili2311:

#### What does this PR do?
In preparation for introducing a different sequential fix strategy
refactor the existing code into re-usable blocks needed to add sequential
fix. 
Added some extra optional fields that can be sent in changes so we have all the data we need in case we want to output is as --json or consume it another way

packages/snyk-fix/src/plugins/python/handlers/poetry/update-dependencies/index.ts

### DIFF
--- a/packages/snyk-fix/src/plugins/python/handlers/poetry/update-dependencies/index.ts
+++ b/packages/snyk-fix/src/plugins/python/handlers/poetry/update-dependencies/index.ts
@@ -6,8 +6,8 @@ import * as poetryFix from '@snyk/fix-poetry';
 
 import { PluginFixResponse } from '../../../../types';
 import {
-  DependencyPins,
   EntityToFix,
+  FixChangesError,
   FixChangesSummary,
   FixOptions,
 } from '../../../../../types';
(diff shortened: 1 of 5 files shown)


packages/snyk-fix/src/plugins/python/handlers/poetry/update-dependencies/index.ts
```diff
@@ -128,56 +111,82 @@ function throwPoetryError(stderr: string, command?: string) {
...
+      changes,
+    });
   } catch (error) {
     debug(
       `Failed to fix ${entity.scanResult.identity.targetFile}.\nERROR: ${error}`,
     );
     handlerResult.failed.push({
       original: entity,
       error,
-      tip: poetryCommand ? `Try running \`${poetryCommand}\`` : undefined,
```
aarlaud, line 179: we just nuke that tip thing? I'm not seeing it moved elsewhere

aarlaud: nevermind ;)

aarlaud (approved):

packages/snyk-fix/src/plugins/python/handlers/poetry/update-dependencies/index.ts
lili2311: good catch, we do want to keep it. Let me see what `tip` makes sense here as we lose the context of the command we ran.

lili2311: added the tip via the Error thrown

PR #2244, opened 2021-09-10
fix: exit code when IaC scans are empty

craigfurman:

#### What does this PR do?

**fix: exit code when IaC scans are empty**

According to
https://support.snyk.io/hc/en-us/articles/360003812578#UUID-c88e66cf-431c-9ab1-d388-a8f82991c6e0,
the Snyk CLI should return 3 when there were no supported files in a
scan. This allows user to distinguish this situation from a complete
scan, and from when errors encountered during the scan.

We were actually returning 2, which is also returned when errors are
encountered during a scan.

See self-review comments.

Customer issue: https://snyk.zendesk.com/agent/tickets/13733

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -30,6 +30,7 @@ import {
   CustomError,
   NoSupportedSastFiles,
 } from '../lib/errors';
+import { IaCErrorCodes } from './commands/test/iac-local-execution/types';
 import stripAnsi from 'strip-ansi';
 import { ExcludeFlagInvalidInputError } from '../lib/errors/exclude-flag-invalid-input';
 import { modeValidation } from './modes';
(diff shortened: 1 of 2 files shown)


src/cli/index.ts
```diff
@@ -96,8 +97,9 @@ async function handleError(args, error) {
     'Could not detect supported target files in',
   );
   const noSupportedSastFiles = error instanceof NoSupportedSastFiles;
+  const noSupportedIaCFiles = error.code === IaCErrorCodes.NoFilesToScanError;
```
craigfurman, line 100: I initially started by mimicking the SAST case: `error instanceof NoFilesToScanError`, but this didn't do anything! This might just be me being inexperienced in node, but I started to suspect some problem of type erasure at runtime. I didn't validate whether or not the SAST case works, I might ask that team to start with.

craigfurman:
I pushed a smoke/ branch, having changed the smoke tests.

src/cli/index.ts
craigfurman: The SAST case indeed exits with 3 apparently. I don't know why I can't seem to get the `instanceof` check working, but tbh if we're going to carry around this `error.code` metadata, which kind of acts as a shadow type, we may as well use it in logic.

Avishagp (approved):

src/cli/index.ts
teodora-sandu: I think maybe the `error` we're sending is post-formatted so that's why it doesn't work. But I don't see anything wrong with this implementation anyway

teodora-sandu (approved):
Thanks for fixing this! 🥇

PR #2245, opened 2021-09-12
feat: support analytics from plugins

anthogez:
We are introducing the capability to allow the cli of receive analytics
data from plugins. In this initial interaction it's being added only for
async tests for the new unmanaged product line (c/c++ support) during
the `test` phase. See snyk-cpp-plugin pr https://github.com/snyk/snyk-cpp-plugin/pull/35


Avishagp (approved):
lgtm

PR #2247, opened 2021-09-12
feat: replace source cmd with unmanaged

anthogez:
BREAKING CHANGE: replacing `snyk source test` and `snyk source monitor`
commands with `snyk unmanaged test` and `snyk unmanaged monitor`

### DIFF
--- a/test/jest/unit/all-projects-from-plugins.spec.ts
+++ b/test/jest/unit/all-projects-from-plugins.spec.ts
@@ -35,33 +35,13 @@ describe('Detect extra projects available that could be tested using --all-proje
     expect(actualResult).toBe(expectedResult);
   });
 
-  it('should return `undefined` when `source` command for cpp is being used', async () => {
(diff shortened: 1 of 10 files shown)


Avishagp (commented):
No one's using `snyk source test` according to our dashboard, so lgtm

Avishagp (approved):

test/jest/unit/all-projects-from-plugins.spec.ts
```diff
@@ -35,33 +35,13 @@ describe('Detect extra projects available that could be tested using --all-proje
...
-    const actualResult = await getExtraProjectCount(
-      root,
-      options,
-      inspectResult,
-    );
-    const expectedResult = undefined;
-    expect(actualResult).toBe(expectedResult);
-  });
-
-  it('should return `undefined` when `docker` command is being used', async () => {
```
anthogez, line 54: unnecessary test as already exists one in line 38, and on top of that it was not testing a `docker` case but `source` again

PR #2250, opened 2021-09-13
fix: error to handle when no fixes applied

lili2311:

#### What does this PR do?
Ensure we separately throw when nothing was changed vs when
100% of fixes failed. And send appropriate message.

Revert a change that was outputting the whole pipenv stderr and only
show a short summary for those caught errors

### DIFF
--- a/packages/snyk-fix/src/plugins/python/handlers/pipenv-pipfile/update-dependencies/index.ts
+++ b/packages/snyk-fix/src/plugins/python/handlers/pipenv-pipfile/update-dependencies/index.ts
@@ -5,7 +5,6 @@ import * as pipenvPipfileFix from '@snyk/fix-pipenv-pipfile';
 import { PluginFixResponse } from '../../../../types';
 import {
   EntityToFix,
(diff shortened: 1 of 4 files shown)


packages/snyk-fix/src/plugins/python/handlers/pipenv-pipfile/update-dependencies/index.ts
```diff
@@ -52,14 +51,14 @@ function throwPipenvError(stderr: string, command?: string) {
...
+    incompatibleDeps,
     lockingFailed,
     versionNotFound,
-    incompatibleDeps,
   ].map((e) => e.toLocaleLowerCase());
 
   for (const error of errorsToBubbleUp) {
     if (errorStr.includes(error)) {
-      throw new CommandFailedError(stderr, command);
+      throw new CommandFailedError(error, command);
```
lili2311, line 61: don't print whole `stderr` is super long and unreadable

aarlaud (approved):

PR #2252, opened 2021-09-13
feat: sequential fix

lili2311:

#### What does this PR do?

Support optionally upgrading 1 package at a time, this
is much slower but can provide better success in upgrading
vulnerable package versions
Some combination of package upgraded might not be compatible, allow upgrading 1 at a time to increase the % of fixed issues. This is much slower 🐌 

`snyk fix --sequential` will trigger the behavior to try applying package upgrades via `poetry` and `pipenv` 1 by 1.

#### Screenshots
Can't see a difference in the screenshot unless one of the upgrades fails. See the added test to show what to expect.

Actual project run
CleanShot 2021-09-14 at 13 32 06@2x

From a test that crafts the right scenario, when running with `--sequential-fix` the whole fix does not fail.
CleanShot 2021-09-14 at 13 34 45@2x

### DIFF
--- a/packages/snyk-fix/src/plugins/python/handlers/pipenv-pipfile/update-dependencies/index.ts
+++ b/packages/snyk-fix/src/plugins/python/handlers/pipenv-pipfile/update-dependencies/index.ts
@@ -7,7 +7,7 @@ import {
   FixOptions,
 } from '../../../../../types';
 
-import { ensureHasUpdates } from '../../ensure-has-updates';
+import { failIfNoUpdatesApplied } from '../../fail-if-no-updates-applied';
 import { NoFixesCouldBeAppliedError } from '../../../../../lib/errors/no-fixes-applied';
 import { generateUpgrades } from './generate-upgrades';
 import { pipenvAdd } from './pipenv-add';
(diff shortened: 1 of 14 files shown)


packages/snyk-fix/src/plugins/python/handlers/pipenv-pipfile/update-dependencies/index.ts
```diff
@@ -46,14 +31,14 @@ async function fixAll(
...
   };
-  const { upgrades } = await generateUpgrades(entity);
-  if (!upgrades.length) {
-    throw new NoFixesCouldBeAppliedError(
-      'Failed to calculate package updates to apply',
-    );
-  }
   const changes: FixChangesSummary[] = [];
   try {
+    const { upgrades } = await generateUpgrades(entity);
```
lili2311, line 36: move back into try/catch to not prevent further fixes & to let the spinner finish

aarlaud (approved):

jan-stehlik (approved):
LGTM, nice one @lili2311 :raised_hands:

PR #2254, opened 2021-09-14
fix: standardize poetry package names before comparing

lili2311:

#### What does this PR do?
When generating upgrades make sure to standardize both package
names and the remediation package names before comparing
and choosing to do the upgrade for prod / dev dep.


packages/snyk-fix/test/unit/plugins/python/handlers/poetry/generate-upgrades.spec.ts
```diff
@@ -105,7 +105,7 @@ describe('generateUpgrades', () => {
 
     [tool.poetry.dependencies]
     python = "*"
-    django = "1.1.1"
+    Django = "1.1.1"
```
lili2311, line 108: this should still match `django` in the pins recommendation coming back from back end

aarlaud (approved):

PR #2255, opened 2021-09-14
feat: support print-deps cmd

anthogez:
Support `--print-dep-paths` and `print-deps` args

To allow the user of have different kind of UI outputs

`--print-dep-paths`: includes the  dependencies that have been identified
and the each dependency filePaths in the test result output

`--print-deps`: includes the dependencies that have been identified in
the test result output

Also introducing depsFilePaths type for `unmanaged c/c++` product line

See https://github.com/snyk/snyk-cpp-plugin/pull/37

Screen Shot 2021-09-14 at 21 19 55

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -69,6 +69,7 @@ export interface Options {
   severityThreshold?: SEVERITY;
   dev?: boolean;
   'print-deps'?: boolean;
+  'print-dep-paths'?: boolean;
   'remote-repo-url'?: string;
   scanAllUnmanaged?: boolean;
   allProjects?: boolean;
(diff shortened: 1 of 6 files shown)


anthogez:
Will be merged once this one  https://github.com/snyk/snyk-cpp-plugin/pull/37  is released

src/lib/types.ts
```diff
@@ -69,6 +69,7 @@ export interface Options {
   severityThreshold?: SEVERITY;
   dev?: boolean;
   'print-deps'?: boolean;
+  'print-dep-paths'?: boolean;
```
JackuB, line 72: Should these be in the `--help` as well?

JackuB (approved):

src/lib/types.ts
anthogez: no yet as we are still in beta, thanks for raising it!

PR #2260, opened 2021-09-16
feat: replace YAML parser with library call [CC-1012]

teodora-sandu:
#### What does this PR do?
This PR replaces the `yaml` library with a call to `@snyk/cloud-config-parser`, where we have moved our customised YAML parsing logic. This way when we have parsing issues that we need to fix, we can fix them in just one place.

Run:
1. `npm run build`
2. ` snyk-dev iac test ./test/fixtures/iac/kubernetes/pod-privileged.yaml`
3. `snyk-dev iac test ./test/fixtures/iac/kubernetes/pod-valid.json`

https://snyksec.atlassian.net/browse/CC-1012

#### Screenshots

Screenshot 2021-09-16 at 16 45 45
Screenshot 2021-09-16 at 16 45 49

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "dependencies": {
         "@open-policy-agent/opa-wasm": "^1.2.0",
         "@snyk/cli-interface": "2.11.0",
-        "@snyk/cloud-config-parser": "^1.10.2",
+        "@snyk/cloud-config-parser": "^1.11.0",
         "@snyk/code-client": "^4.2.3",
         "@snyk/dep-graph": "^1.27.1",
         "@snyk/fix": "file:packages/snyk-fix",
@@ -80,8 +80,7 @@
         "tar": "^6.1.2",
         "tempy": "^1.0.1",
         "uuid": "^8.3.2",
-        "wrap-ansi": "^5.1.0",
-        "yaml": "^1.10.2"
+        "wrap-ansi": "^5.1.0"
       },
       "bin": {
         "snyk": "bin/snyk"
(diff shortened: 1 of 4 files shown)


craigfurman (approved):
Nice! Reviewer note: I haven't read the cloud config parser code again, I'm relying on our tests in both of these repos catching any regressions.

That is a really massive diff to the package-lock though, are we sure that's done the right thing?

teodora-sandu:
Good shout on the package-lock.json. I followed the instructions in this thread to fix it: https://snyk.slack.com/archives/C1NREQSG0/p1631814339032800

anthogez (approved):
maybe remove the `^` symbol from `"@snyk/cloud-config-parser": "^1.11.0"` to avoid breaking changes unless it's a desired behavior from your team

jan-stehlik (approved):
nice one @teodora-sandu and amazing PR description :fire:

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "snyk",
```
jahed-snyk, line 8: I've seen this line come and go a lot. I wonder why it happens... version differences? 🤔 Which version of npm are you using? Might be worth putting this line back in to avoid merge conflicts (in this or future PRs).

ghost (approved):

package-lock.json
teodora-sandu: I'm using:
```
❯ node -v
v16.6.1
❯ npm -v
7.20.3
```

I'll consider adding it back but it feels weird to have to manually change the package-lock. Whenever I get conflicts in this file I just delete the lock file and re-generate it like it says here: https://mannes.tech/merge-lock-file/. Is this bad practice?

jahed-snyk: Manually changing the lockfile is definitely bad practice 😬. I've looked into the issue, looks like if you upgrade your npm version to latest, it should fix that line.

Is deleting and regenerated the lockfile bad practice? IMO, Yes. The lockfile is meant to "lock" dependencies in place so that everyone's using the same deps to develop on. Deleting lock files and regenerating them can introduce unrelated changes which makes it difficult to review and can make dev environments go out of sync. In this PR's case, that didn't happen, but it can depending on the freshness of `node_modules` (as seen in past PRs).

I guess deleting is fine for now, until we start encountering those problematic issues more often. We have a bunch of issues with our current lock file setup so hopefully we can iron things out over time 👍

PR #2265, opened 2021-09-21
test: migrate fake-server to express

ghost:
# Why

It seems Restify is unable to correctly process the JSON payloads we send, at least going through `snyk monitor`. This blocks tests for #2264.

The issue seems to be a race condition between Restify’s bodyReader (which assigns req.body) and jsonBodyParser (which converts req.body to a js object). jsonBodyParser receives an empty req.body and fails to parse it. However, bodyReader clearly assigns the correct req.body — eventually. No idea if/why/how the other test endpoints are working. They look identical.

Restify is pretty inactive and we get deprecation warnings for it so instead of digging further, I'll try migrating us to Express and see if that fixes it. We want to use Express anyway as it's a lot more popular so has a familiar API.

I can confirm that these changes fixes the body parsing issues I had for #2264.

# Changes

The main change is in `fake-server.ts`. I've migrated `restify` to `express`.

Due to how we previously assigned additional variables and functions on the `server` object, it was difficult to make the code in `fake-server` typesafe. So I've changed a few things to make that possible and avoid the inheritance we previously relied upon for `FakeServer` which coupled us to Restify.

- Variables like `server.requests` and `server.depGraphResponse` are no longer directly exposed outside the closure. You'll need to use the new setter methods.
  - We should only expose functions so that future migrations are easier.
- Created wrappers around `server.listen` and `server.close` to avoid changing heavily used public APIs.
  - In the future, we can change these wrappers to use promises and get rid of the callbacks in our test hooks.

Restify cannot be removed yet. We need to migrate tests using the old `cli-server.js` to `fake-server.ts`. This isn't trivial, it needs some authorisation logic migrated over too. I did remove `@types/restify` as `cli-server.js` isn't TypeScript.

I've also fixed a test in `fix.spec.ts` which was relying on fake-server's previously broken behaviour. Now fake-server is setup in the test to explicitly return a 500. While I was there I migrated all of the tests in that file to use async/await instead of callbacks. I also adding a positive test case so that it's easier to see what success looks like.

## Side Fixes

I've fixed the `dangerfile` to properly list all known problematic files at once. Previously it was only returning the first file. I've also excluded `.js` files as those can't use import/export syntax (which includes the dangerfile itself).

Related, I've fixed the import/export issues with the files DangerJS flagged.


ipapast (approved):
Get Port function for CloudConfig is fine according to the proposed changes ✅

maxjeffos (approved):
Looks great

test/acceptance/fake-server.ts
```diff
@@ -1,64 +1,91 @@
...
-  server.popRequest = () => {
-    return server.requests.pop()!;
+  const getRequests = () => {
+    return requests;
   };
-  server.popRequests = (num: number) => {
-    return server.requests.splice(server.requests.length - num, num);
+
+  const popRequest = () => {
+    return requests.pop()!;
```
maxjeffos, line 42: What's the `!` for?

maxjeffos (approved):

test/acceptance/fake-server.ts
maxjeffos: 👍

PR #2270, opened 2021-09-23
wip: support Yarn workspaces in `--all-projects`

lili2311:

#### What does this PR do?

Spike to show how we can start scanning yarn workspaces behind --all-projects flag as well.

### DIFF
--- a/src/lib/plugins/get-multi-plugin-result.ts
+++ b/src/lib/plugins/get-multi-plugin-result.ts
@@ -14,6 +14,7 @@ import { convertMultiResultToMultiCustom } from './convert-multi-plugin-res-to-m
 import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
 import { CallGraph } from '@snyk/cli-interface/legacy/common';
 import { FailedToRunTestError } from '../errors';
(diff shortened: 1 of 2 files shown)


src/lib/plugins/get-multi-plugin-result.ts
```diff
@@ -32,6 +33,7 @@ export interface MultiProjectResultCustom
   extends cliInterface.legacyPlugin.MultiProjectResult {
   scannedProjects: ScannedProjectCustom[];
   failedResults?: FailedProjectScanError[];
+  processedFiles?: string[];
```
lili2311, line 36: need to refactor to the workspaces handling to a function so there is not a need to change this inyterface

src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
```diff
@@ -65,6 +65,7 @@ export async function processYarnWorkspaces(
       runtime: process.version,
     },
     scannedProjects: [],
+    processedFiles: [],
```
lili2311, line 68: needs refactoring so we return the plugin response separately to process projects as to not add new props to the interface

PR #2279, opened 2021-09-26
chore: allow all code owners to update package.json

admons:

#### What does this PR do?
Gives the ability to all teams that are code owners of this repo to update the package.json file to release a new version.


JackuB (changes requested):
This file is not controlling only versions, but many other properties we are responsible for as well. I'd prefer to keep this line like it is now

PR #2280, opened 2021-09-26
chore: bump gradle plugin version

Avishagp:

#### What does this PR do?
Bumps version of `snyk-gradle-plugin` to 3.16.2 to include this fix (presented through `java-call-graph-builder`): https://github.com/snyk/java-call-graph-builder/pull/57

Use temporary file copied to the local filesystem to ensure the separate process that reads init.gradle can find it in binary releases.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "snyk-cpp-plugin": "2.9.0",
         "snyk-docker-plugin": "4.24.0",
         "snyk-go-plugin": "1.18.0",
-        "snyk-gradle-plugin": "3.16.1",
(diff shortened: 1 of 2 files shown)


anthogez (commented):
That’s a “fix: gradle java-call-graph binary” not a chore as it’s introducing changes in the binary releases. Which means that binary users needs to be aware of those changes.

anthogez (commented):
Also could you include the gh message into the commit’s message body? Thanks!

anthogez (approved):

maxjeffos (approved):

PR #2289, opened 2021-09-30
feat: Analytics for IaC Custom Rules

ofekatr:

#### What does this PR do?
- For users that have provided us with custom rules,  it adds IaC custom-rules-based metrics to the analytics sent to `registry`.

```
src/cli/commands/test/iac-local-execution/index.ts:test
```

- Generate a custom rules tarball with the [IaC custom rules bundles generator CLI](https://github.com/snyk/snyk-iac-rules)
- In the Snyk CLI, run the following command to test an IaC file in debug mode: 
```
snyk iac test -d --rules=<path-to-generated-tarball> <path-to-iac-file-to-test>
```
- In the logged analytics, the following analytics have been added:
    - `iac-issues-from-custom-rules-count`
    - `iac-issues-from-custom-rules-percentage`
    - `iac-custom-rules-utilized-count`

The Snyk CLI can provide us with built-in analytics for the `--rules` flag (where customers would be providing their bundles).
We'd like to ensure we capture as much analytics about the utilization of custom rules as possible.

- [Ticket CFG-322](https://snyksec.atlassian.net/browse/CFG-322)

#### Additional information
- [Related PR in snyk/registry repository](https://github.com/snyk/registry/pull/23877)

### DIFF
--- a/src/cli/commands/test/iac-local-execution/analytics.ts
+++ b/src/cli/commands/test/iac-local-execution/analytics.ts
@@ -1,5 +1,6 @@
 import { FormattedResult } from './types';
 import * as analytics from '../../../../lib/analytics';
+import { calculatePercentage } from './math-utils';
 
 export function addIacAnalytics(
   formattedResults: FormattedResult[],
@@ -7,12 +8,15 @@ export function addIacAnalytics(
   ignoredIssuesCount: number,
 ) {
   let totalIssuesCount = 0;
+  const customRulesIdsFoundInIssues: { [customRuleId: string]: true } = {};
+  let issuesFromCustomRulesCount = 0;
   const issuesByType: Record<string, object> = {};
   const packageManagers = Array<string>();
 
   formattedResults.forEach((res) => {
     totalIssuesCount =
       (totalIssuesCount || 0) + res.result.cloudConfigResults.length;
+
     const packageManagerConfig = res.packageManager;
     packageManagers.push(packageManagerConfig);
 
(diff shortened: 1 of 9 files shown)


src/cli/commands/test/iac-local-execution/analytics.ts
```diff
@@ -21,15 +25,33 @@ export function addIacAnalytics(
...
+      if (policy.isGeneratedByCustomRule) {
+        totalIssuesFromCustomRulesCount++;
+        customRulesIdsFoundInIssues[policy.publicId] = true;
+      }
     });
   });
 
+  const customRulesUtilizedCount: number = Object.keys(
+    customRulesIdsFoundInIssues,
+  ).length;
```
ipapast, line 38: I am not sure I understand the difference between this metric and the `totalIssuesFromCustomRulesCount`. Could you explain your thinking behind it?

ipapast (commented):
Nice!! I tested it locally as well and it works as expected. I left one comment because I don't fully understand the thinking behind the `customRulesUtilizedCount` metric.

Also, I am not 100% sure about namings of the new metrics, but I can't think of better names at the moment. We could ask in the channel if these make sense to others before merging (they can not easily be renamed retrospectively; if the changes go live from registry, the fields will be deployed to BQ as new columns and it's tricky to delete them). 
So ideally we want a name that we will be happy with.

Can we add a small test to verify that `isGeneratedByCustomRule` is catching correctly the Custom Engine rules only?

Note that this should be merged after the relevant Registry PR: https://github.com/snyk/registry/pull/23877/files

ipapast (approved):

src/cli/commands/test/iac-local-execution/analytics.ts
ofekatr: The naming will be changed to be more clear, but the difference is as follows:
- `totalIssuesFromCustomRulesCount `: Holds the number of issues that were generated due to a custom rule.
- `customRulesUtilizedCount`: Holds the number of unique custom rules that generated issues.

Avishagp (approved):

dekelund (commented):
Looks good for Tundras code owner file.

PR #2292, opened 2021-10-01
feat: support target reference for test command

ghost:
We need to support `--target-reference` for `snyk test` as it does a project lookup to apply remote ignores setup on projects via the Web UI. These projects are created using `snyk monitor` which provides the option to use `--target-reference` since #2264.

I split implementation and docs into separate PRs as doc generation is a bit noisy.


maxjeffos (approved):

ipapast (approved):
approving to unblock, we'll review our smoke tests in IaC

PR #2294, opened 2021-10-03
fix: better error when org is not found for snyk code

saark-snyk:

#### What does this PR do?
This PR shows better error when org is not found for user (instead of showing "Snyk code is not enabled for org")

run `snyk code test --org=foo` with this PR

https://snyksec.atlassian.net/browse/COD-556?atlOrigin=eyJpIjoiMTBlZjliODkxMjM4NGVkNjgzMGEyZmE5M2NkMDExMWMiLCJwIjoiaiJ9

### DIFF
--- a/src/lib/plugins/sast/validate.ts
+++ b/src/lib/plugins/sast/validate.ts
@@ -1,10 +1,12 @@
 import { Options } from '../../types';
 import config from '../../config';
+import { getSastSettingsForOrg, trackUsage } from './checks';
 
 import {
   AuthFailedError,
   FailedToRunTestError,
   FeatureNotSupportedForOrgError,
+  NotFoundError,
 } from '../../errors';
 
 export async function validateCodeTest(options: Options) {
@@ -16,7 +18,6 @@ export async function validateCodeTest(options: Options) {
   }
 
   // TODO: We would need to remove this once we fix circular import issue
-  const { getSastSettingsForOrg, trackUsage } = require('./checks');
 
   const sastSettingsResponse = await getSastSettingsForOrg(org);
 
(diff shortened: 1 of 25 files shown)


src/lib/plugins/sast/validate.ts
```diff
@@ -15,9 +17,6 @@ export async function validateCodeTest(options: Options) {
     throw new FeatureNotSupportedForOrgError(org);
   }
 
-  // TODO: We would need to remove this once we fix circular import issue
```
neonnoon, line 18: I assume that this circular import issue still persists, so moving this to `import` will break things?

neonnoon (changes requested):
The actual change looks good, moving `require` to `import` either needs to be reverted or circular imports have to be fixed.

src/lib/plugins/sast/validate.ts
saark-snyk: For now it doesn't seem like it... all tests but mine pass.
Do you have any specific thing that might point on failure?

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -17,7 +17,7 @@ import { jsonStringifyLargeObject } from '../../../../src/lib/json';
 import { ArgsOptions } from '../../../../src/cli/args';
 
 const { getCodeAnalysisAndParseResults } = analysis;
-const osName = require('os-name');
+import osName = require('os-name');
```
Arvi3d, line 20: this looks like a typo. import and require in one line

saark-snyk: no typo. in order to cope with non es6 modules.

PR #2297, opened 2021-10-06
feat: Pull remote OCI Artifact with custom rules [CFG-1039]

ipapast:
#### What does this PR do?

We currently allow users to specify the path of a local custom rules bundle and use their own security rules written in Rego to scan their files, along with our own Security Rules. We now want to allow users to write their own security rules and use them. Users can store their bundles in a remote Registry repository, which they will provide to us and we are then going to download their rules, evaluate them and scan files against them.

This PR implements the "pull" functionality of an OCI Artifact. 

When running `snyk iac test` (if the right credentials are provided), we will try to download a remote OCI artifact.

- `index.ts` in the iac-local-execution directory does the checks for the credentials and handles errors, then calls the `pull()` from `oci-pull.ts`.
- `oci-pull.ts` handles the calls to the `docker-registry-v2-client` to get the layers and the manifests for the specified artifact. Then, we write the blob to a tarball, extract it as usual and follow the normal flow of evaluating rules and test them.

#### For non-IaC files:
- The `package-lock.json` and `package.json` have been updated to include the new version of the `docker-registry-v2-client`

Locally:
export OCI_REGISTRY_URL= url
export OCI_REGISTRY_USERNAME= username
export OCI_REGISTRY_PASSWORD = password

This PR is also dependent on the contentType extension of the docker-registry-client library: https://github.com/snyk/docker-registry-v2-client/pull/77. This is already merged and the new version is included in the `package.json` file.

Then running `snyk-dev iac test file.tf` should work and get all the custom rules from the OCI Registry.

OCI registries are typically used for container images, but it’s actually possible to use them for storing and distributing other types of data. There are a couple techniques for doing this, and one of them is commonly referred to as “OCI Artifacts”.

Info on OCI Artifacts: https://www.notion.so/snyk/OCI-Artifacts-2aca3ef2d1ce45c3915c828bff50f468

[CFG-1039]

#### Screenshots
image

#### Additional questions
- Acceptance tests have to be completed and pass the env vars to the job in circleCI - currenty on draft

[CFG-1039]: https://snyksec.atlassian.net/browse/CFG-1039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@snyk/cloud-config-parser": "^1.11.1",
         "@snyk/code-client": "^4.4.0",
         "@snyk/dep-graph": "^1.27.1",
+        "@snyk/docker-registry-v2-client": "^2.4.0",
         "@snyk/fix": "file:packages/snyk-fix",
         "@snyk/gemfile": "1.2.0",
         "@snyk/graphlib": "^2.1.9-patch.3",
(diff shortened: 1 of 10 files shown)


ofekatr (approved):

aron (commented):
Looking really good, nice work. I've left some comments inline, but nothing blocking. I'll try and find some time to run this locally tomorrow then approve.

Avishagp (approved):

PR #2300, opened 2021-10-08
Project attributes and tags: update flags, docs, improve error UX

jaysh:

#### What does this PR do?

1. Changes flags to have a `--project` prefix per a comment in this PR
2. Updates the docs which I missed originally to show the new CLI options (this required regenerating them all so unfortunately introduces a bunch of unrelated changes, but discussed with Jahed offline so checking them all in)
3. Creates a new `ValidationError` so when throwing them the output is in red like other types of errors we raise (I couldn't see an existing one to enable this, maybe I missed it?)

There are two separate commits, so I'd suggest looking at it separately.

Pass the `--project-business-criticality`, `--project-environment`, `--project-lifecycle`, `--project-tags` with invalid values and see the errors in red.


ghost (approved):
Looks good. 👍 Made some minor docs suggestions.

PR #2306, opened 2021-10-15
test: re-enable --all-projects tests

lili2311:

#### What does this PR do?
Re-enable the `--all-projects` tests. They were disabled as they were flaky however we have made changes since and other tests use the same setup and work fine.

Some tests were deleted that were mostly testing the right plugin is loaded & called.

Any tests that was testing output was converted to jest, but the tests that call the `cli` as a function can't be converted to Jest yet as we are doing hot loading which is not supported by Jest.


jan-stehlik (approved):
nice one @lili2311 , you are my hero :100:

PR #2309, opened 2021-10-19
feat: c/c++ unmanaged scan ui improvements 

dekelund:

#### What does this PR do?

This PR updates the Snyk CPP-plugin to v2.12.3, it includes following changes:

  - feat: add support for uhash 
  - feat: capitalize vuln severities in cli output
  - fix: punctuation
  - fix: show path for test target
  - feat(uhash): Don't include UTF8 BOM in hash calculation  ⚠️ Internal Experiment
  - perf(uhash): Speed up uhash calculations  ⚠️ Internal Experiment
  - fix: display relative paths based on project path

The main change is in package.json, where we updated the CPP dependency
to make use of v2.12.3.

- TUN-85

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
         "rimraf": "^2.6.3",
         "semver": "^6.0.0",
         "snyk-config": "4.0.0",
-        "snyk-cpp-plugin": "2.9.0",
+        "snyk-cpp-plugin": "2.12.2",
         "snyk-docker-plugin": "4.25.2",
         "snyk-go-plugin": "1.18.0",
         "snyk-gradle-plugin": "3.16.2",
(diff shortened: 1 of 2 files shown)


anthogez (commented):
Hey @dekelund nice work! Could you please update the commit's prefix?

feat: add support for uhash is misleading to cli users.

I would suggest to go with something like "feat: c/c++ unmanaged scan ui improvements"

Is true that we are introducing uHash, but we are not making it `usable` either by the cli and our customers, it's an experimental code flow, so it should not be announced as the main change.

I would suggest to add near

- feat(uhash): Don't include UTF8 BOM in hash calculation
- perf(uhash): Speed up uhash calculations

something like

- feat(uhash): Don't include UTF8 BOM in hash calculation ⚠️ Internal Experiment
- perf(uhash): Speed up uhash calculations  ⚠️ Internal Experiment

What do you think?

florindumitrascu (approved):

anthogez (commented):
@dekelund rebase then generate a new lockfile should help with conflicting files

regarding the approval I would recommend to reach-out team hammer 
as they are the owners of `package.json` and `package-lock.json`

JackuB (approved):
--

I'm updating codeowners file to not pull all the teams into a package-lock.json change PR https://github.com/snyk/snyk/pull/2310

PR #2329, opened 2021-10-28
feat: add --gradle-accept-legacy-config-roles flag

muscar:

#### What does this PR do?

This patch adds the `--gradle-accept-legacy-config-roles` flag, and uses a version of the Gradle plugin that accepts it. When set by the user, the plugin will use a more permissive dependency resolution process.

See https://github.com/snyk/snyk-gradle-plugin/pull/191/files for more details.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "snyk-cpp-plugin": "2.12.3",
         "snyk-docker-plugin": "4.26.1",
         "snyk-go-plugin": "1.18.0",
-        "snyk-gradle-plugin": "3.16.2",
+        "snyk-gradle-plugin": "3.17.0",
         "snyk-module": "3.1.0",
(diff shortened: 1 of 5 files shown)


anthogez (commented):
Why not `--gradle-support-all-config-roles`?

dtrunley-snyk (approved):

ghost (dismissed):
package-lock.json is out of sync and the diff is huge. You'll need to rollback and run `npm ci` before adding your package.json changes. Also to avoid weird bugs, best to use the latest npm@7 (if you're not already).

JackuB (changes requested):

orsagie (approved):

JackuB (approved):
Anything we should test on the CLI side or is the gradle-plugin testing enough in this case?

dekelund (approved):

PR #2330, opened 2021-10-29
support tags/attributes in Docker projects & alias --project-tags to --tags

jaysh:

#### What does this PR do?

This PR adds support for `snyk container monitor` to use `--project-tags` (and now `--tags`) as well as the project attributes (`--project-lifecycle`, `--project-business-criticality`, `--project-environment`).

Example command line:

```
snyk container monitor alpine:3.14 --tags=container2=monitor5 --project-lifecycle=sandbox --project-environment=backend --project-business-criticality=critical
```

### DIFF
--- a/src/lib/ecosystems/monitor.ts
+++ b/src/lib/ecosystems/monitor.ts
@@ -4,7 +4,7 @@ import chalk from 'chalk';
 import config from '../config';
 import { isCI } from '../is-ci';
 import { makeRequest } from '../request/promise';
-import { MonitorResult, Options } from '../types';
+import { MonitorResult, Options, ProjectAttributes, Tag } from '../types';
 import * as spinner from '../../lib/spinner';
 import { getPlugin } from './plugins';
 import { BadResult, GoodResult } from '../../cli/commands/monitor/types';
@@ -26,6 +26,10 @@ import {
 import { findAndLoadPolicyForScanResult } from './policy';
 import { getAuthHeader } from '../api-token';
 import { resolveAndMonitorFacts } from './resolve-monitor-facts';
+import {
+  generateProjectAttributes,
+  generateTags,
+} from '../../cli/commands/monitor';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 
@@ -35,6 +39,10 @@ export async function monitorEcosystem(
   options: Options,
 ): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
   const plugin = getPlugin(ecosystem);
+
+  const tags = generateTags(options);
+  const attributes = generateProjectAttributes(options);
+
   const scanResultsByPath: { [dir: string]: ScanResult[] } = {};
   for (const path of paths) {
     try {
@@ -63,6 +71,8 @@ export async function monitorEcosystem(
     ecosystem,
     scanResultsByPath,
     options,
+    attributes,
+    tags,
   );
   return [monitorResults, errors];
 }
@@ -71,16 +81,20 @@ async function selectAndExecuteMonitorStrategy(
   ecosystem: Ecosystem,
   scanResultsByPath: { [dir: string]: ScanResult[] },
   options: Options,
+  attributes: ProjectAttributes,
+  tags: Tag[] | undefined,
 ): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
   const isUnmanagedEcosystem = ecosystem === 'cpp';
   return isUnmanagedEcosystem
     ? await resolveAndMonitorFacts(scanResultsByPath, options)
-    : await monitorDependencies(scanResultsByPath, options);
+    : await monitorDependencies(scanResultsByPath, options, attributes, tags);
 }
 
 export async function generateMonitorDependenciesRequest(
   scanResult: ScanResult,
   options: Options,
+  attributes: ProjectAttributes,
+  tags: Tag[] | undefined,
 ): Promise<MonitorDependenciesRequest> {
   // WARNING! This mutates the payload. The project name logic should be handled in the plugin.
   scanResult.name =
@@ -95,6 +109,8 @@ export async function generateMonitorDependenciesRequest(
     scanResult,
     method: 'cli',
     projectName: options['project-name'] || config.PROJECT_NAME || undefined,
+    tags: tags,
+    attributes: attributes,
   };
 }
 
@@ -103,6 +119,8 @@ async function monitorDependencies(
     [dir: string]: ScanResult[];
   },
   options: Options,
+  attributes: ProjectAttributes,
+  tags: Tag[] | undefined,
 ): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
   const results: EcosystemMonitorResult[] = [];
   const errors: EcosystemMonitorError[] = [];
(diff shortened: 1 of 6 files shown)


src/lib/ecosystems/monitor.ts
```diff
@@ -35,6 +39,10 @@ export async function monitorEcosystem(
   options: Options,
 ): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
   const plugin = getPlugin(ecosystem);
+
+  const tags = generateTags(options);
+  const attributes = generateProjectAttributes(options);
```
darscan, line 44: `options` is already passed all the way through to `generateMonitorDependenciesRequest`, so consider doing this inside that function instead of extracting these here and weaving them through all the intermediate function calls.

src/lib/ecosystems/types.ts
```diff
@@ -133,4 +133,6 @@ export interface MonitorDependenciesRequest {
   projectName?: string;
   policy?: string;
   method?: 'cli';
+  tags?: Tag[];
+  attributes?: ProjectAttributes;
```
jahed-snyk, line 137: Since `projectName` (above) is deprecated in favour of `ScanResult.name`, I think `tags` and `attributes` should also go under `ScanResult` as it's similar in model to a `Project` as these properties are only useful under a project context rather than a monitor one.

src/lib/ecosystems/monitor.ts
jaysh: The downside of that approach (now) is it means it'll only do the validation then as well so if it's a big project it will do a bunch of work only to say you've used a CLI flag wrong, hence this approach, which is the pattern used elsewhere already.

src/lib/ecosystems/types.ts
jaysh: Intuitively shouldn't `ScanResult` be used to hold the results of a scan?

darscan: I also don't think these belong in the `ScanResult` - it's additional, user supplied metadata

src/lib/ecosystems/monitor.ts
jaysh: I understand, and I would've loved to do that but the original implementation (in regular `snyk monitor`) followed the existing pattern (which is already merged) and this is supposed to only be an additive/wiring PR - I think it's going to be hard to justify that refactor right now. It'll require refactoring the entire implementation across `container monitor` and regular `monitor` and retesting everything.

If you are really strong on that I can possibly just do this here, but then we will end up with two different approaches to achieve the same thing - which for a feature that has only been out a few weeks - is not great.

darscan: I'm really not suggesting anything that drastic. Either way, what you have is fine.. I just hate seeing parameter explosion.. and I dislike seeing "options" not being the final parameter in function calls.

src/lib/ecosystems/types.ts
jahed-snyk: I agree. I guess the question is where do they belong then? Placing them at the top level, we end up conflating monitoring and project management. There is also no project identity at the top level so the hierachy to me doesn't make sense. 😕

src/lib/ecosystems/monitor.ts
jaysh: It will require changing every file I've already touched across multiple PRs unfortunately as it's a fundamentally changing the approach. I can re-order the params to put `options` last though, that makes sense. I'll do that now.

src/lib/ecosystems/types.ts
jaysh: I don't see the problem, it's called `MonitorDependenciesRequest` - putting at the top level seems fair enough? We could add another layer of grouping but I don't see any clear boundary there that will make it immediately obvious here a new property should go, resulting in potential future inconsistency/debt.

src/lib/ecosystems/monitor.ts
```diff
@@ -71,16 +81,20 @@ async function selectAndExecuteMonitorStrategy(
   ecosystem: Ecosystem,
   scanResultsByPath: { [dir: string]: ScanResult[] },
   options: Options,
+  attributes: ProjectAttributes,
+  tags: Tag[] | undefined,
 ): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
   const isUnmanagedEcosystem = ecosystem === 'cpp';
   return isUnmanagedEcosystem
     ? await resolveAndMonitorFacts(scanResultsByPath, options)
```
ivanstanev, line 89: CPP won't have attributes and tags?

jaysh: It will, but in a separate PR. We've setup the tickets by vertical slice so CPP (`snyk unmanaged`) is separate.

src/lib/ecosystems/types.ts
jahed-snyk: As Hammer doesn't own the endpoint, I don't think I should block on this. We don't really have a long-term vision for these endpoints on our end. If Open Source group are good with it server-side (they own the endpoint), it should be fine.

ghost (approved):

src/lib/ecosystems/types.ts
darscan: > If Open Source group are good with it server-side (they own the endpoint), it should be fine.

Do we have anyone from the Open Source group looking at this PR?

src/lib/ecosystems/monitor.ts
anthogez: > this is all internal and can be tidied up later
I think we should tidied up later avoid this, we know that it will not happen so let's go as @darscan suggested above. We already have enough mess into the cli space, if we can reduce it by passing these params through options we will not have to rethink later on and/or give space to others add further params.

anthogez: > This is all internal and can be tidied up later

I think we should avoid tidied up this later, we know that it will not happen I would suggest to go as as @darscan suggested above. We already have enough mess into the cli space, if we can reduce it by passing these params through options we will not have to rethink later on and/or give space to others add further params.

src/lib/ecosystems/types.ts
jaysh: @anthogez can you opine here please?

src/lib/ecosystems/monitor.ts
jaysh: Yeah I get it - I was just highlighting earlier that it will create it's own inconsistency as it's an alternative to the precedent used elsewhere, but it sounds like simplifying this one is the lesser of the two evils so I will go with that and get it done now. Thanks for the feedback!

src/lib/ecosystems/types.ts
darscan: I realise that we all have battle scars from the top-level explosion that happened in the old flows. However, in the new flow we have a good place for things related to the scan itself (which would cover most of those old random properties, as these would become concrete "facts"), so the risk of top-level explosion is now much lower. I don't think we should introduce a generic top-level bucket for stuff (e.g. "meta") as this would just encourage people to use that instead of creating proper "facts". TLDR: I'm OK with `tags` and `attributes` living at the top level of the payload.

src/lib/ecosystems/monitor.ts
jaysh: This is done - the best approach I could see is basically to re-alias the method. The rationale here is the validation logic is deep: we need to split strings then validate so repeating that logic twice (once for `validate`, once for `get`) seemed pretty redundant, but having an explicit 'validate' method which is an alias makes the intent much more clear (i.e. not just calling the `get` variant and relying on a side-effect).

darscan: Thanks for doing that. Agree with the approach

PR #2332, opened 2021-11-01
fix: match package.json to workspace definition

lili2311:

#### What does this PR do?
Re-enable windows tests as they have been hiding the fact that the
feature broke on Windows.

Update the file matching logic to support both Windows & Linux file names
as well as different glob types.
e.g. 'package/*' is an example given in Yarn docs

https://github.com/snyk/snyk/compare/fix/yarn-workspaces-windows?expand=1#diff-58a7b63707ecc3959ab8c5f5ec8cfb77660fc185f22cb4c2677b5af04180bba4R165

Disabled tests did not catch that Windows support had a bug

https://github.com/snyk/snyk/issues/1560

### DIFF
--- a/src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
+++ b/src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
@@ -7,7 +7,6 @@ import * as micromatch from 'micromatch';
 
 const debug = baseDebug('snyk-yarn-workspaces');
 import * as lockFileParser from 'snyk-nodejs-lockfile-parser';
-import * as path from 'path';
 import { NoSupportedManifestsFoundError } from '../../errors';
 import {
   MultiProjectResultCustom,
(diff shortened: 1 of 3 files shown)


src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
```diff
@@ -178,20 +162,23 @@ export function getWorkspacesMap(file: {
...
   const workspacesGlobs = (
     yarnWorkspacesMap[workspaceRoot].workspaces || []
-  ).map((p) => path.join(workspaceRootFolder, p));
-  const filePath = path.normalize(packageJsonFileName);
+  ).map((workspace) => pathUtil.join(workspaceRootFolder, workspace));
 
   const match = micromatch.isMatch(
-    filePath,
-    workspacesGlobs.map((p) => (p.endsWith('/**') ? p : p + '/**')),
+    packageJsonFileName.replace(/\\/g, '/'),
```
lili2311, line 178: normalize all paths to `/` as `\\` does not allow path.parse to identify the folder correctly

jan-stehlik (approved):
thanks @lili2311 :heart:

PR #2333, opened 2021-11-02
fix: fix combination of rubygems and --all-projects

admons:

#### What does this PR do?
today all rebygems projects have the same project name with --all-projects flag.
this adds the nested dir name into the project name.

### DIFF
--- a/src/lib/plugins/rubygems/index.ts
+++ b/src/lib/plugins/rubygems/index.ts
@@ -1,17 +1,19 @@
 import { inspectors, Spec } from './inspectors';
 import { MissingTargetFileError } from '../../errors/missing-targetfile-error';
 import gemfileLockToDependencies = require('./gemfile-lock-to-dependencies');
-const get = require('lodash.get');
+import * as get from 'lodash.get';
 import { MultiProjectResult } from '@snyk/cli-interface/legacy/plugin';
+import * as types from '../types';
 
 export async function inspect(
   root: string,
   targetFile: string,
+  options: types.Options = {},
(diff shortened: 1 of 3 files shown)


src/lib/plugins/rubygems/index.ts
```diff
@@ -1,17 +1,19 @@
...
 import gemfileLockToDependencies = require('./gemfile-lock-to-dependencies');
-const get = require('lodash.get');
+import * as get from 'lodash.get';
 import { MultiProjectResult } from '@snyk/cli-interface/legacy/plugin';
+import * as types from '../types';
 
 export async function inspect(
   root: string,
   targetFile: string,
+  options: types.Options = {},
```
admons, line 11: We already send the options when we run `inspect`, but never used it here.

muscar (approved):

src/lib/plugins/rubygems/inspectors/gemfile.ts
```diff
@@ -28,8 +33,11 @@ export async function gatherSpecs(root: string, target: string): Promise<Spec> {
   );
 
   if (gemfileLock) {
+    const basePackageName = path.basename(root);
     return {
-      packageName: path.basename(root),
+      packageName: options.allSubProjects
```
admons, line 38: Because the name of the package here is the actual name of the project, and the project name is a part of the identifier, we only want to change this in the problematic area we're facing of 2 or more projects with the '--all-projects` flag.

jan-stehlik (approved):
nice one @admons :man_dancing:

PR #2339, opened 2021-11-03
Feat/add iac custom rules entitlement

ofekatr:
### What this does
- Refactors the pull flow in the `iac test` command.
- Wraps the `--rules` flag with the `iacCustomRulesEntitlement` entitlement.
- Wraps the custom-rules bundles pulling operation with the `iacCustomRules` flag.
- Wraps the custom-rules bundles pulling operation with the `iacCustomRulesEntitlement` entitlement.

### Notes for the reviewer

<details>
  <summary><b>Background context</b></summary>

The custom-rules feature and `--rules` flag in the Snyk CLI is currently behind a feature flag called `iacCustomRules`. 

We have introduced a new entitlement called `iacCustomRulesEntitlement`, which will soon replace the deprecated feature flag and will be the only gate after GA that will be used to make sure only specific customers can use this feature.

</details>

<details>
  <summary><b>How to run locally</b></summary>

  ##### The `--rules` Flag:
1. Prepare a custom-rules bundle generated by the [Snyk custom-rules bundles generator](https://github.com/snyk/snyk-iac-rules). 
2. Make sure your org has the `iacCustomRules` and the `infrastructureAsCode` feature flags.
3. Make sure your org has the `iacCustomRulesEntitlement` entitlement.
4. In the DB, make sure to not have any preconfigured OCI registry configurations in `org.cloudConfig.customRules`.
5. In the Snyk CLI, run `snyk-dev iac test <file> --rules=<path-to-bundle>`.
6. Verify that:
    - No errors were thrown.
    - The defined custom rules generated the expected issues successfully.
7. Through the org's admin page, disable the `iacCustomRulesEntitlement` entitlement.
8. In the Snyk CLI, run `snyk-dev iac test <file> --rules=<path-to-bundle>`.
9. Verify that the flag was not available due to the disabled entitlement.

  ##### Custom-Rules Bundles Pulling:
1. Prepare a custom-rules bundle generated by the [Snyk custom-rules bundles generator](https://github.com/snyk/snyk-iac-rules). 
2. Store the generated bundle in your preferred OCI registry.
3. Export the following env variables:
    - `OCI_REGISTRY_USERNAME`: The username for the desired registry.
    - `OCI_REGISTRY_PASSWORD`: The password for the desired registry. 
4. Make sure your org has the `iacCustomRules` and the `infrastructureAsCode` feature flags.
5. Make sure your org has the `iacCustomRulesEntitlement` entitlement.
6. Make sure your user has the following org permission: `READ`, `IAC_SETTINGS_READ` `IAC_SETTINGS_EDIT`.
7. Navigate to the custom-rules configurations section in the org IaC settings page on `/org/ofekatr/manage/cloud-config`.
8. Update the OCI registry URL and tag for the desired custom-rules bundle.
9. In the Snyk CLI, run `snyk-dev iac test <file>`.
10. Verify that:
    - No errors were thrown.
    - The bundle was pulled successfully into the `.iac-data` directory.
    - The defined custom rules generated the expected issues successfully.
11. Through the org's admin page, disable the `iacCustomRulesEntitlement` entitlement.
12. In the Snyk CLI, run `snyk-dev iac test <file>`.
13. Verify that the operation was not available due to the disabled entitlement.
14. Through the org's admin page, enable the `iacCustomRulesEntitlement` entitlement.
15. Through the org's admin page, disable the `iacCustomRules` feature flag.
16. Verify that the operation was not available due to the disabled feature flag.

</details>

### More information

- [Jira ticket CFG-1086](https://snyksec.atlassian.net/browse/CFG-1086)
- [Related PR on snyk/registry](https://github.com/snyk/registry/pull/24450)

### DIFF
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -40,7 +40,7 @@ function getFlagName(key: string) {
 }
 
 export class FlagError extends CustomError {
-  constructor(key: string, featureFlag: string) {
+  constructor(key: string, featureFlag?: string) {
     const flag = getFlagName(key);
     let msg;
     if (featureFlag) {
@@ -65,6 +65,18 @@ export class FlagValueError extends CustomError {
     this.code = IaCErrorCodes.FlagValueError;
     this.strCode = getErrorStringCode(this.code);
     this.userMessage = msg;
+  }
+}
+
+export class UnsupportedEntitlementFlagError extends CustomError {
+  constructor(key: string, entitlementName: string) {
+    const flag = getFlagName(key);
+    super(
+      `Unsupported flag: ${flag} - Missing the ${entitlementName} entitlement`,
+    );
+    this.code = IaCErrorCodes.UnsupportedEntitlementFlagError;
+    this.strCode = getErrorStringCode(this.code);
+    this.userMessage = `Flag "${flag}" is currently not supported for this org. To enable it, please contact snyk support.`;
   }
 }
 
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -7,8 +7,10 @@ import {
   IaCTestFlags,
   layerContentType,
   manifestContentType,
+  OCIRegistryURLComponents,
   SafeAnalyticsOutput,
   TestReturnValue,
+  IacOrgSettings,
 } from './types';
 import { addIacAnalytics } from './analytics';
 import { TestLimitReachedError } from './usage-tracking';
@@ -27,18 +29,27 @@ import {
   trackUsage,
 } from './measurable-methods';
 import { isFeatureFlagSupportedForOrg } from '../../../../lib/feature-flags';
-import { FlagError } from './assert-iac-options-flag';
+import {
+  UnsupportedEntitlementFlagError,
+  FlagError,
+} from './assert-iac-options-flag';
 import { config as userConfig } from '../../../../lib/user-config';
 import config from '../../../../lib/config';
 import { findAndLoadPolicy } from '../../../../lib/policy';
-import { CustomError } from '../../../../lib/errors';
+import {
+  CustomError,
+  UnsupportedFeatureFlagError,
+} from '../../../../lib/errors';
 import { getErrorStringCode } from './error-utils';
 import {
-  extractURLComponents,
+  extractOCIRegistryURLComponents,
   FailedToBuildOCIArtifactError,
   InvalidManifestSchemaVersionError,
   InvalidRemoteRegistryURLError,
+  UnsupportedEntitlementPullError,
+  UnsupportedFeatureFlagPullError,
 } from './oci-pull';
+import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-entitlement-error';
 
 // this method executes the local processing engine and then formats the results to adapt with the CLI output.
 // this flow is the default GA flow for IAC scanning.
@@ -47,59 +58,28 @@ export async function test(
   options: IaCTestFlags,
 ): Promise<TestReturnValue> {
   try {
-    const org = options.org ?? config.org;
-    const iacOrgSettings = await getIacOrgSettings(org);
-    const customRulesPath = await customRulesPathForOrg(options.rules, org);
+    let customRulesPath: string | undefined;
 
-    const OCIRegistryURL =
-      (iacOrgSettings.customRules?.isEnabled &&
-        iacOrgSettings.customRules?.ociRegistryURL) ||
-      userConfig.get('oci-registry-url');
+    const orgPublicId = options.org ?? config.org;
 
-    if (OCIRegistryURL && customRulesPath) {
+    const iacOrgSettings = await getIacOrgSettings(orgPublicId);
+
+    if (options.rules) {
+      await assertRulesFlagAvailable(orgPublicId, iacOrgSettings);
+      customRulesPath = options.rules;
+    }
+
+    const isOCIRegistryURLProvided = checkOCIRegistryURLProvided(
+      iacOrgSettings,
+    );
+
+    if (isOCIRegistryURLProvided && customRulesPath) {
       throw new FailedToExecuteCustomRulesError();
     }
 
-    if (OCIRegistryURL) {
-      if (!isValidURL(OCIRegistryURL)) {
-        throw new InvalidRemoteRegistryURLError();
-      }
-
-      const URLComponents = extractURLComponents(OCIRegistryURL);
-      const username = userConfig.get('oci-registry-username');
-      const password = userConfig.get('oci-registry-password');
-
-      const opt = {
-        username,
-        password,
-        reqOptions: {
-          acceptManifest: manifestContentType,
-          acceptLayer: layerContentType,
-          indexContentType: '',
-        },
-      };
-
-      try {
-        await pull(URLComponents, opt);
-      } catch (err) {
-        if (err.statusCode === 401) {
-          throw new FailedToPullCustomBundleError(
-            'There was an authentication error. Incorrect credentials provided.',
-          );
-        } else if (err.statusCode === 404) {
-          throw new FailedToPullCustomBundleError(
-            'The remote repository could not be found. Please check the provided URL.',
-          );
-        } else if (err instanceof InvalidManifestSchemaVersionError) {
-          throw new FailedToPullCustomBundleError(err.message);
-        } else if (err instanceof FailedToBuildOCIArtifactError) {
-          throw new FailedToBuildOCIArtifactError();
-        } else if (err instanceof InvalidRemoteRegistryURLError) {
-          throw new InvalidRemoteRegistryURLError();
-        } else {
-          throw new FailedToPullCustomBundleError();
-        }
-      }
+    if (isOCIRegistryURLProvided) {
+      await assertPullAvailable(orgPublicId, iacOrgSettings);
+      await pullIaCCustomRules(iacOrgSettings);
     } else {
       await initLocalCache({ customRulesPath });
     }
@@ -113,7 +93,7 @@ export async function test(
     );
 
     // Duplicate all the files and run them through the custom engine.
-    if (customRulesPath || OCIRegistryURL) {
+    if (customRulesPath || isOCIRegistryURLProvided) {
       parsedFiles.push(
         ...parsedFiles.map((file) => ({
           ...file,
@@ -163,22 +143,6 @@ export async function test(
   }
 }
 
-async function customRulesPathForOrg(
-  customRulesPath: string | undefined,
-  publicOrgId: string,
-): Promise<string | undefined> {
-  if (!customRulesPath) return;
-
-  const isCustomRulesSupported = (
-    await isFeatureFlagSupportedForOrg('iacCustomRules', publicOrgId)
-  ).ok;
-  if (isCustomRulesSupported) {
-    return customRulesPath;
-  }
-
-  throw new FlagError('rules', 'iacCustomRules');
-}
-
 export function removeFileContent({
   filePath,
   fileType,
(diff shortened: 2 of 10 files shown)


src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -46,59 +57,30 @@ export async function test(
...
-      }
+    const isOCIRegistryURLProvided = !!ociRegistryURLComponents;
+
+    if (isOCIRegistryURLProvided && customRulesPath) {
+      throw new FailedToExecuteCustomRulesError();
+    }
+
+    if (isOCIRegistryURLProvided) {
+      await assertPullAvailable(orgPublicId, orgIaCSettings);
+      await pullIaCCustomRules(ociRegistryURLComponents!);
```
teodora-sandu, line 83: is `ociRegistryURLComponents!` an accidental change here? It's showing me a warning in VSCode but if I take the `!` away it's okay, so I don't know if we need this

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -192,16 +158,148 @@ export function removeFileContent({
...
   return url.protocol === 'http:' || url.protocol === 'https:';
 }
 
+/**
+ * Gets the OCI registry URL components from either the env variables or the IaC org settings.
+ */
+function getOCIRegistryURLComponents(
+  iacOrgSettings?: IaCOrgSettings,
+): OCIRegistryURLComponents | undefined {
+  if (process.env.OCI_REGISTRY_URL) {
```
teodora-sandu, line 177: Correct me if I'm wrong but I believe you're changing the behaviour here - before, the org settings would override the environment variables. Now, the environment variables take precedence.  Was that on purpose?

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -192,16 +158,148 @@ export function removeFileContent({
...
+    const settingsOCIRegistryURL = iacOrgSettings.customRules.ociRegistryURL;
+
+    if (!isValidURL(settingsOCIRegistryURL)) {
+      throw new InvalidRemoteRegistryURLError();
+    }
+
+    return {
+      ...extractOCIRegistryURLComponents(settingsOCIRegistryURL),
+      tag: iacOrgSettings.customRules.ociRegistryTag || 'latest',
+    };
```
teodora-sandu, line 200: Instead of duplicating this code a bit, can you do something like:
```
function getOCIRegistryURLComponents(
  iacOrgSettings?: IaCOrgSettings,
): OCIRegistryURLComponents | undefined {
    let settingsOciRegistryURL;
     if (
        !!iacOrgSettings?.customRules?.isEnabled &&
        iacOrgSettings?.customRules?.ociRegistryURL
    ) {
        settingsOciRegistryURL = iacOrgSettings?.customRules?.ociRegistryURL + (iacOrgSettings?.customRules?.ociRegistryTag || 'latest');
    }
    const ociRegistryUrl = settingsOciRegistryURL || process.env.OCI_REGISTRY_URL;
    if (!isValidURL(ociRegistryUrl)) {
      throw new InvalidRemoteRegistryURLError();
    }
    return extractOCIRegistryURLComponents(ociRegistryUrl);
}
```

I haven't run this but hopefully it makes sense. Also, good job catching that the `ociRegistryTag` wasn't being read

src/cli/commands/test/iac-local-execution/oci-pull.ts
```diff
@@ -110,3 +116,23 @@ export class InvalidRemoteRegistryURLError extends CustomError {
       'The Remote Registry URL is invalid, or does not include a http/https protocol. Please check it again.';
   }
 }
+
+export class UnsupportedFeatureFlagPullError extends UnsupportedFeatureFlagError {
+  constructor(
+    featureFlag: string,
+    userMessage = `Pulling custom-rules bundles is not supported for this org - The feature flag '${featureFlag}' is not currently enabled. To enable, please contact snyk support`,
```
teodora-sandu, line 123: Applies to this error and the next one. Do you think it might confusing to a customer to see an error about pulling custom rules when they're not actually doing that? I feel like maybe it'd be better if we just gave a general "custom rules are not supported" error, on top of instructions for how to enable it

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -68,16 +68,21 @@ export type FormattedResult = {
 
 export type IacCustomPolicies = Record<string, { severity?: string }>;
 
-export interface IaCCustomRulesConfigs {
+export interface IaCOrgSettingsCustomRulesConfigs {
```
teodora-sandu, line 71: nit: `IaCCustomRules`

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -68,16 +68,21 @@ export type FormattedResult = {
...
   isEnabled?: boolean;
   ociRegistryURL?: string;
   ociRegistryTag?: string;
 }
 
-export interface IacOrgSettings {
+export interface IaCOrgSettingsEntitlements {
```
teodora-sandu, line 77: nit: `IaCEntitlements`
(`customPolicies` below has type `IacCustomPolicies`) and the fact they're org settings or not is not as useful to a developer looking at this code, since they can see they're part of the interface `IaCOrgSettings`

src/cli/commands/test/iac-local-execution/types.ts
teodora-sandu: Actually, I would suggest not using the `IaC` capitalisation, just to be consistent with everything else in this file

src/cli/commands/test/iac-local-execution/oci-pull.ts
```diff
@@ -110,3 +116,23 @@ export class InvalidRemoteRegistryURLError extends CustomError {
...
+  }
+}
+
+export class UnsupportedEntitlementPullError extends UnsupportedEntitlementError {
+  constructor(
+    entitlement: string,
+    userMessage = `Pulling custom-rules bundles is not supported for this org - The entitlement '${entitlement}' is not currently enabled. To enable, please contact snyk support`,
+  ) {
+    super(entitlement, userMessage);
+    this.userMessage = userMessage;
```
teodora-sandu, line 136: Make sure to add a `code` and `strCode` field to all new errors and the corresponding field unde `IaCErrorCodes `. These are used by us to differentiate errors in analytics:
https://github.com/snyk/snyk/blob/9b531be8bada5e3c2dd351a43ec9354cfb22dd70/src/cli/commands/test/iac-local-execution/file-loader.ts#L94
https://github.com/snyk/snyk/blob/07fa0cdb7043ea8942adaeab8db7ff3e9b64948d/src/cli/commands/test/iac-local-execution/types.ts#L253


https://www.notion.so/snyk/Snyk-CLI-IaC-related-error-codes-de1c8d673b614d00a1404713eaf4ea3e

test/jest/acceptance/iac/custom-rules.spec.ts
```diff
@@ -52,22 +52,33 @@ describe('iac test --rules', () => {
...
+  let teardown: () => Promise<unknown>;
 
   beforeAll(async () => {
     const result = await startMockServer();
     run = result.run;
     teardown = result.teardown;
   });
 
-  afterAll(async () => teardown());
+  afterAll(async () => await teardown());
```
teodora-sandu, line 81: nit: I don't think you need the `await` here?

src/lib/errors/unsupported-feature-flag-error.ts
```diff
@@ -1,13 +1,16 @@
...
+  public readonly featureFlag: string;
+
   private static ERROR_CODE = 403;
 
   constructor(
     featureFlag: string,
     userMessage = `Feature flag '${featureFlag}' is not currently enabled for your org, to enable please contact snyk support`,
   ) {
     super(userMessage);
+    this.featureFlag = featureFlag;
```
teodora-sandu, line 13: Why do you need this? I think it worked without setting `featureFlag`?

src/lib/errors/unsupported-entitlement-error.ts
```diff
@@ -0,0 +1,17 @@
+import { CustomError } from './custom-error';
```
teodora-sandu, line 1: I'm wondering if we need to set a new owner for this file? I was wondering why so many teams have to review the PR. We don't own entitlements, but we could either move this to our `iac-local-execution` folder or change the codeowner, since we added this functionality for us

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -192,16 +158,148 @@ export function removeFileContent({
...
+): Promise<void> {
+  const isCustomRulesEnabled = !!(
+    await isFeatureFlagSupportedForOrg('iacCustomRules', orgPublicId)
+  ).ok;
+
+  if (!isCustomRulesEnabled) {
+    throw new UnsupportedFeatureFlagError('iacCustomRules');
+  }
+
+  if (!orgIaCSettings) {
```
teodora-sandu, line 258: I think the `orgIaCSettings` would always be passed to this function through  the `assertRulesFlagAvailable` and `assertPullAvailable` calls in the `test` function. So you don't need to check here and call `getIacOrgSettings` again

src/cli/commands/test/iac-local-execution/index.ts
ofekatr: I've considered this, but decided to go with the implementation you see for 3 reasons:
1. In the suggested implementation, what ends up happening is that the URL is built and then decomposed in `extractOCIRegistryURLComponents`
2. The URL may not have been provided and as such the function had have a flow where returning undefined is a valid option.
3. It felt more readable in my opinion.

What do you think? I'll refactor the code if you think it's the preferred choice

src/cli/commands/test/iac-local-execution/index.ts
ofekatr: It was done more for flexibility, in case we use it again, so we wouldn't have to get the settings before calling the function on all occasions.

What do you think?

src/lib/errors/unsupported-feature-flag-error.ts
ofekatr: This was added for accessibility in the `assertPullAvailable` function so that when it catches this error, it will not use hard-coded feature-flag names.

However, after the refactor I'm doing for the pull being unsupported they will not be needed

test/jest/acceptance/iac/custom-rules.spec.ts
ofekatr: Based on [this line](https://github.com/snyk/snyk/blob/1c0fad330f7b79e3a98409c6203e67dbbe4a97ab/test/jest/acceptance/iac/helpers.ts#L39) in the `startMockServer` I believe it is needed

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
```diff
@@ -68,6 +68,17 @@ export class FlagValueError extends CustomError {
   }
 }
 
+export class UnsupportedEntitlementFlagError extends CustomError {
+  constructor(key: string, entitlementName: string) {
+    const flag = getFlagName(key);
+    const msg = `Flag "${flag}" is only supported for users with the '${entitlementName}' entitlement.`;
```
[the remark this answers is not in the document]
ofekatr: So should the existing flag error regarding the feature-flag should be changed accordingly as well?
Should the new message be something like:
```
Flag "${flag}" is currently not supported for this org. To enable, please contact snyk support`;
```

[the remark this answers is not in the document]
teodora-sandu: The feature flag is under the customer's control - they'll be able to enable it using the Snyk Preview page. The entitlement is assigned by us internally so that's why we should just tell them here what they can do to get this feature (i.e. get in touch)

src/cli/commands/test/iac-local-execution/index.ts
teodora-sandu: I think it's a bit confusing for this function to sometimes retrieve the org settings, sometimes not 🤔  Plus, it's unlikely that we'll need to call this function from somewhere else but if that does happen we would be better re-organising the code so that we get the org settings only once and then pass it as a function argument. The org settings make an API call to `registry` and we want to minimise the number of API calls

test/jest/acceptance/iac/custom-rules.spec.ts
teodora-sandu: Because this is an arrow function without curly brackets, I believe the `await` is implicit: https://masteringjs.io/tutorials/fundamentals/arrow. Which is (I think) why all the other tests in this file work without the `await`

src/lib/errors/unsupported-feature-flag-error.ts
ofekatr: After further consideration they can still be utilized for the internal error message to give more context, so it might be worthwhile to keep these.

src/cli/commands/test/iac-local-execution/index.ts
teodora-sandu: I'm happy to leave it as is - readability is important and it's not that much code duplication 😄

src/cli/commands/test/iac-local-execution/types.ts
ofekatr: I wanted to maintain the context in case other entitlements related types are created in the future. But that's a good point

src/lib/errors/unsupported-entitlement-error.ts
ofekatr: Due to the generic context of the error, as well as the error code for it, I think it's better to keep It in the current ownership. Suffice to say, the derived errors that are relevant to IaC are in the `iac-local-execution` directory.

src/cli/commands/test/iac-local-execution/index.ts
ofekatr: The same could be said about calling the `isFeatureFlagSupportedForOrg`, as it also makes a call to `registry` on every execution. Should the flag be set prior, and passed as a parameter?

src/cli/commands/test/iac-local-execution/index.ts
ofekatr: You're right, this is unintentional. Fixed.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -46,59 +57,28 @@ export async function test(
...
+    if (options.rules) {
+      await assertRulesFlagAvailable(orgPublicId, iacOrgSettings);
+      customRulesPath = options.rules;
     }
 
-    if (OCIRegistryURL) {
-      if (!isValidURL(OCIRegistryURL)) {
-        throw new InvalidRemoteRegistryURLError();
-      }
+    const isOCIRegistryURLProvided = checkOCIRegistryURLProvided(
```
teodora-sandu, line 71: This function is also called by `checkOCIRegistryURLExistsInSettings`, which is called by `getOCIRegistryURLComponents`, inside the `pullIaCCustomRules` function. I think the flow was easier to understand before, when we called `getOCIRegistryURLComponents` here and then passed those through. WDYT?

test/jest/acceptance/iac/custom-rules.spec.ts
```diff
@@ -189,7 +189,7 @@ describe('custom rules pull from a remote OCI registry', () => {
 
     expect(exitCode).toBe(2);
     expect(stdout).toContain(
-      `Pulling custom-rules bundles is not supported for this org - The entitlement 'iacCustomRulesEntitlement' is not currently enabled. To enable, please contact snyk support`,
+      `The custom-rules feature is currently not supported for this org. To enable it, please contact snyk support.`,
```
teodora-sandu, line 192: This applied everywhere, but `custom rules` should not have a `-`

src/cli/commands/test/iac-local-execution/index.ts
ofekatr: It's only called once here. The function that is called in two different places is `checkOCIRegistryURLExistsInSettings`:

1. In `checkOCIRegistryURLProvided` - To check if it was provided in the org IaC settings or env variables
2. In `getOCIRegistryURLComponents` - To know where to get the URL components from.

It's following the comment earlier to prevent `getOCIRegistryURLComponents` from returning `undefined`, and was done in order to separate the existence check from the `get` function.

teodora-sandu (approved):
LGTM 🎉

src/cli/commands/test/iac-local-execution/index.ts
teodora-sandu: The difference is that `isFeatureFlagSupportedForOrg` calls another endpoint to `getIacOrgSettings`, so it has to be separate

ofekatr: Correct, but the requests for the feature flag could also be reduced if the result is received at an early stage and then passed into this assertion method, so I wanted to check

Avishagp (approved):

jan-stehlik (approved):
nice one @ofekatr :tada:

PR #2341, opened 2021-11-03
feat: support yarn workspaces projects in --all-projects

lili2311:

#### What does this PR do?

Support detecting & scanning Yarn workspace projects `--all-projects`.
Process all projects as yarn workspaces first since many files are related to each other and could belong to a workspace, then process the rest as usual.

https://github.com/snyk/snyk/pull/2341/files#diff-2194ffc6b9f5b3533db3a8c636c6fcde3d7f163e92de6d915a5dc542a576fb82R48

Run `snyk test --all-projects` in a yarn workspaces projects

https://github.com/snyk/snyk/issues/1561

##### Scenario 1:
`snyk test --all-projects` detects all the Yarn workspace projects + the non-workspace project.
CleanShot 2021-11-15 at 16 44 40

##### Scenario 2:

`snyk test --all-projects --json` has the same output all the Yarn workspace projects via `--yarn-workspaces` with the addition of the non-workspace project.

CleanShot 2021-11-15 at 16 47 29

##### Scenario 3:

Out of sync error during `--yarn-workspaces` scan is resulting in the same error code and behaviour as an error in another project type during `--all-projects`

**Note**: `--all-projects` by default does not log the errors and instead shows them in debug mode. So the difference in the error seen is expected.

CleanShot 2021-11-19 at 10 18 42

No supported manifests error during `--yarn-workspaces` scan is resulting in the same error code and message as during `--all-projects`

##### Scenario 4:

CleanShot 2021-11-19 at 14 49 47

##### Scenario 5:
`snyk-dev test --all-projects  --strict-out-of-sync=false` and `snyk-dev test --yarn-workspaces  --strict-out-of-sync=false` both detect and scan all the projects in a yarn workspace monorepo

CleanShot 2021-11-19 at 15 00 17

##### Scenario 6 ⚠️ :
`--all-projects` does not fail if 1 project fails, it logs it and keeps going. So scanning a out of sync workspace now follows the same login when scanned with `--all-projects` to ensure backwards compatible & expected behaviour. The same behaviour as with `--yarn-workspaces` can be achieved by adding `--fail-fast` flag.

CleanShot 2021-11-19 at 15 54 29

##### Scenario 7:
Exit codes match when 100% of the projects fail to scan via `--all-projects` and `--yarn-workspaces`

CleanShot 2021-11-19 at 16 12 45

##### Scenario 8:
`--all-projects --dev --json` matches the data from `--yarn-workspaces --dev --json` + finds an extra non workspace projects.

⚠️ need to check why the patched date has a slight diff.

CleanShot 2021-11-19 at 16 17 17

##### Scenario 9:
`snyk monitor --all-projects` matches the same projects  from `snyk monitor --yarn-workspaces` + applies all the Snyk org ignores

CleanShot 2021-11-22 at 11 12 11

### DIFF
--- a/src/lib/plugins/get-multi-plugin-result.ts
+++ b/src/lib/plugins/get-multi-plugin-result.ts
@@ -147,6 +147,7 @@ async function processYarnWorkspacesProjects(
     );
     return { scannedProjects, unprocessedFiles };
   } catch (e) {
+    debug('ERROR: ', e)
     return { scannedProjects: [], unprocessedFiles: targetFiles };
   }
 }
--- a/src/lib/plugins/get-multi-plugin-result.ts
+++ b/src/lib/plugins/get-multi-plugin-result.ts
@@ -147,7 +147,7 @@ async function processYarnWorkspacesProjects(
     );
     return { scannedProjects, unprocessedFiles };
   } catch (e) {
-    debug('ERROR: ', e)
+    debug('Error during detecting or processing Yarn Workspaces: ', e);
     return { scannedProjects: [], unprocessedFiles: targetFiles };
   }
 }
--- a/src/lib/plugins/get-multi-plugin-result.ts
+++ b/src/lib/plugins/get-multi-plugin-result.ts
@@ -1,5 +1,7 @@
 const cloneDeep = require('lodash.clonedeep');
-import * as path from 'path';
+import * as pathLib from 'path';
+import sortBy = require('lodash.sortby');
+import groupBy = require('lodash.groupby');
 import * as cliInterface from '@snyk/cli-interface';
 import chalk from 'chalk';
 import { icon } from '../theme';
@@ -14,6 +16,7 @@ import { convertMultiResultToMultiCustom } from './convert-multi-plugin-res-to-m
 import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
 import { CallGraph } from '@snyk/cli-interface/legacy/common';
 import { errorMessageWithRetry, FailedToRunTestError } from '../errors';
+import { processYarnWorkspaces } from './nodejs-plugin/yarn-workspaces-parser';
 
 const debug = debugModule('snyk-test');
 export interface ScannedProjectCustom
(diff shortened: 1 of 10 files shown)


src/lib/plugins/get-multi-plugin-result.ts
```diff
@@ -107,3 +120,67 @@ export async function getMultiPluginResult(
...
+      base: string;
+      dir: string;
+    }>;
+  } = groupBy(sorted, 'dir');
+
+  const scanned = scannedProjects.map((p) => p.targetFile!);
+  const targetFiles: string[] = [];
+
+  for (const directory of Object.keys(targetFilesByDirectory)) {
+    const { base } = targetFilesByDirectory[directory][0];
```
admons, line 170: The current behavior will ignore other files in the same directory, so if you have a `requirements.txt` file in the same directory, it will not return in the result.

What do you think about this implementation for removing scanned files:
```typescript
  const scannedFiles = scannedProjects.map((p) => [
    path.resolve(options.path, p.targetFile!),
    path.resolve(options.path, p.targetFile!.replace('package.json', 'yarn.lock')),
  ]).flat();

  return allTargetFiles
    .filter(f => !scannedFiles.includes(f));
```
1. map the scanned projects to their targetFile (which is always a package.json)
2. add to this list a dummy "yarn.lock" file in the same place that the package.json is
3. remove those files from the original list (allTargetFiles)

admons (commented):
Please see my comments

src/lib/plugins/get-multi-plugin-result.ts
lili2311: hmm weird I fixed that somewhere, let me find and push 😓

lili2311: fixed

lili2311:
Ready for review, still have some manual verification left to do before it can be merged

lili2311:
QA completed, please see the PR description for tested scenarios & their results

JamesPatrickGill (approved):

src/lib/plugins/get-multi-plugin-result.ts
```diff
@@ -78,16 +89,18 @@ export async function getMultiPluginResult(
...
+    } catch (error) {
+      const errMessage =
+        error.message ?? 'Something went wrong getting dependencies';
       // TODO: propagate this all the way back and include in --json output
       failedResults.push({
         targetFile,
-        error: err,
-        errMessage: err.message || 'Something went wrong getting dependencies',
+        error,
+        errMessage: errMessage,
```
jahed-snyk, line 99: Key and value are the same. Might as well squash this like the other keys.

jahed-snyk (approved):
Not much here to review from a Hammer POV. The ownership rules aren't quite right as Hammer shouldn't be owning the package-lock.json within the fixture. Hammer will fix this.

candrews:
It sounds like, with this change, `-all-projects` implies `--yarn-workspaces`

If so, can you update the documentation at https://docs.snyk.io/features/snyk-cli/guides-for-our-cli/cli-reference#options to explain that `-all-projects` shouldn't be used with `--yarn-workspaces` because `-all-projects` always behaves as if `--yarn-workspaces` is provided?

PR #2353, opened 2021-11-10
feat: Improve output for dockerfile instructions

pecodez:

#### What does this PR do?
This PR allows for use of the installCommand, extracted from a RUN instruction in newer versions of the SDP, to provide a better output for dockerFile user instructions.

https://snyksec.atlassian.net/browse/CAP-280

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -445,7 +445,7 @@ async function parseRes(
       const dockerfilePackage = dockerfilePackages[vuln.name.split('/')[0]];
       if (dockerfilePackage) {
         (vuln as DockerIssue).dockerfileInstruction =
-          dockerfilePackage.instruction;
+          dockerfilePackage.installCommand;
       }
       (vuln as DockerIssue).dockerBaseImage = res.docker!.baseImage;
       return vuln;
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
         "semver": "^6.0.0",
         "snyk-config": "4.0.0",
         "snyk-cpp-plugin": "2.13.0",
-        "snyk-docker-plugin": "4.26.1",
+        "snyk-docker-plugin": "4.27.0",
         "snyk-go-plugin": "1.18.0",
         "snyk-gradle-plugin": "3.17.0",
         "snyk-module": "3.1.0",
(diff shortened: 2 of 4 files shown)


agatakrajewska:
I think we actually could release the SDP first, and add that bump here, so we only raise the cli PR once

admons (approved):

src/lib/snyk-test/run-test.ts
```diff
@@ -445,7 +445,7 @@ async function parseRes(
       const dockerfilePackage = dockerfilePackages[vuln.name.split('/')[0]];
       if (dockerfilePackage) {
         (vuln as DockerIssue).dockerfileInstruction =
-          dockerfilePackage.instruction;
+          dockerfilePackage.installCommand ?? dockerfilePackage.instruction;
```
jahed-snyk, line 448: Why do we only use "instruction" if "installCommand" is defined? Do we need a test for this to explain the scenario for future changes/reviewers?

It seems like this logic should be in "snyk-docker-plugin" rather than here.

pecodez: Hey, yeah you're right! Initially this was backwards compatible and then we decided to include the commit for the SDP bump (so no longer needed). I have amended the commit and also the associated tests.

src/lib/snyk-test/run-test.ts
```diff
@@ -445,7 +445,7 @@ async function parseRes(
       const dockerfilePackage = dockerfilePackages[vuln.name.split('/')[0]];
       if (dockerfilePackage) {
         (vuln as DockerIssue).dockerfileInstruction =
-          dockerfilePackage.instruction;
+          dockerfilePackage.installCommand;
```
jahed-snyk, line 448: Should we mass find/replace `dockerfileInstruction` to `dockerfileInstallCommand` to avoid confusion?

jahed-snyk (approved):
Looks good! I'd suggest find/replacing all usages of "instruction" in this context to "install command" in test names, variable names, etc. It's confusing/inconsistent.

src/lib/snyk-test/run-test.ts
pecodez: I think it's ok in this instance. The install command is extracted from the instruction and simply provides a cleaner output, so is still within context here.

jahed-snyk: Sounds good. 👍

PR #2366, opened 2021-11-16
refactor: migrate spinner to use import/export

jahed-snyk:
Considering NodeJS more or less supports ES modules now natively, I was checking to see how close we are to building CLI using ES modules. The only thing stopping us TypeScript-wise seems to be the `spinner` module. So I've migrated that. The only thing left after this is the various JS modules.

## Notes

There's some odd logic around `isRequired`. I think this was for tests and import flows which we no longer provide. So it's no longer needed. See: #2371 

The `isCI` check is also kind of weird. It looks like we can check the environment if it's TTY instead to decide whether or not to render the spinner. Kind of like how we render colours.

I don't want to get too deep into improving `spinner`'s code/types as we may want to replace it with a library. Especially if we want to show better progress updates (e.g. percentages, fractions, etc.).


src/lib/spinner.ts
```diff
@@ -1,16 +1,13 @@
...
-const debug = debugModule('snyk:spinner');
 import { isCI } from './is-ci';
+
+const debug = debugModule('snyk:spinner');
 const spinners = {};
 let sticky = false;
 let handleExit = false;
 
-function createSpinner(label: string): Promise<void> {
+function addSpinner(label: string): Promise<void> {
```
maxjeffos, line 10: Should explicitly tag this function with `async` IMO

src/lib/spinner.ts
```diff
@@ -57,15 +56,15 @@ createSpinner.clear = <T>(label): ((valueToPassThrough: T) => T) => {
...
   Object.keys(spinners).map((lbl) => {
-    createSpinner.clear<void>(lbl)();
+    addSpinner.clear<void>(lbl)();
   });
 };
 
 // taken from http://git.io/vWdUm and modified
-function spinner(opt: SpinnerOptions) {
-  if (module.exports.isRequired || isCI()) {
+function createSpinner(opt: SpinnerOptions) {
```
maxjeffos, line 66: Could give this function a return type.

maxjeffos (approved):
Made a couple suggestions but approved it.

Arvi3d (approved):

PR #2367, opened 2021-11-16
refactor: remove main cli export

jahed-snyk:
We no longer export "main" in our package.json so we don't need this export.


maxjeffos (approved):
I guess you don't really need this PR anymore since https://github.com/snyk/snyk/pull/2368/files includes this same change but approved it anyway so you can merge as you prefer.

PR #2368, opened 2021-11-16
fix: handle unexpected errors

jahed-snyk:
Ensure unexpected errors are handled with Exit Code 2 rather than the NodeJS default 1. We use 1 for vulns found.

Node 16 now uses Exit Code 1 for unhandledRejections (rather than only logging), so this also makes us consistent across different Node versions.

https://nodejs.org/en/blog/release/v15.0.0/#throw-on-unhandled-rejections-33021

## Note for Reviewers

When reviewing, it's best to review through each commit.

I've moved a file and placed a new file in its old place. To avoid losing history, these two steps are in separate commits.

If I didn't do this, git will squash the two changes to look like I deleted a bunch of code and created a new file with a bunch of new code. Obviously, that's not correct. This is what GitHub's PR interface also shows, but it's not how it looks in history with separate commits.

## Notes

- [x] Add a test for it. Probably can't test via CLI acceptance tests or unit tests as it requires "unexpected" errors and the process needs to be exiting. Probably can write some sort of separate acceptance test which uses the same helper function.


maxjeffos (approved):
Great change

src/lib/unexpected-error.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Ensures a given function does not throw any errors, including unexpected ones
+ * outside of its execution.
+ *
+ * This function can only be used once in the same process. If you have multiple
+ * callables needing this, compose them into a single callable.
+ */
+export async function callHandlingUnexpectedErrors(
```
jahed-snyk, line 8: @maxjeffos how about "safelyCall". Seems like a most minimal solution to naming this (vs. inventing a DSL)

src/lib/unexpected-error.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Do NOT remove 'uncaughtException' and 'unhandledRejection' handlers
+   * at any point. There's never a point when everything's "done" and its safe
+   * to remove them.
+   *
+   * These events only get called when NodeJS is about to exit the process. i.e.
+   * when there's nothing left to execute. Adding steps to remove them gives
+   * NodeJS something to execute. So they'll never get called when they need to!
+   */
```
jahed-snyk, line 52: Need to improve this explanation. Provide a sequence of events to make the problem clearer.

jahed-snyk (commented):
Self-review based on in-person discussions.

PR #2372, opened 2021-11-16
refactor: use NodeJS env var for insecure mode

jahed-snyk:
Currently we're using global state and needle's option to setup insecure mode.

However, since Needle uses TLS module defaults, we can apply the same logic by using NodeJS's NODE_TLS_REJECT_UNAUTHORIZED env var.

https://nodejs.org/docs/latest/api/cli.html#node_tls_reject_unauthorizedvalue

The benefit is that we don't need to test it across multiple modules. This is a NodeJS feature which we can switch on globally. This also applies to any requests not going through our Needle wrapper so it's more thorough.


src/cli/args.ts
```diff
@@ -277,7 +271,12 @@ export function args(rawArgv: string[]): Args {
   }
 
   if (argv.insecure) {
```
RA80533, line 273: Just out of curiosity, what is the use case for not validating TLS certificates?

src/cli/args.ts
```diff
@@ -277,7 +271,12 @@ export function args(rawArgv: string[]): Args {
...
-    global.ignoreUnknownCA = true;
+    // https://nodejs.org/docs/latest/api/cli.html#node_tls_reject_unauthorizedvalue
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
RA80533, line 275: Is the logic here meant to be the other way around? E.g., if `NODE_TLS_REJECT_UNAUTHORIZED` is set to be disabled, `insecure` mode is on?

jahed-snyk: @RA80533 `NODE_TLS_REJECT_UNAUTHORIZED` by default is on. Rejecting unauthorized certs is the safe approach. So `0` turns it off for customers who need it.

Yes, module load order matters, but this was the previously the case too. We already ensure args.ts is loaded first to avoid this. Fixing module order requirements and reducing side effects is a larger issue to resolve, beyond the scope of this PR.

src/cli/args.ts
jahed-snyk: Node.js's CA handling is limited, especially on Windows. So some customers may prefer to use it but we don't recommend it. See: https://support.snyk.io/hc/en-us/articles/360000925358-How-can-I-use-Snyk-behind-a-proxy-

darscan:
Closing as stale

PR #2374, opened 2021-11-17
fix: Remove http protocols validation in custom rules pull.

ofekatr:
#### What does this PR do?

- Changes the existence of protocols in OCI registry URLs to be optional

- `src/cli/commands/test/iac-local-execution/index.ts`

1. Make sure your org has the `iacCustomRules` and the `infrastructureAsCode` feature flags.
2. Make sure your org has the `iacCustomRulesEntitlement` entitlement.
3. Prepare a custom-rules bundle generated by the [Snyk custom-rules bundles generator](https://github.com/snyk/snyk-iac-rules). 
4. Push the generated bundle to an OCI registry. You can refer to [our docs](https://docs.snyk.io/products/snyk-infrastructure-as-code/custom-rules/getting-started-with-the-sdk/pushing-a-bundle) to learn how.
. In the Snyk CLI, run `snyk-dev iac test <file> --path=<path-to-remote-bundle>`, where the path does not contain a protocol.
5. Verify that:
    - No errors were thrown.
    - The defined custom rules generated the expected issues successfully.

At the moment we impose that customers provide an OCI registry URL with the http:// or https:// protocol in front. When we push, on the SDK side, we don’t have that. Customers may copy the URL from the push into the UI/CLI, which won’t work. We don’t need the protocol actually so this task is to remove the need for it. We should still allow customers to provide.

- [CFG-1183](https://snyksec.atlassian.net/browse/CFG-1183)

#### Additional notes for reviewer

- During the task it's been discovered that our current URL validation method - using the [WHATWG URL API](https://nodejs.org/api/url.html#the-whatwg-url-api) is very basic, and validates most URLs passed to it. Some examples are:
    - `http://    $  `
    - `z://6`
    - `a://`

    Meanwhile, valid URIs like www.google.com will not pass due to the missing protocol.

### DIFF
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -50,6 +50,7 @@ import {
   UnsupportedFeatureFlagPullError,
 } from './oci-pull';
 import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-entitlement-error';
+import { isValidUrl } from './url-utils';
 
 // this method executes the local processing engine and then formats the results to adapt with the CLI output.
 // this flow is the default GA flow for IAC scanning.
@@ -157,16 +158,6 @@ export function removeFileContent({
   };
 }
 
-export function isValidURL(str: string): boolean {
-  let url;
-  try {
-    url = new URL(str);
-  } catch (e) {
-    return false;
-  }
-  return url.protocol === 'http:' || url.protocol === 'https:';
-}
-
 /**
  * Checks if the OCI registry URL has been provided.
  */
@@ -197,10 +188,6 @@ function getOCIRegistryURLComponentsFromSettings(
 ) {
   const settingsOCIRegistryURL = iacOrgSettings.customRules!.ociRegistryURL!;
 
-  if (!isValidURL(settingsOCIRegistryURL)) {
-    throw new InvalidRemoteRegistryURLError();
-  }
-
   return {
     ...extractOCIRegistryURLComponents(settingsOCIRegistryURL),
     tag: iacOrgSettings.customRules!.ociRegistryTag || 'latest',
@@ -213,7 +200,7 @@ function getOCIRegistryURLComponentsFromSettings(
 function getOCIRegistryURLComponentsFromEnv() {
   const envOCIRegistryURL = userConfig.get('oci-registry-url')!;
 
-  if (!isValidURL(envOCIRegistryURL)) {
+  if (!isValidUrl(envOCIRegistryURL)) {
     throw new InvalidRemoteRegistryURLError();
   }
 
--- a/src/cli/commands/test/iac-local-execution/oci-pull.ts
+++ b/src/cli/commands/test/iac-local-execution/oci-pull.ts
@@ -22,15 +22,31 @@ export function extractOCIRegistryURLComponents(
   OCIRegistryURL: string,
 ): OCIRegistryURLComponents {
   try {
-    const url = new URL(OCIRegistryURL);
-    const registryBase = url.hostname;
-    // with or without a path, there will at least be a / in the pathname
-    const repoWithTag = url.pathname.substring(1);
+    const urlWithoutProtocol = OCIRegistryURL.includes('://')
+      ? OCIRegistryURL.split('://')[1]
+      : OCIRegistryURL;
+
+    const firstSlashIdx = urlWithoutProtocol.indexOf('/');
+    if (firstSlashIdx === -1) {
+      throw new InvalidRemoteRegistryURLError(OCIRegistryURL);
+    }
+
+    const [registryHost, repoWithTag] = [
+      urlWithoutProtocol.substring(0, firstSlashIdx),
+      urlWithoutProtocol.substring(firstSlashIdx + 1),
+    ];
+    if (!registryHost || !repoWithTag) {
+      throw new InvalidRemoteRegistryURLError(OCIRegistryURL);
+    }
 
     const [repo, tag = 'latest'] = repoWithTag.split(':');
-    return { registryBase, repo, tag };
+    if (!repo) {
+      throw new InvalidRemoteRegistryURLError(OCIRegistryURL);
+    }
+
+    return { registryBase: registryHost, repo, tag };
   } catch {
-    throw new InvalidRemoteRegistryURLError();
+    throw new InvalidRemoteRegistryURLError(OCIRegistryURL);
   }
 }
 
@@ -106,12 +122,13 @@ export class InvalidManifestSchemaVersionError extends CustomError {
 }
 
 export class InvalidRemoteRegistryURLError extends CustomError {
-  constructor(message?: string) {
-    super(message || 'Invalid URL for Remote Registry');
+  constructor(url?: string) {
+    super('Invalid URL for Remote Registry');
     this.code = IaCErrorCodes.InvalidRemoteRegistryURLError;
     this.strCode = getErrorStringCode(this.code);
-    this.userMessage =
-      'The Remote Registry URL is invalid, or does not include a http/https protocol. Please check it again.';
+    this.userMessage = `The provided remote registry URL${
+      url ? `: "${url}"` : ''
+    } is invalid. Please check it again.`;
   }
 }
 
--- a/src/cli/commands/test/iac-local-execution/url-utils.ts
+++ b/src/cli/commands/test/iac-local-execution/url-utils.ts
@@ -1,0 +1,8 @@
+const URL_REGEX = /^((?:https?:\/\/)?[^./]+(?:\.[^./]+)+(?:\/.*)?)$/;
+
+/**
+ * Checks if the provided URL string is valid.
+ */
+export function isValidUrl(urlStr: string): boolean {
+  return URL_REGEX.test(urlStr);
+}
--- a/test/jest/unit/iac-unit-tests/oci-pull.spec.ts
+++ b/test/jest/unit/iac-unit-tests/oci-pull.spec.ts
@@ -22,6 +22,18 @@ describe('extractOCIRegistryURLComponents', () => {
       tag: 'latest',
     });
   });
+
+  it('extracts components from URL without protocol', async () => {
+    const expected = extractOCIRegistryURLComponents(
+      'gcr.io/user/repo-test:0.5.2',
+    );
+    expect(expected).toEqual({
+      registryBase: 'gcr.io',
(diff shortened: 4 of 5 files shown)


ipapast:
I am a bit concerned of the code complexity we are adding with this approach as we've discussed yesterday - I added @teodora-sandu to have a look at it for a third opinion if that's ok?

ofekatr:
> I am a bit concerned of the code complexity we are adding with this approach as we've discussed yesterday - I added @teodora-sandu to have a look at it for a third opinion if that's ok?

Absolutely.

ipapast (commented):
For visibility: 
My current thought for this is that we can not control what is a valid URL 100%.
- If users provide an https protocol, we can validate it via the URL constructor.
- If users don't provide a protocol, we can not be sure. Let's assume they type giberrish. It will be a similar case to when a user made a typo. In both cases (giberrish url, a typo in the url) we won't know until we call the docker registry client and try to get the manifests. this will fail and we will return an error in which we prompt the user to check the URL provided (FailedToPullCustomBundleError).

So I think my suggestion is to treat the URL without a protocol the same as a URL with a typo and maybe update the error message to be something more meaningful.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -195,9 +186,11 @@ function checkOCIRegistryURLExistsInSettings(
 function getOCIRegistryURLComponentsFromSettings(
   iacOrgSettings: IacOrgSettings,
 ) {
-  const settingsOCIRegistryURL = iacOrgSettings.customRules!.ociRegistryURL!;
+  let settingsOCIRegistryURL = iacOrgSettings.customRules!.ociRegistryURL!;
+
+  settingsOCIRegistryURL = normalizeUrl(settingsOCIRegistryURL);
```
teodora-sandu, line 191: Small note before I forget, but not something introduced in this PR. We actually already validate this in the UI/API, which is the only way we'd have these settings. So we're doing extra work here by validating I think

teodora-sandu:
I've given this some thought and I think it might be easier to reason with if we don't add the protocol ourselves and so remove the URL validation altogether. We then leave the validation to the registry client, like Ilianna says. We get rid of the protocol anyway for it and so the flow of appending the protocol only to then remove it is a bit unintuitive.

If we let the registry client throw the error and we just improve the error a bit by printing out the configured URL and making it clear it's invalid, then we can remove the URL validation altogether and just modify `extractOCIRegistryURLComponents` to remove the protocol if provided.

Also, since we already have validation in the UI and API, validating it here as well is redundant. We can remove this custom code and make our code easier to read for future developers. WDYT?

ofekatr:
> I've given this some thought and I think it might be easier to reason with if we don't add the protocol ourselves and so remove the URL validation altogether. We then leave the validation to the registry client, like Ilianna says. We get rid of the protocol anyway for it and so the flow of appending the protocol only to then remove it is a bit unintuitive.
> 
> If we let the registry client throw the error and we just improve the error a bit by printing out the configured URL and making it clear it's invalid, then we can remove the URL validation altogether and just modify `extractOCIRegistryURLComponents` to remove the protocol if provided.
> 
> Also, since we already have validation in the UI and API, validating it here as well is redundant. We can remove this custom code and make our code easier to read for future developers. WDYT?

1. Since the validation for the UI and API will be changed in a similar fashion, it's problematic to assume a validation at these stages.

2. Since the URL constructor is used the `extractOCIRegistryURLComponents` throws an error when a protocol does not exist, this parsing stage would have to be either:
- Adding a dummy protocol for the sake of being able to use this constructor.
- Creating our own logic for parsing the required parts for the URL.
Which of these would be more adequate?

teodora-sandu:
We can discuss it in a call tomorrow if it helps. The way I'm looking at it is we have two parts of the workflow:
1. configuring the custom rules settings: UI, API, `snyk config set` -> we want to keep the validation in the UI and API and just make it less restrictive on there being a protocol; `snyk config set` has no validation sadly
2. running the `snyk` CLI, so this PR -> the validation in this part of the code is more useful for the `snyk config set` scenario, but since we already have errors from the downstream and we don't gain anything extra by adding our own validation (other than complexity without even being able to guarantee that we can validate every bad edge case), I think we would make our lives easier by keeping this simple

We only use the `URL` constructor in `extractOCIRegistryURLComponents` for simplicity of pulling out the hostname and pathname but that can be done with a regex, like it was before: https://github.com/snyk/snyk/pull/2337/files

src/cli/commands/test/iac-local-execution/url-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+export function normalizeUrl(
+  urlStr: string,
+  { protocol = 'https' }: NormalizeURLOptions = {},
+): string {
+  if (!urlStr.match(HTTP_PROTOCOLS_REGEX)) {
+    urlStr = `${protocol}://${urlStr}`;
+  }
+
+  return urlStr;
+}
```
ipapast, line 31: here we are mutating the user's input, even though we allow them to send us a URL without a protocol in the first place. It doesn't feel right to do this for me.

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -211,9 +204,11 @@ function getOCIRegistryURLComponentsFromSettings(
  * Extracts the OCI registry URL components from the environment variables.
  */
 function getOCIRegistryURLComponentsFromEnv() {
-  const envOCIRegistryURL = userConfig.get('oci-registry-url')!;
+  let envOCIRegistryURL = userConfig.get('oci-registry-url')!;
+
+  envOCIRegistryURL = normalizeUrl(envOCIRegistryURL);
 
-  if (!isValidURL(envOCIRegistryURL)) {
+  if (!validateUrl(envOCIRegistryURL)) {
```
ipapast, line 211: nit: The usual convention to name methods that return boolean is to prefix verbs like 'is' or 'has' to the predicate as a question.`isValidURL` used to conform to that. any reason for the rename?

ofekatr: It was dones out of the attempt to follow a convention of naming functions with clear verbs like `get`, `check`, etc., but after reading further into it it seems the more common convention for boolean functions is verbs like `is` and `has` as you say.
Good call, will be changed to `isValidUrl`

src/cli/commands/test/iac-local-execution/url-utils.ts
ofekatr: Removed, the parsing was adjusted to not require this

src/cli/commands/test/iac-local-execution/oci-pull.ts
```diff
@@ -22,12 +23,19 @@ export function extractOCIRegistryURLComponents(
   OCIRegistryURL: string,
 ): OCIRegistryURLComponents {
   try {
-    const url = new URL(OCIRegistryURL);
-    const registryBase = url.hostname;
-    // with or without a path, there will at least be a / in the pathname
-    const repoWithTag = url.pathname.substring(1);
+    if (!isValidUrl(OCIRegistryURL)) {
```
teodora-sandu, line 26: Why do we validate here after validating in `getOCIRegistryURLComponentsFromEnv` before we call this function?

And what happens when no protocol is provided here? Will `OCIRegistryURL.split('://')[1]` not throw an error?

ofekatr: Validation was removed in the latest commit, the case described was handled and a test case has been added 👍

test/jest/unit/iac-unit-tests/oci-pull.spec.ts
```diff
@@ -57,7 +69,7 @@ describe('extractOCIRegistryURLComponents', () => {
 
   it('throws an error if URL is invalid', () => {
     expect(() => {
-      extractOCIRegistryURLComponents('url/not/valid');
```
ofekatr, line 60: Changed, as this function no longer validates URLs with the URL constructor, but rather fails if there're no slashes between the hostname and the path.

ofekatr, line 60: Changed, as this function no longer validates URLs with the URL constructor, but rather fails if there're no slashes between the hostname and the path.

src/cli/commands/test/iac-local-execution/oci-pull.ts
```diff
@@ -22,15 +22,31 @@ export function extractOCIRegistryURLComponents(
   OCIRegistryURL: string,
 ): OCIRegistryURLComponents {
   try {
-    const url = new URL(OCIRegistryURL);
-    const registryBase = url.hostname;
-    // with or without a path, there will at least be a / in the pathname
-    const repoWithTag = url.pathname.substring(1);
+    const urlWithoutProtocol = OCIRegistryURL.includes('://')
```
teodora-sandu, line 25: nit: I think what this code is saying is we want to avoid `urlWithoutProtocol` being either `''`, `'/blah'` or `'blah/'`. So if you change the `if (firstSlashIdx === -1) {` condition to `if (firstSlashIdx <= 0 || firstSlashIdx === urlWithoutProtocol.length - 1) {` then we are guaranteed that `registryHost` and `repoWithTag` will not be empty - so you won't need to check `if (!registryHost || !repoWithTag) {` as well later.

test/jest/unit/iac-unit-tests/url-utils.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { isValidUrl } from '../../../../src/cli/commands/test/iac-local-execution/url-utils';
+
+describe('url-utils.ts', function() {
+  describe('isValidUrl', function() {
+    describe('When given a valid URL', function() {
```
teodora-sandu, line 5: Could you also add a test that shows a URL with the tag at the end works? Since we're using a regular expression it'd be easier in future to look at these tests and see exactly what URLs are accepted

teodora-sandu (approved):
Thanks for fixing everything, looks great! ⭐

src/cli/commands/test/iac-local-execution/oci-pull.ts
ofekatr: This is a great observation, although it wouldn't capture cases where a colon is provided for the tag, e.g., `https://registry/:`, `https://registry/:latest`.
I think the `if (!registryHost || !repoWithTag)` condition also  makes it slightly more readable.
For the sake of not missing unique cases, I'd prefer not to change this for the time being. Would that be alright?

PR #2378, opened 2021-11-19
fix: unify the retry & contact support errors across commands

lili2311:

#### What does this PR do?
Use consistent message for retry & contact where we ask the user to try again but with debug enabled.

src/lib/common.ts

### DIFF
--- a/src/lib/plugins/get-multi-plugin-result.ts
+++ b/src/lib/plugins/get-multi-plugin-result.ts
@@ -13,8 +13,7 @@ import { convertSingleResultToMultiCustom } from './convert-single-splugin-res-t
 import { convertMultiResultToMultiCustom } from './convert-multi-plugin-res-to-multi-custom';
 import { PluginMetadata } from '@snyk/cli-interface/legacy/plugin';
 import { CallGraph } from '@snyk/cli-interface/legacy/common';
-import { FailedToRunTestError } from '../errors';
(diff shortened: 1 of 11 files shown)


src/lib/plugins/get-multi-plugin-result.ts
```diff
@@ -95,7 +96,7 @@ export async function getMultiPluginResult(
 
   if (!allResults.length) {
     throw new FailedToRunTestError(
-      `Failed to get dependencies for all ${targetFiles.length} potential projects. Run with \`-d\` for debug output and contact support@snyk.io`,
+      `Failed to get dependencies for all ${targetFiles.length} potential projects.\n${reTryMessage}. ${contactSupportMessage}`,
```
jahed-snyk, line 99: There's a `.` after retryMessage here (the others don't have that).

jahed-snyk (approved):
Looks good. Instead of exporting multiple strings and concatenating, maybe a single function will help maintain consistency, avoid typos and make it easier to modify in the future. Something like:

```ts
export function createRetryMessage(reason: string): string {
  return [reason, 'retry ...', 'contact ...'].join('\n');
}
```

lili2311:
yes sure can look at adding a function that creates the string. But will keep is as a string literal instead of an array join which is harder to scan.

lili2311:
@jahed-snyk updated

PR #2386, opened 2021-11-24
feat: IaC Snapshot PoC [CFG-1311]

YairZ101:

### DIFF
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -22,6 +22,7 @@ const keys: (keyof IaCTestFlags)[] = [
   'quiet',
   'scan',
   'legacy',
+  'report',
 
   // PolicyOptions
   'ignore-policy',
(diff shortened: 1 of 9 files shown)


src/cli/commands/test/iac-local-execution/analytics.ts
```diff
@@ -55,6 +56,7 @@ export function addIacAnalytics(
     computeCustomRulesBundleChecksum(),
   );
   analytics.add('iac-custom-rules-coverage-count', uniqueCustomRulesCount);
+  analytics.add('iac-cli-report', cliReport);
```
ipapast, line 59: I'd argue if this is neccessary addition, as we can already get the flag's presence from the `args`. Everytime we add a new analytics field it ends up on a new column in BigQuery. I think as we only check for a boolean value here and we don't calculate anyhting else, we can leave it simple and get it from the args. 

What do you think?

e.g.:

```
 snyk analytics {
  "args": [
    "test/fixtures/iac/terraform/sg_open_ssh.tf",
    {
      "report": true,
      "debug": true,
      "iac": true
    }
  ],
...
}
```

YairZ101: great suggestion, definitely agree with you.

ipapast (commented):
This looks good! We tested it and discussed some things with @YairZ101:
- add some tests 
- add an error message if the report flag is not present 
- share the URL of the projects page in the end of the scan when we use the --report flag, 

otherwise we think it's in a very good state 🪄

PR #2395, opened 2021-11-29
test: migrate json-file-output to jest

Avishagp:
Migrate `json-file-output` test from tap to jest.


test/jest/unit/json-file-output.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+});
+
+afterAll(() => {
+  cleanupOutputDirsAndFiles();
+});
+
+describe('json file output test', () => {
+  it('createDirectory returns true if directory already exists - non-recursive', () => {
+    // initially create the directory
+    fs.mkdirSync(testOutputFull);
```
jahed-snyk, line 73: We should use `createProject` instead. It'll handle setup/cleanup.

test/jest/unit/json-file-output.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    expect(fs.existsSync(testOutputFull)).toBe(true);
+  });
+
+  it('createDirectory creates directory - recursive', () => {
+    // attempt to create the directory requiring recursive
+
+    // recursive should fail (return false) for node < 10 LTS and pass (return true) for node >= 10 LTS
+    // if node >= 10, verify that the deep folder was created
+    // if node 8 verify that the deep folder was not created
```
jahed-snyk, line 86: We don't support Node 8 so we can remove the branching logic.

test/jest/unit/json-file-output.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    // this should throw an error within writeContentsToFileSwallowingErrors but that error should be caught, logged, and disregarded
+    await writeContentsToFileSwallowingErrors(
+      testOutputFileFull,
+      'fake-contents',
+    );
+    expect(fs.existsSync(testOutputFileFull)).toBe(true);
+  });
+
+  test('writeContentsToFileSwallowingErrors captures any errors when attempting to write to a readonly directory', async () => {
+    if (!iswindows) {
```
jahed-snyk, line 110: Not sure why we need this. Might be due to setup/teardown. If we use `createProject`, this shouldn't be needed.

test/jest/unit/json-file-output.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        // create a directory without write permissions
+        fs.mkdirSync(readonlyFull, 0o555);
+
+        const outputPath = pathLib.join(readonlyFull, 'test-output.json');
+
+        await writeContentsToFileSwallowingErrors(outputPath, 'fake-contents');
+        expect(fs.existsSync(outputPath)).toBe(false);
+      } catch (err) {
+        throw new Error('Should not throw exception'); // we expect to not get an error even though we can't write to this folder
```
jahed-snyk, line 123: This re-throw is pointless. We can remove the try/catch and it'll do the same thing with a more useful failure message. Alternatively, 

```
await expect(writeContentsToFileS...).resolves.toBeUndefined()
```

should work too.

test/jest/unit/json-file-output.spec.ts
```diff
@@ -0,0 +1,127 @@
+/* eslint-disable jest/no-conditional-expect */
```
jahed-snyk, line 1: We should avoid adding these as it's like skipping tests and can make us ambivalent towards the linter. Unless if we absolutely have to, in which case, we should include a reason comment and put it close to the source rather than skipping the entire file.

jahed-snyk (approved):
Looks good from a Tap -> Jest perspective as it's improving what we have.

I think we should do more here to reduce the complexity and avoid it polluting other Jest tests. Otherwise, we are kind of transfering a disease from Tap to Jest.

I've left some comments on how we can achieve that. Whether in this PR or a follow-up one.

test/jest/unit/json-file-output.spec.ts
Avishagp: Kinda stuck on migrating the test to use `createProject`. We can pair on that tomorrow if you can

Avishagp: Looking at the test and at `json-file-output' I'm not sure why do you think that'd be necessary, let's pair on that when you're here

jahed-snyk (approved):

PR #2396, opened 2021-11-29
feat: add snyk apps command and create sub-command

love-bhardwaj:

#### What does this PR do?

Adds `$snyk apps` command to the Snyk CLI. This command is available only under the `--experimental` flag. Example: `$snyk apps create --experimental`

The current PR only focuses on the `create` sub-command. There are two modes of the command 
1. `interactive` mode available with the `--interactive` or the `-i` flag. Example: `$ snyk apps create -i --experimental`
2. `scriptable` mode which is the default but has required options. Example: `$ snyk apps create --experimental --org=<your-org-id> --name=<your-snyk-app-name> --redirect-uris=https://example1.com,HTTP://example2.com` --scopes=apps:beta` 

Changes that this PR makes:
- `create` command to create a Snyk App
- `config` value `V3endpoint` to configure the Snyk V3 API endpoint base URL
- Initial acceptance to test the `create` command

- `src/cli/commands/apps`

We at `#team-moose` are starting our investigation to see how to provide users with a CLI to manage Snyk Apps. We have had a discussion with `#team-hammer` in the past and this is our first effort at doing so.

### DIFF
--- a/help/cli-commands/apps.md
+++ b/help/cli-commands/apps.md
@@ -1,0 +1,55 @@
+# snyk apps -- Create and manage your Snyk Apps
+
+# Usage
+
+`snyk apps <COMMAND> [<OPTIONS>]`
+
+## Description
+
+Snyk Apps are integrations that extend the functionality of the Snyk platform. They provide you with an opportunity to mould your Snyk experience to suit your specific needs.
+
+[For more information see our user docs](https://docs.snyk.io/features/integrations/snyk-apps)
+
+## Commands
+
+**_Note: All `apps` commands are only accessible behind the `--experimental` flag and the behaviour can change at any time, without prior notice. You are kindly advised to use all the commands with caution_**
+
+### `create`
+
+Create a new Snyk App.
+
+## Options
+
+### `--interactive`
+
+Use the command in interactive mode.
+
+### `--org=<ORG_ID>`
+
+(Required for the `create` command)
+Specify the `<ORG_ID>` to create the Snyk App under.
+
+### `--name=<SNYK_APP_NAME>`
+
+(Required for the `create` command)
+The name of Snyk App that will be displayed to the user during the authentication flow.
+
+### `--redirect-uris=<REDIRECT_URIS>`
+
+(Required for the `create` command)
+A comma separated list of redirect URIs. This will form a list of allowed redirect URIs to call back after authentication.
+
+### `--scopes=<SCOPES>`
+
+(Required for the `create` command)
+A comma separated list of scopes required by your Snyk App. This will for a list of scopes that your app is allowed to request during authorization.
+
+## Examples
+
+### `Create Snyk App`
+
+\$ snyk apps create --experimental --org=48ebb069-472f-40f4-b5bf-d2d103bc02d4 --name='My Awesome App' --redirect-uris=https://example1.com,https://example2.com --scopes=apps:beta
+
+### `Create Snyk App Interactive Mode`
+
+\$ snyk apps create --experimental --interactive
--- a/src/cli/commands/apps/create-app.ts
+++ b/src/cli/commands/apps/create-app.ts
@@ -1,0 +1,57 @@
+import * as Debug from 'debug';
+import {
+  EAppsURL,
+  getAppsURL,
+  handleCreateAppRes,
+  handleV3Error,
+  ICreateAppRequest,
+  ICreateAppResponse,
+  SNYK_APP_DEBUG,
+} from '../../../lib/apps';
+import { makeRequestV3 } from '../../../lib/request/promise';
+import { spinner } from '../../../lib/spinner';
+
+const debug = Debug(SNYK_APP_DEBUG);
+
+/**
+ * Function to process the app creation request and
+ * handle any errors that are request error and print
+ * in a formatted string. It throws is error is unknown
+ * or cannot be handled.
+ * @param {ICreateAppRequest} data to create the app
+ * @returns {String} response formatted string
+ */
+export async function createApp(
+  data: ICreateAppRequest,
+): Promise<string | void> {
+  debug('App data', data);
+  const {
+    orgId,
+    snykAppName: name,
+    snykAppRedirectUris: redirectUris,
+    snykAppScopes: scopes,
+  } = data;
+  const payload = {
+    method: 'POST',
+    url: getAppsURL(EAppsURL.CREATE_APP, { orgId }),
+    body: {
+      name,
+      redirectUris,
+      scopes,
+    },
+    qs: {
+      version: '2021-08-11~experimental',
+    },
+  };
+
+  try {
+    await spinner('Creating your Snyk App');
+    const response = await makeRequestV3<ICreateAppResponse>(payload);
+    debug(response);
+    spinner.clearAll();
+    return handleCreateAppRes(response);
+  } catch (error) {
+    spinner.clearAll();
+    handleV3Error(error);
+  }
+}
--- a/src/cli/commands/apps/index.ts
+++ b/src/cli/commands/apps/index.ts
@@ -1,0 +1,47 @@
+import * as Debug from 'debug';
+import { MethodArgs } from '../../args';
+import { processCommandArgs } from '../process-command-args';
+import {
+  EValidSubCommands,
+  validAppsSubCommands,
+  SNYK_APP_DEBUG,
+  ICreateAppOptions,
+  AppsErrorMessages,
+} from '../../../lib/apps';
+
+import { createApp } from './create-app';
+// import * as path from 'path';
+import {
+  createAppDataInteractive,
+  createAppDataScriptable,
+} from '../../../lib/apps/create-app';
+import help from '../help';
+
+const debug = Debug(SNYK_APP_DEBUG);
+
+export default async function apps(
+  ...args0: MethodArgs
+): Promise<string | undefined | any> {
+  debug('Snyk apps CLI called');
+
+  const { options, paths } = processCommandArgs<ICreateAppOptions>(...args0);
+  debug(options, paths);
+
+  const commandVerb1 = paths[0];
+  const validCommandVerb =
+    commandVerb1 && validAppsSubCommands.includes(commandVerb1);
+  if (!validCommandVerb) {
+    // Display help md for apps
+    debug(`Unknown subcommand: ${commandVerb1}`);
+    return help('apps');
+  }
+  // Check if experimental flag is being used
+  if (!options.experimental) throw new Error(AppsErrorMessages.useExperimental);
+
+  if (commandVerb1 === EValidSubCommands.CREATE) {
+    const createAppData = options.interactive
+      ? await createAppDataInteractive()
+      : createAppDataScriptable(options);
+    return await createApp(createAppData);
+  }
+}
--- a/src/lib/apps/constants.ts
+++ b/src/lib/apps/constants.ts
@@ -1,0 +1,57 @@
+import chalk from 'chalk';
+
+export const SNYK_APP_NAME = 'snykAppName';
+export const SNYK_APP_REDIRECT_URIS = 'snykAppRedirectUris';
+export const SNYK_APP_SCOPES = 'snykAppScopes';
+export const SNYK_APP_CLIENT_ID = 'snykAppClientId';
+export const SNYK_APP_ORG_ID = 'snykAppOrgId';
+export const SNYK_APP_DEBUG = 'snyk:apps';
+
+export enum EValidSubCommands {
+  CREATE = 'create',
+}
+
+export enum EAppsURL {
+  CREATE_APP,
+}
+
+export const validAppsSubCommands = Object.values<string>(EValidSubCommands);
+
+export const AppsErrorMessages = {
+  orgRequired: `Option '--org' is required! For interactive mode, please use '--interactive' or '-i' flag. For more information please run the help command 'snyk apps --help' or 'snyk apps -h'.`,
+  nameRequired: `Option '--name' is required! For interactive mode, please use '--interactive' or '-i' flag. For more information please run the help command 'snyk apps --help' or 'snyk apps -h'.`,
+  redirectUrisRequired: `Option '--redirect-uris' is required! For interactive mode, please use '--interactive' or '-i' flag. For more information please run the help command 'snyk apps --help' or 'snyk apps -h'.`,
+  scopesRequired: `Option '--scopes' is required! For interactive mode, please use '--interactive' or '-i' flag. For more information please run the help command 'snyk apps --help' or 'snyk apps -h'.`,
+  useExperimental: `\n${chalk.redBright(
+    "All 'apps' commands are only accessible behind the '--experimental' flag.",
+  )}\n 
+The behaviour can change at any time, without prior notice. 
+You are kindly advised to use all the commands with caution.
+  
+${chalk.bold('Usage')}
+  ${chalk.italic('snyk apps <COMMAND> --experimental')}\n`,
+};
+
+export const CreateAppPromptData = {
+  SNYK_APP_NAME: {
+    name: SNYK_APP_NAME,
+    message: `Name of the Snyk App (visible to users when they install the Snyk App)?`,
+  },
+  SNYK_APP_REDIRECT_URIS: {
+    name: SNYK_APP_REDIRECT_URIS,
+    message: `Your Snyk App's redirect URIs (comma seprated list. ${chalk.yellowBright(
+      ' Ex: https://example1.com,https://example2.com',
+    )})?: `,
+  },
+  SNYK_APP_SCOPES: {
+    name: SNYK_APP_SCOPES,
+    message: `Your Snyk App's permission scopes (comma separated list. ${chalk.yellowBright(
+      ' Ex: apps:beta',
+    )})?: `,
+  },
+  SNYK_APP_ORG_ID: {
+    name: SNYK_APP_ORG_ID,
+    message:
+      'Please provide the org id under which you want to create your Snyk App: ',
+  },
+};
--- a/src/lib/apps/create-app/index.ts
+++ b/src/lib/apps/create-app/index.ts
@@ -1,0 +1,73 @@
+import {
+  AppsErrorMessages,
+  createAppPrompts,
+  ICreateAppRequest,
+  ICreateAppOptions,
+  SNYK_APP_NAME,
+  SNYK_APP_REDIRECT_URIS,
+  SNYK_APP_SCOPES,
+  SNYK_APP_ORG_ID,
+  validateUUID,
+  validateAllURL,
+} from '..';
+import * as inquirer from '@snyk/inquirer';
+import { ValidationError } from '../../errors';
+
+/**
+ * Validates and parsed the data required to create app.
+ * Throws error if option is not provided or is invalid
+ * @param {ICreateAppOptions} options required to create an app
+ * @returns {ICreateAppRequest} data that is used to make the request
+ */
+export function createAppDataScriptable(
+  options: ICreateAppOptions,
+): ICreateAppRequest {
+  if (!options.org) {
+    throw new ValidationError(AppsErrorMessages.orgRequired);
+  } else if (typeof validateUUID(options.org) === 'string') {
+    // Combines to form "Invalid UUID provided for org id"
+    throw new ValidationError(`${validateUUID(options.org)} for org id`);
+  } else if (!options.name) {
+    throw new ValidationError(AppsErrorMessages.nameRequired);
+  } else if (!options['redirect-uris']) {
+    throw new ValidationError(AppsErrorMessages.redirectUrisRequired);
+  } else if (typeof validateAllURL(options['redirect-uris']) === 'string') {
+    throw new ValidationError(
+      validateAllURL(options['redirect-uris']) as string,
+    );
+  } else if (!options.scopes) {
+    throw new ValidationError(AppsErrorMessages.scopesRequired);
+  } else {
+    return {
+      orgId: options.org,
+      snykAppName: options.name,
+      snykAppRedirectUris: options['redirect-uris']
+        .replace(/\s+/g, '')
+        .split(','),
+      snykAppScopes: options.scopes.replace(/\s+/g, '').split(','),
+    };
+  }
+}
+
+// Interactive format
+export async function createAppDataInteractive(): Promise<ICreateAppRequest> {
+  // Proceed with interactive
+  const answers = await inquirer.prompt(createAppPrompts);
+  // Process answers
+  const snykAppName = answers[SNYK_APP_NAME].trim() as string;
+  const snykAppRedirectUris = answers[SNYK_APP_REDIRECT_URIS].replace(
+    /\s+/g,
+    '',
+  ).split(',') as string[];
+  const snykAppScopes = answers[SNYK_APP_SCOPES].replace(/\s+/g, '').split(
+    ',',
+  ) as string[];
+  const orgId = answers[SNYK_APP_ORG_ID].trim() as string;
+  // POST: to create an app
+  return {
+    orgId,
+    snykAppName,
+    snykAppRedirectUris,
+    snykAppScopes,
+  };
+}
--- a/src/lib/apps/input-validator.ts
+++ b/src/lib/apps/input-validator.ts
@@ -1,0 +1,60 @@
+import * as uuid from 'uuid';
+
+/**
+ *
+ * @param {String} input of space separated URL/URI passed by
+ * user for redirect URIs
+ * @returns { String | Boolean } complying with inquirer return values, the function
+ * separates the string on space and validates each to see
+ * if a valid URL/URI. Return a string if invalid and
+ * boolean true if valid
+ */
+export function validateAllURL(input: string): string | boolean {
+  const trimmedInput = input.trim();
+  let errMessage = '';
+  for (const i of trimmedInput.split(',')) {
+    if (typeof validURL(i) == 'string')
+      errMessage = errMessage + `\n${validURL(i)}`;
+  }
+
+  if (errMessage) return errMessage;
+  return true;
+}
+
+/**
+ * Custom validation logic which takes in consideration
+ * creation of Snyk Apps and thus allows localhost.com
+ * as a valid URL.
+ * @param {String} input of URI/URL value to validate using
+ * regex
+ * @returns {String | Boolean } string message is not valid
+ * and boolean true if valid
+ */
+export function validURL(input: string): boolean | string {
+  try {
+    new URL(input);
+    return true;
+  } catch (error) {
+    return `${input} is not a valid URL`;
+  }
+}
+
+/**
+ * Function validates if a valid UUID (version of UUID not tacken into account)
+ * @param {String} input UUID to be validated
+ * @returns {String | Boolean } string message is not valid
+ * and boolean true if valid
+ */
+export function validateUUID(input: string): boolean | string {
+  return uuid.validate(input) ? true : 'Invalid UUID provided';
+}
+
+/**
+ * @param {String} input
+ * @returns {String | Boolean } string message is not valid
+ * and boolean true if valid
+ */
+export function validInput(input: string): string | boolean {
+  if (!input) return 'Please enter something';
+  return true;
+}
--- a/src/lib/apps/utils.ts
+++ b/src/lib/apps/utils.ts
@@ -1,0 +1,7 @@
+import * as fs from 'fs';
+import { renderMarkdown } from '../../cli/commands/help/markdown-renderer';
+
+export function readAppsHelpMarkdown(filename: string): string {
+  const file = fs.readFileSync(filename, 'utf8');
+  return renderMarkdown(file);
+}
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -9,6 +9,7 @@ interface Config {
   MAX_PATH_COUNT: number;
   API: string;
   api: string;
+  API_V3_URL: string;
   disableSuggestions: string;
   org: string;
   ROOT: string;
--- a/src/lib/request/promise.ts
+++ b/src/lib/request/promise.ts
@@ -1,3 +1,4 @@
+import { getAuthHeader } from '../api-token';
 import * as request from './index';
 
 export async function makeRequest<T>(payload: any): Promise<T> {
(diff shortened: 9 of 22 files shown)


mitechie:
I think that any and all apps commands should require a --beta argument to function for now to acknowledge that this UX/experience can change. I'm not sure on "create" vs "submit" or "add" or some other function. I'd like to map out the workflow of using the CLI from both an app developer and app consumer perspective and make sure the language the CLI makes sense and distinguishes between the two user stories and their workflows nicely before doing anything that isn't --beta or noted as unfinished/etc.

love-bhardwaj:
> I think that any and all apps commands should require a --beta argument to function for now to acknowledge that this UX/experience can change. I'm not sure on "create" vs "submit" or "add" or some other function. I'd like to map out the workflow of using the CLI from both an app developer and app consumer perspective and make sure the language the CLI makes sense and distinguishes between the two user stories and their workflows nicely before doing anything that isn't --beta or noted as unfinished/etc.

I believe this will be behind a feature flag before we release it. So the `--beta` flag won't be required. As for the verbs used for the CLI commands. The idea is to build an interactive CLI with user prompts. Tomorrow if we add features for both consumers and developers that we could further prompt users to select options and work accordingly. 

For example, a user wants to list apps. We can ask follow-up questions like List apps created by your org? or list apps you have installed on your org.

src/lib/apps/inputValidator.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *
+ * @param {String} input of URI/URL value to validate using
+ * regex
+ * @returns {String | Boolean } string message is not valid
+ * and boolean true if valid
+ */
+export function validURL(input: string): boolean | string {
+  const pattern = new RegExp(
+    '^(http(s)?:\\/\\/)?' + // protocol
```
JackuB, line 33: What is the intent for this URL validator? What kind of URLs are considered valid? It mentions URI in the function comment as well - should be following the [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)?

Because here it's marking protocol as optional with `?`, thus allowing strings like `test.com`.

Node.js `URL` might be helpful here https://nodejs.org/docs/latest-v10.x/api/url.html if we just want to validate a general shape of the string.

src/lib/apps/inputValidator.ts
```diff
@@ -0,0 +1,63 @@
...
+  return isValid ? true : `${input} is not a valid URL`;
+}
+
+/**
+ * @param {String} input UUID to be validated
+ * @returns {String | Boolean } string message is not valid
+ * and boolean true if valid
+ */
+export function validateUUID(input: string): boolean | string {
+  return uuid.validate(input) ? true : 'Invalid UUID provided';
```
JackuB, line 52: Do we care about the version of UUID we are validating, or any will do? https://www.npmjs.com/package/uuid#uuidvalidatestr

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!options.experimental) throw new Error(AppsErrorMessages.useExperimental);
+
+  const commandVerb1 = paths[0];
+  const validCommandVerb =
+    commandVerb1 && validAppsSubCommands.includes(commandVerb1);
+
+  if (!validCommandVerb) {
+    // Display what is available
+    debug(`Not a valid sub command ${commandVerb1}`);
```
JackuB, line 38: I'd consider updating the error message a bit:

```suggestion
    debug(`Unknown subcommand: ${commandVerb1}`);
```

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      ) as string[];
+      const orgId = answers[SNYK_APP_ORG_ID];
+      // POST: to create an app
+      const res = await createApp({
+        orgId,
+        snykAppName,
+        snykAppRedirectUris,
+        snykAppScopes,
+      });
+      if (res) return res;
```
JackuB, line 62: Wouldn't it be better if `createApp` always returned something? When is it a valid case that it will return `undefined`? The `handleV3Error` always throw, and would trigger a `catch`? Then you could lose this `if`, as it serves no purpose right now.

jgresty:
Should moose be codeowners of the apps commands?

JackuB (commented):
Left some comments and suggestions 👍

src/lib/apps/constants.ts
```diff
@@ -0,0 +1,104 @@
...
+      'The permission scopes for your Snyk App',
+    ],
+  ],
+  useExperimental: [[`Usage: $snyk apps <sub-command> --experimental`]],
+};
+
+export const AppsErrorMessages = {
+  availableCommands: createDisplayErrTable(
+    'Missing a subcommand or invalid sub-command for Snyk Apps!',
+    'The available sub-commands are:',
```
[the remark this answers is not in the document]
love-bhardwaj: Thanks and agreed we should replicate the rest of the commands to maintain consistency.

src/cli/commands/apps/create-app.ts
```diff
@@ -0,0 +1,55 @@
...
+      scopes,
+    },
+    qs: {
+      version: '2021-08-11~experimental',
+    },
+  };
+
+  try {
+    await spinner('Creating your Snyk App');
+    const response = await makeRequestV3<ICreateAppRes, IV3ErrorResponse>(
```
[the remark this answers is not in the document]
DarrellMozingo: V3 as a nae is being communicated widely & to customers, afaik. It's also a major version in the url path, eg `api.snyk.io/v3`, so quite front & centre. There might be a v4 one day (hopefully not 😅 ). Also helps in support in the future imo, like Googling. Knowing a post/article is for Snyk API V3 is more clear if there ever is a v4, instead of just "Snyk API" and not being sure.

test/jest/util/runCommand.ts
```diff
@@ -41,4 +41,62 @@ const runCommand = (
...
+  // This code is heavily inspired (if not blantantly copied)
+  // from inquirer-test package
+  const loop = (inputs) => {
+    if (!inputs.length) {
+      childProcess.stdin.end();
+      return;
+    }
+
+    setTimeout(() => {
+      childProcess.stdin.write(inputs[0]);
```
DarrellMozingo, line 64: Worth a console/logger writ of what `inputs[0]` is being sent in the loop?

test/jest/acceptance/snyk-apps/create-app.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+      expect(stdout).toContain('Snyk App created successfully!');
+      expect(stdout).toContain(`${testData.appName}`);
+      expect(stdout).toContain(`${testData.redirectURIs.split(' ').join(',')}`);
+      expect(stdout).toContain(`${testData.scopes.split(' ').join(',')}`);
+    });
+
+    // Interactive mode with flag shortcut (-i)
+    it('should prompt users for input with flag shortcut (-i)', async () => {
+      const { stdout } = await runSnykCLIWithUserInputs(
+        'apps create --interactive --experimental',
```
DarrellMozingo, line 161: Should this be `-i`?

test/jest/acceptance/snyk-apps/create-app.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+    const baseApi = '/v3';
+    env = {
+      ...process.env,
+      SNYK_API: 'http://localhost:' + port + baseApi,
+      SNYK_API_V3: 'http://localhost:' + port + baseApi,
+      SNYK_HOST: 'http://localhost:' + port,
+      SNYK_TOKEN: '123456789',
+      SNYK_DISABLE_ANALYTICS: '1',
+    };
+    server = fakeServer(baseApi, env.SNYK_TOKEN);
```
DarrellMozingo, line 25: This is just stubbing responses, no? Do we ever assert the data we're sending to the api to ensure we're parsing values correctly?

src/lib/apps/inputValidator.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *
+ * @param {String} input of space separated URL/URI passed by
+ * user for redirect URIs
+ * @returns { String | Boolean } complying with inquirer return values, the function
+ * separates the string on space and validates each to see
+ * if a valid URL/URI. Return a string if invalid and
+ * boolean true if valid
+ */
+export function validateAllURL(input: string): string | boolean {
```
DarrellMozingo, line 12: Could all these validation methods be unit tested? Heavily for the regex ones below if we can't leverage oss alternatives that are tested.

test/jest/acceptance/snyk-apps/create-app.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+      // Assert to print error message when no name provided
+      expect(stdout).toContain('Please enter something');
+    });
+
+    it('should prompt for redirect uris and print error if not provided', async () => {
+      const {
+        stdout,
+      } = await runSnykCLIWithUserInputs(
+        'apps create --interactive --experimental',
+        [testData.appName, ENTER, 'something-invalid', ENTER],
```
DarrellMozingo, line 84: This is a neat testing strategy! I'm curious to see how flakey it might become...

test/jest/acceptance/snyk-apps/create-app.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+        { env },
+      );
+      // Assert
+      // Currently not able to test the whole printed text as it is wrapped to next line
+      // TODO: look for way to increase the line length
+      expect(stdout).toContain(
+        "Your Snyk App's redirect URIs (comma seprated list.",
```
DarrellMozingo, line 91: Single quotes? Not sure if there's a reason for double, or if linting is setup for the CLI...

DarrellMozingo (commented):
Muy bueno, nice work Love! 👏   Left lots of questions & nits below!

src/lib/apps/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+        alignment: 'left',
+      },
+      border: getBorderCharacters('void'),
+    };
+  }
+
+  return table(data, tableConfig as any);
+}
+
+export function createDisplayErrTable(
```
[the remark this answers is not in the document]
love-bhardwaj: Maybe I should have asked this question earlier. `consistency` was the reason I was using `table` for `apps` related command since I could find helpers functions to display messages.

src/cli/commands/apps/create-app.ts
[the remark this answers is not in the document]
love-bhardwaj: Adding to Darrell's comment. We could have used `API` but the thing is it is already being used in other places like `snyk config set api=<token>` and most of the code still relies on `v1`. So I wanted to make sure we don't confuse it and have a distinctive understanding of both.

src/lib/apps/inputValidator.ts
love-bhardwaj: This is a pretty custom regex that supports `localhost` and other URL that we could think of. That the user may use. I believe, for the API `localhost:3000` works as well. We are open to suggestions if there is a better way to do so 🙂 .

src/lib/apps/inputValidator.ts
love-bhardwaj: Good questions, I don't have a definitive answer but whatever versions the `orgId` are off, which I am not sure.

src/lib/request/promise.ts
```diff
@@ -16,3 +16,20 @@ export async function makeRequest<T>(payload: any): Promise<T> {
     });
   });
 }
+
+export async function makeRequestV3<T, E>(payload: any): Promise<T> {
+  return new Promise((resolve, reject) => {
+    request.makeRequest(payload, (error, res, body) => {
```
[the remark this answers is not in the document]
love-bhardwaj: Correct, request module has the debugger to `debug(payload)` 🙂

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    return AppsErrorMessages.availableCommands;
+  }
+
+  if (commandVerb1 === EValidSubCommands.CREATE) {
+    if (options.interactive) {
+      // Proceed with interactive
+      const answers = await inquirer.prompt(createAppPrompts);
+      // Process answers
+      const snykAppName = answers[SNYK_APP_NAME].trim() as string;
+      const snykAppRedirectUris = answers[SNYK_APP_REDIRECT_URIS].trim().split(
```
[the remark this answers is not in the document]
love-bhardwaj: Haven't tested that and don't think we have taken that into consideration. Although we do explicitly tell it is a comma-separated list. But this is a good point, so I will add this to our `todo` list.

src/lib/apps/constants.ts
```diff
@@ -0,0 +1,104 @@
...
+  CREATE = 'create',
+}
+
+export enum EAppsURL {
+  CREATE_APP,
+}
+
+export const validAppsSubCommands = Object.values<string>(EValidSubCommands);
+
+const tableData = {
```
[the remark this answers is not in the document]
love-bhardwaj: Completely agree as it will also add `consistency`. Would it be alright to create a markdown for all `apps` commands such as `apps create`, `apps list` and so on and use those? Haven't seen this for other commands so would like to confirm before proceeding.

src/cli/commands/apps/create-app.ts
```diff
@@ -0,0 +1,55 @@
...
+      payload,
+    );
+    debug(response);
+    spinner.clearAll();
```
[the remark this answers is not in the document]
love-bhardwaj: Sorry if I misunderstood, but the problem there was that you would see it spinning even after the text was printing to the terminal and disappear causing a glitchy effect 😓 .

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!validCommandVerb) {
+    // Display what is available
+    debug(`Not a valid sub command ${commandVerb1}`);
+    return AppsErrorMessages.availableCommands;
+  }
+
+  if (commandVerb1 === EValidSubCommands.CREATE) {
+    if (options.interactive) {
+      // Proceed with interactive
```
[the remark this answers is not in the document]
love-bhardwaj: Great idea. I was looking for something similar but haven't been able to focus on it. Added to my `todo` list.

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      const res = await createApp(createAppData);
+      if (res) return res;
+    }
+  } else {
+    debug(`Not a valid sub command ${commandVerb1}`);
+    return AppsErrorMessages.availableCommands;
+  }
+}
+
+function validateProcessCreateAppOpts(
```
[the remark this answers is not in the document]
love-bhardwaj: Yes, that is part of the plan 😅 but have been delaying it to work on the feedback provided. Another one for the `todo` list 😄

src/cli/commands/apps/create-app.ts
```diff
@@ -0,0 +1,55 @@
...
+    orgId,
+    snykAppName: name,
+    snykAppRedirectUris: redirectUris,
+    snykAppScopes: scopes,
+  } = data;
+  const payload = {
+    method: 'POST',
+    url: getAppsURL(EAppsURL.CREATE_APP, { orgId }),
+    json: true,
+    headers: getV3Headers(),
```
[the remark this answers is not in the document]
love-bhardwaj: `json: true` can surely go in there but not sure about `headers` as it has the authorization header. Will look into this and see if it can be moved 🙂

src/lib/request/promise.ts
```diff
@@ -16,3 +16,20 @@ export async function makeRequest<T>(payload: any): Promise<T> {
     });
   });
 }
+
+export async function makeRequestV3<T, E>(payload: any): Promise<T> {
```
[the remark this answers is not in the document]
love-bhardwaj: A good idea will try that. This was me trying to play with generics to type cast error response.

src/lib/apps/v3utils.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (error.code) {
+    if (error.code === 400) {
+      // Bad request
+      const responseJSON: IV3ErrorResponse = error.body;
+      const errString = errorsToDisplayString(responseJSON);
+      throw new Error(errString);
+    } else if (error.code === 401) {
+      // Unauthorized
+      throw AuthFailedError(
+        'Unauthorized: the request requires an authentication token or a token with more permissions.',
```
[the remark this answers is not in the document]
love-bhardwaj: I think it does. Sorry, this was implemented a while back at least in terms of coding, but not sure. I will check and confirm. But as far as I remember it does.

src/cli/commands/apps/create-app.ts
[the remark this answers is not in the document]
DarrellMozingo: Ah you're right, sorry! I read too fast, missed there was text in that handle method afterwards :)

test/jest/acceptance/snyk-apps/create-app.spec.ts
love-bhardwaj: The reason for double quotes is that the content inside the string has a single quote (apostrophe i.e App's) 🙂

DarrellMozingo: Totally missed that, sorry 😄

src/lib/apps/inputValidator.ts
love-bhardwaj: `v3` API validates the `UUID` without taking into account the version. So for consistency, we too don't take the version of UUID into account.

test/jest/util/runCommand.ts
love-bhardwaj: My only worry it will lead to a lot of unnecessary data being printed to logs/console.

test/jest/acceptance/snyk-apps/create-app.spec.ts
love-bhardwaj: The stub response has data that is directly returned in the response from the request body. So in other words, we are asserting on the response that the data we sent was sent in the correct format.

test/jest/acceptance/snyk-apps/create-app.spec.ts
love-bhardwaj: This was the only reasonable strategy I could find for testing the `--interactive` commands 😅 I took am curious to see the flakiness of it. Initial doubt was that it may not work on Windows, but has been working fine till now 😄

src/cli/commands/apps/index.ts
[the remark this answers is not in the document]
love-bhardwaj: This has been added 🙂

src/cli/commands/apps/index.ts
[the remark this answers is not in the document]
love-bhardwaj: ✅

src/lib/request/promise.ts
```diff
@@ -16,3 +17,31 @@ export async function makeRequest<T>(payload: any): Promise<T> {
...
+ * @returns
+ */
+export async function makeRequestV3<T>(payload: any): Promise<T> {
+  return new Promise((resolve, reject) => {
+    payload.headers = payload.headers ?? {
+      'Content-Type': 'application/vnd.api+json',
+      authorization: getAuthHeader(),
+    };
+    payload.json = true;
+    request.makeRequest(payload, (error, res, body) => {
```
jahed-snyk, line 34: IIRC, if you don't provide a callback, `makeRequest` will return a promise instead so you won't need the promise wrapper.

src/lib/config.ts
```diff
@@ -64,4 +65,6 @@ if (!config.ROOT) {
   config.ROOT = apiUrl.protocol + '//' + apiUrl.host;
 }
 
+config.API_V3 = userConfig.get('v3endpoint') || 'https://api.snyk.io/v3';
```
jahed-snyk, line 68: You can set `API_V3`'s default in `config.default.json`. That's also where `API`'s default is set and picked up by snyk-config. `process.env.SNYK_API_V3` is also picked up automatically via snyk-config, however, there are some precedence issues when combined with userConfig (config misalignment in general is a known issue).

If you rename `v3endpoint` to `api-v3`, then `process.env.SNYK_CONF_API_V3` will be picked up automatically by userConfig so you don't need to separately define an override later on.

So precedence will be like:

SNYK_CONF_API_V3 > api-v3 in user-config > SNYK_API_V3 > API_V3 in snyk-config

I don't think we need to tie ourselves too much to "endpoint" as it's too ambiguous and inconsistent.

src/lib/apps/input-validator.ts
```diff
@@ -0,0 +1,66 @@
...
+ * user for redirect URIs
+ * @returns { String | Boolean } complying with inquirer return values, the function
+ * separates the string on space and validates each to see
+ * if a valid URL/URI. Return a string if invalid and
+ * boolean true if valid
+ */
+export function validateAllURL(input: string): string | boolean {
+  const trimmedInput = input.trim();
+  let errMessage = '';
+  for (const i of trimmedInput.split(',')) {
```
jahed-snyk, line 15: Might be useful to trim each URL after the split as people may put spaces after each comma. Just a common pet peeve of mine...

jahed-snyk (commented):
I've left some thoughts on a few things but nothing major other than the package-lock.json (which as you mentioned will likely not need to be modified once this is ready).

src/lib/apps/constants.ts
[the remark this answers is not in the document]
JackuB: I'd start with having a single `apps` help page and we can go from there if the doc gets too large. Definitely use markdown rather than loose text strings in the code for such docs

src/lib/apps/utils.ts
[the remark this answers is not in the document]
JackuB: I'd recommend getting rid of the `table` altogether and keep using Markdown for help text and plain strings for errors like the rest of the commands.

help/cli-commands/apps.md
```diff
@@ -0,0 +1,86 @@
...
+
+Silence all output.
+
+#### `--version`, `-v`
+
+Prints versions.
+
+#### `--help [<COMMAND>]`, `[<COMMAND>] --help`, `-h`
+
+Prints a help text. You may specify a `<COMMAND>` to get more details.
```
JackuB, line 67: We've recently cleaned up this repeated section of Help files, making them considerably shorter - see for example changes to `snyk auth` command: https://github.com/snyk/snyk/pull/2512/files

You can drop these repeated flags.

help/cli-commands/apps.md
```diff
@@ -0,0 +1,86 @@
...
+\$ snyk apps create --experimental --interactive
+
+## Exit codes
+
+Possible exit codes and their meaning:
+
+**0**: success, no vulns found<br />
+**1**: action_needed, vulns found<br />
+**2**: failure, try to re-run command<br />
+**3**: failure, no supported projects detected<br />
```
JackuB, line 86: Same for this, as it's not relevant for `apps` command

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  createAppDataInteractive,
+  createAppDataScriptable,
+} from '../../../lib/apps/create-app';
+
+const debug = Debug(SNYK_APP_DEBUG);
+
+export default async function apps(
+  ...args0: MethodArgs
+): Promise<string | undefined | any> {
```
JackuB, line 24: `any` here looks like a code smell, especially for a new command/function that this PR is adding

src/lib/apps/create-app/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    );
+  } else if (!options.scopes) {
+    throw new ValidationError(AppsErrorMessages.scopesRequired);
+  } else {
+    return {
+      orgId: options.org,
+      snykAppName: options.name,
+      snykAppRedirectUris: options['redirect-uris']
+        .replace(/\s+/g, '')
+        .split(','),
```
JackuB, line 46: Not sure if it's an issue in this case, but `,` commas are actually allowed characters in the URL, although uncommon. I'd consider adding an edge case to at least check we error out as expected on this.

src/lib/apps/input-validator.ts
JackuB: I'd add to this, that we are doing this trim&split(&trim) check on a few places, it might be useful to use this single function and always return `string[]`?

JackuB:
Don't mind me, just screwed up a git command when working on #2591. Reverted to the original commit :p

src/lib/apps/create-app/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    };
+  }
+}
+
+// Interactive format
+export async function createAppDataInteractive(): Promise<ICreateAppRequest> {
+  // Proceed with interactive
+  const answers = await inquirer.prompt(createAppPrompts);
+  // Process answers
+  const snykAppName = answers[SNYK_APP_NAME].trim() as string;
```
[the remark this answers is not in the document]
love-bhardwaj: Yes for sure. That will be really great. We lack types for our forked inquirer.

test/jest/util/runCommand.ts
```diff
@@ -41,4 +41,62 @@ const runCommand = (
...
+  // in order to get results from the tool
+  // This code is heavily inspired (if not blantantly copied)
+  // from inquirer-test package
+  const loop = (inputs) => {
+    if (!inputs.length) {
+      childProcess.stdin.end();
+      return;
+    }
+
+    setTimeout(() => {
```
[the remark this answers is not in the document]
love-bhardwaj: Agreed and thanks for your input on this. Timeouts can be removed, but this is the best I could find to test the interactive mode. If there are better ways to test user prompts like interactive modes, please provide an example I would be happy to incorporate that 🙂 

It would be great! if there is a testing API in the codebase to test such interactive commands. It will help massively going forward as we plan to implement both interactive and scriptable modes.

src/lib/request/promise.ts
love-bhardwaj: We are using the `callback` to actually make the response be parsed as an object using `JSON.parse` for all v3 requests 🙂 . So this format actually helps.

src/lib/config.ts
love-bhardwaj: I tried doing what was suggested but there seem to be some type issues. I may not have understood the working of it completely but I am making changes based on your suggestions 🙂

src/lib/apps/input-validator.ts
love-bhardwaj: For input validation using the `new URL` spaces doesn't seem to affect the validation. In general, we are removing all the spaces from the input using a regex.

test/jest/acceptance/snyk-apps/create-app.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+      const {
+        stdout,
+      } = await runSnykCLIWithUserInputs(
+        'apps create --interactive --experimental',
+        [testData.appName, ENTER, 'something#invalid', ENTER],
+        { env },
+      );
+      // Assert
+      // Currently not able to test the whole printed text as it is wrapped to next line
+      // TODO: look for way to increase the line length
```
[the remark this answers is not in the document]
love-bhardwaj: Have tried snapshot testing but got flaky tests with it. Presumably due to width of the screen is not always constant. So to took this approach to reduce the flakiness of the tests 🙂

jahed-snyk (changes requested):
You haven't changed any dependencies right? If so, you'll need to revert package-lock.json. Happy to approve otherwise. I've left some other comments related to some duplication and improving input validation.

src/cli/commands/apps/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    commandVerb1 && validAppsSubCommands.includes(commandVerb1);
+  if (!validCommandVerb) {
+    // Display help md for apps
+    debug(`Unknown subcommand: ${commandVerb1}`);
+    const appsHelpFile = path.resolve(
+      __dirname,
+      '../../help/cli-commands',
+      'apps.md',
+    );
+    return `\n${readAppsHelpMarkdown(appsHelpFile)}`;
```
jahed-snyk, line 41: Instead of duplicating this logic, have you looked into re-using `commands/help`?

src/lib/apps/create-app/index.ts
jahed-snyk: We should validate each entry immediately after the split and fail on invalid ones. That should cover any commas in URLs.

Ideally to avoid conflicts like this, we should support passing the same flagging multiple times for array values. But we don't have that yet.

src/lib/apps/create-app/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  } else if (!options.scopes) {
+    throw new ValidationError(AppsErrorMessages.scopesRequired);
+  } else {
+    return {
+      orgId: options.org,
+      snykAppName: options.name,
+      snykAppRedirectUris: options['redirect-uris']
+        .replace(/\s+/g, '')
+        .split(','),
+      snykAppScopes: options.scopes.replace(/\s+/g, '').split(','),
```
jahed-snyk, line 47: Should we validate these entrys to make sure they're correct scopes? Same question for the interactive format.

test/jest/util/runCommand.ts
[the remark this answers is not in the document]
jahed-snyk: Just to update with out of thread discussions. We'll be working on this in a future PR.

src/cli/commands/apps/index.ts
love-bhardwaj: Thanks, I will update that code to reuse the help command logic.

src/lib/apps/create-app/index.ts
love-bhardwaj: Scopes validation could be a bit tricky at the moment. Since we are doing work in this area and they change frequently. So updating the list of valid scopes in the CLI code could be cumbersome. Ideally, you would want to fetch the valid scope list from API but this isn't available at the moment. So we are going ahead with whatever users provide and letting the API/Backend to the validations in this case.

We are actually validating the URL after splitting on `,` refer to the code [here](https://github.com/snyk/snyk/pull/2396/files#diff-aa08bc92fb0e1b39e7d207aa3b9e37cff15b2775d7bec3f4914a23984e0c77e9R34).

jahed-snyk (approved):
🎉

src/lib/apps/utils.ts
```diff
@@ -0,0 +1,7 @@
+import * as fs from 'fs';
+import { renderMarkdown } from '../../cli/commands/help/markdown-renderer';
+
+export function readAppsHelpMarkdown(filename: string): string {
+  const file = fs.readFileSync(filename, 'utf8');
+  return renderMarkdown(file);
+}
```
jahed-snyk, line 7: Is this still needed?

love-bhardwaj: There are a few minor improvements that need to be pushed. I will push in a separate PR 😄

PR #2397, opened 2021-11-29
test: delete policy-ignore test

Avishagp:
Delete `ignore-policy test since it's testing something already covered in `snyk-policy`: https://github.com/snyk/policy/search?q=ignored+vulns+do+not+turn+up+in+tests


test/jest/unit/policy-ignore.spec.ts
```diff
@@ -0,0 +1,26 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as policy from 'snyk-policy';
+
+import { extendExpiries } from '../../utils';
+
+describe('policy-ignore test', () => {
+  it('ignored vulns do not turn up in tests', async () => {
+    const dir = path.resolve(__dirname, '../../fixtures/jsbin-policy');
```
jahed-snyk, line 9: We can use `createProject('jsbin-policy')` here which can then simplify the rest of the test.

jahed-snyk (approved):
Looks good in terms of Tap -> Jest migration.

I don't think we need this test. It's testing 'snyk-policy' which is an external dependency so this scenario should be covered there instead.

https://github.com/snyk/policy/search?q=ignored+vulns+do+not+turn+up+in+tests

From a CLI perspective, we probably have a test which ensures `snyk test` takes expired ignores into account (if we don't, we should write one).

test/jest/unit/policy-ignore.spec.ts
Avishagp: Not so sure since we have to call `extendExpiries(config);` and we get `config` from `policy.load(dir)`

PR #2399, opened 2021-11-30
fix: enable checksum metric for local scans

YairZ101:

#### What does this PR do?
Lets us send the checksum of local scans for analytics.


src/cli/commands/test/iac-local-execution/analytics.ts
```diff
@@ -49,12 +49,10 @@ export function addIacAnalytics(
     'iac-custom-rules-issues-percentage',
     calculatePercentage(issuesFromCustomRulesCount, totalIssuesCount),
   );
-  if (!isLocalCustomRules) {
-    analytics.add(
-      'iac-custom-rules-checksum',
-      computeCustomRulesBundleChecksum(),
-    );
-  }
+  analytics.add(
```
teodora-sandu, line 52: `isLocalCustomRules` is not required anyore, so could you remove it from the function signature and the functional call too please? 🙏

PR #2403, opened 2021-11-30
feat: preventing code upload if LCE is on

metju90:
If LCE settings are on, it throws and error and disable SAST


shinin-light (approved):

shinin-light:
Please test it against production before merging

PR #2404, opened 2021-12-01
test: migrate cli-monitor-utils to jest

Avishagp:
Migrate `cli-monitor-utils` test from tap to jest.


test/jest/unit/cli-monitor-utils.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  'policy-path': '',
+  'project-name': 'project-name-override',
+  isDocker: true,
+  prune: false,
+};
+
+const stubPluginMeta: PluginMetadata = {
+  name: 'my-plugin',
+  targetFile: '/tmp2/package.json',
+};
```
jahed-snyk, line 53: We should move these into functions so that the object/files are only created/read when the test needing it executes. It will also avoid any possible state pollution across tests (i.e. a test wrongly modifying the object causing the next test to get a different object).

jahed-snyk (approved):
Looks good.

PR #2441, opened 2021-12-09
refactor: remove references to experimental local execution flow

ipapast:
This is a tiny cleanup.

This feature is GA for many months, so this PR removes all references to 'experimental'. 

We also do not provide the option to have an `--experimental` flag anymore when running a scan, so any tests that used to have that reference were updated too.

## IaC 
Also removed some references for the flag in Snyk Preview (not used) in Registry: https://github.com/snyk/registry/pull/25196

## Hammer & Open Source

There is only a cleanup of references in tests and comments.


aron (approved):
👍 Nice!

jahed-snyk (approved):

PR #2442, opened 2021-12-09
refactor: Load IaC Files Flow

ofekatr:

#### What does this PR do?
Adds the following refactors for loading IaC files in the `test` flow:
- Instead of fetching all files in the given directory and then filtering the IaC files, fetch only IaC files.

- `src/cli/commands/test/iac-local-execution/file-loader.ts`

Currently in the `test` flow for IaC, when we load IaC files from the provided directory, we do so in the following steps:
1. Fetch all file paths inside the directory.
2. Filter the IaC files.
3. Load the IaC files.

This PR suggests to combine steps 1 and 2 into 1 step, where only paths for IaC files are fetched.
The new flow would be:
1. Fetch all IaC file paths inside the directory.
2. Load the IaC files.

The primary reasoning for this was an attempt to improve the flow for loading files in `loadFiles`:
- Initially, **all** directory file paths are retrieved in the `getFilePathsFromDirectory` function.
- Afterwards, the content for each of these is retrieved in `tryLoadFileData`. If it's not an IaC file, the returned content is `null` and that file is filtered out.

This PR offers an alternative implementation in which the filtering happens while loading the directory file paths, and as such having only IaC files when loading their content.
This results in fewer iterations at the content loading step.

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-loader.ts
+++ b/src/cli/commands/test/iac-local-execution/file-loader.ts
@@ -18,64 +18,65 @@ export async function loadFiles(
   pathToScan: string,
   options: IaCTestFlags = {},
 ): Promise<IacFileData[]> {
-  let filePaths = [pathToScan];
+  let iacFilesData: IacFileData[] = [];
 
   if (isLocalFolder(pathToScan)) {
-    filePaths = getFilePathsFromDirectory(pathToScan, {
+    iacFilesData = await getIacFilesData(pathToScan, {
       maxDepth: options.detectionDepth,
     });
-  }
-
-  const filesToScan: IacFileData[] = [];
-  for (const filePath of filePaths) {
-    try {
-      const fileData = await tryLoadFileData(filePath);
-      if (fileData) {
-        filesToScan.push(fileData);
-      }
-    } catch (e) {
-      throw new FailedToLoadFileError(filePath);
+  } else {
+    const iacFileData = await tryLoadFileData(pathToScan);
+    if (iacFileData) {
+      iacFilesData.push(iacFileData);
     }
   }
 
-  if (filesToScan.length === 0) {
+  if (iacFilesData.length === 0) {
     throw new NoFilesToScanError();
   }
-  return filesToScan.filter((file) => file.fileContent !== '');
+
+  return iacFilesData.filter((fileData) => fileData.fileContent);
 }
 
-function getFilePathsFromDirectory(
+async function getIacFilesData(
   pathToScan: string,
   options: { maxDepth?: number } = {},
-): string[] {
+): Promise<IacFileData[]> {
   const directoryPaths = makeDirectoryIterator(pathToScan, {
     maxDepth: options.maxDepth,
   });
 
-  const directoryFilePaths: string[] = [];
+  const filesToScan: IacFileData[] = [];
   for (const filePath of directoryPaths) {
-    directoryFilePaths.push(filePath);
+    const fileData = await tryLoadFileData(filePath);
+    if (fileData) {
+      filesToScan.push(fileData);
+    }
   }
-  return directoryFilePaths;
+  return filesToScan;
 }
 
 export async function tryLoadFileData(
   pathToScan: string,
 ): Promise<IacFileData | null> {
-  const fileType = getFileType(pathToScan);
-  if (!VALID_FILE_TYPES.includes(fileType)) {
-    return null;
+  try {
+    const fileType = getFileType(pathToScan);
+    if (!VALID_FILE_TYPES.includes(fileType)) {
+      return null;
+    }
+
+    const fileContent = (
+      await fs.readFile(pathToScan, DEFAULT_ENCODING)
+    ).toString();
+
+    return {
+      filePath: pathToScan,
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/iac-local-execution/file-loader.ts
```diff
@@ -18,64 +18,65 @@ export async function loadFiles(
...
-  const filesToScan: IacFileData[] = [];
-  for (const filePath of filePaths) {
-    try {
-      const fileData = await tryLoadFileData(filePath);
-      if (fileData) {
-        filesToScan.push(fileData);
-      }
-    } catch (e) {
-      throw new FailedToLoadFileError(filePath);
+  } else {
```
teodora-sandu, line 27: This is now a bit harder to understand, in my opinion. The code here reads as:
- if it's a folder, call a function to retrieve all IaC data by iterating through the files inside the folder, call `tryLoadFileData` and then store the results in `iacFilesData`
- if it's not, then just call `tryLoadFileData` directly here and add the result to `iacFilesData`

Can you explain why you're doing this refactoring a bit more? You're combing those two steps in your description but only by moving them from one function to another, so it's not gaining anything performance-wise as far as I can tell.

ofekatr: > This is now a bit harder to understand, in my opinion. The code here reads as:
> 
> if it's a folder, call a function to retrieve all IaC data by iterating through the files inside the folder, call tryLoadFileData and then store the results in iacFilesData
> if it's not, then just call tryLoadFileData directly here and add the result to iacFilesData

This is the current behavior in [the CLI](https://github.com/snyk/snyk/blob/52531267904792f7a9614d5c58856395d69e952e/src/cli/commands/test/iac-local-execution/file-loader.ts#L21) as well:
- If the provided path is a folder, all files paths in the directory are collected.
- If it's a file, only the provided path is collected.

The code here simply makes this more explicit.
The primary reason for this refactor was an attempt to improve the flow for loading files.
In `loadFiles`:
- Initially, **all** directory file paths are retrieved in the `getFilePathsFromDirectory` function.
- Afterwards, the content for each of these is retrieved in `tryLoadFileData`. If it's not an IaC file, the returned content is `null` and that file is thus filtered out.

This PR offers an alternative implementation in which the filtering happens while loading the directory file paths, and as such having only IaC files when loading their content.
In terms of performance, it simply means fewer iterations at the content loading step.

teodora-sandu: > This is the current behavior in the CLI as well:

It's the same behaviour but written differently. The current CLI first builds up all the file paths based on whether the provided path is a folder or not, then calls `tryLoadFileData`. What I was trying to say is that the proposed changes make this flow a bit weirder, because the `tryLoadFileData` is called from different places:
Screenshot 2021-12-20 at 11 17 28

Also, the performance improvement is O(N) instead of O(2*N), which are considered the same: https://stackoverflow.com/questions/25777714/which-algorithm-is-faster-on-or-o2n


I wouldn't want to sacrifice the readability of the code for that much of a performance improvement, but one way to do this is to move `isLocalFolder` to `getIacFilesData` like so:
```
async function getIacFilesData(
  pathToScan: string,
  options: { maxDepth?: number } = {},
): Promise<IacFileData[]> {
  let directoryPaths: string[] = [pathToScan];

  if (isLocalFolder(pathToScan)) {
    directoryPaths = makeDirectoryIterator(pathToScan, {
      maxDepth: options.maxDepth,
    });
  }

  const filesToScan: IacFileData[] = [];
  for (const filePath of directoryPaths) {
    const fileData = await tryLoadFileData(filePath);
    if (fileData) {
      filesToScan.push(fileData);
    }
  }
  return filesToScan;
}
```

Then in `loadFiles` you can have something like:
```
export async function loadFiles(
  pathToScan: string,
  options: IaCTestFlags = {},
): Promise<IacFileData[]> {
  const iacFilesData = getIacFilesData(pathToScan, {
      maxDepth: options.detectionDepth,
    });

  if (iacFilesData.length === 0) {
    throw new NoFilesToScanError();
  }

  return iacFilesData.filter((fileData) => fileData.fileContent);
}
```

It's a bit easier to read, I think 🙂

ofekatr: @teodora-sandu Thank you for the clarification.
Excellent input, I've applied the suggested code modifications.
Regarding the note on performance - in terms of complexity, it's true that these 2 implementations represent a linear growth rate. The only part that is removed in the PR's implementation, is the redundant function calls to `tryLoadFileData`. As such, they should definitely still have the same asymptotic complexity. 
However, these are all equivalent when n approaches infinity.
While they may be the same in terms of their complexity, in terms of execution time it is still valuable to attempt to minimize the runtime of each step of the flow. This improves the overall runtime and applies minor optimizations to the user experience.

TL;DR - From the discussion here, it does seem like the difference may not be as significant as initially seemed, so it'd also be totally acceptable to close this PR. What do you think?

ofekatr:
Will soon be replaced by a PR with a simplified version of this refactor which more closely resembles the current implementation in the CLI

PR #2456, opened 2021-12-12
Feat/json trying test with tar

shaninja:

#### What does this PR do?

We have a test that timed out for windows, because it was pulling an image.
This PR updates the test so it's using a fixture instead, and so we can re-enable the test for windows.

The fixture is a clean Alpine image and `debug` in the package.json.

Original test was added in this PR: https://github.com/snyk/snyk/pull/2455

### DIFF
--- a/test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
+++ b/test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
@@ -10,17 +10,11 @@ import {
 import { fakeServer } from '../../../acceptance/fake-server';
 import { runSnykCLI } from '../../util/runSnykCLI';
 
-const isWindows =
-  require('os-name')()
-    .toLowerCase()
-    .indexOf('windows') === 0;
-
 describe('container test projects behavior with --app-vulns, --file and --exclude-base-image-vulns flags', () => {
   const fixturePath = path.normalize('test/fixtures/container-projects');
 
   function readFixture(filename: string) {
     const filePath = path.join(fixturePath, filename);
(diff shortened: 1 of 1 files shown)


test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
```diff
@@ -66,64 +60,62 @@ describe('container test projects behavior with --app-vulns, --file and --exclud
...
+  });
 
-    it('returns a json with the --experimental flags', async () => {
-      const { code, stdout } = await runSnykCLI(
-        `container test snykgoof/os-app:node-snykin --app-vulns --json --experimental`,
-        {
-          env,
-        },
-      );
+  it('returns a json with the --experimental flags', async () => {
```
shaninja, line 94: question for hammer: would you say that it's perhaps better to leave the `container test snykgoof/os-app:node-snykin --app-vulns --json` for non-windows, and have this variation only for Windows? And then we're getting a wider range of the end-to-end functionality.

Avishagp (approved):

test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
snaftaly: Adding @snyk/magma's take on this - we discussed it this morning and since this test doesn't focus on the pull part, then we think it makes sense to be able to cover more OSs, rather then covering the e2e (with pull) functionality.

michael-c-snyk (approved):

PR #2483, opened 2021-12-20
fix: remove false positive results from log4shell

danlucian:
We had to remove the false positive results from the output of
the log4shell command. We rely stricly on JndiLookup class now
for the matchings while we do non-jar lookups.
We also added diffing for the DoS vs Log4Shell vulns.

#### What does this PR do?

Limits the amount of false positives by making JndiLookup the main source of truth 
in case of file matching. 
It also splits the DoS vs Log4Shell outputs.


src/cli/commands/log4shell.ts
```diff
@@ -3,16 +3,19 @@ import { promises, Stats } from 'fs';
...
-import { vulnerableHashes } from './log4shell-hashes';
+import * as semver from 'semver';
+import { HashFootprint, vulns } from './log4shell-hashes';
 
 const readFile = promises.readFile;
 const readDir = promises.readdir;
 const stat = promises.stat;
 const MAX_FILE_SIZE = 2 * 1024 * 1024 * 1024 - 1;
 
+type ExploitType = 'Log4Shell' | 'DOS' | 'Unknown';
```
anthogez, line 14: MS-DOS? Small typo here 🤗

danlucian: This is DOS: Denial of Service. Maybe DoS would be more suitable 🤗

src/cli/commands/log4shell.ts
```diff
@@ -49,43 +52,61 @@ export default async function log4shell(...args: MethodArgs): Promise<void> {
...
+    return;
+  }
+
   signatures.forEach((signature) => {
-    const path = signature.path.replace(
+    signature.path = signature.path.replace(
       /(.*org\/apache\/logging\/log4j\/core).*/,
       '$1',
     );
-    results.add(path);
```
anthogez, line 58: Why mutation rather than save the values in a new variable?

danlucian: Worthless saving in a new variable since we don't need the original.

src/cli/commands/log4shell.ts
```diff
@@ -49,43 +52,61 @@ export default async function log4shell(...args: MethodArgs): Promise<void> {
...
-    results.forEach((path) => {
-      console.log(`\t ${path}`);
-    });
-    console.log(`\n We highly recommend fixing this vulnerability. If it cannot be fixed by upgrading, see mitigation information here:
-      \t- https://security.snyk.io/vuln/SNYK-JAVA-ORGAPACHELOGGINGLOG4J-2314720
-      \t- https://snyk.io/blog/log4shell-remediation-cheat-sheet/`);
+  const log4ShellIssues = signatures.filter(
+    (signature) => signature.exploitType === 'Log4Shell',
+  );
+  const dosIssues = signatures.filter(
```
anthogez, line 71: log4DoS maybe? To keep the consistency with log4Shell. Not a blocker

danlucian: Maybe rceIssues and dosIssues would be better? So we use consistent terms (and not selling names like log4shell)?

anthogez: That’s also fine, was suggesting the above as it felt more domain related i.e. log4j and not e.g. Jackson or any other JVM lib and we have no intention to introduce those at this time.

saying so RceIssues and DosIssues it’s :)

src/cli/commands/log4shell.ts
```diff
@@ -49,43 +52,61 @@ export default async function log4shell(...args: MethodArgs): Promise<void> {
...
+  if (signatures.length == 0) {
+    console.log('No known vulnerable version of Log4J was detected');
+    return;
+  }
+
   signatures.forEach((signature) => {
-    const path = signature.path.replace(
+    signature.path = signature.path.replace(
       /(.*org\/apache\/logging\/log4j\/core).*/,
       '$1',
```
anthogez, line 64: I am from my phone… albeit I think we could move line 68 to 71

‘’’

  const rceIssues = signatures.filter(
    (signature) => signature.exploitType === 'Log4Shell',
  );
  const dosIssues = signatures.filter(
    (signature) => signature.exploitType === 'DoS',
  );

‘’’

To inside of this forEach and reduce the complexity by removing 2 loops

what do you think?

danlucian: That's a good idea! We sacrifice a bit of single responsability but the number of found signatures might be big 🤔

src/cli/commands/log4shell-hashes.ts
```diff
@@ -1,205 +1,245 @@
...
-  'iFaGU1RTWaznU/GacrGCCA',
-  'dmv2t1Wt7mc4OP35aMFQeQ',
-  'c3tDD6xsrvfEhcnEfw+RBA',
-  'iFaGU1RTWaznU/GacrGCCA',
-  'iFaGU1RTWaznU/GacrGCCA',
-  'ZB/XrnbpWzXwLFX/v0MOaw',
-  'TLOgJx93wC/S3jFEpymrcA',
-  'cZs0M1ZG9Y0Moqm1zHcSow',
-];
+export type HashFootprint = {
```
anthogez, line 1: I am struggling to understand why we are introducing HashFootprint here
We already used fingerprints elsewhere that we deprecated as terminology in favour of fileSignatures...
Maybe HashFileInsights or HashFileSignaturesInsights? something that recalls our already existing domain?

src/cli/commands/log4shell.ts
```diff
@@ -3,21 +3,56 @@ import { promises, Stats } from 'fs';
...
-type Path = string;
 type FileContent = Buffer;
+type FilePath = string;
 type Digest = string;
 
+type InMemoryPath = {
+  path: FilePath;
+  content: FileContent;
+};
+class InMemoryPaths {
```
anthogez, line 28: I think this file is becoming a bit large, maybe is the time to create 
its own module and have separated files per purpose?

src/cli/commands/log4shell.ts
```diff
@@ -151,12 +209,92 @@ async function traverse(path: Path, handle: FileHandler) {
...
+
+function isJavaArchive(path: FilePath) {
   return path.endsWith('.jar') || path.endsWith('.war') || path.endsWith('ear');
 }
 
-function isJvmFile(path: Path) {
-  return path.endsWith('.java') || path.endsWith('.class');
+function isTraceable(path: FilePath) {
+  return (
+    isJavaArchive(path) ||
```
anthogez, line 273: What do you think of the suggestion below?

```
function isTraceable(path: FilePath) {
  return isJavaArchive(path) || ['JndiManager', 'JndiLookup'].includes(path);
}
```

src/cli/commands/log4shell.ts
```diff
@@ -39,88 +74,111 @@ export default async function log4shell(...args: MethodArgs): Promise<void> {
...
   const signatures: Array<Signature> = new Array<Signature>();
-  const paths: Path[] = await find('.');
   const spinner = await startSpinner();
 
-  for (const path of paths) {
-    const content = await readFile(path);
-    await handleContent(content, path, signatures);
+  const paths: FilePath[] = await find('.');
+  try {
+    await parsePaths(new OnDiskPaths(paths), signatures);
```
anthogez, line 81: I think we might be overusing mutation here...

src/cli/commands/log4shell.ts
```diff
@@ -151,12 +209,92 @@ async function traverse(path: Path, handle: FileHandler) {
...
+
+  if (evaluate(footprint, semver.satisfies, '2.16.x')) {
+    return 'DoS';
+  }
+
+  return 'Unknown';
+}
+
+function evaluate(
+  footprint: HashFootprint,
```
anthogez, line 227: as I've raised above, we should use a terminology that reasons our already existing for unmanaged

src/cli/commands/log4shell.ts
```diff
@@ -151,12 +209,92 @@ async function traverse(path: Path, handle: FileHandler) {
...
 
-function isJavaArchive(path: Path) {
+async function computeExploitType(
+  footprint: HashFootprint,
+): Promise<ExploitType> {
+  if (evaluate(footprint, semver.lt, '2.16.0')) {
+    return 'Log4Shell';
+  }
+
+  if (evaluate(footprint, semver.satisfies, '2.16.x')) {
```
anthogez, line 219: Do we really need this evaluate fn to coerce versions?
We cannot just go with the sample below? Let me know if I am missing some context here thanks!

```
async function computeExploitType(
  footprint: HashFootprint,
): Promise<ExploitType> {

  footprint.versions.forEach((version) => {
    if (semver.lt(version, '2.16.0')) {
      return 'Log4Shell';
    }
    if (semver.satisfies(version, '2.16.x')) {
      return 'DoS';
    }
  });

  return 'Unknown';
}
```

src/cli/commands/log4shell.ts
```diff
@@ -39,88 +74,111 @@ export default async function log4shell(...args: MethodArgs): Promise<void> {
...
-    }
-  }
+        content,
+        accumulator,
+      );
 
-  if (!isJavaArchive(path)) {
-    return;
+      if (isJavaArchive(path) && !isVulnerable) {
+        await parsePaths(InMemoryPaths.from(content, path), accumulator);
```
dekelund, line 148: I guess it's the wrong cache here?

src/cli/commands/log4shell.ts
```diff
@@ -39,88 +74,111 @@ export default async function log4shell(...args: MethodArgs): Promise<void> {
...
-  content: FileContent,
-  path: string,
+async function parsePaths(
+  ctx: OnDiskPaths | InMemoryPaths,
   accumulator: Array<Signature>,
-): Promise<void> {
-  if (isJvmFile(path) || isJavaArchive(path)) {
-    const hash = await computeDigest(content);
+) {
+  if (ctx instanceof OnDiskPaths) {
```
dekelund, line 134: should we really check it like this. I thought we where going to make use of an functional interface instead, at least after the discussion we had yesterday

src/cli/commands/log4shell.ts
```diff
@@ -151,12 +209,92 @@ async function traverse(path: Path, handle: FileHandler) {
...
+  }
+}
+
+function isJavaArchive(path: FilePath) {
   return path.endsWith('.jar') || path.endsWith('.war') || path.endsWith('ear');
 }
 
-function isJvmFile(path: Path) {
-  return path.endsWith('.java') || path.endsWith('.class');
+function isTraceable(path: FilePath) {
```
dekelund, line 271: I would probably rename the function, but I don't have a better suggestion (yet).
it's like a `mightBeVuln` check that we do, but we can't make use of the word "might" in names, so that's not a good name either. Hmm we need to think, so let's keep the `isTraceable` till we know a better name.

src/cli/commands/log4shell.ts
dekelund: @anthogez hmm if path is something else like a absolute path to a file, then check would fail, wouldn't it?

anthogez: Oh I see, did we check if it’s not failing the other way too?

src/cli/commands/log4shell.ts
```diff
@@ -3,21 +3,56 @@ import { promises, Stats } from 'fs';
...
   hash: string;
   path: string;
+  exploitType: ExploitType;
 };
-type Path = string;
 type FileContent = Buffer;
+type FilePath = string;
 type Digest = string;
 
+type InMemoryPath = {
```
dekelund, line 24: I don't think we need these types, do we?
We only need to define how the entries look like, and make an adapter for the files from disk.

The interface that we could use would be something like:
```typescript
interface Entry {
    getData:() => byte[];
}
```

then we are able to remove the if/else from the recursive function. We just apply a function to embed each path in the adapter on line 79. Then we have the same kind of call for both types, right?

anthogez:
(I've deleted a comment I did that is no longer relevant)

anthogez (approved):

PR #2490, opened 2021-12-21
refactor: Load IaC files flow

ofekatr:
#### What does this PR do?
- Updates the `getFilePathsFromDirectory` function to return only IaC file paths.

- `src/cli/commands/test/iac-local-execution/file-loader.ts`

- This PR hosts a simplified implementation for [this previous PR](https://github.com/snyk/snyk/pull/2442).
- Currently in the `test` flow for IaC, when loading IaC files from the provided directory, we do so in the following steps:
    1. Fetch **all** file paths (IaC and non-IaC) inside the directory.
    2. Load all files:
        ii.a If IaC file - Return the loaded file's data.
        ii.b If not IaC file - Return `null`.
    3. Filter out files with `null` as their loaded data.
    4. Return the remaining files' data.

    This PR's implementation offers the following alternative flow:
    1. Fetch **only** IaC file paths inside the directory.
    2. Load these IaC files.
    3. Return the files' data.

    In this implementation, the filtering happens while fetching the directory file paths, and as 
such results with having only IaC files when loading their content.
    This results in fewer iterations at the content loading step, and fewer calls to `tryLoadFileData`.

#### Additional Information
- [Replaces this PR](https://github.com/snyk/snyk/pull/2442)

### DIFF
--- a/src/cli/commands/test/iac-local-execution/file-loader.ts
+++ b/src/cli/commands/test/iac-local-execution/file-loader.ts
@@ -18,54 +18,61 @@ export async function loadFiles(
   pathToScan: string,
   options: IaCTestFlags = {},
 ): Promise<IacFileData[]> {
-  let filePaths = [pathToScan];
+  const filePaths = getFilePathsFromDirectory(pathToScan, {
+    maxDepth: options.detectionDepth,
+  });
 
-  if (isLocalFolder(pathToScan)) {
-    filePaths = getFilePathsFromDirectory(pathToScan, {
-      maxDepth: options.detectionDepth,
-    });
-  }
-
-  const filesToScan: IacFileData[] = [];
-  for (const filePath of filePaths) {
-    try {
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/iac-local-execution/file-loader.ts
```diff
@@ -18,54 +18,61 @@ export async function loadFiles(
...
+    throw new NoFilesToScanError();
   }
 
-  const filesToScan: IacFileData[] = [];
-  for (const filePath of filePaths) {
-    try {
-      const fileData = await tryLoadFileData(filePath);
-      if (fileData) {
-        filesToScan.push(fileData);
+  const loadedFiles: IacFileData[] = await Promise.all(
```
teodora-sandu, line 29: Great catch making this async!

teodora-sandu (approved):
🎉

PR #2494, opened 2021-12-21
fix: remove pro tip for base image remediation

ChristinaDara:

#### What does this PR do?
remove the pro tip for the base image remediation

https://snyksec.atlassian.net/browse/CAP-443

### DIFF
--- a/src/lib/formatters/test/format-test-results.ts
+++ b/src/lib/formatters/test/format-test-results.ts
@@ -170,7 +170,11 @@ export function getDisplayedOutput(
   const fixAdvice = fixTip ? `\n\n${fixTip}` : '';
 
   const dockerfileWarning = getDockerfileWarning(res.scanResult);
-  const dockerSuggestion = getDockerSuggestionText(options, config);
+  const dockerSuggestion = getDockerSuggestionText(
+    options,
+    config,
+    res?.docker?.baseImage,
+  );
 
   const vulns = res.vulnerabilities || [];
   const groupedVulns = groupVulnerabilities(vulns);
@@ -244,7 +248,7 @@ export function dockerUserCTA(options) {
   return '';
 }
 
-function getDockerSuggestionText(options, config): string {
+function getDockerSuggestionText(options, config, baseImageRes): string {
   if (!options.docker || options.isDockerUser) {
     return '';
   }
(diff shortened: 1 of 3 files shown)


jahed-snyk (approved):
It would be good to move all the "Docker"-related functions in `format-test-results.ts` into their own file so that Team Capsule can take ownership via CODEOWNERS. Then you wouldn't need to wait for Hammer's approval for this sort of change.

michael-c-snyk (approved):

src/lib/formatters/test/format-test-results.ts
```diff
@@ -244,7 +243,7 @@ export function dockerUserCTA(options) {
   return '';
 }
 
-function getDockerSuggestionText(options, config): string {
+function getDockerSuggestionText(options, config, res): string {
```
shaninja, line 246: suggestion: since you're using only `res.docker.baseImage`, you can pass only that, and not the entire `res`.

test/jest/unit/lib/formatters/test/__snapshots__/display-result.spec.ts.snap
```diff
@@ -46,8 +46,8 @@ Tested 40 dependencies for known issues, found 3 issues.
 
 Tip: Detected multiple supported manifests (3), use --all-projects to scan all of them at once.
 
-Pro tip: use \`--file\` option to get base image remediation advice.
-Example: $ snyk test --docker src --file=path/to/Dockerfile
+Snyk wasn’t able to auto detect the base image, use \`--file\` option to get base image remediation advice.
+Example: $ snyk container test --docker src --file=path/to/Dockerfile
```
shaninja, line 50: 👏  thank you for improving by changing from `snyk --docker` to `snyk container`.

however (required:) I believe that if we're specifying `snyk container` then there is no need for `--docker`. Please remove `--docker`. (but of course verify that what I am saying is true!)

shaninja (commented):
Great job on your first PR! it wasn't a trivial one.

Some general comments:

1. the commit message is "remove pro tip for base image remediation", but this is not what the PR does. The PR removes the tip on how to get base image remediation advice, if the base image is already detected.
Please rephrase.

 2. I think @jahed-snyk 's suggestion is a great one -- if not in this PR (though in a different commit), please add it in another one which you'll raise after this is merged.

 3. This is a functionality change, but there's no amendment to tests. Please add a test that asserts your change.
 I believe that the right file for it is `test/jest/unit/lib/formatters/test/display-result.spec.ts` (but there could be a better place!)
  
4. I am a huge fan of screenshots (and even GIFs, when applicable!) to help reviewers review. Please add a before / after screenshot, to demonstrate your work 🙏 

 5. There is a checkbox in the PR description "Follows CONTRIBUTING rules". If you're read them and followed them, please ✔️ . 🤗

ChristinaDara:
Screen Shot 2021-12-28 at 10 41 53

shaninja (approved):
Great job on the test! Gives us confidence in the change. It fails if we revert the logic. 👌  which means that it checks exactly this change, and it's what we want.

PR #2502, opened 2021-12-29
fix: Remove scanResult from json output [MAGMA-1280]

snaftaly:

#### What does this PR do?

This PR removes the scanResult object from the json output.
It also refactors the code that creates the json output, therefore it's recommended to review it commit by commit.

Run: `snyk container test node:apline`

In a previous PR (https://github.com/snyk/snyk/pull/2093) the scanResult object was added in order to calculate the appropriate dockerfile analysis warning message shown in the textual output.
As a result the scanResult was also shown in the json output, although it was not intended to.

https://snyksec.atlassian.net/browse/MAGMA-1280


jan-stehlik (approved):
thanks @snaftaly :ship:

src/cli/commands/test/index.ts
```diff
@@ -162,18 +162,12 @@ export default async function test(
...
-      ? errorMappedResults[0]
-      : errorMappedResults;
-
   const {
     stdout: dataToSend,
     stringifiedData,
     stringifiedJsonData,
     stringifiedSarifData,
-  } = extractDataToSendFromResults(results, jsonData, options);
+  } = extractDataToSendFromResults(results, errorMappedResults, options);
```
jahed-snyk, line 170: To reduce the changes in this PR and I think `errorMappedResults` should be renamed to `jsonData`. Both don't have much meaning, we'll need to do a larger refactor to reduce the amount of intermediate types in this giant function.

src/lib/formatters/test/format-test-results.ts
```diff
@@ -100,7 +103,7 @@ export function extractDataToSendFromResults(
     : stringifiedJsonData;
 
   return {
-    stdout: dataToSend, // this is for the human-readable stdout output and is set (but not used) even if --json or --sarif is set
+    stdout: dataToSend, // this is for the human-readable stdout output and is set (but not used) even if --json or --sarif is not set
```
jahed-snyk, line 106: The previous comment makes sense to me. The human-readable stdout is set, even if json/sarif is set and it isn't used. Am I misunderstanding it? 🤔 

It could be phrased better though. Or renamed to "stringifiedHumanReadableData"... or something to avoid confusion. Again, probably a broader refactor in the future. We don't need to calculate all of this stuff if it's not even output.

jahed-snyk (approved):
Looks good. Left some comments. Might be worth getting Snyk Open Source and/or Envelope reps to review as mentioned in the Slack thread (if they haven't already).

src/cli/commands/test/index.ts
snaftaly: @jahed-snyk I think it would make sense to maybe call it mappedResults, or jsonMappedResults rather than jsonData, just to indicate that this is a results object (and not just any general data object as it was before). Wdyt?

src/lib/formatters/test/format-test-results.ts
snaftaly: I think I misunderstood it before, I'll revert the comment.
Anyway, not sure why it says it's not used, because it is 🤔

jahed-snyk: Yeah, since the comment is talking about behaviour outside the scope of its function, it's bound to go out of sync. We could even remove these comments to avoid future confusion.

src/cli/commands/test/index.ts
jahed-snyk: That makes sense. My initial suggestion was more to reduce the amount of changes in this PR as I think renaming variables doesn't improve much here. Ideally we should only have 1 result type at any given moment, rather than multiple.

But, I agree removing the "error" in the name would be an improvement. 👍

snaftaly: @jahed-snyk Cool, renamed it. 
Just wanted to clarify that this refactor was done in order to not have to deal with a single result/array twice, which imho was making the code a bit complex, and instead , only have this logic once after calling createJsonResultOutput function.
If you think it's better to have it in a separate PR I'm ok with that as well. 🙏

src/lib/formatters/test/format-test-results.ts
snaftaly: I kept the comment for now, as it does help, but removed the "(but not used)" part in the comment.

src/cli/commands/test/index.ts
jahed-snyk: I think it's fine to have it in this PR. 👍 

To give some more context for this suggestion: I found the fixture changes difficult to follow since the diff didn't recognise the rename as it also included content changes. It presented the diff as though the files were deleted and new files created (rather than renamed and edited). So I suggested keeping the names the same to avoid the confusing diff history. If needed, a future commit can then apply the renames and git won't get confused.

snaftaly: Let me try to see if I can get git to understand that with different commits in the same PR, otherwise will create a separate PR for that 🙏

snaftaly: So what I did in the end is split the rename of the files to a separate commit, and not to a separate PR.
When looking at the PR changes as a whole it still looks like files were created and removed, but looking at the changes commits by commit helps understanding it.

jahed-snyk: Thanks! That array refactor commit's a lot more readable now. 😄

PR #2504, opened 2021-12-30
feat: add link to docs for base image remediation

SteveShani:

#### What does this PR do?

To try to minimize the questions to support regarding how we give remediation advice, we add a link to the documentation page explaining it.

Run `snyk container test ...`

#### Screenshots

image

### DIFF
--- a/src/lib/formatters/test/format-test-results.ts
+++ b/src/lib/formatters/test/format-test-results.ts
@@ -177,6 +177,7 @@ export function getDisplayedOutput(
     config,
     res?.docker?.baseImage,
   );
+  const dockerDocsLink = getDockerRemediationDocsLink(dockerAdvice, config);
 
   const vulns = res.vulnerabilities || [];
   const groupedVulns = groupVulnerabilities(vulns);
@@ -239,6 +240,7 @@ export function getDisplayedOutput(
     dockerAdvice +
     dockerfileWarning +
     dockerSuggestion +
+    dockerDocsLink +
     dockerCTA
   );
 }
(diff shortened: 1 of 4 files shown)


pecodez:
suggestion: How about adding some tests to check no link is rendered when there is no advice or when disableSuggestions is set to true ? Would give us better coverage 🤔

SteveShani:
> suggestion: How about adding some tests to check no link is rendered when there is no advice or when `disableSuggestions` is set to `true` ? Would give us better coverage 🤔

There are existing tests that check that no link is rendered in the case of no advice 💯 

Regarding `disableSuggestions`, I looked into it and it isn't trivial to add (because "config" gets set when the app loads, and changing the configuration in runtime doesn't change the `config` object.
I vote to skip that test (I manually tested and it works) because it doesn't really have much benefit. If you disagree let me know and I'll find a way to add it 🥲

shaninja:
@SteveShani If you call `config.set('disableSuggestions', 'true');` before calling `displayResult` it should work.

so, test should be something like this:
```
import { config as userConfig } from '??/../src/lib/user-config';
...
describe('disableSuggestions true', ()=> { // in a different describe, so you can have a beforeEach and afterEach, to clear up the config

beforeEach(() => {
userConfig.set('disableSuggestions', 'true');
});

it('should not show suggestion if disableSuggestions is set to true', () => {
  // pretty much the same test as you've added, just different results
});

afterEach(() => {
userConfig.delete('disableSuggestions');
});

```

shaninja:
ok, tested it with @SteveShani and my suggestion actually does not work as I anticipated 😬 
I think our other option is to unit-test `getDockerRemediationDocsLink`, but I am not a big fan of changing code (exporting the func) just for the sake of tests.
not sure if matters greatly. I think I'm ok with not testing `disableSuggestions` for now, but if others think it's important than we can add the unit test (or other idea).

shaninja (approved):

Avishagp (approved):

PR #2510, opened 2022-01-02
fix: exclude base image vulns filtering all vulns

yaronschwimmer:

#### What does this PR do?
A previous [change](https://github.com/snyk/snyk/pull/2138) moved the filtering out of base image vulns (based on
the existence of `dockerfileInstruction`) to be before
`dockerfileInstruction` is assigned to the vulns, so all vulns were
filtered out in case we used the `exclude-base-image-vulns` option.

This fix moves the filtering back to where it was before, and
recalculates the value of `ok`, to account for the reason it was moved
in the first place.

https://snyk.zendesk.com/agent/tickets/16714

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -1,5 +1,5 @@
 import * as fs from 'fs';
-const get = require('lodash.get');
+import * as get from 'lodash.get';
 import * as path from 'path';
 import * as pathUtil from 'path';
 import * as debugModule from 'debug';
@@ -451,6 +451,24 @@ async function parseRes(
       (vuln as DockerIssue).dockerBaseImage = res.docker!.baseImage;
       return vuln;
     });
+  }
+  if (
+    options.docker &&
+    res.docker?.baseImage &&
+    options['exclude-base-image-vulns']
+  ) {
+    const filteredVulns = res.vulnerabilities.filter(
+      (vuln) => (vuln as DockerIssue).dockerfileInstruction,
+    );
+    // `exclude-base-image-vulns` might have left us with no vulns, so `ok` is now `true`
+    if (
+      res.vulnerabilities.length !== 0 &&
+      filteredVulns.length === 0 &&
+      !res.ok
+    ) {
+      res.ok = true;
+    }
+    res.vulnerabilities = filteredVulns;
   }
 
   res.uniqueCount = countUniqueVulns(res.vulnerabilities);
--- a/test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
+++ b/test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
@@ -1,63 +1,60 @@
-import { createFromJSON, DepGraphData } from '@snyk/dep-graph';
-import * as fs from 'fs';
 import * as path from 'path';
-import * as legacy from '../../../../src/lib/snyk-test/legacy';
-import {
-  Options,
-  SupportedProjectTypes,
-  TestOptions,
-} from '../../../../src/lib/types';
 import { fakeServer } from '../../../acceptance/fake-server';
 import { runSnykCLI } from '../../util/runSnykCLI';
 
 describe('container test projects behavior with --app-vulns, --file and --exclude-base-image-vulns flags', () => {
-  const fixturePath = path.normalize('test/fixtures/container-projects');
-
-  function readFixture(filename: string) {
-    const filePath = path.join(fixturePath, filename);
-    return fs.readFileSync(filePath, 'utf8');
-  }
-
-  function readJsonFixture(filename: string) {
-    const contents = readFixture(filename);
-    return JSON.parse(contents);
-  }
-
-  interface fixture {
-    res: legacy.TestDepGraphResponse;
-    depGraph: DepGraphData;
-    packageManager: SupportedProjectTypes;
-    options: Options & TestOptions;
-  }
-
-  const mockApkFixture = readJsonFixture(
-    'app-vuln-apk-fixture.json',
-  ) as fixture;
-  const mockNpmFixture = readJsonFixture(
-    'app-vuln-npm-fixture.json',
-  ) as fixture;
-
-  it('should return no vulnerability for apk fixture', async () => {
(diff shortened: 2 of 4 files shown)


test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
```diff
@@ -1,63 +1,60 @@
...
-import * as legacy from '../../../../src/lib/snyk-test/legacy';
-import {
-  Options,
-  SupportedProjectTypes,
-  TestOptions,
-} from '../../../../src/lib/types';
 import { fakeServer } from '../../../acceptance/fake-server';
 import { runSnykCLI } from '../../util/runSnykCLI';
 
 describe('container test projects behavior with --app-vulns, --file and --exclude-base-image-vulns flags', () => {
```
jahed-snyk, line 5: Can these tests use fake server? Similar to the tests below it. It'd make them faster and less dependent on Registry uptime. You shouldn't need the 10 second timeout once that's done.

test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
```diff
@@ -1,63 +1,60 @@
...
+    expect(jsonOutput[0].ok).toEqual(false);
+    expect(jsonOutput[0].uniqueCount).toEqual(12);
+    expect(jsonOutput[1].ok).toEqual(false);
+    expect(jsonOutput[1].uniqueCount).toEqual(4);
+    expect(code).toEqual(1);
+  }, 10000);
+  it('should return only dockerfile instructions vulnerabilities when excluding base image vulns', async () => {
+    const dockerfilePath = path.normalize(
+      'test/fixtures/container-projects/Dockerfile-vulns',
+    );
```
jahed-snyk, line 30: You can use `createProjectFromFixture` instead so that this path continues to work in the future even if the fixture location changes.

src/lib/snyk-test/run-test.ts
```diff
@@ -452,6 +452,17 @@ async function parseRes(
...
+  if (
+    options.docker &&
+    res.docker?.baseImage &&
+    options['exclude-base-image-vulns']
+  ) {
+    res.vulnerabilities = res.vulnerabilities.filter(
+      (vuln) => (vuln as DockerIssue).dockerfileInstruction,
+    );
+    // `exclude-base-image-vulns` might have left us with no vulns, so `ok` is now `true`
+    res.ok = res.vulnerabilities.length === 0;
```
jahed-snyk, line 464: This seems brittle. Could there be other reasons why a result is not "ok"? e.g. license issues, errors, etc.

Also, can this logic be done in snyk-docker-plugin? Is there a missing API or lack of data that prevents it?

yaronschwimmer: The value of `ok` is decided [here](https://github.com/snyk/snyk/blob/fix/exclude-base-image-vulns-empty/src/lib/snyk-test/legacy.ts#L440) and I didn't see something else that changes the ok-ness in this flow

test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
yaronschwimmer: Initially I tried that, but then these tests failed (consistently) for some reason. Removing the fake server made the tests pass

jahed-snyk: If you still have the failing test code, feel free to share it so that we can troubleshoot and make fake server easier to work with.

yaronschwimmer: git remembers everything 🙂 
https://github.com/snyk/snyk/compare/4c23e11ad0464399cb210a6ee291129ee9fcc4ff..d8f0389b2321341dbb4a37bbf3ed9fad222fbaa2

src/lib/snyk-test/run-test.ts
yaronschwimmer: The bug I'm trying to fix was created when there was an attempt to move the filtering to before we create the result. The problem is that it's also before there's something to base filtering on. I actually moved it back to where it was (+ added the `ok` change to fix the original problem).

What I can do is maybe add some defences, like checking the number of vulns before and after the filter, and only change the `ok` status if that number has changed

jahed-snyk (approved):

src/lib/snyk-test/run-test.ts
jahed-snyk: I'll approve this PR so that the fix can be released. It seems the things we're talking about here might need broader changes.

To clarify my thought process a bit:

We want to be in a position where the API does most of the work. The CLI output being a view of what the API returns. This would allow consistency across other products using the same API without needing to re-implement features (like filtering base image vulns).

Looking at the larger set of mutations done to the response object here, I think both of the docker-specific if-statements here should be moved over to the snyk-docker-plugin (to build the correct request) and related registry endpoint (for the correct response) rather than doing additional post-processing in the CLI. This would help avoid bugs like this as the docker-specific logic will be less fragmented.

So, I'm wondering if there's any blockers that's preventing this logic from being moved to those locations and if there's anything we can do to help unblock that. 🤔

test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts
```diff
@@ -1,63 +1,60 @@
...
+    const { code, stdout } = await runSnykCLI(
+      `container test docker-archive:test/fixtures/container-projects/os-packages-and-app-vulns.tar --json --experimental --app-vulns`,
     );
-    expect(result.vulnerabilities).toHaveLength(10);
-    expect(result.ok).toEqual(false);
-  });
+    const jsonOutput = JSON.parse(stdout);
+
+    expect(jsonOutput[0].ok).toEqual(false);
+    expect(jsonOutput[0].uniqueCount).toEqual(12);
```
ivanstanev, line 22: This might flake as new vulns are introduced. Could these comparisons be somewhat loose like `greaterThan(0)` or similar?

ivanstanev (approved):

michael-c-snyk (approved):

PR #2513, opened 2022-01-04
docs: update code command help

awileysnyk:
This file has all revisions through January 3, 2022, for review by the Hammer team. Question: For commands with the sub-command test, should the test command be explained the same as as the snyk test command: test a project for open source vulnerabilities and license issues? At this time, the code and iac commands say: test for any known issue. The container command says: test for any known vulnerabilities.

### DIFF
--- a/help/cli-commands/code.md
+++ b/help/cli-commands/code.md
@@ -1,4 +1,4 @@
-# snyk code -- Find security issues using Static code analysis
+# snyk code -- find security issues using static code analysis
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


help/cli-commands/code.md
```diff
@@ -1,114 +1,58 @@
...
+Use the `-d` to output the debug logs.
 
-### `SNYK_REGISTRY_PASSWORD`
+## Options
 
-Specify a password to use when connecting to a container registry. Note that using the `--password` flag will override this value. This will be ignored in favour of local Docker binary credentials when Docker is present.
+### `--org=<ORG_NAME>`
 
-### Connecting to Snyk API
+Specify the `<ORG_NAME>`, that is, an Organization in your Snyk account, to run Snyk commands tied to a specific organization. The Organization ID influences private test limits.
```
Avishagp, line 38: I thought the original sentence was more clear. Plus wanted to ask about the usage of `organization ID` instead of `organization name` at the end of the sentence.

maxjeffos, line 38: There's a difference between organization name and organization ID. Your text @awileysnyk mentions them both, which is confusing. We should also make sure we know for sure:
- does it really support an org name? And if so is that the friendly org name or the code-y org name?
- does it support org ID?

For clarification, see picture

image

maxjeffos (changes requested):

help/cli-commands/code.md
awileysnyk: Removed "that is, an Organization in your Snyk account." Changed to: The `<ORG_NAME>` influences private test limits. Organization ID may have come from conversation or another help file. Originally the text said "this influences," presumably meaning specifying the `<ORG_NAME>` or more simply, the `<ORG_NAME>`.

awileysnyk: We thought this would be an organization name found in the user's Snyk account. Which form of the name would be found there?

Avishagp (approved):

PR #2523, opened 2022-01-04
refactor: Refactor Docker-related functions in format-test-results.ts

ChristinaDara:

#### What does this PR do?
Refactor the format-test-results script and split all docker related functions

docker-format-test-results script

The reason why moving this functions is the format-test-results.ts owned by Hammer team and each time we change any function related to docker we need Hammer approval. 
https://snyksec.atlassian.net/browse/CAP-621


.github/CODEOWNERS
```diff
@@ -46,5 +47,4 @@ lerna.json @snyk/hammer
 /tsconfig.json @snyk/hammer
 /tsconfig.settings.json @snyk/hammer
 CODEOWNERS @snyk/hammer
-/docker @snyk/hammer
-
+/docker @snyk/hammer
```
shaninja, line 50: question: is this change intentional?

.github/CODEOWNERS
```diff
@@ -20,6 +20,7 @@ src/lib/iac/ @snyk/group-infrastructure-as-code
 src/lib/snyk-test/iac-test-result.ts @snyk/group-infrastructure-as-code
 src/lib/snyk-test/payload-schema.ts @snyk/group-infrastructure-as-code
 src/lib/snyk-test/run-iac-test.ts @snyk/group-infrastructure-as-code
+src/lib/formatters/docker/docker-format-test-results.ts @snyk/capsule
```
shaninja, line 23: @michael-c-snyk , @agatakrajewska , @RotemS -- do you think it might make sense for Capsule to own all of `src/lib/formatters/docker/`?

shaninja (commented):
Good job refactoring those functions out!

Since our aim is to reduce our dependency on team Hammer when we want to make docker-specific changes, there's another change we should do: extract the tests. Because imagine that we'd want to make a change in the code, and as a result, the test would have to change. if we need Capsule approval for the code but Hammer approval for the test -- then we're still dependant on Hammer.

I know that the functions that you extracted aren't unit-tested directly (neither `groupVulnerabilities` nor `getDisplayedOutput`), but when looking at the tests for `displayResult`, there are test cases there that are distinct for dockers. So my ask is that you'd take `test/jest/unit/lib/formatters/test/display-result.spec.ts` and move all the docker-specific tests out of there and into a new fule.
Keep the new file in the same path (because it's supposed to mirror the code path, and `displayResult` is a function you haven't moved), and add that file also in the CODEOWNERS to have us as it's owners.

Thank you!!

.github/CODEOWNERS
ChristinaDara: I just removed a space

shaninja: yes. I was asking if you intentionally did that, if there was a reason for it. Unfortunately, we don't have linting for CODEOWNER files, and I do not know if it expects a new line at the end (we could check).

(speaking from experience, because I once added a bad character in Registry's CODEOWNER file and it broke all the ownership without anyone noticing for days 😅 🙈 )

ChristinaDara: when I added a line then the regression tests start failing so , I removed the space at the end removed the line and push then added the line again then I pushed it again

shaninja: very strange 🤔  that feels unrelated. maybe hammer would have something to shed some light.

.github/CODEOWNERS
ChristinaDara: I was thinking the same

.github/CODEOWNERS
ChristinaDara: done with refactoring the test file

src/lib/formatters/docker/docker-format-test-results.ts
```diff
@@ -0,0 +1,286 @@
...
+  const multipleVulnsText = res.licensesPolicy ? 'issues' : 'vulnerabilities';
+
+  // Text will look like so:
+  // 'found 232 vulnerabilities, 404 vulnerable paths.'
+  let vulnCountText =
+    `found ${res.uniqueCount} ` +
+    (res.uniqueCount === 1 ? singleVulnText : multipleVulnsText);
+
+  // Docker is currently not supported as num of paths is inaccurate due to trimming of paths to reduce size.
+  if (options.showVulnPaths && !options.docker) {
```
RotemS, line 91: If I understand correctly these results formatter is always container, so we don't need this?

.github/CODEOWNERS
michael-c-snyk: If I'm reading the diff correctly, this PR seems to _add_ a space at the end of 51? What if that new space became a newline?

ChristinaDara: I removed the space at the end and added 2 lines

michael-c-snyk: Screenshot 2022-01-05 at 16 35 04
Ignore me, I was confused by GH diff's view showing a space here, but pulled locally and it's not there

michael-c-snyk: However my IDE shows the diff as being TWO blank lines being removed from the end 🤔 
Screenshot 2022-01-05 at 16 40 28

src/lib/formatters/docker/docker-format-test-results.ts
```diff
@@ -0,0 +1,286 @@
...
+import { ScanResult } from '../../ecosystems/types';
+
+export function groupVulnerabilities(
+  vulns,
+): {
+  [vulnId: string]: GroupedVuln;
+} {
+  return vulns.reduce((map, curr) => {
+    if (!map[curr.id]) {
+      map[curr.id] = {};
```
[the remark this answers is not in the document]
ChristinaDara: done , totally agree fixed that

src/lib/formatters/docker/docker-format-test-results.ts
```diff
@@ -0,0 +1,286 @@
...
+  facts as dockerFacts,
+} from 'snyk-docker-plugin/dist';
+import { ScanResult } from '../../ecosystems/types';
+
+export function groupVulnerabilities(
+  vulns,
+): {
+  [vulnId: string]: GroupedVuln;
+} {
+  return vulns.reduce((map, curr) => {
```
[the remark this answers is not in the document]
ChristinaDara: done , totally agree fixed that

src/lib/formatters/test/format-test-results.ts
```diff
@@ -126,211 +101,6 @@ export function createErrorMappedResultsForJsonOutput(results) {
...
-
-  return (
-    chalk.white('\n\nLearn more: ') +
-    chalk.white.underline(
-      'https://docs.snyk.io/products/snyk-container/getting-around-the-snyk-container-ui/base-image-detection',
-    )
-  );
-}
-
 export function groupVulnerabilities(
```
shaninja, line 104: @ChristinaDara , can you please check if this function should also be removed? I see it in the new file, but it's also still here. probably a rebase error? and then I believe `metadataForVuln` should be removed as well.

shaninja (approved):

shaninja:
Approved after seeing that 2 of Rotem's requests were addressed. [The third one](https://github.com/snyk/snyk/pull/2523#discussion_r778908617)  needs to be investigated, and I believe it can be done and changed in a different PR, if needed (and also, we should now be sole code owners for said PR, so yay!)

src/lib/formatters/test/format-test-results.ts
```diff
@@ -125,265 +96,3 @@ export function createErrorMappedResultsForJsonOutput(results) {
 
   return errorMappedResults;
 }
-
-export function getDisplayedOutput(
```
Avishagp, line 129: Why were `getDisplayedOutput`, `groupVulnerabilities` and `metadataForVuln` moved to `docker-format-test-results`?
`getDisplayedOutput` uses all the other functions, and is used in `display-results.ts` where it's not specific only for docker scans.

src/lib/formatters/test/display-result.ts
```diff
@@ -23,7 +23,7 @@ import {
 import {
   dockerUserCTA,
   getDisplayedOutput,
-} from '../../../lib/formatters/test/format-test-results';
```
Avishagp, line 26: See my comment on `format-test-results.ts`. `getDisplayedOutput` function is not only used for docker, it's not right to move it to a docker specific file.

Avishagp (changes requested):
getDisplayedOutput function is not only used for docker, it's not right to move it to a docker specific file

PR #2524, opened 2022-01-04
feat: IaC SARIF output improvements [CFG-1313] [CFG-1314]

YairZ101:

#### What does this PR do?
The snyk CLI generate SARIF output, which is then used for [code scanning by GitHub]( https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning).
We would like to improve our SARIF output which will enhance the UX when using GH code scanning.

1. Test an IaC file/directory and save the output using `snyk-dev iac test {path to the file/directory} --sarif > sarif.snyk`
2. Create a test repo and upload the `sarif.snyk` file to it
3. Configure the Snyk IaC scanning tool in your repo
4. Commit any changes and check the scanning results in the Securit tab.

The [following Notion page](https://www.notion.so/snyk/SARIF-output-for-GitHub-Security-Alerts-65a89cfa82a74ee79e5d41dfbe452cef) provides additional context.

- [CFG-1313](https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-1313)
- [CFG-1314](https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-1314)

#### Screenshots
Before
image
After
image

### DIFF
--- a/src/lib/formatters/iac-output.ts
+++ b/src/lib/formatters/iac-output.ts
@@ -4,6 +4,8 @@ import * as Debug from 'debug';
 import * as pathLib from 'path';
 import { pathToFileURL } from 'url';
 import upperFirst = require('lodash.upperfirst');
+import camelCase = require('lodash.camelcase');
+import { execSync } from 'child_process';
 
 import {
   IacTestResponse,
@@ -136,11 +138,16 @@ export function createSarifOutputForIac(
   const basePath = isLocalFolder(iacTestResponses[0].path)
     ? pathLib.resolve('.', iacTestResponses[0].path)
     : pathLib.resolve('.');
+  const repoRoot = getRepoRoot();
   const issues = iacTestResponses.reduce((collect: ResponseIssues, res) => {
     if (res.result) {
       // targetFile is the computed relative path of the scanned file
       // so needs to be cleaned up before assigning to the URI
-      const targetPath = res.targetFile.replace(/\\/g, '/');
+      const targetPath = getPathRelativeToRepoRoot(
+        repoRoot,
+        basePath,
+        res.targetFile,
+      );
       const mapped = res.result.cloudConfigResults.map((issue) => ({
         issue,
         targetPath,
@@ -152,18 +159,23 @@ export function createSarifOutputForIac(
 
   const tool: sarif.Tool = {
     driver: {
-      name: 'Snyk Infrastructure as Code',
+      name: 'Snyk IaC',
+      fullName: 'Snyk Infrastructure as Code',
+      informationUri:
+        'https://docs.snyk.io/products/snyk-infrastructure-as-code',
       rules: extractReportingDescriptor(issues),
     },
   };
   return {
+    $schema:
+      'https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json',
     version: '2.1.0',
     runs: [
       {
         // https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317498
         originalUriBaseIds: {
           [PROJECT_ROOT_KEY]: {
-            uri: pathToFileURL(pathLib.join(basePath, '/')).href,
+            uri: pathToFileURL(repoRoot).href,
             description: {
               text: 'The root directory for all project files.',
             },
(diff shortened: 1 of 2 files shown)


src/lib/formatters/iac-output.ts
```diff
@@ -274,3 +288,30 @@ export function mapIacTestResponseToSarifResults(
     };
   });
 }
+
+function toPascalCase(string) {
```
teodora-sandu, line 292: The CLI already uses (lodash.camelcase](https://github.com/snyk/snyk/blob/b0efa8257907977ea3dcda7d13f0f5ea9e7f5953/src/cli/commands/test/iac-test-shim.ts#L8) so could we maybe replace this with what this link does https://medium.com/@robertsavian/javascript-case-converters-using-lodash-4f2f964091cc

````
const camelCase = require('lodash.camelcase');
startCase(camelCase(str)).replace(/ /g, '');
```

src/lib/formatters/iac-output.ts
```diff
@@ -251,6 +264,7 @@ export function mapIacTestResponseToSarifResults(
       : '';
     return {
       ruleId: issue.id,
+      level: getIssueLevel(issue.severity),
```
teodora-sandu, line 267: This property is correct in theory, but it should be different to `defaultConfiguration.level`. So `defaultConfiguration.level` should be the original/defined level (the one from the bundle) and the former the overriding/actual level (from the custom policies configured by the customer): https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#result-object. Sometimes they might be the same (if there are no custom policies), other times they might be different.

It seems we can't get the original level easily so I'm wondering if adding this optional field is then more confusing than useful at this point. WDYT?

src/lib/formatters/iac-output.ts
```diff
@@ -274,3 +288,30 @@ export function mapIacTestResponseToSarifResults(
...
+  const stdout = execSync('git rev-parse --show-toplevel', {
+    encoding: 'utf8',
+    cwd,
+  });
+  return stdout.trim() + '/';
+}
+
+function getPathRelativeToRepoRoot(basePath: string, filePath: string) {
+  const fullPath = pathLib.resolve(basePath, filePath).replace(/\\/g, '/');
+  const repoRoot = getRepoRoot();
```
teodora-sandu, line 315: Unless I saw it wrong, I think we're calling `getRepoRoot` in the same function we call this function. Could we pass the `repoRoot` as an argument instead of making this call again?

Also, do you think it might be safer to move the `.replace(/\\/g, '/')` to be the last function called in this function? Because it's been added for Windows (AFAIR) and I'm not sure what `getRepoRoot` will return for Windows. So:
```
return fullPath.replace(repoRoot, '').replace(/\\/g, '/');
```

src/lib/formatters/iac-output.ts
YairZ101: great proposal, I've changed it.

src/lib/formatters/iac-output.ts
YairZ101: I think it's unnecessary to use both of them if at the end of the line GH will use and display only one value.
I guess we can drop `defaultConfiguration.level` because it gets overwritten, do you agree?

src/lib/formatters/iac-output.ts
YairZ101: I checked with @ipapast (the queen of Windows VMs) the output of this command and it returned the correct output with the expected format.
In addition to that we cannot move the `.replace(/\\/g, '/')` to be the last function called cause we need to match the path format for the `fullPath` and the `repoRoot`, so if we want to create some kind of safeguard what we'll need to add `.replace(/\\/g, '/')` to the repoRoot, do you think it is necessary?

src/lib/formatters/iac-output.ts
teodora-sandu: Hmm I would have removed the one that was added in this PR. Mostly because this is "additional" to `defaultConfiguration.level` and we were already using this successfully

src/lib/formatters/iac-output.ts
teodora-sandu: If it works then I don't think it's necessary 👍

src/lib/formatters/iac-output.ts
YairZ101: done :)

teodora-sandu (approved):
LGTM 🎉

PR #2531, opened 2022-01-10
feat: use same folder and script name as before for docker-desktop tars

StefanScherer:
Signed-off-by: Stefan Scherer <stefan.scherer@docker.com>

#### What does this PR do?

This PR changes the contents of the docker-desktop tars a bit to keep them compatible with the previous `docker-mac-signed-bundle.tar.gz` to make updates easier.

Build snyk, then build the tarball with eg. `./docker-desktop/build.sh darwin arm64`

The old tarball `docker-mac-signed-bundle.tar.gz` has this content:

```
docker/
docker/package.json
docker/...
docker/snyk-mac.sh
```

The new tarballs have a platform specific subfolder, here an example for Mac arm64:

```
snyk-for-docker-desktop-darwin-arm64/
snyk-for-docker-desktop-darwin-arm64/package.json
snyk-for-docker-desktop-darwin-arm64/...
snyk-for-docker-desktop-darwin-arm64/entrypoint.sh
```

When we looked at the new platform specific Mac tarballs they look different and this makes the update scenario for our users more problematic. We would have to change the `~/.docker/scan/config.json` file and adjust the path to the snyk shell script.

We thought about renaming the contents a bit to keep it compatible, but then we came to the conclusion it would make more sense to produce the new tarballs already with this structure, so we don't hide our change.

With this PR the two tarballs are still created at the same place, so your CI should pick them up correctly. Only the contents of the two tarballs now look again like the old tarball:

```
docker/
docker/package.json
docker/...
docker/snyk-mac.sh
```

Follow-up of #2444

### DIFF
--- a/docker-desktop/README.md
+++ b/docker-desktop/README.md
@@ -32,5 +32,5 @@ To test it, you can do the following:
 ```sh
 cd ./binary-releases
 tar xzf snyk-for-docker-desktop-darwin-x64.tar.gz
-./snyk-for-docker-desktop-darwin-x64/entrypoint.sh woof
+./docker/snyk-mac.sh woof
 ```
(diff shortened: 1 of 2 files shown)


StefanScherer:
Thank you. I see that there is 1.823.0 with the updated tarballs. We'll integrate this version into Docker Desktop.

PR #2532, opened 2022-01-10
[ASK-HAMMER] add drift command for iac [CFG-1299] 

moadibfr:
Hi, just creating this PR to show an exemple of what we want to do and ask questions.

For driftctl integration into snyk cli ([CFG-1299](https://snyksec.atlassian.net/browse/CFG-1299)) we want to have new `snyk iac drift` subcommand. That command will need to have it's own subcommand like `snyk iac drift scan` or `snyk iac drift test` for example. These subcommand will also need to have their own flags.

We looked into how we could be integrating that into how our cli and arg parsing in currently working and this show what we thought will be the best way to do.
So to add this subcommand with its own subcommands and flags we are planning to add a drift global command that will error when called without the `iac` mode. We would then have to do with the option given to drift command and do our own parsing.

It sounds odd to us to have a global drift command and we will need to look how we will do the help display. So please tell us if you think there is a better way to do this.

### DIFF
--- a/src/cli/commands/drift.ts
+++ b/src/cli/commands/drift.ts
@@ -1,0 +1,12 @@
+import { MethodArgs } from '../args';
+import { processCommandArgs } from './process-command-args';
+
+export default async function drift(...args: MethodArgs): Promise<any> {
+  const { options, paths } = processCommandArgs(...args);
+  console.log(options);
+  console.log(paths);
+  if (options.iac != true) {
+    throw new Error('BAD MODE');
+  }
+  console.log('snyk iac drift');
+}
--- a/src/cli/commands/index.js
+++ b/src/cli/commands/index.js
@@ -9,6 +9,7 @@ async function callModule(mod, args) {
 const commands = {
   auth: async (...args) => callModule(import('./auth'), args),
   config: async (...args) => callModule(import('./config'), args),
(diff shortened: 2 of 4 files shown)


craigfurman:
The method in this PR is designed to be a temporary measure to get what we need (3rd and 4th level subcommands) in place at all. WDYT hammer? In the longer term, we're cooking up a proposal for CLI plugins (we know there's likely prior art here), but I don't want to distract from this PR with that. We need _some_ drift integration before any such idea is likely to be ready.

It seems like we might be able to (ab)use options and paths to do what we need:

```
❯ snyk-dev iac drift scan --foo=bar --a-unary-flag a-dir-to-scan
]<ref *1> {
  _: [ 'scan', 'a-dir-to-scan', [Circular *1] ],
  foo: 'bar',
  'a-unary-flag': true,
  iac: true
}
[ 'scan', 'a-dir-to-scan' ]
snyk iac drift
```

We can grab the drift subcommand (scan or gen-driftignore) from the path array, and use the rest of that array and the options map to figure out what we need to do.

Help pages are still a bit of an unknown. We'll want separate help pages for the existing `snyk iac test`, and the new `snyk iac drift ...`, without exposing `snyk drift` as a non-iac-mode subcommand to anyone.

PR #2533, opened 2022-01-10
feat: Change getVersion() from async to sync

YairZ101:

#### What does this PR do?
Currently the `getVersion` function is an async function that is being used like a sync one ( by using `await` everywhere).
This PR removes the need for `await` by changing the function to be a sync function.

### DIFF
--- a/src/lib/version.ts
+++ b/src/lib/version.ts
@@ -1,35 +1,14 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import { executeCommand } from './exec';
 
-export async function getVersion(): Promise<string> {
+export function getVersion(): string {
   const root = path.resolve(__dirname, '../..');
 
   const { version } = JSON.parse(
     fs.readFileSync(path.join(root, 'package.json'), 'utf8'),
   );
 
-  if (version && version !== '0.0.0') {
-    return version;
-  }
-
-  try {
-    const [branchName, commitStr, filesCount] = await Promise.all([
-      executeCommand('git rev-parse HEAD', root),
-      executeCommand('git rev-parse --abbrev-ref HEAD', root),
-      executeCommand('git diff --shortstat', root),
-    ]);
-
-    const dirtyCount = parseInt(filesCount as string, 10) || 0;
-    let curr = `${branchName}: ${commitStr}`;
-    if (dirtyCount !== 0) {
-      curr += ` (${dirtyCount} dirty files)`;
-    }
-
-    return curr;
-  } catch (_err) {
-    return 'unknown';
-  }
+  return version;
 }
 
 /**
--- a/src/cli/commands/version.ts
+++ b/src/cli/commands/version.ts
@@ -1,7 +1,7 @@
 import { getVersion, isStandaloneBuild } from '../../lib/version';
 
 export default async function version() {
-  let version = await getVersion();
+  let version = getVersion();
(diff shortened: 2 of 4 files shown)


anthogez (commented):
I understand your point of view regarding this being a refactor, although we are changing the flow from "async" to "sync" would suggest `feat` rather than `refactor`

YairZ101:
> I understand your point of view regarding this being a refactor, although we are changing the flow from "async" to "sync" would suggest `feat` rather than `refactor`

Changed it to `feat`.
Other than that, do you agree / disagree with the change? I'm looking to get some feedback :)

jahed-snyk (changes requested):
As I understand it, this change is for dev convenience, right? To me, we shouldn't be removing async code only to avoid having to write async code. NodeJS is built around async so we should be designing for it. Sync code will cause perf issues and block event loops. So we should be reducing sync flows and optimising for async flows. For example, that `fs.readFileSync` for the `package.json` should be `fs.readFile`.

If an async `getVersion` is causing inconveniences for devs, we should look into improving the way it's retrieved rather than making it synchronous. For example, I noticed we are retrieving it on every HTTP request. A filesystem read and/or 3 git executions on every HTTP call. That's excessive and it should really be gated, retrieve once, stored and reused. This will allow other flows beyond the gate to retrieve it synchronously (since it's just stored state) and solve the dev inconvenience.

---

Side note, I don't think the git-related logic after the version check is ever executed. As far as I know, the package.json version is never 0.0.0 so that branch will always return early.

YairZ101:
> If an async `getVersion` is causing inconveniences for devs, we should look into improving the way it's retrieved rather than making it synchronous. For example, I noticed we are retrieving it on every HTTP request. A filesystem read and/or 3 git executions on every HTTP call. That's excessive and it should really be gated, retrieve once, stored and reused. This will allow other flows beyond the gate to retrieve it synchronously (since it's just stored state) and solve the dev inconvenience.

What is your opinion on saving the version in the config?



> Side note, I don't think the git-related logic after the version check is ever executed. As far as I know, the package.json version is never 0.0.0 so that branch will always return early.

how can I verify that? if that's the case we can simply delete this part...

YairZ101:
> Delete it and see which tests fail, if any. It looks like a dev thing so I doubt we need it. If we can delete it then I think it's fine to change getVersion to sync since that's what it will be without the exec stuff.

Deleted it and it looks good, from your perspective we can merge this change?

src/lib/version.ts
```diff
@@ -13,23 +12,7 @@ export async function getVersion(): Promise<string> {
...
-    let curr = `${branchName}: ${commitStr}`;
-    if (dirtyCount !== 0) {
-      curr += ` (${dirtyCount} dirty files)`;
-    }
-
-    return curr;
-  } catch (_err) {
-    return 'unknown';
-  }
+  return 'unknown';
```
jahed-snyk, line 15: We can remove the "if" and just return version I think. We'll always have package.json and a version. "0.0.0" can be considered "unknown" implictly if it ever does show up.

jahed-snyk (approved):
I think we can simplify it a little bit more. But other than that, looks good. Thanks for bring this up!

src/lib/version.ts
YairZ101: Done, thank you for all your insights during this review process! :)

PR #2535, opened 2022-01-10
fix: block user from scanning IAC files if not entitled [CFG-1090]

wbeuil:
#### What does this PR do?

The `snyk iac test` command is now looking at the `infrastructureAsCode` entitlement before scanning.

1. Verify that `infrastructureAsCode` is true (by default).
2. Run `snyk iac test` command without any errors.
3. Change your `infrastructureAsCode` entitlement to false.
4. Re-run the above command, you should have now an error.

#### Background context

Will copy-paste what I wrote in registry:

We don’t check in the Snyk CLI if a customer has the infrastructureAsCode entitlement. Thus, for customers not entitled to use IAC, they can't import via the web UI but they still can test IAC files via the CLI.

#### More

- [Jira ticket CFG-1090](https://snyksec.atlassian.net/browse/CFG-1090)
- [Registry PR](https://github.com/snyk/registry/pull/25610)

### DIFF
--- a/src/cli/commands/test/iac-local-execution/index.ts
+++ b/src/cli/commands/test/iac-local-execution/index.ts
@@ -28,6 +28,7 @@ import {
   scanFiles,
   trackUsage,
 } from './measurable-methods';
+import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-entitlement-error';
 import { UnsupportedEntitlementFlagError } from './assert-iac-options-flag';
 import { config as userConfig } from '../../../../lib/user-config';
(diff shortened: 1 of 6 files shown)


ipapast:
our smoke tests are broken because we introduced an entitlement check. However, the entitlement is enabled for the snyk org that the smoke tests are using. Checking again to validate what's the root cause

ipapast (approved):
The tests passed now, I also ran it locally with/without entitlement against registry. Confirmed it works like a charm 🙌

jahed-snyk (approved):

PR #2620, opened 2022-01-23
feat: modify analytics [CFG-1446]

YairZ101:

#### What does this PR do?
Today a user can test multiple files/directories by using the command `snyk {product} test {path1} {path2}`.

When doing so we are saving some of the analytics incorrectly, some examples:
`"iac-test-count": [1,2],` (should be a numeric value and not an array)
`"packageManager": ["cloudformationconfig",["terraformconfig","cloudformationconfig"]],` (shouldn't be a nested array)
Those are some IaC examples but the problem is shared across all products.

Objects should probably be handled in [registry](https://github.com/snyk/registry/blob/cdc684900ca34379eedb823e5fe0c0deff4b39d1/src/lib/analytics/cli.ts#L200) by each team separately.

Run a test with multiple args and use the `-d` flag to check the metadata field at the bottom.

https://snyksec.atlassian.net/browse/CFG-1446

#### Screenshots
Before:
image

After:
image

### DIFF
--- a/src/lib/analytics/index.ts
+++ b/src/lib/analytics/index.ts
@@ -98,7 +98,7 @@ async function postAnalytics(
 }
 
 /**
- * Adds a key-value pair to the analytics data `metadata` field. This doesn't send the analytis, just stages it for
+ * Adds a key-value pair to the analytics data `metadata` field. This doesn't send the analytics, just stages it for
  * sending later (via the {@link addDataAndSend} function).
  * @param key
  * @param value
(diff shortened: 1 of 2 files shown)


YairZ101:
> This really needs a test, as it's not clear what the change is doing for the interface.

Done.
If you think I should add anything to the test please let me know.


> Use case like this should also be documented, if it's something we want to support https://github.com/snyk/snyk/blob/master/help/cli-commands/iac.md (documented on gitbook ideally, doesn't have to be part of this PR)

Updating the docs is part of the following [ticket](https://snyksec.atlassian.net/browse/CFG-1477).

I would also like to mention again (just to make sure that we are aligned) the fact that Objects are not taking care of here and instead dealing with them will be in [registry](https://github.com/snyk/registry/blob/cdc684900ca34379eedb823e5fe0c0deff4b39d1/src/lib/analytics/cli.ts#L200) (IaC for example will open a PR to deal with our metrics that we are saving as objects).

JackuB (approved):

PR #2623, opened 2022-01-24
chore(ci): move windows nodejs to choco config

jahed-snyk:
So that it can benefit from caching.

The previous config was parameterised using `parameters.node_version` but this is actually more like a global constant since we don't pass in any parameters and it falls back to defaults.

Progresses #2599

### DIFF
--- a/.circleci/chocolatey.config
+++ b/.circleci/chocolatey.config
@@ -4,4 +4,5 @@
     <package id="gradle" version="6.8.3" />
     <package id="sbt" version="1.5.5" />
     <package id="awscli" version="2.4.12" />
+    <package id="nodejs" version="16.13.2" />
 </packages>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,11 @@ commands:
                   - << parameters.npm_cache_directory >>
   install_sdks_windows:
     steps:
+      - run:
+          name: Removing pre-installed NodeJS
+          command: |
+            $current_node_version = node --version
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -68,6 +68,11 @@ commands:
                   - << parameters.npm_cache_directory >>
   install_sdks_windows:
     steps:
+      - run:
+          name: Removing pre-installed NodeJS
+          command: |
+            $current_node_version = node --version
+            nvm uninstall $current_node_version
```
JackuB, line 75: This looks odd - can't we just override the default? Or use some different image?

jahed-snyk: Windows uses a VM image, so we can't. I tried keeping the pre-installed one but it takes precedence. To change that, we'd require even more brittle setup logic to change PATH ordering.

It's possible in the future once we've got the split builds and testing done, we can remove our custom NodeJS install completely and use the default one that comes with the Windows image.

To clarify, this logic existed before in a separate command, I moved it into a single command.

JackuB (approved):

.circleci/chocolatey.config
```diff
@@ -3,4 +3,5 @@
     <package id="maven" version="3.8.2" />
     <package id="gradle" version="6.8.3" />
     <package id="sbt" version="1.5.5" />
+    <package id="nodejs" version="16.13.2" />
```
JackuB, line 6: Also, I guess this can't be pulled from some global CI config?

PR #2626, opened 2022-01-25
test: added a unit test to validate LCE url is baseURL [NEBULA-303]

metju90:
Added a unit test as per https://snyksec.atlassian.net/browse/NEBULA-303

### DIFF
--- a/test/jest/unit/snyk-code/snyk-code-test.spec.ts
+++ b/test/jest/unit/snyk-code/snyk-code-test.spec.ts
@@ -521,6 +521,33 @@ describe('Test snyk code', () => {
     },
   );
 
+  it('Local Code Engine - Always calls code-client with url coming from sastSettings', async () => {
+    const sastSettings = {
+      sastEnabled: true,
+      localCodeEngine: {
+        url: 'http://lce:31111/api',
+        allowCloudUpload: false,
+        enabled: true,
+      },
+    };
+
+    const analyzeFoldersSpy = analyzeFoldersMock.mockResolvedValue(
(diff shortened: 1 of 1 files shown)


ArturSnyk (approved):

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -521,6 +521,29 @@ describe('Test snyk code', () => {
     },
   );
 
+  it("Local Code Engine - Always calls code-client with url coming from sastSettings", async () => {
```
ArturSnyk, line 524: [nitpicking] 
```suggestion
  it('Local Code Engine - Always calls code-client with url coming from sastSettings', async () => {
```

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -521,6 +521,29 @@ describe('Test snyk code', () => {
...
+
+    const sastSettings = {
+       sastEnabled: true,
+      localCodeEngine: { url: 'http://lce:31111/api', allowCloudUpload: false, enabled: true },
```
ArturSnyk, line 528: [nitpicking]: i think we can make the string as a constant in this test `'http://lce:31111/api'`

metju90: The juice is not worth the squeeze for this case. It's used just in two places.

Besides, It's clearer what `'http://lce:31111/api'` is vs what `validatedLCEurl` is

miiila (approved):

PR #2630, opened 2022-01-26
feat: added a new option: --no-markdown for sarif output

saark-snyk:
#### What does this PR do?
Adding a new flag to options, `--no-markdown`, which removes the `markdown` property from `result.message` when using `sarif` output.

`snyk code test --sarif --no-markdown`

Slack discussion: https://snyk.slack.com/archives/C01U14WSN73/p1642928960104800

This is a result of a support ticket: https://snyk.zendesk.com/agent/tickets/18450

### DIFF
--- a/help/cli-commands/code.md
+++ b/help/cli-commands/code.md
@@ -53,6 +53,10 @@ Print results in JSON format.
 
 Return results in SARIF format.
 
+## `--no-markdown`
+
+Should be used when using `--sarif`. Will remove the `markdown` field from the `result.message` object. Might help if parsing `arguments` is not working properly.
+
 ### `--severity-threshold=low|medium|high|critical`
 
 Report only vulnerabilities at the specified level or higher. Note that the Snyk Code configuration issues do not currently use the `critical` severity level.
--- a/help/cli-commands/code.md
+++ b/help/cli-commands/code.md
@@ -55,7 +55,7 @@ Return results in SARIF format.
 
 ## `--no-markdown`
 
-Should be used when using `--sarif`. Will remove the `markdown` field from the `result.message` object. Might help if parsing `arguments` is not working properly.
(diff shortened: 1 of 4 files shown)


src/lib/plugins/sast/index.ts
```diff
@@ -40,9 +41,13 @@ export const codePlugin: EcosystemPlugin = {
       }
       const numOfIssues = sarifTypedResult!.runs?.[0].results?.length || 0;
       analytics.add('sast-issues-found', numOfIssues);
-
       if (options.sarif || options.json) {
         if (numOfIssues > 0) {
+          if (options['no-markdown']) {
+            sarifTypedResult.runs?.[0].results?.forEach((result) => {
+              result.message = omit(result.message, ['markdown']);
```
pkey, line 48: Apparently `omit` is not great for performance... Would it be better to just use either `delete` or object spread? I'm aware that this code will only be run if `no-markdown` is passed.

robcresswell, line 48: Would be better not to use lodash here. You could call `delete` since you're modifying in place anyway.

robcresswell (approved):
Fine from SOS pov, but I'm not a huge fan; feels like a hack to work around a bug elsewhere?

pkey (approved):

help/cli-commands/code.md
```diff
@@ -53,6 +53,10 @@ Print results in JSON format.
 
 Return results in SARIF format.
 
+## `--no-markdown`
+
+Should be used when using `--sarif`. Will remove the `markdown` field from the `result.message` object. Might help if parsing `arguments` is not working properly.
```
JackuB, line 58: Docs should be written in active voice.
Also, I'm not sure about the "Might help if" part, but up to you.

```suggestion
Removes the `markdown` field from the `result.message` object. Should be used when using `--sarif`. 
```

JackuB (approved):

PR #2633, opened 2022-01-26
fix: detect non-workspace Yarn projects

lili2311:

#### What does this PR do?
There is a bug in the Yarn workspaces processing that exist if:
- test is run with `--all-projects`
- there is a Yarn workspaces project that includes inside of it a non workspace Yarn project

Before this fix a non-workspace Yarn project is skipped for processing due to a bug in the matching logic that checks if we have scanned this already. 

The fix is rewriting the matching + adding tests around that function specifically + acceptance test

https://github.com/snyk/snyk/compare/fix/detect-non-workspace-yarn-projects?expand=1#diff-2194ffc6b9f5b3533db3a8c636c6fcde3d7f163e92de6d915a5dc542a576fb82L155

Use the new fixture added in this PR to see the bug


jan-stehlik (approved):
nice one @lili2311 :tada:

ebickle:
Gave this a try on my sample project and everything looks great now. Thank you so much for all the hard work on the Yarn Workspace support in `--all-projects`. This will go a long way towards boosting our security 😎

PR #2634, opened 2022-01-26
[CFG-1299] add drift subcommand to the iac command

moadibfr:

#### What does this PR do?
This PR is meant to add a new subcommand `snyk iac drift scan`. As a first step integrating driftctl into snyk this will only launch driftctl and mirror driftctl command, scan arguments and return code.

Current limitation are:
- snyk arg parsing does not support variadic args but driftctl support multiple --from. For now running it through snyk cli will only support one from argument.
- snyk cli help was not meant to support help for subcommand so we will need to find a way to show how the new command work.
- currently I only mirrored arguments for drifctl's scan command. If we want to support more command we should check their args.

- Added a new command, that can be found in `src/cli/commands/drift.ts`. It will error using legacy-error if the mode is not iac.
- The command call `parseArgs` and `driftctl` from `src/lib/iac/drift.ts` where most of the code live. `driftctl` function will try to find driftctl executable and download it to a cache directory if needed.

launching `snyk iac drift scan` should show an error like:
```
error configuring Terraform AWS Provider: no valid credential sources for Terraform AWS Provider found.

Please see https://registry.terraform.io/providers/hashicorp/aws
for more information about providing credentials.

Error: NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```
To make it work correctly it will be useful to check [driftctl doc](https://docs.driftctl.com/)

https://snyksec.atlassian.net/browse/CFG-1299

#### Screenshots
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,6 +33,7 @@
         "configstore": "^5.0.1",
         "debug": "^4.1.1",
         "diff": "^4.0.1",
+        "env-paths": "^2.0.0",
         "glob": "^7.1.7",
         "global-agent": "^2.1.12",
         "lodash.assign": "^4.2.0",
@@ -6825,6 +6826,14 @@
       "version": "0.3.0",
       "resolved": "https://registry.npmjs.org/endian-reader/-/endian-reader-0.3.0.tgz",
       "integrity": "sha1-hOykNrgK7Q0GOcRykTOLky7+UKA="
+    },
+    "node_modules/env-paths": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/env-paths/-/env-paths-2.2.1.tgz",
+      "integrity": "sha512-+h1lkLKhZMTYjog1VEpJNG7NZJWcuc2DDk/qsqSTRRCOXiLjeQ1d1/udrUGhqMxUgAlwKNZ0cf2uqan5GLuS2A==",
+      "engines": {
+        "node": ">=6"
+      }
     },
     "node_modules/envinfo": {
       "version": "7.8.1",
@@ -25471,6 +25480,11 @@
       "version": "0.3.0",
       "resolved": "https://registry.npmjs.org/endian-reader/-/endian-reader-0.3.0.tgz",
       "integrity": "sha1-hOykNrgK7Q0GOcRykTOLky7+UKA="
+    },
+    "env-paths": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/env-paths/-/env-paths-2.2.1.tgz",
+      "integrity": "sha512-+h1lkLKhZMTYjog1VEpJNG7NZJWcuc2DDk/qsqSTRRCOXiLjeQ1d1/udrUGhqMxUgAlwKNZ0cf2uqan5GLuS2A=="
     },
     "envinfo": {
       "version": "7.8.1",
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "configstore": "^5.0.1",
     "debug": "^4.1.1",
     "diff": "^4.0.1",
+    "env-paths": "^2.0.0",
     "glob": "^7.1.7",
     "global-agent": "^2.1.12",
     "lodash.assign": "^4.2.0",
--- a/src/cli/commands/drift.ts
+++ b/src/cli/commands/drift.ts
@@ -1,0 +1,32 @@
+import { MethodArgs } from '../args';
+import { processCommandArgs } from './process-command-args';
+import { driftctl, parseArgs } from '../../lib/iac/drift';
+import { UnsupportedEntitlementCommandError } from './test/iac-local-execution/assert-iac-options-flag';
+import { getIacOrgSettings } from './test/iac-local-execution/measurable-methods';
+import config from '../../lib/config';
+
+const legacyError = require('../../lib/errors/legacy-errors');
+
+export default async function drift(...args: MethodArgs): Promise<any> {
+  const { options, paths } = processCommandArgs(...args);
+
+  if (options.iac != true) {
+    return legacyError('drift');
+  }
+
+  const orgPublicId = options.org ?? config.org;
+  const iacOrgSettings = await getIacOrgSettings(orgPublicId);
+
+  if (!iacOrgSettings.entitlements?.iacDrift) {
+    throw new UnsupportedEntitlementCommandError('drift', 'iacDrift');
+  }
+
+  try {
+    const args = parseArgs(paths, options);
+    const ret = await driftctl(args);
+    process.exit(ret);
+  } catch (e) {
+    const err = new Error('Error running `iac drift` ' + e);
+    return Promise.reject(err);
+  }
+}
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -80,6 +80,17 @@ export class UnsupportedEntitlementFlagError extends CustomError {
   }
 }
 
+export class UnsupportedEntitlementCommandError extends CustomError {
+  constructor(key: string, entitlementName: string) {
+    super(
+      `Unsupported command: ${key} - Missing the ${entitlementName} entitlement`,
+    );
+    this.code = IaCErrorCodes.UnsupportedEntitlementFlagError;
+    this.strCode = getErrorStringCode(this.code);
+    this.userMessage = `Command "${key}" is currently not supported for this org. To enable it, please contact snyk support.`;
+  }
+}
+
 /**
  * Validates the command line flags passed to the snyk iac test
  * command. The current argument parsing is very permissive and
--- a/src/lib/iac/drift.ts
+++ b/src/lib/iac/drift.ts
@@ -1,0 +1,324 @@
+import * as debugLib from 'debug';
+import * as child_process from 'child_process';
+import * as os from 'os';
+import envPaths from 'env-paths';
+import * as fs from 'fs';
+import { spinner } from '../spinner';
+import { makeRequest } from '../request';
+import config from '../../lib/config';
+
+const cachePath = config.CACHE_PATH ?? envPaths('snyk').cache;
+const debug = debugLib('drift');
+
+export const driftctlVersion = 'v0.19.0';
+const driftctlChecksums = {
+  'driftctl_windows_386.exe':
+    '0336132cc0c24beaef2535e0a129d146c1a267f6296d5d6bcc7fbe0b02f0bc78',
+  driftctl_darwin_amd64:
+    '07eae8f9e537183031bb78203c1c50a0ca80e114716598946b76baadcbca26566',
+  driftctl_linux_386:
+    '4b83a5644ce72d3eabd915ffc1bba13ad1d61914984801800f598b35db2fe054',
+  driftctl_linux_amd64:
+    '4bfd536e2123667e01b6e5d54acc27733be0db5a00b7fda7f41fabcd9e910e1d',
+  driftctl_linux_arm64:
+    '50bbf8f47ec7dcb9cc09a628444b093a0305aa9b4accc84b7c366a2297390881',
+  'driftctl_windows_arm64.exe':
+    '6eee390fb97998f309ff0491b893d8727f90f45b98f42fd1aa3ebef29fd7fc5b',
+  driftctl_darwin_arm64:
+    '6f37d9f2e385ef81adea9a1f68f2c85f859608dd854416d8ecf629e626bc8b0c',
+  'driftctl_windows_arm.exe':
+    '7ce916deaad289f41c874a4ddff41e9e6195cd5f78821a5769f8e66434be5465',
+  driftctl_linux_arm:
+    'a73193472fb33744f0a344a5f3a5663bd0f4791c393a68ea1b4c219b02eda8f1',
+  'driftctl_windows_amd64.exe':
+    'be423648c164f816ea98eae9694aa7c6679d0a6c33305aceb435cd75821bc730',
+};
+
+const dctlBaseUrl = 'https://github.com/snyk/driftctl/releases/download/';
+const driftctlPath = cachePath + '/driftctl_' + driftctlVersion;
+
+interface DriftCTLOptions {
+  quiet?: true;
+  filter?: string;
+  output?: string;
+  to?: string;
+  headers?: string;
+  'tfc-token'?: string;
+  'tfc-endpoint'?: string;
+  'tf-provider-version'?: string;
+  strict?: true;
+  deep?: true;
+  driftignore?: string;
+  'tf-lockfile'?: string;
+  'config-dir'?: string;
+  from?: string; // TODO We only handle one from at a time due to snyk cli arg parsing
+}
+
+export function parseArgs(
+  commands: string[],
+  options: DriftCTLOptions,
+): string[] {
+  const args: string[] = commands;
+
+  if (options.quiet) {
+    args.push('--quiet');
+  }
+
+  if (options.filter) {
+    args.push('--filter');
+    args.push(options.filter);
+  }
+
+  if (options.output) {
+    args.push('--output');
+    args.push(options.output);
+  }
+
+  if (options.headers) {
+    args.push('--headers');
+    args.push(options.headers);
+  }
+
+  if (options['tfc-token']) {
+    args.push('--tfc-token');
+    args.push(options['tfc-token']);
+  }
+
+  if (options['tfc-endpoint']) {
+    args.push('--tfc-endpoint');
+    args.push(options['tfc-endpoint']);
+  }
+
+  if (options['tf-provider-version']) {
+    args.push('--tf-provider-version');
+    args.push(options['tf-provider-version']);
+  }
+
+  if (options.strict) {
+    args.push('--strict');
+  }
+
+  if (options.deep) {
+    args.push('--deep');
+  }
+
+  if (options.driftignore) {
+    args.push('--driftignore');
+    args.push(options.driftignore);
+  }
+
+  if (options['tf-lockfile']) {
+    args.push('--tf-lockfile');
+    args.push(options['tf-lockfile']);
+  }
+
+  let configDir = cachePath;
+  createIfNotExists(cachePath);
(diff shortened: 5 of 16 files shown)


package.json
```diff
@@ -78,10 +78,13 @@
     "@snyk/inquirer": "^7.3.3-patch",
     "@snyk/snyk-cocoapods-plugin": "2.5.2",
     "@snyk/snyk-hex-plugin": "1.1.4",
+    "@types/isexe": "^2.0.1",
     "@types/marked": "^4.0.0",
+    "@types/which": "^2.0.1",
     "abbrev": "^1.1.1",
     "adm-zip": "^0.5.9",
     "ansi-escapes": "3.2.0",
+    "cache-directory": "^2.0.0",
```
karniwl, line 87: @moadibfr I was able to click through the failed license check.
so first, I added you to the relevant org, check your email inbox to accept the invite.
secondly, the failed check is due to this added package. Now the question is if you think we can use an alternative package? otherwise we'd need to check with the Snyk's OSS license policy to see if we can make an exception for this or not
image

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
```diff
@@ -80,6 +80,17 @@ export class UnsupportedEntitlementFlagError extends CustomError {
...
 }
 
+export class UnsupportedEntitlementCommandError extends CustomError {
+  constructor(key: string, entitlementName: string) {
+    super(
+      `Unsupported command: ${key} - Missing the ${entitlementName} entitlement`,
+    );
+    this.code = IaCErrorCodes.UnsupportedEntitlementFlagError;
+    this.strCode = getErrorStringCode(this.code);
+    this.userMessage = `Command "${key}" is currently not supported for this org. To enable it, please contact snyk support.`;
```
craigfurman, line 90: I think it's a product matter as to what exact message to put here, but I suspect that we don't necessarily want to direct folks towards support for something that's very experimental. That then might change as we move more towards GA. Let's ask our PM about this.

src/lib/iac/drift.ts
```diff
@@ -0,0 +1,245 @@
...
+      debug('Found driftctl in $DRIFTCTL_PATH: %s', dctlPath);
+      return dctlPath;
+    }
+  }
+
+  // Lookup in user path
+  dctlPath = which.sync('driftctl', { all: false, nothrow: true });
+  if (dctlPath != null) {
+    debug('Found driftctl in $PATH: %s', dctlPath);
+    console.log('Found driftctl in $PATH: %s', dctlPath);
```
craigfurman, line 172: ✂️

test/jest/unit/lib/iac/drift.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  it('findDriftctl fallback to $PATH when DRIFTCTL_PATH is not set', async () => {
+    jest.spyOn(isExe, 'sync').mockReturnValue(true);
+    jest.spyOn(which, 'sync').mockReturnValue('value from $PATH');
+    const path = await findDriftCtl();
+    expect(path).toEqual('value from $PATH');
+  });
+
+  it('findDriftctl fallback to cache when DRIFTCTL_PATH is not set and there is nothing in $PATH', async () => {
```
craigfurman, line 79: I think the path precedence is wrong, at least based on our early discussions (let me know if that's changed). Last I checked, we wanted: DRIFTCTL_PATH > local cache, of a pinned version of driftctl - never $PATH.

src/lib/iac/drift.ts
```diff
@@ -0,0 +1,245 @@
...
+  args.push(to);
+
+  debug(args);
+
+  return args;
+}
+
+export async function driftctl(args: string[]): Promise<number> {
+  debug('driftctl(%s)', args);
+
```
craigfurman, line 117: nit: the newlines here don't assist readability IMO

src/cli/commands/drift.ts
```diff
@@ -0,0 +1,35 @@
...
+    return legacyError('drift');
+  }
+
+  const orgPublicId = options.org ?? config.org;
+  const iacOrgSettings = await getIacOrgSettings(orgPublicId);
+
+  if (!iacOrgSettings.entitlements?.iacDriftEntitlement) {
+    throw new UnsupportedEntitlementCommandError(
+      'drift',
+      'iacDriftEntitlement',
```
craigfurman, line 23: if we end up changing the entitlement name here (https://github.com/snyk/registry/pull/25994#discussion_r797687130), we'll need to change it in this PR too

craigfurman (commented):
This seems to work well 😁 

A few notes/questions, but this is very nearly there.

test/jest/unit/lib/iac/drift.spec.ts
moadibfr: sorry I might have overthink this one. Do you think I should remove the path lookup ?

src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
moadibfr: Checked with @sjourdan it should be ok to display an entitlement error.

test/jest/unit/lib/iac/drift.spec.ts
moadibfr: removes

src/cli/commands/drift.ts
moadibfr: 👍

src/lib/iac/drift.ts
moadibfr: 😓

test/jest/unit/lib/iac/drift.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      '--to',
+      'to',
+    ]);
+  });
+
+  it('findDriftctl precedence to env var', async () => {
+    jest.spyOn(isExe, 'sync').mockReturnValue(true);
+    process.env.DRIFTCTL_PATH = 'var env path';
+    const path = await findDriftCtl();
+    delete process.env.DRIFTCTL_PATH;
```
jahed-snyk, line 71: We shouldn't modify globals in unit tests as it affects other tests within the same Jest runtime. This should be covered with acceptance tests instead.

package.json
```diff
@@ -74,17 +74,20 @@
     "@snyk/inquirer": "^7.3.3-patch",
     "@snyk/snyk-cocoapods-plugin": "2.5.2",
     "@snyk/snyk-hex-plugin": "1.1.4",
+    "@types/isexe": "^2.0.1",
     "@types/marked": "^4.0.0",
     "abbrev": "^1.1.1",
     "adm-zip": "^0.5.9",
     "ansi-escapes": "3.2.0",
+    "cachedir": "^2.3.0",
```
jahed-snyk, line 82: env-paths is more active and popular.

https://snyk.io/advisor/npm-package/env-paths
https://snyk.io/advisor/npm-package/cachedir

ghost (dismissed):
Have you looked into writing acceptance tests for this?

`findDriftCtl` unit tests aren't very useful IMO as it's tied to the platform's configuration and relies on mocks and filesystems to work.

It'd be more useful to test `snyk iac drift` actually downloads and launches correct with some args. We can then test against this on various OS/archs.

To write those tests, we can override the cache dir using `XDG_CACHE_HOME` (for test isolation) and provide a mock endpoint and executable for the download similar to how we mock Snyk API requests (to avoid external service dependencies).

test/jest/unit/lib/iac/drift.spec.ts
craigfurman: Good catch - I agree that modifying the environment in a test is not a good idea, but moving the test level to acceptance is not the only solution. We could inject the env value into the findDriftCtl function instead, and pass in empty string / undefined / a valid path in various tests.

src/lib/iac/drift.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  let ext = '';
+  switch (os.platform()) {
+    case 'win32':
+      ext = '.exe';
+      break;
+  }
+
+  return `${dctlBaseUrl}/${driftctlVersion}/driftctl_${platform}_${arch}${ext}`;
```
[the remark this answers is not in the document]
craigfurman: By messing with the function to return a version that can't exist, I get this error end-to-end:

```
❯ node ~/workspace/github.com/snyk/snyk/dist/cli iac drift scan
Error running `iac drift` 404
```

I think it's ok for this function to remain "dumb", but we should catch the subsequent 404 from github and return a more verbose error to the user, e.g. "driftctl_linux_amd64_v0.18.0 (or whatever) not found".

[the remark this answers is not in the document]
moadibfr: Yeah I thought it would be ok leaving the erroring to the part that actually download the tools.

moadibfr:
> `findDriftCtl` unit tests aren't very useful IMO as it's tied to the platform's configuration and relies on mocks and filesystems to work.
Yeah I thought testing the precedence could be a good idea but originally I also was looking into local $PATH. That might have less sense now...

> It'd be more useful to test `snyk iac drift` actually downloads and launches correct with some args. We can then test against this on various OS/archs.
I'll look into this you are totally right that this king of test would be useful. I think we would likely test with only one fake binary regardless of arch and just test the overall wiring. Driftctl have it's own acceptance tests and this is meant to be a wrapper.

src/lib/iac/drift.ts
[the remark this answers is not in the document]
jahed-snyk: Okay, just thought I'd bring this up as a way to provide clearer error messages to the customer. 👍

rontalx:
@craigfurman @moadibfr 
The Arch/OS functionality looks cool, back then in Local-Exec we didn't want to go this way to keep things simple and avoid having to distribute and download multiple Arch/OS binaries of our OPA Bundle. Although it was a viable option, we eventually download WASM from S3 and use a GopherJS compiled version of our HCL parser.
I'm curious to hear if this version of shelling out to a binary feels robust to your folks till now from you experience here and if there any drawbacks you've encountered?
It might be an opportunity for us in Config team as well to think of switching from WASM (for the OPA bundle with the rules) and GopherJS (for HCL Parsing) approaches and adopt the child-process Golang binary approach if it works well for you.

rontalx:
@moadibfr 
one more thing to bare in mind, I've noticed this uses the output engine from DriftCTL, and not any Snyk generated output.
Do you have any plans of aligning it with the new UX Prototype by Esra? (a project we're soon going to prioritize in Config team)

src/lib/iac/drift.ts
```diff
@@ -0,0 +1,236 @@
...
+  debug('driftctl not found');
+
+  return '';
+}
+
+async function download(url, destination: string): Promise<boolean> {
+  debug('downloading driftctl into %s', destination);
+
+  return new Promise<boolean>((resolve, reject) => {
+    needle.get(url, { output: destination, follow: 3 }, function(
```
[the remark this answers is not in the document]
moadibfr: I tried that. Thanks for the help ! I was not really aware of this function and the only examplea of download I found were with needle for tests purposes

package.json
```diff
@@ -74,17 +74,20 @@
...
     "chalk": "^2.4.2",
     "cli-spinner": "0.2.10",
     "configstore": "^5.0.1",
     "debug": "^4.1.1",
     "diff": "^4.0.1",
     "glob": "^7.1.7",
     "global-agent": "^2.1.12",
+    "isexe": "^2.0.0",
```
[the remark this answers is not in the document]
moadibfr: I did my own isExe with fs.access 👍

package.json
moadibfr: changed 👍 
Althought the last version of env path do not work for us due to ESM it seems

test/jest/unit/lib/iac/drift.spec.ts
moadibfr: I deleted the findDriftctl tests as they have little value now that there is only 2 places to look for.

moadibfr:
> Regarding the Arch/OS detection functionality, back then in Local-Exec for IaC we didn't want to go this way to keep things simple and avoid having to distribute and download multiple Arch/OS binaries of our OPA Bundle. Although it was a viable option, we eventually used cross-platform WASM, download it from S3, and used a GopherJS compiled version of our HCL parser. I'm curious to hear if this version of shelling out to a binary feels robust to your folks till now from you experience here and if there any drawbacks you've encountered? 

I think major difference here would be that driftctl is already packaged so it make sense to just use it ? I'm not sure what the limitation of GopherJS actually are. But driftctl need golang plugins (terraform providers) to work and will download them too so I'm not sure of the benefits to recompile it into js.

> It might be an opportunity for us in Config team as well to think of switching from WASM (for the OPA bundle with the rules) and GopherJS (for HCL Parsing) approaches and adopt the child-process Golang binary approach if it works well for you.

We have plan to have metrics on this command run/failure maybe it will give us some insight for you.

> one more thing to bare in mind, I've noticed this uses the output engine from DriftCTL, and not any Snyk generated output.
Do you have any plans of aligning it with the new UX Prototype by Esra? (a project we're soon going to prioritize in Config team)

This PR is the first step for the integration it has very rough edges. But yeah eventually we will use driftctl json output and reformat so snyk cli owns the ux. We still have some work and discussions to have on this point.

moadibfr:
> Have you looked into writing acceptance tests for this?

Made two of them to test the entitlement and that we actually launch an external tool correctly. Sadly env-pats only follows XDG_CACHE_HOME for linux, so the only way to test a regular run reliably on all os would mean overriding XDG_CACHE_HOME, HOME and USERDATA. I choose to not do that as we have a dev purprose env var DRIFTCTL_PATH that allows the test to still run the echo-args bash script I made.

> It'd be more useful to test `snyk iac drift` actually downloads and launches correct with some args. We can then test against this on various OS/archs.

I did not push the real download for the same reason, we cannot control where it's downloaded. Also it might make the test unreliable because we could encounter network issues and GitHub issues that are not linked to our code. Let me know if that's ok.

craigfurman (approved):
Great first iteration, this should give us a base to build on 🎊

package.json
jahed-snyk: Yeah, once we've removed Node 10 support in March, we'll be able to upgrade a bunch of deps.

src/lib/iac/drift.ts
```diff
@@ -0,0 +1,263 @@
...
+    } catch (err) {
+      return Promise.reject(err);
+    }
+  }
+  return dctl;
+}
+
+export async function findDriftCtl(): Promise<string> {
+  // lookup in custom path contained in env var DRIFTCTL_PATH
+  let dctlPath: string | null | undefined = process.env.DRIFTCTL_PATH;
```
jahed-snyk, line 158: For consistency, consider using `./config.default.json` and `src/lib/config.ts` to define configuration. These parameters can be overidden by the customer using `SNYK_<parameter>` env vars.

ghost (commented):
There are some package-lock.json issues which need fixing.

I've left some suggestions about configuration and mocking the download. But not blocking on those.

moadibfr:
> My suggestion involves pointing the download to the local mock server too which would avoid external network calls. So this line could also be configurable like `DRIFTCTL_PATH` and point to the mock server and send back a mock binary:
> 
> ```
> const dctlBaseUrl = 'https://github.com/snyk/driftctl/releases/download/';
> ```

Ok sorry I did not understood that but that would also mean adding another en var to override the cache (as env-paths make it difficult to override it's returned path). I'm usually not really into adding production code that only exists for testing purpose. Do you think that there is enough value in testing the actual downloading part so I do that ?

src/lib/iac/drift.ts
```diff
@@ -0,0 +1,263 @@
...
+import * as child_process from 'child_process';
+import * as os from 'os';
+import envPaths from 'env-paths';
+import * as fs from 'fs';
+import { spinner } from '../spinner';
+import { makeRequest } from '../request';
+
+const paths = envPaths('snyk');
+const debug = debugLib('drift');
+export const driftctlVersion = 'v0.19.0';
```
p0tr3c, line 11: have we considered doing simple checksum validation ? It appears we are pinning specific version, and this should catch any network based MITM attacks.

package-lock.json
```diff
@@ -16513,9 +16522,9 @@
       "integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A=="
     },
     "node_modules/snyk-docker-plugin": {
-      "version": "4.33.0",
-      "resolved": "https://registry.npmjs.org/snyk-docker-plugin/-/snyk-docker-plugin-4.33.0.tgz",
-      "integrity": "sha512-ha85npGG86Ef6ygCtNt1KtOVW+IuoIKjFOIS84ZQJ3uHDSjjlOjdMENW2sN7OQntYtWs01mp73IJW0sbKDRwaw==",
+      "version": "4.34.5",
```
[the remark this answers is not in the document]
moadibfr: yeah it seems that the package got bump in package.json on master so my `npm install` does that...

src/lib/iac/drift.ts
moadibfr: I did that but just not put any default value as we do not want that the default url and default path is computed depending on platform.

src/lib/iac/drift.ts
moadibfr: ok added a checksum check. For now the sum are hardcoded we might want to find a better way to pin version and hash in the future.

moadibfr:
@jahed-snyk 
> On mocked endpoints: Do the current acceptance tests make network calls to GitHub? If so, I think it's worth mocking that endpoint to avoid flakey tests and to avoid spamming GitHub on every test run.

No I just didn't added the download test.

> On providing an env var just for testing: I think it's worth us providing a SNYK_CACHE_PATH (or similar) if the customer doesn't not want to use `env-paths`'s system-default behaviour. Customisable cache paths are common in CLI tools (e.g. for optimising caches in CI) so I don't think this is a test-only feature.

I did that and added the download tests that should be correctly isolated.

gitphill (approved):

ghost (approved):

PR #2638, opened 2022-01-27
refactor: Move smoke tests to acceptance jest tests

ipapast:
This commit converts shellspec tests to jest tests. All tests arerelevant to Terraform IaC scanning.


aron (approved):
👏 interesting that we now need to accomodate Windows paths 😅 so perhaps we update the assertions to only include the filename portion?

ipapast:
> 👏 interesting that we now need to accomodate Windows paths 😅 so perhaps we update the assertions to only include the filename portion?

yes, so for the depth-detection ones where we would print relevant paths, I changed the assertion to include just the filename.

For the legacy tests: I brought back the smoke tests that were testing the legacy command for the time being. 
I think it's the same trade-off of deleting those when we delete the legacy flag 🤷 What do you think @aron?  If not, I can open another PR for updating the fakeServer api calls for the legacy flow (more reviewers).

aron:
> I think it's the same trade-off of deleting those when we delete the legacy flag 🤷 What do you think @aron? If not, I can open another PR for updating the fakeServer api calls for the legacy flow (more reviewers).

I think it's fine to leave them, if it's still going to trigger false smoke test failures when the policies are updated, then I'd prioritize removing the flow.

PR #2640, opened 2022-01-28
feat: added option --sarif-file-output for snyk code test [NEBULA-157]

patricia-v:
### What this does

This PR adds support to the `--sarif-file-output` flag in `snyk code test`.
The commands `snyk test` and `snyk iac test` support this flag, while `snyk code test` only supports `--sarif` but not writing to a file directly.

### Notes for the reviewer

Run `snyk code test --sarif-file-output` in a repository to generate a file with the sarif result. This can be used simultaneously with other flags like `--sarif` and `--no-markdown`.

### More information

- [Jira ticket NEBULA-157](https://snyksec.atlassian.net/browse/NEBULA-157)
- [Zendesk ticket 15314](https://snyk.zendesk.com/agent/tickets/15314)
- [Zendesk ticket 15028](https://snyk.zendesk.com/agent/tickets/15028)

### DIFF
--- a/src/lib/plugins/sast/index.ts
+++ b/src/lib/plugins/sast/index.ts
@@ -40,28 +40,29 @@ export const codePlugin: EcosystemPlugin = {
       }
       const numOfIssues = sarifTypedResult!.runs?.[0].results?.length || 0;
       analytics.add('sast-issues-found', numOfIssues);
-      if (options.sarif || options.json) {
-        if (numOfIssues > 0) {
-          if (options['no-markdown']) {
-            sarifTypedResult.runs?.[0].results?.forEach(({ message }) => {
-              delete message.markdown;
-            });
-          }
-          hasIssues(jsonStringifyLargeObject(sarifTypedResult));
-        }
-        return { readableResult: jsonStringifyLargeObject(sarifTypedResult) };
-      }
       const meta = getMeta(options, path);
       const prefix = getPrefix(path);
-      const readableResult = getCodeDisplayedOutput(
+      let readableResult = getCodeDisplayedOutput(
         sarifTypedResult!,
         meta,
         prefix,
       );
-      if (numOfIssues > 0) {
-        hasIssues(readableResult);
+      let sarifResult;
+      if (numOfIssues > 0 && options['no-markdown']) {
+        sarifTypedResult.runs?.[0].results?.forEach(({ message }) => {
+          delete message.markdown;
+        });
       }
-      return { readableResult };
+      if (options['sarif-file-output']) {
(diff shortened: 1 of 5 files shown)


src/lib/plugins/sast/index.ts
```diff
@@ -58,6 +58,13 @@ export const codePlugin: EcosystemPlugin = {
         meta,
         prefix,
       );
+      if (options['sarif-file-output']) {
```
ArturSnyk, line 61: Notice that on line 45 we are checking if there is an additional flag for `no-markdown`, which we use only when showing a sarif output. i think it's worth considering having this flag to be used in this flow as well.

ArturSnyk, line 61: [question] With this flow would it be possible to *both*, show the output as sarif *and* to save it as an output, by running `snyk code test -- sarif --sarif-file-output`.

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -258,6 +258,46 @@ describe('Test snyk code', () => {
...
 
+  it('should create sarif result when `--sarif-file-output` is used', async () => {
+    const options: ArgsOptions = {
+      path: '',
+      traverseNodeModules: false,
+      showVulnPaths: 'none',
+      code: true,
+      _: [],
+      _doubleDashArgs: [],
+      'sarif-file-output': 'test',
```
ArturSnyk, line 269: [nitpicking]  since sarif is a JSON-based format, i would name the file to reflect that it is indeed a json file
```suggestion
      'sarif-file-output': 'test.json',
```

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -258,6 +258,46 @@ describe('Test snyk code', () => {
...
+      localCodeEngine: {
+        enabled: false,
+      },
+    });
+    trackUsageSpy.mockResolvedValue({});
+
+    try {
+      await snykTest('some/path', options);
+    } catch (error) {
+      // check if error has sarif result
```
ArturSnyk, line 284: `// check if error has sarif result` The error case, in general, is a bit misleading. it's not that we have functional use of an error. here it's more of indicating that we have found vulnerabilities that you indeed check later on (`expect(error.code).toBe('VULNS');`).
I would suggest either removing the comment or updating to something that indicates that we're nearly checking that we have a stringified sarif result

ArturSnyk (approved):

robcresswell (approved):

Avishagp (approved):

PR #2641, opened 2022-01-28
fix: add support for --API to iac test command [CFG-1513]

aron:

#### What does this PR do?

Adds support for the `--API` flag to the `snyk iac test` command.

Start in assert-iac-options-flag.ts and verify the types.ts file. We just needed to add the flag to the safelist of commands supported by the `snyk iac test` flow.

We also have to be careful _not_ to support the lowercase `--api` flag because due to our unique arg parsing and config objects we parse the command line arguments both in `args.ts` _and_ in the snyk-config library. Meaning that command line flags will also end up in the `config` object and we need to be careful to avoid conflicts here.

The flag `--api=foo` will override the `config.api` value which is the auth token saved in the Snyk config file under the `api` key NOT the endpoint. But `--API=foo` will set the `config.API` value which is the _endpoint_ saved in the Snyk config file under the `endpoint` key. What a world we live in.

./bin/snyk iac test --API=<your api> test/fixtures/iac/kubernetes/pod-privileged-multi.yaml

[CFG-1513]

[CFG-1513]: https://snyksec.atlassian.net/browse/CFG-1513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
+++ b/src/cli/commands/test/iac-local-execution/assert-iac-options-flag.ts
@@ -4,6 +4,7 @@ import { getErrorStringCode } from './error-utils';
 import { IaCErrorCodes, IaCTestFlags, TerraformPlanScanMode } from './types';
 
 const keys: (keyof IaCTestFlags)[] = [
+  'API',
   'org',
   'debug',
   'insecure',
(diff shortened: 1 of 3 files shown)


craigfurman (approved):
Nice

aron:
Closing this as on further consideration even supporting this as a first class interface is nonsense.

PR #2666, opened 2022-01-31
chore(ci): run windows tests against snyk-win.exe

jahed-snyk:
This ensures our Windows builds are valid before release.

I've removed `npm_version` from analytics as it's only relevant for OSS NPM scans rather than being "standard data".

Also made general analytics test changes so that they're more flexible to different test environments.


karniwl (approved):
lgtm 👌

maxjeffos (approved):

PR #2692, opened 2022-02-02
fix: remove 'unauthorized' msg for non 403/1 stautscodes [COD-634] [NEBULA-334]

metju90:
When code-client was throwing an error, regardless what type of error, users were seeing this:

```
Unauthorized: Failed to run 'code test'
```

With this PR when running with `-d`, users will see:

image

### DIFF
--- a/src/lib/plugins/sast/index.ts
+++ b/src/lib/plugins/sast/index.ts
@@ -12,7 +12,8 @@ import { FailedToRunTestError, NoSupportedSastFiles } from '../../errors';
 import { jsonStringifyLargeObject } from '../../json';
 import * as analytics from '../../analytics';
 
-const debug = debugLib('snyk-code-test');
+debugLib.enable('snyk-code');
+const debug = debugLib('snyk-code');
 
 export const codePlugin: EcosystemPlugin = {
   // We currently don't use scan/display. we will need to consolidate ecosystem plugins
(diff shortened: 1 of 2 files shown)


ArturSnyk (approved):

pkey (approved):
.?Looks good! One question: would we see connection refused error in the `--debug` mode

robcresswell (dismissed):
Please update the codeowners separately and rebase, which will mean you no longer need our review.

metju90:
@robcresswell I did rebase. But now, since you requested changes, I need your approval again 😬

robcresswell:
You can dismiss reviews :)

src/lib/plugins/sast/index.ts
```diff
@@ -12,7 +12,8 @@ import { FailedToRunTestError, NoSupportedSastFiles } from '../../errors';
 import { jsonStringifyLargeObject } from '../../json';
 import * as analytics from '../../analytics';
 
-const debug = debugLib('snyk-code-test');
+debugLib.enable('snyk-code');
+const debug = debugLib('snyk-code');
```
metju90, line 16: @pkey  so apparently we were always attempting to log more meaningful stuff, but the `debug()` function used further down in the file wasn't enabled.

Now, with -d, it look like this:

image

robcresswell:
I don't see any change to codeowners that you've rebased on that would fix this?

robcresswell:
@metju90 Those codeowners are not for the file you are working in

https://github.com/snyk/snyk/blob/master/.github/CODEOWNERS#L22

metju90:
Thank you, @robcresswell  🙏

PR #2708, opened 2022-02-03
chore: change of ownership order

metju90:


robcresswell (approved):
Nice one. As a general note, please try and write commits that explain what the reasoning is for a change, not just a text explanation of the change. Its fairly obvious that the order is changing, but it may not be obvious _why_ (which is that we believe ordering matters, and this is causing SAST to not own their own files)

jahed-snyk (approved):

PR #2709, opened 2022-02-03
perf: Fix OOM issue caused by group-issues

snaftaly:

#### What does this PR do?

This PR fixes an OOM issue caused by group-issues.

Group issues mechanism was extending the "from" and "name" arrays in the current vuln it was iterating on, and only then assigning it to acc[vuln.id]. This meant that needs to be grouped, an array larger by 1 (of "from" and of "name") is added to the vuln. In large vulns results this may cause may OOM issues, because of duplicate data.

This fix changes the the mechanism so that when acc[vuln.id] exists, it add the current vuln's 'from' and 'name' values to the existing acc[vuln.id].from and name arrays.

**Please note**: In order to keep the "from" and "name" arrays order as it was before, and in order to use the last vuln's data (e.g. upgradePath) in the output, we need to iterate over the reversed array. We then reverse the order back, in order to preserve the overall (grouped) vulns order.
Changing the order of the "from" and "name" arrays, and using the first vuln's data rather than the first was [discussed and agreed with product](https://snyk.slack.com/archives/CBXS33GJY/p1643899263725699), but will be done in a separate PR, as this is not related to the performance issue fixed in this PR.

This can be tested with the following image:
```
# export NODE_OPTIONS="--max-old-space-size=4096"
# snyk container test --app-vulns --json --experimental --group-issues gpuci/rapidsai@sha256:2b84312b0cffa046c7767c3db0baafa18e58d5d7a7aeda5bb87fbcd535ba4f69
```

https://snyksec.atlassian.net/browse/MAGMA-1300

### DIFF
--- a/src/lib/formatters/test/format-test-results.ts
+++ b/src/lib/formatters/test/format-test-results.ts
@@ -45,17 +45,26 @@ function createJsonResultOutput(jsonResult, options: Options) {
 
 function formatJsonVulnerabilityStructure(jsonResult, options: Options) {
   if (options['group-issues']) {
+    // Note: we have to reverse the array to keep the existing behavior so that the json output will stay the same.
+    // Since the entire array is reversed before grouping, we reverse it back after grouping to preserve the grouped vulns order.
+    const reversedVulnerabilities = jsonResult.vulnerabilities
+      ? jsonResult.vulnerabilities.slice().reverse()
+      : [];
     jsonResult.vulnerabilities = Object.values(
-      (jsonResult.vulnerabilities || []).reduce((acc, vuln): Record<
-        string,
-        any
-      > => {
-        vuln.from = [vuln.from].concat(acc[vuln.id]?.from || []);
-        vuln.name = [vuln.name].concat(acc[vuln.id]?.name || []);
-        acc[vuln.id] = vuln;
+      reversedVulnerabilities.reduce((acc, vuln): Record<string, any> => {
+        if (!acc[vuln.id]) {
+          acc[vuln.id] = {
+            ...vuln,
+            from: [vuln.from],
+            name: [vuln.name],
+          };
+        } else {
+          acc[vuln.id].from.push(vuln.from);
+          acc[vuln.id].name.push(vuln.name);
+        }
         return acc;
       }, {}),
-    );
+    ).reverse();
   }
 
   if (jsonResult.vulnerabilities) {


src/lib/formatters/test/format-test-results.ts
```diff
@@ -43,16 +43,25 @@ function createJsonResultOutput(jsonResult, options: Options) {
...
   if (options['group-issues']) {
     jsonResult.vulnerabilities = Object.values(
       (jsonResult.vulnerabilities || []).reduce((acc, vuln): Record<
         string,
         any
       > => {
-        vuln.from = [vuln.from].concat(acc[vuln.id]?.from || []);
-        vuln.name = [vuln.name].concat(acc[vuln.id]?.name || []);
-        acc[vuln.id] = vuln;
+        if (!acc[vuln.id]) {
```
Avishagp, line 53: Can we please lose the "WARNING" message? It causes needless confusion and stress. If you think a comment is needed, either make the code more intuitive or add a simple, short comment like `initialize new vuln.id` in the if statement, `add more vuln paths` in the else.

test/fixtures/basic-apk/resultJsonDataGrouped.json
```diff
@@ -72,40 +72,38 @@
...
         [
           "docker-image|alpine@3.12.1",
-          "openssl/libcrypto1.1@1.1.1g-r0",
+          "musl/musl-utils@1.1.24-r9",
           "musl/musl@1.1.24-r9"
         ],
         [
           "docker-image|alpine@3.12.1",
-          "alpine-baselayout/alpine-baselayout@3.2.0-r7",
+          "zlib/zlib@1.2.11-r3",
```
lili2311, line 90: it would be better this is not changed to confirm this is all the same and instead use `.sort()` in the test assertions

test/fixtures/npm/issue-grouping/multiProjectResultJsonDataGrouped.json
```diff
@@ -529,20 +527,18 @@
...
           [
             "cli-grouping@1.0.0",
             "express@2.5.11",
+            "connect@1.9.2",
             "qs@0.4.2"
           ]
         ],
         "upgradePath": [
           false,
-          "express@2.5.11",
```
lili2311, line 543: we can change upgradePath, this completely changed remediation we give this is not right. This change should have no effect on remediation

snaftaly: this PR changes the upgrade path because it takes it now from the first vuln in the list rather that the last vuln in the list with the same ID (which from what I see may have a different upgrade path). 
Does the order here matters?

test/fixtures/npm/issue-grouping/multiProjectResultJsonDataGrouped.json
```diff
@@ -529,20 +527,18 @@
...
-          "connect@1.9.2",
-          "qs@1.0.0"
```
darscan, line 545: This looks wrong. i.e. The package with the vuln is `qs`.. but this is no longer in the upgrade path at all

src/lib/formatters/test/format-test-results.ts
```diff
@@ -43,16 +43,23 @@ function createJsonResultOutput(jsonResult, options: Options) {
...
+          // initialize new vuln.id with a 'from' and 'name' arrays
+          acc[vuln.id] = vuln;
+          acc[vuln.id].from = [vuln.from];
+          acc[vuln.id].name = [vuln.name];
```
darscan, line 57: Not a fan of mutating the original vuln. Consider:
```suggestion
          acc[vuln.id] = {
            ...vuln,
            from: [vuln.from],
            name: [vuln.name],
          };
```

src/lib/formatters/test/format-test-results.ts
```diff
@@ -43,16 +43,23 @@ function createJsonResultOutput(jsonResult, options: Options) {
...
+        } else {
+          // add current vuln's 'from' and 'name' to the existing vuln 'from' and 'name' arrays
+          acc[vuln.id].from.push(vuln.from);
+          acc[vuln.id].name.push(vuln.name);
```
darscan, line 61: To keep the old upgradePath behaviour consider adding:
```
  acc[vuln.id].upgradePath = vuln.upgradePath;
```
i.e. use the "final" upgradePath which is what was happening before.

darscan: Although.. if we do that then the upgradePath might not match the vuln properly (i.e. vuln package version might not match.. which I think might already be happening)

test/fixtures/npm/issue-grouping/multiProjectResultJsonDataGrouped.json
snaftaly: This is coming from the original non-grouped stubbed data.. See: https://github.com/snyk/snyk/blob/master/test/fixtures/npm/issue-grouping/multiProjectMappedResults.json#L668

src/lib/formatters/test/format-test-results.ts
snaftaly: The main difference here which is not performance-related is that:
1. Before we were always taking the upgrade path of the last vuln with the same ID, where now here we take the first one
2. Before we were taking appending every new vuln of the same ID "from" value to the beginning of the "from" array, so that the orderof the array was reversed.

Looking at the code shared by @lili2311 [here](https://snyk.slack.com/archives/CFTAUCYQ7/p1644414092031229?thread_ts=1644239753.163269&cid=CFTAUCYQ7), the order now is more similar with what's done in Registry, but there is still an issue with the upgrade path not being shown for all vulns, but just for the first one.

In order to keep the same order (and keep the same behavior as before for 1 & 2) I can iterate over the reversed array, but I'm not sure if the order matters here 🤔 

As for fixing the grouping itself - will be happy to understand about what needs to be fixed (but this would be done in a separate task/PR).

darscan: Yep. Understood. What I was pointing out is that now we're showing an upgrade path for the wrong version. Perviously, we essentially returned the last vuln, so the upgrade path matched the vuln package version. Now we're taking the package version from the first vuln, but the upgrade path from the last vuln.

test/fixtures/npm/issue-grouping/multiProjectResultJsonDataGrouped.json
snaftaly: After a discussion with @darscan we have decided to keep the same output as before, and I will create a separate PR make a change in the output's order.

test/fixtures/basic-apk/resultJsonDataGrouped.json
snaftaly: After a discussion with @darscan we have decided to keep the same output as before, and I will create a separate PR make a change in the output's order.

robcresswell (approved):

PR #2747, opened 2022-02-10
fix acceptance test for `snyk iac drift` on windows

moadibfr:

#### What does this PR do?
Fix acceptance test for `snyk iac drift` on windows.
Test was using `/tmp` and custom concatenated path so it could not work.
But the test also run a bash script that echos it's own args. As I'm not sure how to handle this I disabled the test on windows to restore or ability to run acceptance tests.

### DIFF
--- a/src/lib/iac/drift.ts
+++ b/src/lib/iac/drift.ts
@@ -6,6 +6,8 @@ import * as fs from 'fs';
 import { spinner } from '../spinner';
 import { makeRequest } from '../request';
 import config from '../../lib/config';
+import * as path from 'path';
+import * as crypto from 'crypto';
 
 const cachePath = config.CACHE_PATH ?? envPaths('snyk').cache;
 const debug = debugLib('drift');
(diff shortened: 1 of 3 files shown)


src/lib/iac/drift.ts
```diff
@@ -15,7 +17,7 @@ const driftctlChecksums = {
   'driftctl_windows_386.exe':
     '0336132cc0c24beaef2535e0a129d146c1a267f6296d5d6bcc7fbe0b02f0bc78',
   driftctl_darwin_amd64:
-    '07eae8f9e537183031bb78203c1c50a0ca80e114716598946b76baadcbca26566',
+    '07eae8f9e537183031bb78203c1c50a0ca80e114716598946b76baadcbca2656',
```
wbeuil, line 20: Is this normal?

moadibfr: https://github.com/snyk/driftctl/releases/download/v0.19.0/driftctl_SHA256SUMS

wbeuil (approved):

jahed-snyk (approved):

PR #2748, opened 2022-02-11
feat: observability in snyk code [NEBULA-399]

metju90:
https://snyksec.atlassian.net/browse/NEBULA-339

Ignore the request ID difference in one of the screenshots 🙃 

image
image
image

### DIFF
--- a/src/lib/plugins/sast/index.ts
+++ b/src/lib/plugins/sast/index.ts
@@ -1,5 +1,6 @@
 import chalk from 'chalk';
 import * as debugLib from 'debug';
+import { v4 as uuidv4 } from 'uuid';
 import { getCodeAnalysisAndParseResults } from './analysis';
 import { getSastSettings } from './settings';
 import {
@@ -24,15 +25,19 @@ export const codePlugin: EcosystemPlugin = {
     return '';
   },
   async test(paths, options) {
+    const requestId = uuidv4();
+    debug(`Analysis request ID: ${requestId}`);
     try {
       analytics.add('sast-scan', true);
       const sastSettings = await getSastSettings(options);
       // Currently code supports only one path
       const path = paths[0];
+
       const sarifTypedResult = await getCodeAnalysisAndParseResults(
         path,
         options,
         sastSettings,
+        requestId,
       );
 
       if (!sarifTypedResult) {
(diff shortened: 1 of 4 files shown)


src/lib/plugins/sast/index.ts
```diff
@@ -12,8 +13,7 @@ import { FailedToRunTestError, NoSupportedSastFiles } from '../../errors';
 import { jsonStringifyLargeObject } from '../../json';
 import * as analytics from '../../analytics';
 
-debugLib.enable('snyk-code');
-const debug = debugLib('snyk-code');
```
pkey, line 16: question: why change this? not sure what it does though.

metju90: The removed code fixed an issue with not printing debugging code from [here](https://github.com/snyk/snyk/pull/2748/files#diff-2284575013fd6588d30774b796013333e67e588d92cf0b4fee2508d926419614L84). but it fixed it too good, so good that it was outputting printing debug info even without the `-d` flag.

Now it's working as intended, debug info only when `-d` is passed.

pkey: gotcha

metju90: TBH I didn't fully understand how our debug library works, but now we have the desired behaviour ✌️

pkey (approved):

pkey:
@metju90 does [this](https://github.com/snyk/snyk/pull/2756) fix the issue you were seeing [here](https://github.com/snyk/snyk/pull/2748/files#diff-2284575013fd6588d30774b796013333e67e588d92cf0b4fee2508d926419614L15)?

metju90:
> @metju90 does [this](https://github.com/snyk/snyk/pull/2756) fix the issue you were seeing [here](https://github.com/snyk/snyk/pull/2748/files#diff-2284575013fd6588d30774b796013333e67e588d92cf0b4fee2508d926419614L15)?

Good catch, I will check it out 👮

metju90:
@pkey so apparently with me using `debug(snyk-code)` and doing `debug.enable` I broke their stuff. 

I've asked the product [here](https://snyk.slack.com/archives/C01KLME457X/p1644929792211759) on whether we should use `snyk-code` or `snyk` to debug when using Snyk Code.

Avishagp (approved):

PR #2751, opened 2022-02-11
refactor: pull out custom rules code

teodora-sandu:
#### What does this PR do?

This PR extracts the custom and internal Snyk rules code from the main `test` function into its own file. For readability and in preparation of adding extended Terraform language support, which will make this function even bigger.

1. `npm run build`
2. Create a custom rules bundle using https://github.com/snyk/snyk-iac-rules: `snyk-iac-rules build ./fixtures/custom-rules -o bundle.tar.gz`
3. Run the CLI on a file that would trigger the rule (just use the fixtures in that repo): `snyk-dev iac test ./fixtures/custom-rules/rules/CUSTOM-1/fixtures`

The acceptance tests would verify this plus the remote bundles, so if the tests pass then this is good to go.


src/cli/commands/test/iac-local-execution/rules.ts
```diff
@@ -0,0 +1,201 @@
...
+import { config as userConfig } from '../../../../lib/user-config';
+import { isValidUrl } from './url-utils';
+import { CustomError } from '../../../../lib/errors';
+import { getErrorStringCode } from './error-utils';
+
+export async function initRules(
+  iacOrgSettings: IacOrgSettings,
+  options: IaCTestFlags,
+): Promise<{
+  isLocalCustomRules: boolean;
```
ofekatr, line 28: This is not a request, but rather an idea open for discussion.

I've given a bit of consideration to the return type of this function.
What we have here seems to be a tri-state value for provided custom rules.
Custom rules are either:
1. Not provided.
2. Provided locally.
3. Provided remotely.

Instead of returning an object with 2 boolean flags, one for flagging rules provided locally, and one for flagging provided remotely, I'd like to propose returning a union type value here:

```
type RulesOrigin = 'local' | 'remote' | 'none';
```

Which can later be utilized in the `addIacAnalytics` function, where these values are needed.

ofekatr (approved):
Excellent initiative, the main flow feels much cleaner after this refactor.
Managed to successfully run it locally 🚀 
Left a small refactor suggestion open for discussion.

src/cli/commands/test/iac-local-execution/rules.ts
teodora-sandu: That was a really good suggestion. Makes the code much easier to ready 🤩

PR #2754, opened 2022-02-14
Chore/test and release iac parser dep

ipapast:


teodora-sandu (approved):
Approving because this is a huge improvement on the existing process 🎉 
The one thing I noticed (and I'm not too fussed about) is that the error colours don't show up:
Screenshot 2022-02-18 at 09 32 58

ipapast:
> Approving because this is a huge improvement on the existing process 🎉 The one thing I noticed (and I'm not too fussed about) is that the error colours don't show up: Screenshot 2022-02-18 at 09 32 58

Ah I think you are pointing when an internal error happened on the download (make failed)? I didn't have the printError function wrapping it there, so I just added it. Something like this looks okay? @teodora-sandu 

image

teodora-sandu:
Yep looks good @ipapast 👍

PR #2755, opened 2022-02-14
refactor: tap tests codeowners

Avishagp:
Add owners to tap tests.
Delete some old tests that are no longer needed, mainly policy tests that test the external package `snyk-policy`.

[JIRA ticket](https://snyksec.atlassian.net/browse/HMMR-500)


test/tap/policy.test.ts
```diff
@@ -1,80 +0,0 @@
-import * as policy from 'snyk-policy';
-import { test } from 'tap';
-import { extendExpiries } from '../utils';
```
JackuB, line 3: You could also delete the `extendExpiries` function, since it's not used anywhere else

test/tap/get-remote-url.test.ts
```diff
@@ -1,123 +0,0 @@
-import { test } from 'tap';
-import * as Proxyquire from 'proxyquire';
-const proxyquire = Proxyquire.noPreserveCache();
-
-test('getInfo returns null for isFromContainer=true', async (t) => {
-  const {
-    getInfo,
-  } = require('../../src/lib/project-metadata/target-builders/git');
```
JackuB, line 8: Actually not sure about deleting this test - seems like this functionality is not tested anywhere else and is actively used in flows like `snyk monitor`

Avishagp: You're right, my bad!

JackuB (approved):

PR #2759, opened 2022-02-16
test: fix windows acceptance tests for iac

teodora-sandu:

#### What does this PR do?

This PR unblocks the `master` pipeline. I will try to investigate a bit more why only this test fails and the other don't, but for now this should go in to unblock the release of the CLI.

The Windows acceptance tests don't run on non-`master` so I didn't catch this edge case. I'm modifying the tests to build the path using `path.join` to get around the problem seen in this [CircleCI run](https://app.circleci.com/pipelines/github/snyk/snyk/10032/workflows/57b0959f-0dae-43c1-9311-21aea081adbd/jobs/77706) where the Windows path uses `\\` instead of `/`.

I'm also renaming a file that had a typo in it..

The fix can be seen in this [Circle run](https://app.circleci.com/pipelines/github/snyk/snyk/10047/workflows/da5e4364-26c5-4c43-b628-9618cc489971/jobs/77824).


YairZ101 (approved):
LGTM

PR #2761, opened 2022-02-16
test: fix terraform plan tests

teodora-sandu:

#### What does this PR do?

This test started failing because it finds less issues. I think it must be related to https://github.com/snyk/cloud-config-opa-policies/pull/720 but it's hard to tell.

See the test failure [here](https://app.circleci.com/pipelines/github/snyk/snyk/10037/workflows/6ad6d181-4509-4b4a-aeac-01306e874151/jobs/77740)

### DIFF
--- a/test/jest/acceptance/iac/test-terraform-plan.spec.ts
+++ b/test/jest/acceptance/iac/test-terraform-plan.spec.ts
@@ -40,7 +40,7 @@ describe('Terraform plan scanning', () => {
     expect(exitCode).toBe(1);
     expect(stdout).toContain('Infrastructure as code issues:');
     expect(stdout).toContain(
-      'Tested ./iac/terraform-plan/tf-plan-create.json for known issues, found 3 issues',
+      'Tested ./iac/terraform-plan/tf-plan-create.json for known issues',
     );
   });
 


aron (approved):
Hmm, I think it's okay to approve this to unblock the tests, but we should understand what happened here, with the 4.x hotfix I wouldn't expect results to drop.

teodora-sandu:
I've asked Pat if he thinks it could be related. I can see that the scan finds `SNYK-CC-TF-124` and `SNYK-CC-TF-127`, which are mentioned in the release notes: https://github.com/snyk/cloud-config-opa-policies/releases/tag/v0.34.0. That's why I thought it could be related. 

The better thing to do would be to remove the count altogether because this will probably happen again, but sometimes knowing the number of issues found is a good test

aron:
Missing rule is:

```
  ✗ S3 server access logging is disabled [Low Severity] [SNYK-CC-TF-45] in S3
    introduced by input > resource > aws_s3_bucket[terra_ci] > logging
```

p0tr3c:
We should drop count because the number may change for correct reasons. I will see why it has dropped in this case to ensure we didn't introduce "unexpected" regression.

p0tr3c:
I have fix PR for the bundle if this can wait 25 minutes ish. But I would remove the count from the output check anyway.

teodora-sandu:
Okay I'll remove the count then, thanks @p0tr3c !

PR #2764, opened 2022-02-17
feat: add org name to snyk code test [NEBULA-195]

patricia-v:
### What does this PR do?

When running `snyk code test`, the organization would appear as `undefined`.
This was solved by passing the `org` from the `SastSettings` API endpoint (which is the default `org` when not specified by the flag or config) and displaying it in the CLI.

### Notes for the reviewer

Check the [PR](https://github.com/snyk/registry/pull/26282) on the registry side as well for the API changes.

### How should this be manually tested?

Run `snyk code test` in a repository and check the printed `Organization` line. It should not be undefined.

Notes:
- Undefined organizations will assume the default organization.
- Specify an `org` by using the flag `--org=org.name` or by setting the `org` in the config by running `snyk config set org=org.name`
- Check if an org is in the config by running `snyk config get org`
- To remove an assigned org in the config, run `snyk config unset org`

### More information

- [Slack thread](https://snyk.slack.com/archives/C01DGQ3AT09/p1635359391196500)
- [Jira ticket NEBULA-195](https://snyksec.atlassian.net/browse/NEBULA-195)

### Screenshots

_Note: The default `org` for this example was `patricia.vale`. The config had no org specified and no flags were used._

**When running `snyk code test`...**

Before:
image

After:
image

### DIFF
--- a/src/lib/plugins/sast/index.ts
+++ b/src/lib/plugins/sast/index.ts
@@ -45,7 +45,11 @@ export const codePlugin: EcosystemPlugin = {
       }
       const numOfIssues = sarifTypedResult!.runs?.[0].results?.length || 0;
       analytics.add('sast-issues-found', numOfIssues);
-      const meta = getMeta(options, path);
+      let newOrg = options.org;
+      if (!newOrg && sastSettings.org) {
+        newOrg = sastSettings.org;
+      }
+      const meta = getMeta({ ...options, org: newOrg }, path);
       const prefix = getPrefix(path);
       let readableResult = getCodeDisplayedOutput(
         sarifTypedResult!,
--- a/src/lib/plugins/sast/types.ts
+++ b/src/lib/plugins/sast/types.ts
@@ -12,6 +12,7 @@ export interface SastSettings {
   error?: string;
   userMessage?: string;
   localCodeEngine: LocalCodeEngine;
+  org?: string;
 }
 
 export interface TrackUsageResponse {
--- a/test/jest/unit/snyk-code/snyk-code-test.spec.ts
+++ b/test/jest/unit/snyk-code/snyk-code-test.spec.ts
@@ -298,6 +298,77 @@ describe('Test snyk code', () => {
     }
   });
 
+  describe('Default org test in CLI output', () => {
+    beforeAll(() => {
+      userConfig.set('org', 'defaultOrg');
+    });
+
+    afterAll(() => {
+      userConfig.set('org', undefined);
+    });
+
+    it('should show the default org in the output when org is not provided', async () => {
+      const options: ArgsOptions = {
+        path: '',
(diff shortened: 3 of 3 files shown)


src/lib/plugins/sast/index.ts
```diff
@@ -45,6 +45,9 @@ export const codePlugin: EcosystemPlugin = {
       }
       const numOfIssues = sarifTypedResult!.runs?.[0].results?.length || 0;
       analytics.add('sast-issues-found', numOfIssues);
+      if (!options.org && sastSettings.org) {
+        options.org = sastSettings.org;
```
ArturSnyk, line 49: I would not mutate the `option.org` object. this might cause some side effects that we don't expect. Instead, we can in every place we check for `options.org` use `options.org || sastSettings.org`

patricia-v: I agree that changing `option.org` is not the best thing. However, to solve this problem we will need the new org in the `meta` const in the next line. I didn't want to change how `getMeta` is called, so what would be the best solution here? I would have to change the arguments of `getMeta` to either include the `options.org || sastSettings.org` (or a new const with this result), or just the `sastSettings/sastSettings.org`. Do you think this is a good solution, or do I need a new approach here?

pkey: @ArturSnyk I think this is not a bad approach. It's basically saying - if no `org` defined in options let that `org` be the default one. Also, `getMeta` is the only consumer of options and all it does is extract the relevant information.

pkey: @patricia-v have you looked at how other `snyk` commands assign default `org`? Like `snyk test`. As far as I know, this issue is handled there so maybe we can actually use that solution?

pkey (changes requested):
Agree with @ArturSnyk's comment, let's have a test to make sure we prevent this issue in the future.

src/lib/plugins/sast/index.ts
patricia-v: @pkey I have, it is done in a very different way, even in the formatting of the output in the CLI. `snyk test` requests from a few APIs that `snyk code test` doesn't, and one of those returns the org. [Somewhere down the line](https://github.com/snyk/snyk/blob/66f09c231f8fa0326bc06c679da045832d5a59e0/src/lib/formatters/format-test-meta.ts#L17) the response of this request is used as an argument to format the output, including the `Organization`.

In this case I think it would be similar to passing the `sastSettings` in the arguments of `getMeta`.

pkey: @patricia-v sounds good then.

pkey:
@patricia-v let me know if you get stuck with the tests we can look together!

pkey (approved):

PR #2774, opened 2022-02-21
test: migrate policy-display to jest

Avishagp:
Migrate `policy-display` test to jest


test/jest/unit/policy-display.spec.ts
```diff
@@ -31,8 +30,8 @@ test('test sensibly bails if gets an old .snyk format', async (t) => {
       .split('\n')
       .slice(3)
       .join('\n');
-    t.equal(result, expected);
-  } catch (error) {
-    t.threw();
+    expect(result).toEqual(expected);
+  } catch (e) {
+    expect(e.code).toEqual('OLD_DOTFILE_FORMAT');
```
jahed-snyk, line 35: Which `expect` are we relying on here? The one in the try block or catch block? It can't be both 🤔

Avishagp: Trying to figure out where did the `OLD_DOTFILE_FORMAT` error come from and I have no idea... got rid of it. Good catch, thank you

PR #2775, opened 2022-02-21
test: migrate dev-count-analysis to jest

Avishagp:
Merge the current dev-count-analysis jest test with the migrated tap test.


test/jest/unit/dev-count-analysis.spec.ts
```diff
@@ -90,4 +103,231 @@ describe('cli dev count via git log analysis', () => {
...
+    } catch (e) {
+      expect(e.stdout).toBeFalsy();
+      if (isWindows) {
+        expect(e.stderr.includes('not recognized')).toBeTruthy();
+        expect(e.exitCode).toEqual(1);
+      } else {
+        expect(e.stderr.includes('not found')).toBeTruthy();
+        expect(e.exitCode).toEqual(127);
+      }
+    }
```
jahed-snyk, line 152: The stuff in the `catch` block won't be called as `rejects.toThrow` will pass if it throws. Jest's defaults are a bit limited here in that you'll need to work around the need to expect an error and then assert on its properties. Might be easier to explain if we pair on this.

jahed-snyk (commented):
I don't think these are unit tests. There's too much reliance on git. If we have acceptance tests for dev count scenarios, We don't need these tests.

test/jest/unit/dev-count-analysis.spec.ts
```diff
@@ -90,4 +103,197 @@ describe('cli dev count via git log analysis', () => {
...
+    expect(lines.includes('line 4')).toBeTruthy();
+  });
+
+  it('can separate lines Windows line endings', () => {
+    const windowsStyleFileData =
+      'line 0\r\n' + 'line 1\r\n' + 'line 2\r\n' + 'line 3\r\n' + 'line 4\r\n';
+
+    const lines = separateLines(windowsStyleFileData);
+    expect(lines.length).toEqual(5);
+    expect(lines.includes('line 0')).toBeTruthy();
```
jahed-snyk, line 293: We can use `toContain` instead of `includes` and `toBeTruthy`

test/jest/unit/dev-count-analysis.spec.ts
```diff
@@ -90,4 +103,197 @@ describe('cli dev count via git log analysis', () => {
...
+
+    const contributors: {
+      email: string;
+      lastCommitDate: string;
+    }[] = stats.getRepoContributors();
+    expect(contributors.length).toEqual(2);
+
+    expect(
+      contributors.map((c) => c.email).includes('someemail-1@somedomain.com'),
+    ).toBeTruthy();
```
jahed-snyk, line 252: We can split this (and similar lines in this file) into a loop with `expect` for each item (and again using `toContain` instead of includes/toBeTruthy. It will give us better failure messages.

test/jest/unit/dev-count-analysis.spec.ts
```diff
@@ -90,4 +103,197 @@ describe('cli dev count via git log analysis', () => {
...
+      '9dabba3667520f7e2baf1ac75fb58369ea16050c_SNYK_SEPARATOR_some-user-2_SNYK_SEPARATOR_someemail-2@somedomain.com_SNYK_SEPARATOR_2020-02-02T23:23:41+02:00\r\n';
+    const stats: GitRepoCommitStats = parseGitLog(gitLogWindowsLineEndings);
+
+    expect(stats.getCommitsCount()).toEqual(3);
+    expect(stats.getUniqueAuthorsCount()).toEqual(2);
+
+    const uniqueAuthors: Set<string> = stats.getUniqueAuthorEmails();
+    expect(uniqueAuthors.size).toEqual(2);
+
+    expect(uniqueAuthors.has('someemail-1@somedomain.com')).toBeTruthy();
```
jahed-snyk, line 227: We can use `toContain` instead of has/toBeTruthy.

jahed-snyk (approved):
Forgot to mention some assertions we can improve so that we get better failure messages like "expected \<value> to contain <substring/item>" rather than stuff like "expected \<value> to be truthy".

test/jest/unit/dev-count-analysis.spec.ts
Avishagp: There's only two items, one containing `someemail-1@somedomain.com` and the other containing `someemail-2@somedomain.com`. I think a loop would make it too long and unclear. Why do you think this change is needed?
I changed it to call `toContain` instead of this whole `toBeTruthy` trick

jahed-snyk: You're right, we don't need a for loop. 👍 For some reason I took `includes` and processed it as `every`.

jahed-snyk (approved):
Looks good!

PR #2777, opened 2022-02-21
fix: support C++-dependencies with unknown version

dekelund:
Updates the CPP-plugin to version [v2.16.5](https://github.com/snyk/snyk-cpp-plugin/releases/tag/v2.16.5), to include following fixes:

- Add support for larger projects to prevent recursion problems
- Adds support for dependencies with omitted version number:
  - Use dependency ID with empty string as version.
  - Use "unknown" in the standard output, if version has been omitted.

Use `snyk unmanaged test` to verify that some specific projects now works without failures due to number of files.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "rimraf": "^2.6.3",
         "semver": "^6.0.0",
         "snyk-config": "4.0.0",
-        "snyk-cpp-plugin": "2.16.3",
+        "snyk-cpp-plugin": "^2.16.5",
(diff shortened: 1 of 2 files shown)


danlucian (approved):

JackuB (changes requested):
Package-lock needs to be updated as well. You should revert this change, run `npm ci` and then `npm install snyk-cpp-plugin@2.16.5`.

I'm worried that this didn't crash the build or anything 🤔 not even Snyk CLI complained for "out-of-sync-lockfile"

JackuB (approved):

PR #2781, opened 2022-02-22
test: migrate cli-json-file-output to jest

Avishagp:
Migrate `cli-json-file-output` test to jest.


jahed-snyk (commented):
Those are acceptance tests rather than unit tests (they're lauching the whole CLI).

I think we have JSON output covered already in Jest here: https://github.com/snyk/snyk/blob/master/test/jest/acceptance/cli-json-file-output.spec.ts

If there's any missing scenarios that those Tap tests covered, we can move them to that existing one.

jahed-snyk (approved):

PR #2807, opened 2022-02-28
[CFG-1510] allow multiple from params to snyk iac drift

moadibfr:

#### What does this PR do?
driftctl support variadic --from to input multiple state source. We decided to go with coma separated value in snyk cli in order to not change too much how the arg parsing is working.

https://github.com/snyk/snyk/pull/2812

### DIFF
--- a/help/cli-commands/iac-describe.md
+++ b/help/cli-commands/iac-describe.md
@@ -196,34 +196,32 @@ the `/tmp` folder.
 Currently, driftctl only supports reading IaC from a Terraform state. We are investigating to support the Terraform code
 as well, as a state does not represent an intention.
 
-Multiple states can be read by passing `--from` flags. You can also use glob patterns to match multiple state files at
+Multiple states can be read by passing `--from` a coma separated list. You can also use glob patterns to match multiple state files at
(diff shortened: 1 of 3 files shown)


eliecharra (approved):
That sounds good to me as a first pragmatic approach 👍🏻 

I'm not sure that the comma is the right character for that because it may be possible to use in within some backend paths. For example [its allowed](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines-special-handling) to use a comma in a key path on S3.
 
Anyway, that limitation is acceptable for now.

moadibfr:
> That sounds good to me as a first pragmatic approach 👍🏻
> 
> I'm not sure that the comma is the right character for that because it may be possible to use in within some backend paths. For example [its allowed](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines-special-handling) to use a comma in a key path on S3.
> 
> Anyway, that limitation is acceptable for now.

totally right. it's also allowed in path in some fs for what I know. But as a first step that will dot it...

PR #2814, opened 2022-03-01
feat: log protect removal message

Avishagp:
Log protect removal message when protect is called.

Screen Shot 2022-03-03 at 10 51 03

### DIFF
--- a/src/cli/commands/protect/index.ts
+++ b/src/cli/commands/protect/index.ts
@@ -1,15 +1,22 @@
 import * as debugModule from 'debug';
 import * as snyk from '../../../lib/';
 import * as types from '../../../lib/types';
-const protect = require('../../../lib/protect');
 import * as analytics from '../../../lib/analytics';
 import * as detect from '../../../lib/detect';
 import * as pm from '../../../lib/package-managers';
 import { CustomError } from '../../../lib/errors';
 import { LegacyVulnApiResult } from '../../../lib/snyk-test/legacy';
 import * as errors from '../../../lib/errors';
+import * as theme from '../../../lib/theme';
 
+const protect = require('../../../lib/protect');
 const debug = debugModule('snyk');
+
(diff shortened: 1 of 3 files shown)


src/cli/commands/protect/index.ts
```diff
@@ -8,9 +8,18 @@ import * as pm from '../../../lib/package-managers';
...
 import { LegacyVulnApiResult } from '../../../lib/snyk-test/legacy';
 import * as errors from '../../../lib/errors';
+import * as theme from '../../../lib/theme';
+import { EOL } from 'os';
 
 const debug = debugModule('snyk');
 
+function addDeprecationMessage(): string {
+  return theme.color.status.warn(
+    `${theme.icon.WARNING} WARNING: The protect command will be deprecated and removed starting from 21 March 2022.\nPlease use the @snyk/protect package instead.\nFor more info: https://github.com/snyk/snyk/tree/master/packages/snyk-protect#readme` +
```
JackuB, line 18: I would use the `snyk.co` link, so we could control it if needed

src/cli/commands/protect/index.ts
```diff
@@ -8,9 +8,18 @@ import * as pm from '../../../lib/package-managers';
...
+      EOL,
```
JackuB, line 19: Nit: any reason why are you combining the string literal `` ` `` and a string concat `+` for the EOL?

src/cli/commands/protect/index.ts
```diff
@@ -66,10 +75,12 @@ export default async function protectFunc(
 
   try {
     const policy = await snyk.policy.load(protectOptions['policy-path']);
+    const msg = addDeprecationMessage();
```
JackuB, line 78: Nit: `msg` is a really generic. Something like `deprecationMessage` might be more descriptive

JackuB (approved):
Noticed a few small things

PR #2820, opened 2022-03-02
[CFG-1639] Add default --no-version-check flag to driftctl

sundowndev:

#### What does this PR do?

Driftctl update checking should be turned off when running driftctl from Snyk CLI. This PR adds `--no-version-check` option to driftctl by default so it doesn't perform this check anymore. This is a global flag in driftctl so I added it for both scan and gen-driftignore commands.

```
$ SNYK_API=http://localhost:8000/api/v1 ./bin/snyk iac describe --service=aws_s3
```

- https://snyksec.atlassian.net/browse/CFG-1639
- https://docs.driftctl.com/0.22.0/usage/flags/version-check

#### Screenshots

image

### DIFF
--- a/src/lib/iac/drift.ts
+++ b/src/lib/iac/drift.ts
@@ -50,6 +50,7 @@ const driftctlChecksums = {
 
 const dctlBaseUrl = 'https://github.com/snyk/driftctl/releases/download/';
 const driftctlPath = path.join(cachePath, 'driftctl_' + driftctlVersion);
+const driftctlDefaultOptions = ['--no-version-check'];
 
 export interface DriftctlGenDriftIgnoreOptions {
   input?: string;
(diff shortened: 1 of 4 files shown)


src/lib/iac/drift.ts
```diff
@@ -119,7 +120,7 @@ export function parseArgs(
 const parseGenDriftIgnoreFlags = (
   options: DriftctlGenDriftIgnoreOptions,
 ): string[] => {
-  const args: string[] = [];
+  const args: string[] = [...driftctlDefaultOptions];
```
eliecharra, line 123: There is something weird here 🤔 
On master, exactly like the scan, there is a `gen-driftignore` key here, check here https://github.com/snyk/snyk/blob/master/src/lib/iac/drift.ts#L88

I don't understand that diff can you double check that ?

eliecharra (changes requested):

src/lib/iac/drift.ts
sundowndev: It's probably just my branch not being up to date with master, I have to rebase it

eliecharra (approved):

PR #2770, opened 2022-02-18
feat: dereference variables from terraform.tfvars and *.auto.tfvars [CFG-1502]

teodora-sandu:
- [x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules

#### What does this PR do?
This PR enables support for dereferencing variables defined in`terraform.tfvars` and `*.auto.tfvars` . It does not aim to support `terraform.tfvars.json` and `*.auto.tfvars.json` files, or files provided via the `-var` and `-var-file`flags. Tfvars files are explained [here](https://www.terraform.io/language/values/variables#variable-definitions-tfvars-files).

It requires https://github.com/snyk/snyk-iac-parsers/pull/12 to be merged first in order to generate a new release of the parser.

#### Where should the reviewer start?
There are only two code changes required in `src/cli/commands/test/iac-local-execution/index.ts`, to start reading `.tfvars` files. The rest of the changes in this PR are tests, making sure we are able to read and dereference variables from both `terraform.tfvars` and `*.auto.tfvars` files, with the right priority.

#### How should this be manually tested?
1. `npm run build`
2. Enable the `iacTerraformVarSupport` feature flag
3. Run `snyk iac test ./test/fixtures/iac/terraform/var_deref` and see only one issue
5. Run `snyk-dev iac test ./test/fixtures/iac/terraform/var_deref` and see four issues

#### What are the relevant tickets?
Jira ticket: https://snyksec.atlassian.net/browse/CFG-1502

#### Screenshots
- before
Screenshot 2022-02-21 at 18 15 20

- after
Screenshot 2022-02-18 at 19 53 11

PR #2832, opened 2022-03-07
feat: update snyk-iac-parsers version

YairZ101:

#### What does this PR do?
Adds support for local variables by updating the `snyk-iac-parsers` version.

Test a TF file that uses local variables and experience  M A G I C!!

https://snyksec.atlassian.net/browse/CFG-1500

### DIFF
--- a/test/jest/acceptance/iac/test-directory.spec.ts
+++ b/test/jest/acceptance/iac/test-directory.spec.ts
@@ -47,7 +47,7 @@ describe('Directory scan', () => {
     expect(stdout).toContain('Failed to parse YAML file');
     expect(stdout).toContain('Failed to parse JSON file');
     expect(stdout).toContain(
-      '22 projects, 16 contained issues. Failed to test 8 projects.',
+      '22 projects, 15 contained issues. Failed to test 8 projects.',
     );
   });
 
--- a/test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
+++ b/test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
@@ -22,22 +22,23 @@ describe('Terraform Language Support', () => {
       const { stdout, exitCode } = await run(
         `snyk iac test ./iac/terraform/var_deref`,
       );
-      expect(exitCode).toBe(1);
+
+      // expect exitCode to be 0 or 1
+      expect(exitCode).toBeLessThanOrEqual(1);
 
       expect(stdout).toContain('Testing sg_open_ssh.tf...');
-      expect(stdout).toContain('Infrastructure as code issues:');
-      expect(stdout).not.toContain('✗ Security Group allows open ingress');
-      expect(stdout).not.toContain(
-        ' input > resource > aws_security_group[allow_ssh] > ingress',
+      expect(stdout.match(/✗ Security Group allows open ingress/g)).toBeNull();
+      expect(stdout).toContain('Tested sg_open_ssh.tf for known issues');
+
+      expect(stdout).toContain(
+        `Testing ${path.join('nested_var_deref', 'sg_open_ssh.tf')}...`,
       );
-      expect(stdout).not.toContain(
-        ' input > resource > aws_security_group[allow_ssh_terraform_tfvars] > ingress',
-      );
-      expect(stdout).not.toContain(
-        ' input > resource > aws_security_group[allow_ssh_a_auto_tfvars] > ingress',
-      );
-      expect(stdout).not.toContain(
-        ' input > resource > aws_security_group[allow_ssh_b_auto_tfvars] > ingress',
+      expect(stdout.match(/✗ Rule allows open egress/g)).toBeNull();
+      expect(stdout).toContain(
+        `Tested ${path.join(
+          'nested_var_deref',
+          'sg_open_ssh.tf',
+        )} for known issues`,
       );
     });
   });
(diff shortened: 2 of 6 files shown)


test/jest/acceptance/iac/test-directory.spec.ts
```diff
@@ -47,7 +47,7 @@ describe('Directory scan', () => {
     expect(stdout).toContain('Failed to parse YAML file');
     expect(stdout).toContain('Failed to parse JSON file');
     expect(stdout).toContain(
-      '22 projects, 16 contained issues. Failed to test 8 projects.',
+      '22 projects, 15 contained issues. Failed to test 8 projects.',
```
YairZ101, line 50: 16 -> 15 because now `nested_var_deref/sg_open_ssh.tf` doesn't include an issue.

test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
```diff
@@ -78,23 +72,17 @@ describe('Terraform Language Support', () => {
...
         expect(stdout).toContain(
-          ' input > resource > aws_security_group[allow_ssh_b_auto_tfvars] > ingress',
+          'Tested sg_open_ssh.tf for known issues, found 5 issues',
         );
 
+        expect(stdout).toContain('Testing nested_var_deref/sg_open_ssh.tf...');
+        expect(stdout.match(/✗ Rule allows open egress/g)).toHaveLength(1);
         expect(stdout).toContain(
-          `Testing ${path.join('nested_var_deref', 'sg_open_ssh.tf')}...`,
+          'Tested nested_var_deref/sg_open_ssh.tf for known issues, found 1 issues',
```
teodora-sandu, line 85: The `path.join` is required for Windows: https://github.com/snyk/cli/pull/2759

And we agreed in a previous PR that we don't check the number of issues like `found x issues` : https://github.com/snyk/cli/pull/2770#discussion_r812741353

test/jest/acceptance/iac/test-directory.spec.ts
teodora-sandu: What is the reason you removed the issue from there?

YairZ101: sry I should have provided more context, I didn't really deleted the issue I simply started using the variable that was configured in the directory.
Check that change I did in `test/fixtures/iac/terraform/var_deref/nested_var_deref/sg_open_ssh.tf` to fully understand why there is one less project that contained an issue.

test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
YairZ101: My memory sucks, thx for reminding me.

test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
```diff
@@ -22,22 +22,16 @@ describe('Terraform Language Support', () => {
       const { stdout, exitCode } = await run(
         `snyk iac test ./iac/terraform/var_deref`,
       );
-      expect(exitCode).toBe(1);
+      expect(exitCode).toBe(0);
```
YairZ101, line 25: What should we do in such cases? when we are expecting 0 issues its fine to test for the exact number?

teodora-sandu: This exit code says there were no issues found. The problem here is that there might be some issues in the future. That's why I purposefully made this test fail. It might be a silly worry though so will leave it up to you to deice

However, I was checking the `stdout` specifically for the var deref issues to make sure those don't get flagged up. What I'd say is instead of saying `found 0 issues` you could do what you're doing in the other test:
```
expect(
          stdout.match(/✗ Security Group allows open ingress/g),
        ).toHaveLength(0);
```

YairZ101: Tried doing this but I received an error:
```
    Matcher error: received value must have a length property whose value must be a number

    Received has value: null

      28 |       expect(
      29 |         stdout.match(/✗ Security Group allows open ingress/g),
    > 30 |       ).toHaveLength(0);
         |         ^
      31 |       expect(stdout).toContain(
      32 |         'Tested sg_open_ssh.tf for known issues, found 0 issues',
      33 |       );
```

used another implementation.

test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
```diff
@@ -22,22 +22,23 @@ describe('Terraform Language Support', () => {
       const { stdout, exitCode } = await run(
         `snyk iac test ./iac/terraform/var_deref`,
       );
-      expect(exitCode).toBe(1);
+
+      // expect exitCode to be 0 or 1
+      expect(exitCode).toBeLessThanOrEqual(1);
```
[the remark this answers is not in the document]
YairZ101: I agree that that in cases where we expect to see an issue we shouldn't use this 0 or 1 solution.
I used it in this cause currently we are not expecting to get any issues but we said that we don't know what the future will hold so I've used this instead of hard coding 0 which might cause a fail in the future.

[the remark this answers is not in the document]
YairZ101: > I think it's safer overall to revert the change you did to test/fixtures/iac/terraform/var_deref/nested_var_deref/sg_open_ssh.tf and leave this check as it was. I don't think we gain anything by using the variable in this file, or if we do then we can at least change its value to one that generates a vulnerabilities

but then what's the purpose of this file? we want to test the vars deref is working inside sub-dirs, before my change we never checked that...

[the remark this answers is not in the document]
YairZ101: that's what I did :|

teodora-sandu (approved):
Approving to unblock! One note though - if the commit is a `chore:` we won't actually have a new version of the CLI with our feature. So could you change it to a `feat:` please 🙇

test/jest/acceptance/iac/test-terraform-var-deref.spec.ts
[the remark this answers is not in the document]
teodora-sandu: I understand what happened. You replaced `cidr_blocks` (which was `["0.0.0.0/0"]`) with the variable (which is `"0.0.0.0/0"`. Notice the lack of square brackets. If you change the variable to an array or the way you assign `cidr_blocks` it'll fix this and the test in `test/jest/acceptance/iac/test-directory.spec.ts`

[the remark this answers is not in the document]
YairZ101: hmmm that's interesting, in both cases (with and without the brackets) we find the issue.
so technically you are correct that the brackets are needed but they don't affect the outcome of the test.

PR #2834, opened 2022-03-07
Always ask driftctl json format and pipe it to `driftctl fmt`

eliecharra:

https://snyksec.atlassian.net/browse/CFG-1646


src/lib/iac/drift.ts
```diff
@@ -240,34 +283,48 @@ export function translateExitCode(exitCode: number) {
...
+    env: { ...process.env, DCTL_IS_SNYK: 'true' },
+  });
 
-    child.on('error', (error) => {
+  let stdout = '';
+  return new Promise<DriftctlExecutionResult>((resolve, reject) => {
+    if (input && p.stdin != null) {
+      p.stdin.write(input);
+      p.stdin.end();
+    }
```
wbeuil, line 314: In typescript you could have done that:

```
if (input) {
      p.stdin?.write(input);
      p.stdin?.end();
    }
p.stdout?.on('data', (data) => {
      stdout += data;
    });
```

No need to check explicitly for null when you can use the `?`

src/lib/iac/drift/types.ts
```diff
@@ -0,0 +1,65 @@
...
+  source?: DriftSource;
+};
+
+export type DriftSource = {
+  source: string;
+  namespace: string;
+  internal_name: string;
+};
+
+export type DriftResourceAttributes = Map<string, any>;
```
wbeuil, line 14: I think we could use the `unknown` type here instead of `any`.

https://mariusschulz.com/blog/the-unknown-type-in-typescript

src/lib/iac/drift/types.ts
```diff
@@ -0,0 +1,65 @@
...
+export type DriftAnalysisDifference = {
+  res: DriftResource;
+  changelog: DriftChange[];
+};
+
+export type DriftChange = {
+  type: string;
+  path: string[];
+  from: any;
+  to: any;
```
wbeuil, line 34: The same here (e.g. `unknown`)

test/fixtures/iac/drift/output/output.json
```diff
@@ -0,0 +1,75 @@
+{
```
wbeuil, line 1: Maybe we should have 2 output files here, one that will be drifted (e.g. the one we have here), and the other one that will be in sync.

wbeuil (changes requested):

src/lib/iac/drift.ts
```diff
@@ -226,7 +269,7 @@ export const parseDescribeFlags = (options: DriftCTLOptions): string[] => {
   return args;
 };
 
-export function translateExitCode(exitCode: number) {
+export function translateExitCode(exitCode: number | null) {
```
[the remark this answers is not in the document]
eliecharra: I don't think we want to handle interrupts signals with such granularity level for know. 
If for some reason the driftctl process is killed, exit code will be null here and we'll return `EXIT_CODES.ERROR`

test/fixtures/iac/drift/args-echo.js
```diff
@@ -0,0 +1,28 @@
+const fs = require("fs")
+
+if (process.env.DCTL_EXIT_CODE) {
```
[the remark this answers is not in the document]
eliecharra: We are not doing any special logic in case of a drifted result. 
I'm curious what could be the reason for us to test that case

test/fixtures/iac/drift/output/output.json
eliecharra: What case does it will cover ? As said above we are not doing anything special regardless if we get drift or not

test/fixtures/iac/drift/args-echo.js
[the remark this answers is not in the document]
wbeuil: I'm not saying we should test that case right now, indeed we don't do any special logic.

It's more about having tests that reflect somewhat the reality in addition to the unknown future where maybe we could right logic for drifted results. And here we could only change the script to `process.exit` only on exit code 2. 

Better now than later they say.

wbeuil (approved):
Amazing work Elie!

PR #2848, opened 2022-03-08
fix: Single files scans of non-tf files behind FF

ipapast:
This commit fixes the issue introduced on a previous PR where a single file scan of a non-Terraform file (.tf, .tfvars) would fail as a non-valid IaC file.

This is because we didn't ignore the error when the file has already successfully been parsed in the file take of the parsing.
We will now ignore the error and continue with the scan.

I also added regression tests that will check each one of the other supported fileTypes for a single file scan with the FF on.
They are duplicate but can be removed when we remove the FF.

Thread: https://snyk.slack.com/archives/C030WL9BLSW/p1646759529069249

image

image


teodora-sandu (approved):
LGTM

PR #3017, opened 2022-03-10
feat: add support for snyk policy

YairZ101:

#### What does this PR do?
This PR adds to the Iac CLI Share Results feature support for the `.snyk` policy file.

1. Run registry locally with [the following branch](https://github.com/snyk/registry/pull/26636).
2. Scan an IaC directory with a `.snyk` file inside that ignores a relevant issue using the command `snyk iac test --report`
3. Watch the issues getting ignored in the platform.

https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-1522

#### Screenshots
image

### DIFF
--- a/src/lib/iac/cli-share-results.ts
+++ b/src/lib/iac/cli-share-results.ts
@@ -4,11 +4,15 @@ import { getAuthHeader } from '../api-token';
 import { IacShareResultsFormat } from '../../cli/commands/test/iac-local-execution/types';
 import { convertIacResultToScanResult } from './envelope-formatters';
 import { AuthFailedError } from '../errors/authentication-failed-error';
+import { Policy } from '../policy/find-and-load-policy';
 
 export async function shareResults(
   results: IacShareResultsFormat[],
+  policy: Policy | undefined,
 ): Promise<Record<string, string>> {
-  const scanResults = results.map(convertIacResultToScanResult);
+  const scanResults = results.map((result) =>
+    convertIacResultToScanResult(result, policy),
+  );
 
   const { res, body } = await makeRequest({
     method: 'POST',
--- a/test/jest/unit/iac/cli-share-results.spec.ts
+++ b/test/jest/unit/iac/cli-share-results.spec.ts
@@ -1,18 +1,20 @@
 import { shareResults } from '../../../../src/lib/iac/cli-share-results';
 import {
   expectedEnvelopeFormatterResults,
-  generateScanResults,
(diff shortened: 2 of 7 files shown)


src/lib/iac/cli-share-results.ts
```diff
@@ -4,11 +4,15 @@ import { getAuthHeader } from '../api-token';
 import { IacShareResultsFormat } from '../../cli/commands/test/iac-local-execution/types';
 import { convertIacResultToScanResult } from './envelope-formatters';
 import { AuthFailedError } from '../errors/authentication-failed-error';
+import { Policy } from '../policy/find-and-load-policy';
 
 export async function shareResults(
   results: IacShareResultsFormat[],
+  policy: Policy | undefined,
```
ipapast, line 11: Alternatively we could make this parameter optional in this function and the `convertIacResultToScanResult` and avoid the undefined type here:

```suggestion
  policy: Policy
```

not much different though.

test/jest/unit/iac/cli-share-results.spec.ts
```diff
@@ -21,15 +21,15 @@ describe('CLI Share Results', () => {
   });
 
   beforeEach(async () => {
-    await shareResults(scanResults);
+    await shareResults(scanResults, undefined);
```
ipapast, line 24: should we add a case where we successfully convert the Policy as well?

ipapast (approved):
I tested a few scenarios in combination with Registry and it looks good.  🙌 

image

Also tested with `snyk-dev iac test file.tf --policy-path=.snyk --report` providing a different path for the policy-path , loads the policy and works as expected.

btw @YairZ101 I see that 2 security checks in the CI are "frozen" - you can close the PR and re-open it below and they will be re-triggered.

test/jest/unit/iac/cli-share-results.spec.ts
YairZ101: I don't have strong opinion about it but I've added a test the checks it cause it's better to have one than not.

src/lib/iac/cli-share-results.ts
YairZ101: My personal preference is the one I implemented cause the policy is not really optional, it just the fact that sometimes the file is not present so its undefined.

ipapast: and that is absolutely fine :)

PR #3018, opened 2022-03-10
[CFG-1658] feat: iac-describe driftignore support in policy

craigfurman:

#### What does this PR do?

`snyk iac describe` enumerates drift between terraform states and cloud
projects using the open-source driftctl tool. This tool can be
configured to ignore certain resources using a pseudo-glob syntax in a
".driftignore" file, as documented here:
https://docs.driftctl.com/next/usage/filtering/driftignore.

This commit adds `.snyk` policy file support for `iac describe` drift
ignores, but does not use the ignore section, instead using the
"exclude" section
(https://www.notion.so/Exclude-files-folders-1bd80d232ca34931a243509af5e2a4bd):

```
version: v1.22.1
exclude:
  iac-drift:
    - '*'
    - '!aws_s3_bucket'
```

Semantically, Snyk ignores discard discovered issues, whereas Snyk
excludes instruct Snyk to not inspect something. See
https://snyksec.atlassian.net/browse/CFG-1658?focusedCommentId=353203
for a justification on using excludes instead of ignores.

This implementation requires no changes to the policy library.

If this PR is approved, I'll document this in gitbook, along with the rest of the policy file docs.

https://snyksec.atlassian.net/browse/CFG-1658

### DIFF
--- a/test/jest/unit/lib/iac/drift.spec.ts
+++ b/test/jest/unit/lib/iac/drift.spec.ts
@@ -6,6 +6,7 @@ import {
   generateArgs,
   parseDriftAnalysisResults,
   translateExitCode,
+  driftignoreFromPolicy,
 } from '../../../../../src/lib/iac/drift';
 import envPaths from 'env-paths';
 import { EXIT_CODES } from '../../../../../src/cli/exit-codes';
@@ -18,6 +19,8 @@ import {
 } from '../../../../../src/lib/iac/types';
 import { addIacDriftAnalytics } from '../../../../../src/cli/commands/test/iac-local-execution/analytics';
 import * as analytics from '../../../../../src/lib/analytics';
+import * as snykPolicy from 'snyk-policy';
+import { Policy } from '../../../../../src/lib/policy/find-and-load-policy';
 
 const paths = envPaths('snyk');
 
@@ -64,6 +67,7 @@ describe('driftctl integration', () => {
       to: 'to',
       'only-managed': true,
       'only-unmanaged': true,
+      ignore: ['*', '!aws_s3_bucket'],
     } as DescribeOptions);
     expect(args).toEqual([
       'scan',
@@ -89,6 +93,8 @@ describe('driftctl integration', () => {
       'driftignore',
       '--tf-lockfile',
       'tflockfile',
+      '--ignore',
+      '*,!aws_s3_bucket',
       '--config-dir',
       'confdir',
       '--from',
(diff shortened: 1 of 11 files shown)


craigfurman:
@sjourdan it'd be good to get product feedback on the proposed interface here. No need to look at the code, just the PR description.

moadibfr:
I think it sounds a little hacky but why not. I like that this preserve everything we can do with a driftignore.
I might have prefer a multiline string for the driftignore like:
```
version: v1.22.1
ignore:
  SNYK-DRIFT: |
    *
    !aws_s3_bucket
```

craigfurman:
@moadibfr I know what you mean, I am flip-flopping on whether this proposal is fine and normal, or whether it is indeed hacky. On one hand, the current structure of the policy file's ignore section is a list of issues, and a list of instances ("paths") of that issue to ignore. If you squint, each resource glob is an instance of an issue to ignore, and so introducing a SNYK-DRIFT pseudo-issue begins to make sense.

I haven't really got any good alternative ideas, but they're certainly welcome!

Re: multi-line string instead of array, I'm not sure I understand why that might be clearer. Semantically, we're passing a list, aren't we?

karniwl:
This seems alright to me and makes sense.
@craigfurman can you explain what you mean by
> As long as "SNYK-DRIFT" does not ever become a real Snyk issue ID
There's no reason it should become a Snyk issue ID, is there? do you mean in the future, for how we might model the data or if ignore functionality somehow treats it as a real issue ID?

craigfurman:
@karniwl this PR proposes to use `SNYK-DRIFT` as a kind of "pseudo-ID". It requires no extensions to the policy file, but the policy library will see it as a real issue ID - just one that is never actually raised by `snyk [code|container|iac] test`.

It should be very unlikely to clash, as far as I can tell. Some examples of current Snyk issue IDsL SNYK-JS-FOLLOWREDIRECTS-2396346 (open source), SNYK-CC-TF-138 (IaC), SNYK-UBUNTU2004-KRB5-1533536 (container). The only clash possibility I can see here is if a programming language named "drift" is introduced, and we introduce a very strange syntax around ignoring all issues for a programming language "SNYK-LANG", rather than "SNYK-LANG*". This all sounds very unlikely, hence why I think it might be safe to do this.

License issues have a very different format. As far as I can tell from our own docs (https://docs.snyk.io/features/fixing-and-prioritizing-issues/issue-management/ignore-issues#ignoring-issues-in-snyk-code), Snyk code uses quite a different format too, although I'm unclear on whether it's even possible to use the .snyk file to ignore code issues locally.

karniwl:
@craigfurman thanks for the explanation, I agree we're safe on that front.

JackuB (approved):

test/jest/unit/lib/iac/drift.spec.ts
```diff
@@ -290,3 +296,37 @@ describe('drift analytics', () => {
...
   });
 });
+describe('driftignoreFromPolicy', () => {
+  const loadPolicy = async (name: string): Promise<Policy> => {
+    const policyPath = path.join(__dirname, 'fixtures', name);
+    const policyText = fs.readFileSync(policyPath, 'utf-8');
+    return await snykPolicy.loadFromText(policyText);
+  };
+
+  it.each([
```
moadibfr, line 306: nice I didn't know that this was something we could do in our ts test

moadibfr (approved):
that looks good to me 👏

PR #3022, opened 2022-03-11
fix: non-tf single files scan successfully in new flow

ipapast:
This commit fixes the issue where we scanned single non-tf files and they failed when using the new flow (behind FF). This is because we were throwing an (non-iac-file) error early in the flow, without checking if files have already been parsed or not.
This is a quick fix for the bug, refactoring to follow.

I also added regression tests that will check each one of the other supported fileTypes for a single file scan with the FF on.
They are duplicate but can be removed when we remove the FF.

This is an "almost" duplicate of https://github.com/snyk/cli/pull/2848#event-6212805181 which I had to revert due to failing tests on master.

Non-TF single file
image

TF single file
image

### DIFF
--- a/src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
+++ b/src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
@@ -39,7 +39,19 @@ export async function loadAndParseTerraformFiles(
       pathToScan,
       currentDirectory,
     );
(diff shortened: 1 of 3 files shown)


teodora-sandu (approved):
This will be refactored in a follow-up PR but for now approving so we unblock beta. Thanks for fixing this 🙇

PR #3029, opened 2022-03-15
SAST - handle 500 errors in case of API failure

metju90:


neonnoon:
@metju90 , why not open a draft then?

metju90:
> @metju90 , why not open a draft then?

I forgot the existence of drafts!

pkey:
not reviewing. You're welcome :D

pkey:
@metju90 what's up with this one? Can you move it to `draft`, if not relevant yet?

metju90:
@pkey thanks for reminding me about this.

I've converted this to a proper PR. Can you have a look?

pkey (approved):
Looks ok, I would add a test as well.

neonnoon:
@metju90 is this still needed?

metju90:
> @metju90 is this still needed?

@neonnoon  Yes, but it will take ~1hr to manually reconfirm this PR is still valid

metju90:
I revisited this PR and came to the conclusion that this is no longer valid. If [this](https://github.com/snyk/cli/blob/f6668bdfc428cc42bac5fb523dbc0defc2f04d5c/src/lib/plugins/sast/checks.ts#L15) API call had to throw 500, the initial call to the same server authenticate users most probably would throw 500 too, rendering the changes in this PR useless.

PR #3032, opened 2022-03-15
`snyk iac describe` show help by default and has to be called with right flags

moadibfr:

#### What does this PR do?
`snyk iac describe` show help by default and has to be called with right flags:
--all, --only-managed or --only-managed

Added a way to validate flags (both exclusive and required). It calls help command to get the help markdown rendered before throwing ab error that ask to add the right flag.

`snyk iac describe --all --only-unmanaged` should throw an error asking to only use one of the flags
`snyk iac describe` should throw an error asking to add one of the required flags

### DIFF
--- a/test/jest/unit/lib/iac/drift.spec.ts
+++ b/test/jest/unit/lib/iac/drift.spec.ts
@@ -3,10 +3,11 @@ import * as mockFs from 'mock-fs';
 import {
   DCTL_EXIT_CODES,
   driftctlVersion,
+  driftignoreFromPolicy,
   generateArgs,
   parseDriftAnalysisResults,
   translateExitCode,
-  driftignoreFromPolicy,
+  validateArgs,
 } from '../../../../../src/lib/iac/drift';
 import envPaths from 'env-paths';
 import { EXIT_CODES } from '../../../../../src/cli/exit-codes';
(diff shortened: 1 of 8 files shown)


moadibfr:
> Please change the ownership so open source is not required on these

@robcresswell could you be more specific ? I thought I already did that. The 2 new files I created are owned by IAC right ?

robcresswell:
🤔 weird that it still pinged me for a review

robcresswell (approved):

Avishagp (approved):

test/jest/unit/lib/iac/drift.spec.ts
```diff
@@ -127,6 +131,36 @@ describe('driftctl integration', () => {
...
+    try {
+      await validateArgs({
+        kind: 'describe',
+        all: true,
+        drift: true,
+      } as DriftCTLOptions);
+    } catch (e) {
+      err = e;
+    }
+    expect(err).toEqual(new DescribeExclusiveArgumentError());
```
eliecharra, line 161: You can use [toThrow](https://jestjs.io/docs/expect#tothrowerror) `expect(() => { someThrowingCode() }).toThrow(expectedError)`

I don't know if we can use it combined with a `.not` for testing that we do not throw

eliecharra (changes requested):

eliecharra (approved):

PR #3034, opened 2022-03-16
Allow to lookup for driftctl binary in `/bin` too

eliecharra:

#### What does this PR do?

That PR allow us to lookup for a driftctl binary in `/bin`.
This enable us to use `snyk/driftctl` as base image for our docker run.
You can see more context here : https://github.com/snyk/snyk-images/pull/26

Probably by reading the [README.md](https://github.com/snyk/snyk-images) of `snyk/snyk-images`

```$
$ touch /bin/driftctl && chmod +x /bin/driftctl
# Be sure that you do not have any cached version
$ rm /home/elie.charra/.cache/snyk-nodejs/driftctl_*
# Be sure that you do not have any overriden binary path
$ env | grep DRIFTCTL_PATH
# Then you can run it with debug to see that we fallback to /bin/driftctl
```

There is currently no way to easily E2E test that since we should be able to have a released version with that PR to test it in docker.

https://github.com/snyk/snyk-images/pull/26

https://snyksec.atlassian.net/browse/CFG-1621

#### Screenshots

image

### DIFF
--- a/src/lib/iac/drift.ts
+++ b/src/lib/iac/drift.ts
@@ -29,14 +29,15 @@ import { Policy } from '../policy/find-and-load-policy';
 const cachePath = config.CACHE_PATH ?? envPaths('snyk').cache;
 const debug = debugLib('drift');
 
-export const driftctlVersion = 'v0.23.0';
-
 export const DCTL_EXIT_CODES = {
   EXIT_IN_SYNC: 0,
   EXIT_NOT_IN_SYNC: 1,
   EXIT_ERROR: 2,
 };
 
+// ⚠ Keep in mind to also update driftctl version used to generate docker images
+// You can edit base image used for snyk final image here https://github.com/snyk/snyk-images/blob/master/alpine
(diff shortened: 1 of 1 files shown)


craigfurman (changes requested):
Isn't this what DRIFTCTL_PATH is for? We can set it to /bin/driftctl in the dockerfile, no need for changes here, right? (since we don't want to use PATH for driftctl)

eliecharra:
> Isn't this what DRIFTCTL_PATH is for? We can set it to /bin/driftctl in the dockerfile, no need for changes here, right? (since we don't want to use PATH for driftctl)

Yes but with the actual docker build pipeline, it will be more complicated to inject an environment variable in the image for that purpose. We are not really using PATH here.
If you think it's a blocker, I can try something on the docker build pipeline to explicitly set that variable.

craigfurman (approved):
Sure, if it's awkward to set up DRIFTCTL_PATH conditionally for tags then this is actually fine, tbh 🙂

PR #3038, opened 2022-03-17
feature: add excluded paths to ignore command

dekelund:
feat: add excluded paths to ignore command

Adds support to alter excluded files in the policy file:

- Extract ignoreVuln from ignore-function, and route CLI command to
  the correct implementation
- Add excludeFilePathPattern to be called by the ignore-implementation.

Background:

  The policy file supports two different groups of file:
  - "code", for files being ignored by Snyk Code
  - "global", for files being ignored by all products supporting the
    exclude field, for instance `snyk test --unmanaged`.

Usage:

  > snyk ignore --file-path='./deps/**.cpp'  # Defaults to the global group
  > snyk ignore --file-path='./deps/**.go' --file-path-group='code'
  > snyk ignore --file-path='./**/vendor/*.java' --file-path-group='global'

### DIFF
--- a/src/cli/commands/ignore.ts
+++ b/src/cli/commands/ignore.ts
@@ -34,74 +34,100 @@ export default function ignore(options): Promise<MethodResult> {
         return;
       }
 
-      if (!options.id) {
-        throw Error('idRequired');
-      }
-      options.expiry = new Date(options.expiry);
-      if (options.expiry.getTime() !== options.expiry.getTime()) {
-        debug('No/invalid expiry given, using the default 30 days');
-        options.expiry = new Date(Date.now() + 30 * 24 * 60 * 60 * 1000);
-      }
-      if (!options.reason) {
-        options.reason = 'None Given';
+      const isFilePathProvided = !!options['file-path'];
+
+      if (isFilePathProvided) {
+        return excludeFilePathPattern(options);
       }
 
-      const isPathProvided = !!options.path;
-      if (!isPathProvided) {
-        options.path = '*';
-      }
-
-      debug(
-        `changing policy: ignore "%s", for %s, reason: "%s", until: %o`,
-        options.id,
-        isPathProvided ? 'all paths' : `path: '${options.path}'`,
-        options.reason,
-        options.expiry,
-      );
-      return policy
-        .load(options['policy-path'])
-        .catch((error) => {
-          if (error.code === 'ENOENT') {
-            // file does not exist - create it
-            return policy.create();
-          }
-          throw Error('policyFile');
-        })
-        .then(async function ignoreIssue(pol) {
-          let ignoreRulePathDataIdx = -1;
-          const ignoreParams = {
-            reason: options.reason,
-            expires: options.expiry,
-            created: new Date(),
-          };
-
-          const ignoreRules: IgnoreRules = pol.ignore;
-
-          const issueIgnorePaths = ignoreRules[options.id] ?? [];
-
-          // Checking if the an ignore rule for this issue exists for the provided path.
-          ignoreRulePathDataIdx = issueIgnorePaths.findIndex(
-            (ignoreMetadata) => !!ignoreMetadata[options.path],
-          );
-
-          // If an ignore rule for this path doesn't exist, create one.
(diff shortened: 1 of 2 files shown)


src/cli/commands/ignore.ts
```diff
@@ -34,74 +34,108 @@ export default function ignore(options): Promise<MethodResult> {
...
+
+      ignoreRules[options.id] = issueIgnorePaths;
+
+      pol.ignore = ignoreRules;
+
+      return await policy.save(pol, options['policy-path']);
+    });
+}
+
+export function excludeFilePathPattern(options): Promise<MethodResult> {
```
dekelund, line 121: This function is the real code change, the rest is just a move of code.

src/cli/commands/ignore.ts
```diff
@@ -34,74 +34,108 @@ export default function ignore(options): Promise<MethodResult> {
...
+      const isFilePathProvided = !!options['file-path'];
+
+      if (isFilePathProvided) {
+        return excludeFilePathPattern(options);
       }
 
-      const isPathProvided = !!options.path;
-      if (!isPathProvided) {
-        options.path = '*';
+      return ignoreIssue(options);
```
dekelund, line 43: the rest of the function was moved to ignoreIssue

src/cli/commands/ignore.ts
```diff
@@ -34,74 +34,108 @@ export default function ignore(options): Promise<MethodResult> {
...
+        // file does not exist - create it
+        return policy.create();
+      }
+
+      throw Error('policyFile');
+    })
+    .then(async (pol) => {
+      pol.addExclude(pattern, group);
+      return policy.save(pol, options['policy-path']);
     });
```
JackuB, line 140: Wouldn't it be better to use async/await here, rather than this syntax? If you kept it only because rest of the file looks like it, I think for readability we should prefer async/await in all new code.

JackuB (commented):
Small style comment.

Is this change also reflected in the Help files on gitbook?

src/cli/commands/ignore.ts
dekelund: I would gladly change it. And yes I tried to be consistent with the rest of the implementation.
I will do the update and come back with a new version.

dekelund:
Regarding "Is this change also reflected in the Help files on gitbook?"
If you are referring to the documentation for the CLI, then yes it will be, but not yet.
I'm working on a seperate PR, that is going to enhance this functionality a bit, and my intention was to publish the documentation when we have added both PR's to the CLI.

dekelund:
Tests will be added soon

danlucian (approved):

dekelund:
Hi @JackuB 

Regarding this message:

> Please make changes to snyk help text in [Gitbook](https://docs.snyk.io/features/snyk-cli/commands). Changes will be automatically synchronised to Snyk CLI as a [scheduled PR](https://github.com/snyk/snyk/actions/workflows/sync-cli-help-to-user-docs.yml).
> For more information, see: [help/README.md](https://github.com/snyk/snyk/tree/master/help/README.md).

If I follow the gitbook-link, it contains a page not found-message.
The help-text has been manually updated in this PR, but I guess that's wrong.

test/jest/unit/snyk-ignore/snyk-ignore-test.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { excludeFilePathPattern } from '../../../../src/cli/commands/ignore';
+import * as policy from 'snyk-policy';
+import * as os from 'os';
+import * as path from 'path';
+const fs = require('fs');
+
+describe('Test snyk ignore, to exclude file-patterns', () => {
+  it('it should add unique file-pattern to snyk-policy', async () => {
```
jahed-snyk, line 8: This test leaks too much details and is reliant on side-effects so its scope goes beyond a unit test. Can we move the scenario to an acceptance test (if it isn't covered already)?

dekelund: I will remove the file since it's covered by the acceptance test

PR #3040, opened 2022-03-18
fix: TF Vars bug when scanning multi-file Kubernetes resource

teodora-sandu:

#### What does this PR do?
This PR fixes a bug in the CLI when `iacTerraformVarSupport` is enabled (by default for free users) and multi-file Kubernetes resources are scanned. 

1. `npm run build`
2. `mkdir empty`
3. `snyk-dev iac test empty` - still generates an empty error
4. `snyk-dev iac test ./test/fixtures/iac/terraform/empty_file.tf` - now generates an empty error, because that's what the old CLI does too and it makes sense to tell the customer if their TF file was empty
5. `snyk-dev iac test ./test/fixtures/iac/kubernetes/pod-privileged.yaml` - still generates vulnerabilities
6. `snyk-dev iac test ./test/fixtures/iac/kubernetes/pod-privileged-multi.yaml` - now generates vulnerabilities instead of an empty error, thus fixing the problem
7. `snyk-dev iac test ./test/fixtures/iac/` - still works for mixed folders, with non-TF and TF files

A fix was delivered in https://github.com/snyk/cli/pull/3022, but we agreed it was a temporary solution for the bug and pending a refactoring. This fix checked if a single file was provided and generated a parsed output in a previous step (for non-TF files). If so, then we wouldn't throw an error if the TF parsing generated no results. However, this solution didn't catch the edge case where a multi file Kubernetes resource was provided, in which case there was more than one parsed output returned for non-TF files.

https://snyk.slack.com/archives/C02JMMTLUF9/p1647607380819509

#### Screenshots
This is covered by the tests but just proof it actually works:
3. 
Screenshot 2022-03-18 at 14 36 55
.4
Screenshot 2022-03-18 at 14 37 43
5.
Screenshot 2022-03-18 at 14 37 04
6.
Screenshot 2022-03-18 at 14 37 14
7.
Screenshot 2022-03-18 at 14 39 20

### DIFF
--- a/src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
+++ b/src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
@@ -1,4 +1,3 @@
-import * as fs from 'fs';
 import { loadContentForFiles, NoFilesToScanError } from './file-loader';
 import * as path from 'path';
 import { parseTerraformFiles } from './file-parser';
@@ -15,23 +14,20 @@ import {
  * loading them, sending them to the parser and getting the parsing results back.
  * Then it concatenates and returns the new parsing results with the existing parsing results.
  * @param pathToScan - the path to scan provided by the user
- * @param parsedFiles - an array that holds all the existing parsedFiles
- * @param failedFiles - an array that holds all the existing failedFiles
  * @param maxDepth? - an optional maxDepth of directories if provided by the detection-depth flag
- * @returns { allParsedFiles, allFailedFiles} - all the parsing results so far
+ * @returns { parsedFiles, failedFiles} - all the parsing results so far
  */
 export async function loadAndParseTerraformFiles(
   pathToScan: string,
-  parsedFiles: IacFileParsed[],
-  failedFiles: IacFileParseFailure[],
   maxDepth?: number,
 ): Promise<{
-  allParsedFiles: Array<IacFileParsed>;
-  allFailedFiles: Array<IacFileParseFailure>;
+  parsedFiles: Array<IacFileParsed>;
+  failedFiles: Array<IacFileParseFailure>;
 }> {
-  let allParsedFiles: IacFileParsed[] = parsedFiles,
-    allFailedFiles: IacFileParseFailure[] = failedFiles;
+  let parsedFiles: IacFileParsed[] = [],
+    failedFiles: IacFileParseFailure[] = [];
   const allDirectories = getAllDirectoriesForPath(pathToScan, maxDepth);
+
   // we load and parse files directory by directory
   // this is because we need all files in the same directory to share the same variable context
   for (const currentDirectory of allDirectories) {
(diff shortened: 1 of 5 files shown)


src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
```diff
@@ -15,61 +15,41 @@ import {
  * loading them, sending them to the parser and getting the parsing results back.
  * Then it concatenates and returns the new parsing results with the existing parsing results.
  * @param pathToScan - the path to scan provided by the user
- * @param parsedFiles - an array that holds all the existing parsedFiles
```
teodora-sandu, line 18: The `parsedFiles` and `failedFiles` were passed in so we could check if non-TF files have been parsed before and so return successfully for single file scans. We don't need those now because we moved the error handling out of here.

src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
```diff
@@ -15,61 +15,41 @@ import {
...
-    )
-      continue;
-    if (
-      filePathsInDirectory.length === 1 &&
-      allParsedFiles.length === 0 &&
-      (!shouldBeParsed(filePathsInDirectory[0]) ||
-        isIgnoredFile(filePathsInDirectory[0]))
-    ) {
-      throw new NoFilesToScanError();
-    }
```
teodora-sandu, line 54: Instead of checking this here, we filter out single non-TF files in `getFilesForDirectory`

src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
```diff
@@ -15,61 +15,41 @@ import {
...
-      } = parseTerraformFiles(tfFilesToParse);
-      allParsedFiles = allParsedFiles.concat(parsedTfFiles);
-      allFailedFiles = allFailedFiles.concat(failedTfFiles);
-    } catch (err) {
-      if (allParsedFiles.length !== 0 && err instanceof NoFilesToScanError) {
-        // ignore this error since we might only have TF files in the folder and we have separated them,
-        // or if we have a single file scan of a non-TF file, and we have already parsed it
-      } else {
-        throw err;
-      }
```
teodora-sandu, line 69: This error handling is not required here because `NoFilesToScanError` is not thrown by anything in this flow

src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
```diff
@@ -15,61 +15,41 @@ import {
...
+      failedFiles: failedTfFiles,
+    } = parseTerraformFiles(tfFilesToParse);
+    parsedFiles = parsedFiles.concat(parsedTfFiles);
+    failedFiles = failedFiles.concat(failedTfFiles);
   }
-  return { allParsedFiles, allFailedFiles };
+
+  if (parsedFiles.length === 0) {
+    throw new NoFilesToScanError();
+  }
```
teodora-sandu, line 50: We throw here a `NoFilesToScanError` if no TF files were parsed, but then in `src/cli/commands/test/iac-local-execution/index.ts` we add a check to ignore it if there are non-TF files to scan

src/cli/commands/test/iac-local-execution/handle-terraform-files.ts
```diff
@@ -86,10 +66,6 @@ export function getAllDirectoriesForPath(
   if (isSingleFile(pathToScan)) {
     return [path.resolve(pathToScan)];
   }
-  // if the path we scanned itself is an empty directory, finish the scan here
-  if (fs.readdirSync(pathToScan).length === 0) {
-    throw new NoFilesToScanError();
-  }
```
teodora-sandu, line 92: No need to have this error handling here, it's caught now by the `parsedFiles.length === 0` check in `loadAndParseTerraformFiles`

src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -88,17 +88,23 @@ export async function test(
...
+          options.detectionDepth,
+        );
+        parsedFiles = parsedFiles.concat(tfParsedFiles);
+        failedFiles = failedFiles.concat(tfFailedFiles);
+      } catch (err) {
+        if (parsedFiles.length > 0 && err instanceof NoFilesToScanError) {
+          // ignore this error since we might not have .tf files in the folder and we have separated them
+        } else {
+          throw err;
+        }
```
aron, line 106: Teo and I discussed this in person, this caught me out initially as I missed that we're relying on `parsedFiles` here to be the length of the array set on line 75 rather than on 99. This logic perhaps indicates the need to refactor this flow in the near future to redesign the flow so there is a single entry point here to a function that rethinks the file discovery and parsing to encapsulate the need for multifile scans.

In the short term changing the comment to something like:
```suggestion
        if (parsedFiles.length > 0 && err instanceof NoFilesToScanError) {
          // ignore this error since we've discovered non-terraform files in the previous block
        } else {
          throw err;
        }
```

aron (approved):

PR #3047, opened 2022-03-22
chore: support sub folders in help command

Avishagp:
Support sub folders when running the help command.
Also solves issues like [this one](https://snyk.slack.com/archives/C0127HWU0E7/p1647938153732809)

### DIFF
--- a/.github/workflows/sync-cli-help-to-user-docs.yml
+++ b/.github/workflows/sync-cli-help-to-user-docs.yml
@@ -20,7 +20,7 @@ jobs:
           gh repo clone snyk/user-docs docs -- --depth=1 --quiet
           git -C ./cli checkout -b docs/automatic-gitbook-update
 
-          cp ./docs/docs/snyk-cli/commands/*.md ./cli/help/cli-commands/
+          cp -R ./docs/docs/snyk-cli/commands/ ./cli/help/cli-commands/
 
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-cli-help-to-user-docs.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           gh repo clone snyk/user-docs docs -- --depth=1 --quiet
           git -C ./cli checkout -b docs/automatic-gitbook-update
 
-          cp ./docs/docs/snyk-cli/commands/*.md ./cli/help/cli-commands/
+          cp ./docs/docs/snyk-cli/commands/ ./cli/help/cli-commands/
```
jahed-snyk, line 23: Does this need a `-R` for recursive?

jahed-snyk (commented):
Should we start adding some acceptances tests for `snyk help` and its various behaviours?

It's getting complicated and difficult to undestand. The implementation can likely be a lot simpler if we had some tests to safely refactor it.

.github/workflows/sync-cli-help-to-user-docs.yml
Avishagp: Believe that's a good idea, thank you!

PR #3052, opened 2022-03-24
chore: add npm version requirement

Avishagp:
Add minimal npm version needed to build CLI.


jahed-snyk (changes requested):
We can't do this as it'll break for customers using `npm install -g snyk` on older NPM versions.

PR #3056, opened 2022-03-24
fix: remove target.branch

YairZ101:

#### What does this PR do?
Currently, when running a scan from a directory that is a part of a git repo we will save inside the `target` filed the repo remote URL and branch.
We are using the remote URL to get information for the contributing developers but we currently don't have any use for the branch.
In addition to that, a side affect of sending the branch inside the `target` is that it gets auto-assigned as the `target-reference` which is not the desired outcome.
Due to all the reasons I mentioned above we've decided to remove the branch from the `target`.

1. Enable the `iacCliShareResults` FF.
2. Use the command `snyk iac test --report` to test your IaC files.
3. Check that the projects have been created successfully in the platform WITHOUT target-reference.

### DIFF
--- a/src/lib/iac/envelope-formatters.ts
+++ b/src/lib/iac/envelope-formatters.ts
@@ -26,7 +26,7 @@ export function convertIacResultToScanResult(
     target:
       Object.keys(gitTarget).length === 0
         ? { name: iacResult.projectName }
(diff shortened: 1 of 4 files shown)


aron (approved):
👏 Nice @YairZ101 that fixed the issue beautifully.

PR #3057, opened 2022-03-24
feat: display policy's exclude entries

danlucian:

#### What does this PR do?

- include the 'exclude' rules on the demunge implementation
- exclude might contain 2 categories/ids: `global` and `code`
- the entries for `exclude.global/code` might be simple paths (strings) or objects (including metadata like reason, creation/expiration dates).

The changes on the policy library were done here: https://github.com/snyk/policy/pull/70

#### Screenshots

Screenshot 2022-03-24 at 18 21 55

### DIFF
--- a/src/lib/display-policy.ts
+++ b/src/lib/display-policy.ts
@@ -4,6 +4,8 @@ import config from './config';
 
 export async function display(policy) {
   const p = demunge(policy, config.ROOT);
+  const delimiter = '\n\n------------------------\n';
+
   let res =
     chalk.bold(
       'Current Snyk policy, read from ' + policy.__filename + ' file',
@@ -13,9 +15,15 @@ export async function display(policy) {
 
   res += p.patch.map(displayRule('Patch vulnerability')).join('\n');
   if (p.patch.length && p.ignore.length) {
-    res += '\n\n------------------------\n';
+    res += delimiter;
   }
+
   res += p.ignore.map(displayRule('Ignore')).join('\n');
+  if (p.ignore.length && p.exclude.length) {
+    res += delimiter;
+  }
+
+  res += p.exclude.map(displayRule('Exclude')).join('\n');
 
   return Promise.resolve(res);
 }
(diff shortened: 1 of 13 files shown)


src/lib/display-policy.ts
```diff
@@ -15,17 +15,32 @@ export async function display(policy) {
...
   return Promise.resolve(res);
 }
 
 function displayRule(title) {
   return (rule, i) => {
     i += 1;
+
+    const formattedTitle =
+      title === 'Exclude'
+        ? chalk.bold('\n#' + i + ' ' + title) +
```
dekelund, line 35: would it be possible to make use of backticket as formatting string instead?

dekelund (approved):

src/lib/display-policy.ts
danlucian: Sure thing! Changed to template literals.

JackuB (approved):

danlucian:
> I can approve the change to the IaC fixture here, but wanted to make sure you're aware of the Snyk IaC functionality under the `exclude` field that was introduced here #3018. It doesn't look from the description like we've tested what the output of this looks like?
> 
> These excluded fields are not file paths but cloud resources, so the text may need to be adapted.

In order to support also the iac-drift entries I updated the message that we use for printing out the items to something like  `Exclude the following <entry_name> items/paths`.

Below is an example:

`.snyk` policy file used as input:

```
# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
# ignores vulnerabilities until expiry date; change duration by modifying expiry date
ignore:
exclude:
  global:
    - some/excluded/file/file-to-exclude.cpp:
        reason: False positive
        expires: 2022-04-17T12:57:47.569Z
        created: 2022-03-18T12:57:47.576Z
    - some/excluded/folder:
        reason: None Given
        expires: 2022-04-17T12:57:47.569Z
        created: 2022-03-18T12:57:47.576Z
  code:
    - some/path
    - some/file
  iac-drift:
     - '*'
     - '!aws_iam_*'
     - 'aws_s3_*'
     - 'aws_s3_bucket.*'
     - 'aws_s3_bucket.name*'
version: v1.22.2
patch: {}
```

the output:
Screenshot 2022-03-25 at 14 36 08


I also updated the tests to include the `iac-drift` entries.
Thanks a lot, @aron ! 🙏

PR #3026, opened 2022-03-13
feat: Add support for `--target-reference` for CLI Share Results [CFG-1521]

ofekatr:
- [X] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/CONTRIBUTING.md) rules

### What does this PR do?

- Adds support for the `--target-reference` flag for the CLI Share Results feature for Snyk IaC

### Where should the reviewer start?

- `src/cli/commands/test/iac-local-execution/index.ts` - Branching into all CLI Share Results related steps.

### How should this be manually tested?

1. Enable the FF named `iacCliShareResults`
2. In the Snyk CLI, test any IaC file/directory with the following commands:
```
snyk-dev iac test {path_to_the_file} --report --target-reference={target_reference}
```
3. The API call to Registry for publishing this project should include a top-level property, named `targetReference`, which includes the provided value to the `--target-reference` flag.

### Any background context you want to provide?

- The `--target-reference` flag provides a reference that differentiates this project, for example, a branch name, or version.
    Projects having the same reference can be grouped based on that reference. Only supported for
    Snyk Open Source. For more information see Separating projects by branch or version.

### Notes for the Reviewer

- For this ticket, the [`ScanResult` model](https://www.notion.so/snyk/The-payload-interface-f713ed7d071a461ca9d1ae3b19505f9c#59372aea0a6a4c56a4e90fa3fe1f24f4) was updated to include a new top-level optional property for the target reference. It was approved in the discussion conducted in [this thread](https://snyk.slack.com/archives/C01342FUD53/p1647173307120089) in [#project-envelope](https://snyk.slack.com/archives/C01342FUD53).
- There are no Help docs updates for the `snyk iac test` command in this PR, based on a discussion with the CLI Share Results feature lead, @YairZ101, who asserted that since this feature has yet to reach beta release, user docs will be added at a later stage closer to the release.

### What are the relevant tickets?

- [CFG-1521](https://snyksec.atlassian.net/browse/CFG-1521)

### Additional information
- [CLI Share Results feature page](https://www.notion.so/snyk/Share-test-results-from-IaC-CLI-6e8287fcc7534a4f8d628d1f8f4187ba)
- [Thread](https://snyk.slack.com/archives/C01342FUD53/p1647173307120089) in [#project-envelope](https://snyk.slack.com/archives/C01342FUD53) about the changes in the `ScanResult` model.
- [Relevant PR in snyk/registry](https://github.com/snyk/registry/pull/26664)

PR #3058, opened 2022-03-24
fix: Honor the --org flag when running snyk iac test --report

aron:

#### What does this PR do?

Adds support for the `--org` flag to the `snyk iac test --report` command. Previous to this change the `--report` flag will always send the share results to the same default org. Now it will respect the value.

Start in src/lib/iac/cli-share-results.ts this passes the `{ org: "<org>" }` value in the query string which is picked up by registry. I've also updated the tests. The mocks were not working before, they were set once in a beforeEach block and then reset after the first test. We also were still making network calls to the backend so the new `mockImplementation()` call also stubs that out.

    % snyk-dev iac test --report --org=<non-default-org> test/fixtures/iac/kubernetes


aron:
Now included in #3026

PR #3062, opened 2022-03-25
chore: Rename references to v3 api to rest api

jgresty:

#### What does this PR do?

The new Snyk rest api was originally released as the v3 api, however
since it contains its own versioning scheme it made no sense to have
another version in each endpoint. Therefore it was renamed to the more
generic "rest api" and the url moved from "/v3" to "/rest". The old url
will continue to work for a while during the transition phase but will
eventually go away.

This patch updates all references from the old v3 naming to the new rest
branding.

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -1,6 +1,6 @@
 {
   "API": "https://snyk.io/api/v1",
-  "API_V3_URL": "https://api.snyk.io/v3",
+  "API_REST_URL": "https://api.snyk.io/rest",
   "devDeps": false,
   "PRUNE_DEPS_THRESHOLD": 40000,
   "MAX_PATH_COUNT": 1500000,
--- a/src/cli/commands/apps/create-app.ts
+++ b/src/cli/commands/apps/create-app.ts
@@ -3,12 +3,12 @@ import {
   EAppsURL,
   getAppsURL,
   handleCreateAppRes,
-  handleV3Error,
+  handleRestError,
   ICreateAppRequest,
   ICreateAppResponse,
   SNYK_APP_DEBUG,
 } from '../../../lib/apps';
-import { makeRequestV3 } from '../../../lib/request/promise';
+import { makeRequestRest } from '../../../lib/request/promise';
 import { spinner } from '../../../lib/spinner';
 
 const debug = Debug(SNYK_APP_DEBUG);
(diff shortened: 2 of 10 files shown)


gitphill (approved):

config.default.json
```diff
@@ -1,6 +1,6 @@
 {
   "API": "https://snyk.io/api/v1",
-  "API_V3_URL": "https://api.snyk.io/v3",
+  "API_REST_URL": "https://api.snyk.io/rest",
```
jahed-snyk, line 3: Changing this will break any existing customers using SNYK_API_V3_URL env var. Is Moose okay with that? 🤔

love-bhardwaj: I would introduce a new variable along with the existing one if that is okay, to make is backwards compatible.

jgresty: Added a backward compatible shim that will use the old key name if set

love-bhardwaj (approved):

config.default.json
jahed-snyk: Do we have some tests for the backwards compat? To avoid regressions.

jgresty: Added

test/jest/unit/lib/config.spec.ts
```diff
@@ -0,0 +1,35 @@
+// Note that the config is loaded at import time, as such to test the config we
+// need to dynamically import the config with require after we perform setup.
+
+describe('API_REST_URL config', () => {
```
jahed-snyk, line 4: Environment variables are a public interface so we should test it through acceptance tests rather than unit tests. Can this be an acceptance test, similar to before using fake-server, but less thorough? That will make sure the env var gets picked up in a real scenario without being tied to specific implementation details and avoids messing with global state.

config.default.json
jahed-snyk: Thanks, left a another review for the tests.

test/jest/unit/lib/config.spec.ts
jgresty: Are you suggesting a test that will stub out a rest api then call the cli?

jahed-snyk: Pretty much. I'm suggesting taking a success scenario in `create-app.spec.ts` and passing it `env.SNYK_API_V3_URL` instead of `env.SNYK_API_REST_URL` to ensure it's backwards compatible.

jahed-snyk (approved):
As backwards compat mostly concerns `snyk apps` customers, I'll leave the rest to Moose and approve the PR to avoid blocking.

Are there any docs in GitBook relating to API v3?  Those will need updating too if so.

PR #3064, opened 2022-03-27
fix: point to the correct projects collection in the platform

YairZ101:
When a user runs a scan inside a git repo we pointed to the wrong projects collection, in this PR we fix this issue.

#### What does this PR do?
When invoking `snyk iac test --report` the command shows a message to the user telling him that the projects are available at his projects page under a dropdown menu with a certain name.
In cases where the user ran the scan inside a directory which is part of a git repo the dropdown menu name is incorrect.
This PR fixes this issue.

1. Run a test using the command `snyk iac test --report` inside of a directory which is part of git repo
2. The output should include the line: "Your test results are available at: http://localhost:8000/org/{org_name}/projects under the name {repo_name}"
3. Run a test using the command `snyk iac test --report` inside of a directory which is NOT part of git repo
4. The output should include the line: "Your test results are available at: http://localhost:8000/org/{org_name}/projects under the name {project_name}"

https://snyksec.atlassian.net/browse/CFG-1740

#### Screenshots
BEFORE
* part of a repo
image
* not part of a repo
image

AFTER
* part of a repo
image
* not part of a repo
image

### DIFF
--- a/src/cli/commands/test/iac-local-execution/types.ts
+++ b/src/cli/commands/test/iac-local-execution/types.ts
@@ -122,6 +122,7 @@ export interface TestMeta {
   ignoreSettings?: IgnoreSettings | null;
   projectId?: string;
   policy?: string;
+  gitRemoteUrl?: string;
 }
 
 export interface OpaWasmInstance {
@@ -399,4 +400,9 @@ export enum PerformanceAnalyticsKey {
   UsageTracking = 'usage-tracking-ms',
   CacheCleanup = 'cache-cleanup-ms',
   Total = 'total-iac-ms',
+}
+
+export interface ShareResultsOutput {
+  projectPublicIds: { [targetFile: string]: string };
+  gitRemoteUrl?: string;
 }
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -6,7 +6,12 @@ import chalk from 'chalk';
 import { MissingArgError } from '../../../lib/errors';
 
 import * as snyk from '../../../lib';
-import { IacFileInDirectory, Options, TestOptions } from '../../../lib/types';
+import {
+  IacFileInDirectory,
+  IacOutputMeta,
+  Options,
+  TestOptions,
+} from '../../../lib/types';
 import { MethodArgs } from '../../args';
 import { TestCommandResult } from '../../commands/types';
 import { LegacyVulnApiResult, TestResult } from '../../../lib/snyk-test/legacy';
@@ -17,7 +22,10 @@ import {
   summariseVulnerableResults,
 } from '../../../lib/formatters';
 import * as utils from './utils';
-import { getIacDisplayErrorFileOutput } from '../../../lib/formatters/iac-output';
+import {
+  getIacDisplayErrorFileOutput,
+  shareResultsOutput,
+} from '../../../lib/formatters/iac-output';
 import { getEcosystemForTest, testEcosystem } from '../../../lib/ecosystems';
 import { hasFixes, hasPatches, hasUpgrades } from '../../../lib/vuln-helpers';
 import { FailOn } from '../../../lib/snyk-test/common';
@@ -43,7 +51,6 @@ import {
   containsSpotlightVulnIds,
   notificationForSpotlightVulns,
 } from '../../../lib/spotlight-vuln-notification';
-import config from '../../../lib/config';
 import { isIacShareResultsOptions } from './iac-local-execution/assert-iac-options-flag';
 import { assertIaCOptionsFlags } from './iac-local-execution/assert-iac-options-flag';
 
@@ -95,6 +102,7 @@ export default async function test(
 
   // Holds an array of scanned file metadata for output.
   let iacScanFailures: IacFileInDirectory[] | undefined;
+  let iacOutputMeta: IacOutputMeta | undefined;
 
   // Promise waterfall to test all other paths sequentially
   for (const path of paths) {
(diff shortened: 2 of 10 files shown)


Avishagp (approved):

src/cli/commands/test/iac-local-execution/types.ts
```diff
@@ -400,3 +401,8 @@ export enum PerformanceAnalyticsKey {
   CacheCleanup = 'cache-cleanup-ms',
   Total = 'total-iac-ms',
 }
+
+export interface shareResultsOutput {
```
ofekatr, line 405: nit: Could you please capitalize the initial letter for this type?

src/lib/iac/cli-share-results.ts
```diff
@@ -30,7 +31,7 @@ export async function shareResults({
   tags?: Tag[];
   attributes?: ProjectAttributes;
   options?: IaCTestFlags;
-}): Promise<Record<string, string>> {
+}): Promise<shareResultsOutput> {
   const gitTarget = (await getInfo(false)) as GitTarget;
```
ofekatr, line 35: nit: Are there any benefits to fetching `gitTarget` in this function and then adding it as part of the return value, rather than providing it as an argument to this function?
What do you think about fetching it from `src/cli/commands/test/iac-local-execution/index.ts` before calling the `formatAndShareResults` function, passing it as an argument, and then re-use it as an argument for `formatScanResults`?

src/lib/formatters/iac-output.ts
```diff
@@ -317,3 +318,15 @@ function getPathRelativeToRepoRoot(
   const fullPath = pathLib.resolve(basePath, filePath).replace(/\\/g, '/');
   return fullPath.replace(repoRoot, '');
 }
+
+export function shareResultsOutput(iacOutputMeta: IacOutputMeta): string {
+  let projectName: string = iacOutputMeta.projectName;
+  if (iacOutputMeta?.gitRemoteUrl) {
+    // from "http://github.com/snyk/cli.git" to "snyk/cli"
+    projectName = iacOutputMeta.gitRemoteUrl.replace(
+      /^https?:\/\/github.com\/(.*)\.git$/,
```
ofekatr, line 327: Could you post here/in a code comment the origin of this regular expression, please?

src/lib/types.ts
```diff
@@ -266,3 +266,9 @@ export interface IacFileInDirectory {
   projectType?: IacProjectTypes;
   failureReason?: string;
 }
+
+export interface IacOutputMeta {
+  projectName: string;
+  org: string;
```
ofekatr, line 272: I know we use it in other places, but the `org` property name is a bit misleading, as it's impossible to tell by the type alone what the content of the `org` string represents, e.g. public ID, name, etc.
Could it be changed to something clearer, please?

src/cli/commands/test/index.ts
```diff
@@ -298,9 +309,8 @@ export default async function test(
 
     if (isIacShareResultsOptions(options)) {
       response +=
-        chalk.bold.white(
-          `Your test results are available at: ${config.ROOT}/org/${resultOptions[0].org}/projects under the name ${resultOptions[0].projectName}`,
-        ) + EOL;
+        chalk.bold.white(shareResultsOutput(iacOutputMeta as IacOutputMeta)) +
```
ofekatr, line 312: The type discrepancy seems to be due to the potential `undefined` value for `iacOutputMeta`. using the `as IacOutputMeta` could also suggest that the passed type is a subset of the properties in `IacOutputMeta`, which is not the case here.
Perhaps we could resolve it in a way more suitable for `undefined`?
Something like:
```suggestion
if (iacOutputMeta && isIacShareResultsOptions(options)) {
      response +=
        chalk.bold.white(shareResultsOutput(iacOutputMeta)) +
        EOL;
    }
```
Or alternatively:
```suggestion
        chalk.bold.white(shareResultsOutput(iacOutputMeta!)) +
```

src/lib/iac/cli-share-results.ts
YairZ101: I semi-agree with you, I don't like the idea of returning the `gitTargetUrl` from this function (because if feels very "forced") but I also don't like the idea of fetching it as part of the main flow (because there is no use for it in the regular testing flow, we need this info only for Contributing Developers in Share Results).
I think we can try and look at better options of where to place it during our CLI refactoring.

src/lib/formatters/iac-output.ts
YairZ101: I created this regex using GitHub Copilot so I don't have an origin for it.

src/cli/commands/test/index.ts
YairZ101: Done, took your 2nd suggestion.

src/lib/types.ts
YairZ101: Done, changed `org` => `orgName`

src/cli/commands/test/iac-local-execution/types.ts
YairZ101: Done

src/lib/types.ts
```diff
@@ -266,3 +266,9 @@ export interface IacFileInDirectory {
   projectType?: IacProjectTypes;
   failureReason?: string;
 }
+
+export interface IacOutputMeta {
```
JackuB, line 270: Any reason to have this change here instead of a iac-specific file?

YairZ101: I checked where another IaC interface (`IacFileInDirectory`) that is being used only at `src/cli/commands/test/iac-local-execution/index.ts` is located and I located my new type at the same place to have consistency.

gitphill (approved):

ofekatr (approved):
LGTM ✅ 🚀

PR #3069, opened 2022-03-28
chore: Added warning when custom rules with the `report` command [CFG-1737]

ofekatr:

#### What does this PR do?

- Adds a warning user message when custom rules are provided for the:
    - `snyk iac report` command.
    - `snyk iac test --report` command.

`src/cli/commands/test/iac-local-execution/rules.ts`

1. Prepare a bundle of custom rules, using the instructions found [here](https://docs.snyk.io/products/snyk-infrastructure-as-code/custom-rules).
2. If your custom rules bundle is local:
    2.1. Run the following in the Snyk CLI:
    ```
    snyk-dev iac report <path-to-iac-file-or-dir> --rules=<path-to-custom-rules-bundle>
    ```
    2.2. Ensure the following message was displayed at the top of the output:
    ```
    Using custom rules to generate misconfigurations.
    Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.
    ```
    2.3. Run the following in the Snyk CLI:
    ```
    snyk-dev iac test <path-to-iac-file-or-dir> --report --rules=<path-to-custom-rules-bundle>
    ```
    2.4. Ensure the following message was displayed at the top of the output:
    ```
    Using custom rules to generate misconfigurations.
    Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.
    ```
3. If your custom rules bundle is remote:
    3.1. Run the following in the Snyk CLI:
    ```
    snyk-dev iac report <path-to-iac-file-or-dir>
    ```
    3.2. Ensure the following message was displayed at the top of the output:
    ```
    Using custom rules to generate misconfigurations.
    Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.
    ```
    3.3. Run the following in the Snyk CLI:
    ```
    snyk-dev iac test <path-to-iac-file-or-dir> --report
    ```
    3.4. Ensure the following message was displayed at the top of the output:
    ```
    Using custom rules to generate misconfigurations.
    Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.
    ```

Today, the CLI Share Results feature does not support custom rules.
We would like to give a small warning to our users that will notify them that their custom rules will not be uploaded to the Snyk platform.

- [CFG-1737](https://snyksec.atlassian.net/browse/CFG-1737)

#### Screenshots

- For `snyk iac report`:
    image
- For `snyk iac test --report`:
    image

### DIFF
--- a/test/jest/acceptance/iac/custom-rules.spec.ts
+++ b/test/jest/acceptance/iac/custom-rules.spec.ts
@@ -67,6 +67,55 @@ describe('iac test --rules', () => {
       `Flag "--rules" is currently not supported for this org. To enable it, please contact snyk support.`,
     );
   });
+
+  describe.each([
+    ['--report flag', 'test --report'],
+    ['report command', 'report'],
+  ])('when used with the %s', (_, testedCommand) => {
+    it('should resolve successfully', async () => {
+      const { stderr, exitCode } = await run(
+        `snyk iac ${testedCommand} --rules=./iac/custom-rules/custom.tar.gz ./iac/terraform/sg_open_ssh.tf`,
+      );
+
+      expect(exitCode).toEqual(1);
+      expect(stderr).toEqual('');
+    });
+
+    it('should display a message informing of the application of custom rules', async () => {
+      const { stdout } = await run(
+        `snyk iac ${testedCommand} --rules=./iac/custom-rules/custom.tar.gz ./iac/terraform/sg_open_ssh.tf`,
+      );
+
+      expect(stdout).toContain(
+        'Using custom rules to generate misconfigurations.',
+      );
+    });
+
+    it('should display a warning message for custom rules not being available on the platform', async () => {
+      const { stdout } = await run(
+        `snyk iac ${testedCommand} --rules=./iac/custom-rules/custom.tar.gz ./iac/terraform/sg_open_ssh.tf`,
+      );
+
+      expect(stdout).toContain(
+        "Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.",
+      );
+    });
+
+    describe.each(['--json', '--sarif'])(
+      'when the %s flag is provided',
+      (testedFormatFlag) => {
+        it('should not display the warning message for the custom rules not being available on the platform', async () => {
+          const { stdout } = await run(
+            `snyk iac ${testedCommand} --rules=./iac/custom-rules/custom.tar.gz ./iac/terraform/sg_open_ssh.tf ${testedFormatFlag}`,
+          );
+
+          expect(stdout).not.toContain(
+            "Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.",
+          );
+        });
+      },
+    );
+  });
 });
 
 describe('custom rules pull from a remote OCI registry', () => {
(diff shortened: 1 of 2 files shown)


test/jest/acceptance/iac/custom-rules.spec.ts
```diff
@@ -67,6 +67,55 @@ describe('iac test --rules', () => {
       `Flag "--rules" is currently not supported for this org. To enable it, please contact snyk support.`,
     );
   });
+
+  describe.each([
+    ['--report flag', 'test --report'],
+    ['report comamnd', 'report'],
```
ipapast, line 73: nit; small typo
```suggestion
    ['report command', 'report'],
```

test/jest/acceptance/iac/custom-rules.spec.ts
```diff
@@ -67,6 +67,55 @@ describe('iac test --rules', () => {
...
+
+    it('should display a warning message for custom rules not being available on the platform', async () => {
+      const { stdout } = await run(
+        `snyk iac ${testedCommand} --rules=./iac/custom-rules/custom.tar.gz ./iac/terraform/sg_open_ssh.tf`,
+      );
+
+      expect(stdout).toContain(
+        "Please note that your custom rules will not be sent to the Snyk platform, and will not be available on the project's page.",
+      );
+    });
```
ipapast, line 102: out of curiosity, is there a benefit of having these two in different tests?
wouldn't the first test's `stdout` contain both strings tested in test (1) and test (2)?

ofekatr: Although both tests perform assertions over `stdout`, I considered these tests to examine 2 distinct parts of the expected behavior for this flow.

One approach would have been to omit the `Using custom rules...` message, as it is already tested in higher level tests here, and only test the `Please note that...` message.
However, I wanted the test to assure that both messages are intended to be displayed in this particular flow, so it felt less ideal here.

Another alternative would have been to include both parts of the `stdout` in the same test. However, in this implementation, the test would have multiple assertions for separate, distinct parts of `stdout`, which goes against our [best practices for testing](https://www.notion.so/snyk/Testing-Best-Practices-2da5494ffbdd4b8daf222d9f734d3bec#12b96ce59504427f98c16f9556d5f774).

test/jest/acceptance/iac/custom-rules.spec.ts
ofekatr: My bad, great catch!

ipapast (approved):
Tested locally with both local and remote bundles and it works 🚢

test/jest/acceptance/iac/custom-rules.spec.ts
ipapast: Cool, thanks!

>Another alternative would have been to include both parts of the stdout in the same test. However, in this implementation, the test would have multiple assertions for separate, distinct parts of stdout, which goes against our [best practices for testing](https://www.notion.so/snyk/Testing-Best-Practices-2da5494ffbdd4b8daf222d9f734d3bec#12b96ce59504427f98c16f9556d5f774).

btw, I think what you pointed in the notion page is different. it's talking about "testing multiple things" whereas the stdout is the same "thing".

ofekatr: Good point there, I think it depends on the context.
Generally, they suggest targeting a single behavior or condition.
On the one hand, it can be said that the printed output for this flow is one behavior as a whole.
On the other hand, it can be said that the printed output could be broken down and each of its components is expected behavior on its own.

It probably comes down to decision per use-case, and making tests more specific whenever it is more significant.

PR #3074, opened 2022-03-29
chore: replace inquirer with enquirer

Avishagp:
Replace `@snyk/inquirer` with `enquirer`.
`@snyk/inquirer` was heavily used in wizard command (removed in https://github.com/snyk/cli/pull/2799), and used in apps. enquirer is an easy replacement, offering a less dependencies.


robcresswell (approved):

jgresty (approved):
+1 for not using a vendored library

My only concern is that this is no longer readline based so it will probably provide a worse experience for some users who expect the shell to act a certain way. If you are fine with this UX then go ahead and make the change

JackuB (approved):

PR #3077, opened 2022-03-29
chore: Wrap new CLI output with `iacCliOutput` feature flag

ofekatr:

### What does this PR do?

- Adds versioning for the IaC output formatting logic.
- Creates a new version, labeled as `v2`, which is currently identical to the original logic, labeled as `v1`.
- Wraps the `v2` version with the newly created `iacCliOutput` feature flag.

### Where should the reviewer start?

   ```
   src/lib/formatters/iac-output/index.ts
   ```

### How should this be manually tested?

At the moment there's no output difference between the 2 versions.
To test it, you can utilize tools like logs or the debugger, to ensure the correct flows are used for each case.

- Without the `iacCliOutput` feature flag, we expect to use the original output formatters, which can be found in:
   ```
   src/lib/formatters/iac-output/v1/index.ts
   ```
- With the `iacCliOutput` feature flag, we expect to use the new output formatters, which can be found in:
   ```
   src/lib/formatters/iac-output/v2/index.ts
   ```

### Any background context you want to provide?
In a recent ticket ([CFG-1565](https://snyksec.atlassian.net/browse/CFG-1565)), we've introduced a new feature flag, named `iacCliOutput`, in order to wrap the implementation for the new CLI test output for Snyk IaC. 
To do so, a new separate flow for formatting the output needs to be created.
As a first step, this new flow will include the duplicated original logic for formatting IaC test results.
**The new flow will only be accessible to users with the `iacCliOutput` feature flag.**

### What are the relevant tickets?

- [CFG-1573](https://snyksec.atlassian.net/browse/CFG-1573)

### Additional information

- Feature page: [New CLI Output](https://www.notion.so/New-CLI-Output-e6ad525e5ab244769af8be5b14d740f7)
- Slack channel: [#feat-iac-cli-output](https://snyk.slack.com/archives/C030UA3U1RB)

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -53,6 +53,7 @@ import {
 } from '../../../lib/spotlight-vuln-notification';
 import { isIacShareResultsOptions } from './iac-local-execution/assert-iac-options-flag';
 import { assertIaCOptionsFlags } from './iac-local-execution/assert-iac-options-flag';
+import { hasFeatureFlag } from '../../../lib/feature-flags';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
@@ -228,11 +229,18 @@ export default async function test(
     throw err;
   }
 
+  const isNewIacOutputSupported = options.iac
+    ? await hasFeatureFlag('iacCliOutput', options)
+    : false;
+
   let response = results
     .map((result, i) => {
       return displayResult(
         results[i] as LegacyVulnApiResult,
-        resultOptions[i],
+        {
+          ...resultOptions[i],
+          isNewIacOutputSupported,
+        },
         result.foundProjectCount,
       );
     })
@@ -253,7 +261,9 @@ export default async function test(
     errorResultsLength = iacScanFailures.length || errorResults.length;
 
     for (const reason of iacScanFailures) {
-      response += chalk.bold.red(getIacDisplayErrorFileOutput(reason));
+      response += chalk.bold.red(
+        getIacDisplayErrorFileOutput(reason, isNewIacOutputSupported),
+      );
     }
   }
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -29,7 +29,7 @@ src/lib/plugins @snyk/snyk-open-source
 src/lib/plugins/sast/ @snyk/nebula
 test/fixtures/sast/ @snyk/nebula
 test/jest/unit/snyk-code/ @snyk/nebula
-src/lib/formatters/iac-output.ts @snyk/group-infrastructure-as-code
+src/lib/formatters/iac-output/ @snyk/group-infrastructure-as-code
 src/lib/iac/ @snyk/group-infrastructure-as-code
 src/lib/snyk-test/iac-test-result.ts @snyk/group-infrastructure-as-code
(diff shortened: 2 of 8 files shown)


src/cli/commands/test/index.ts
```diff
@@ -228,15 +228,18 @@ export default async function test(
...
-    .map((result, i) => {
-      return displayResult(
-        results[i] as LegacyVulnApiResult,
-        resultOptions[i],
-        result.foundProjectCount,
-      );
-    })
-    .join(`\n${SEPARATOR}`);
+  let response = (
+    await Promise.all(
```
ofekatr, line 232: The change in this file only entails updating the calls for the `displayResult` function to use `await`.

src/cli/commands/test/index.ts
```diff
@@ -228,15 +228,18 @@ export default async function test(
...
+      results.map(
+        async (result, i) =>
+          await displayResult(
```
francescomari, line 235: IIUC this function will eventually check if the feature flag is active and dispatch to a different function for actually displaying the output. If that's true, this means that N results will make N HTTP calls to check for the flag. Wouldn't be more efficient to check for the feature flag once, at the beginning of the function, and create a display function depending on the status of the flag?

src/cli/commands/test/index.ts
```diff
@@ -253,7 +256,9 @@ export default async function test(
     errorResultsLength = iacScanFailures.length || errorResults.length;
 
     for (const reason of iacScanFailures) {
-      response += chalk.bold.red(getIacDisplayErrorFileOutput(reason));
+      response += chalk.bold.red(
+        await getIacDisplayErrorFileOutput(reason, options),
```
francescomari, line 260: The same comment made above for `displayResult` applies here as well, where N is the scan failures.

francescomari (changes requested):

src/cli/commands/test/index.ts
francescomari: Incidentally, if my comments below are addressed, we might go back to have synchronous display functions, which will reduce the noise due to the switch to async functions.

src/cli/commands/test/index.ts
ofekatr: This is being discussed further in [this Slack thread](https://snyk.slack.com/archives/C0127HWU0E7/p1648640919993429).

src/cli/commands/test/index.ts
ofekatr: This is being discussed further in [this Slack thread](https://snyk.slack.com/archives/C0127HWU0E7/p1648640919993429).

src/cli/commands/test/index.ts
ofekatr: A new implementation with the `iacCliOutput` feature flag query performed earlier in the `test` function can be found in [the latest commit](https://github.com/snyk/cli/pull/3077/commits/ef842eeab517b9cceb39f72272ccc56f1ebc6184).

In this implementation, the REST call happens only once initially.
Please let me know what you think about this version.

src/cli/commands/test/index.ts
ofekatr: Discussion continues in [another comment](https://github.com/snyk/cli/pull/3077#discussion_r838628347).

Avishagp (approved):

anthogez (approved):

PR #3080, opened 2022-03-30
refactor: Move format and filter results to process-results

ipapast:
Co-authored-by: Ofek Atar <ofek.atar@snyk.io>
Co-authored-by: Francesco Mari <francesco.mari@snyk.io>
Co-authored-by: Teodora Sandu <teodora.sandu@snyk.io>

#### What does this PR do?

Small refactor to extract the “process results” step which mainly consists of extracting the severity threshold filtering from the formatResutls function.


src/cli/commands/test/iac-local-execution/index.ts
```diff
@@ -124,36 +122,19 @@ export async function test(
     }
 
     const scannedFiles = await scanFiles(parsedFiles);
+    // TODO: decide if this should go into scanFiles or stay here
```
YairZ101, line 125: Why not put it inside the `processResults` function?

src/cli/commands/test/iac-local-execution/process-results.ts
```diff
@@ -0,0 +1,43 @@
...
+  options: IaCTestFlags,
+): Promise<{
+  filteredIssues;
+  ignoreCount;
+}> => {
+  let projectPublicIds: Record<string, string> = {};
+  let gitRemoteUrl: string | undefined;
+
+  if (options.report) {
+    ({ projectPublicIds, gitRemoteUrl } = await formatAndShareResults({
```
YairZ101, line 24: Why have we moved the `formatAndShareResults` function to here?

src/cli/commands/test/iac-local-execution/process-results.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const formattedResults = formatScanResults(
+    resultsWithCustomSeverities,
+    options,
+    iacOrgSettings.meta,
+    projectPublicIds,
+    gitRemoteUrl,
+  );
+
+  return filterIgnoredIssues(policy, formattedResults);
```
YairZ101, line 42: Any particular reason that the `formatScanResults` is not the last step before returning the results?

YairZ101 (commented):
Left some comment below.
All of them are basically me trying to understand why we diverged from the [suggested flow](https://miro.com/app/board/uXjVOI3jBlw=/?moveToWidget=3458764520615383560&cot=10) (not saying we shouldn't but if we are going to diverge from it I would like to understand what was the reason behind it).

src/cli/commands/test/iac-local-execution/process-results.ts
teodora-sandu: I think we did this temporarily. We will modify `scanFiles` to return `ScanResults` and then send those directly to `/iac-cli-share-results` (or whatever the endpoints was called). We only had an hour and wanted to keep changes as simple as we can.

src/cli/commands/test/iac-local-execution/process-results.ts
teodora-sandu: Do you see `filterIgnoredIssues` being in the main flow? When we created the flow we included them in `processResults`

src/cli/commands/test/iac-local-execution/process-results/process-results.ts
```diff
@@ -0,0 +1,43 @@
...
+export const processResults = async (
+  resultsWithCustomSeverities: IacFileScanResult[],
+  orgPublicId: string,
+  iacOrgSettings: IacOrgSettings,
+  policy: Policy | undefined,
+  tags: Tag[] | undefined,
+  attributes: ProjectAttributes | undefined,
+  options: IaCTestFlags,
+): Promise<{
+  filteredIssues;
```
teodora-sandu, line 17: Should add typing

src/cli/commands/test/iac-local-execution/index.ts
teodora-sandu: I'm not sure. I remember trying it and then for some reason we started discussing whether it should be part of `scanFiles` or not. In the SCM import case for example (based on the Seldon work going on), these would be applied in the "scanning" phase and then the persistence of the rules (which would happen for example when the `--report` flag is used) doesn't need to care about them

So to me it made sense that this `scanFiles` function would eventually return the `ScanResults` with the actual severities

We could think about all of this though when we refactor `scanFiles` so I don't feel too strongly about it now

src/cli/commands/test/iac-local-execution/process-results.ts
YairZ101: > Do you see `filterIgnoredIssues` being in the main flow? 

Not at all, let me rephrase my question above.
In the suggested flow we've said that `formatScanResults` will be the last step inside `processResults`, currently the last step is `formatScanResults`.
Is it because we've decided to change the order of the functions inside `processResults`? didn't have time to work on it and we plan to do it in the next session?

src/cli/commands/test/iac-local-execution/index.ts
francescomari: @YairZ101 I think there is some tension between this comment and the following about `formatAndShareResults()`. On one hand, according to the suggested flow, `applyCustomSeverities()` should be part of `processResults()`. On the other hand, `formatAndShareResults()` works on the "scan results with custom severities" and not on the plain "scan results", like the diagram suggests.

In other words, the code and the suggested flow are not compatible. Either we move `applyCustomSeverities()` to the "Scan" step, or we move `formatAndShareResults()` to the "Process Results" step. I'm personally more inclined for the second option, because sharing the results sounds like a post-scanning activity to me. What are your preferences?

src/cli/commands/test/iac-local-execution/process-results.ts
francescomari: See [my comment](https://github.com/snyk/cli/pull/3080#discussion_r842495705) above.

src/cli/commands/test/iac-local-execution/process-results.ts
francescomari: We didn't have time to dig deeper in the implementation. As far as I can see, swapping the order of `formatScanResults()` and `filterIgnoredIssues()` might not be trivial. `filterIgnoredIssues()` has been written to accept an array of `FormattedResult` while, according to the suggested flow, should instead work on an array of `IacFileScanResult`. I would better leave this topic for a future session.

src/cli/commands/test/iac-local-execution/process-results/process-results.ts
francescomari: Done at cbb92e5f.

francescomari:
> Small question: do you all think we should be adding the share-results-formatters.spec.ts file in a process-results folder in test to match the src directory?

I'm not sure what's the policy here. If we had to follow the `src/` structure as closely as possible, shouldn't the directory be `test/jest/unit/cli/commands/test/iac-local-execution/process-results/`?.

ipapast:
> > Small question: do you all think we should be adding the share-results-formatters.spec.ts file in a process-results folder in test to match the src directory?
> 
> I'm not sure what's the policy here. If we had to follow the `src/` structure as closely as possible, shouldn't the directory be `test/jest/unit/cli/commands/test/iac-local-execution/process-results/`?.

true. we don't currently follow any policy I think. maybe a chat for another day

src/cli/commands/test/iac-local-execution/index.ts
teodora-sandu: Do we want to leave this TODO comment here? I added it but thought we would address it as part of this refactoring. If we think it's useful to leave because we'll fix it in a future PR then that's okay

src/cli/commands/test/iac-local-execution/process-results/share-results-formatter.ts
```diff
@@ -1,9 +1,9 @@
-import { computePaths } from './file-utils';
+import { computePaths } from '../file-utils';
```
teodora-sandu, line 1: Can we also move the test file `test/jest/unit/iac/share-results-formatters.spec.ts` to `test/jest/unit/iac/process-results/share-results-formatters.spec.ts`

src/cli/commands/test/iac-local-execution/process-results/process-results.ts
```diff
@@ -0,0 +1,48 @@
+import { filterIgnoredIssues } from '../policy';
+import { formatAndShareResults } from './share-results';
+import { formatScanResults } from '../measurable-methods';
```
teodora-sandu, line 3: Can we also move the import in `measure-methods` for `formatScanResults` from `./results-formatter` to `./process-results/results-formatter`. Or was there a reason why this wasn't includes in your commit @ipapast ?

If we end up moving it, could we also move the test file `test/jest/unit/iac/results-formatter.spec.ts` to `test/jest/unit/iac/process-results/results-formatter.spec.ts` and its fixture file

src/cli/commands/test/iac-local-execution/process-results/process-results.ts
```diff
@@ -0,0 +1,48 @@
+import { filterIgnoredIssues } from '../policy';
```
teodora-sandu, line 1: Does it make sense to rename this file to `index.ts` so it's clear that it's the entrypoint for this folder?

teodora-sandu: Also, why not move `src/cli/commands/test/iac-local-execution/policy.ts` to `src/cli/commands/test/iac-local-execution/process-reuslts/policy.ts` too since it's part of the "processResults` flow?

(its test file and fixture file too)

teodora-sandu (changes requested):
Thanks for driving this, I've left some comments but let me know what you think of them!

src/cli/commands/test/iac-local-execution/process-results/share-results-formatter.ts
teodora-sandu: I saw Ilianna also asked this. Although most people don't follow this rule, it's definitely strongly recommended and it will help us in the future when looking for tests

src/cli/commands/test/iac-local-execution/process-results/process-results.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Policy } from '../../../../../lib/policy/find-and-load-policy';
+import { ProjectAttributes, Tag } from '../../../../../lib/types';
+import {
+  FormattedResult,
+  IacFileScanResult,
+  IacOrgSettings,
+  IaCTestFlags,
+} from '../types';
+
+export async function processResults(
```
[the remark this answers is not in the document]
francescomari: I don't think it's worth it. All the functions used here seem to have their own unit tests, and the functionality as a whole has a lot of acceptance tests. We should be covered.

src/cli/commands/test/iac-local-execution/process-results/process-results.ts
francescomari: Fixed both in one of my latest commits.

src/cli/commands/test/iac-local-execution/index.ts
francescomari: I agree on leaving the comment there, for the moment. The refactoring for the proposed flow is not over yet, and this will leave some direction for the future.

src/cli/commands/test/iac-local-execution/process-results/share-results-formatter.ts
francescomari: Fix in one of my latest commits.

teodora-sandu (approved):
LGTM 👍

PR #3083, opened 2022-03-31
feat: log files over size limit [NEBULA-481]

patricia-v:
#### What does this PR do?
Adds logging in debug mode to CLI to warn when files are skipped due to their size and lists which ones were skipped.
This is the CLI side where the debugging was added.

[Link to code-client side](https://github.com/snyk/code-client/pull/141)

**TODO**:
- [x] update CLI to have the correct code-client version

[NEBULA-481](https://snyksec.atlassian.net/browse/NEBULA-481)

#### Screenshots
image

### DIFF
--- a/src/lib/plugins/sast/analysis.ts
+++ b/src/lib/plugins/sast/analysis.ts
@@ -1,4 +1,8 @@
-import { analyzeFolders, AnalysisSeverity } from '@snyk/code-client';
+import {
+  analyzeFolders,
+  AnalysisSeverity,
+  MAX_FILE_SIZE,
+} from '@snyk/code-client';
 import { ReportingDescriptor, Result } from 'sarif';
 import { SEVERITY } from '../../snyk-test/legacy';
 import { api } from '../../api-token';
@@ -13,6 +17,10 @@ import {
 } from './errors';
 import { getProxyForUrl } from 'proxy-from-env';
 import { bootstrap } from 'global-agent';
+import chalk from 'chalk';
+import * as debugLib from 'debug';
+
+const debug = debugLib('snyk-code');
 
 export async function getCodeAnalysisAndParseResults(
   root: string,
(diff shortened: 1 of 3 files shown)


src/lib/plugins/sast/analysis.ts
```diff
@@ -13,6 +13,11 @@ import {
 } from './errors';
 import { getProxyForUrl } from 'proxy-from-env';
 import { bootstrap } from 'global-agent';
+import { MAX_FILE_SIZE } from '@snyk/code-client/dist/constants';
```
metju90, line 16: Lets export the `MAX_FILE_SIZE` in the index file [here](https://github.com/snyk/code-client/blob/95d6601732b5640a6d0b6c1c25fefe83955fad97/src/index.ts#L1).

Benefits of this:

1. If `MAX_FILE_SIZE` is in the said index file, other developers might realise that this variable is used outside the code-client repo and will make them think twice before doing a breaking change (like renaming it)
2. Usually (but not always) we import things from 'library', lets keep the industry standards unless we have a reason why not to.

metju90 (changes requested):
Check my comment

src/lib/plugins/sast/analysis.ts
patricia-v: Since the index file is already exporting the constants [here](https://github.com/snyk/code-client/blob/95d6601732b5640a6d0b6c1c25fefe83955fad97/src/index.ts#L5), wouldn't it be enough to just add to line 1

> import { analyzeFolders, AnalysisSeverity, **_constants_** } from '@snyk/code-client';

and in line 103 add `${constants.MAX_FILE_SIZE}`? 🤔

pkey: @patricia-v in my opinion, that should do, although having variables explicitly exported is nicer.

metju90 (approved):

src/lib/plugins/sast/analysis.ts
```diff
@@ -91,6 +99,17 @@ async function getCodeAnalysis(
     languages: sastSettings.supportedLanguages,
   });
 
+  if (result?.fileBundle.skippedOversizedFiles?.length) {
+    debug(
+      '\n',
+      chalk.yellow(
+        `Warning!\nFiles were skipped in the analysis due to their size being greater than ${MAX_FILE_SIZE}B. Skipped files: ${[
+          ...result.fileBundle.skippedOversizedFiles,
+        ]}`,
```
JackuB, line 108: Wouldn't it be better to do a `.join(', ')` on the files array, to have a better and more predictable formatting?

JackuB (approved):

metju90 (approved):

PR #3095, opened 2022-04-05
refactor: Create a full command for 'snyk iac test'

francescomari:

#### What does this PR do?

The PR takes the current code for the test command, *duplicates* it, and keeps one copy only for running the logic behind `snyk iac test`. This PR also removes every reference to IaC from the original test command, except for a simple check that delegates to the IaC test command if `options.iac` is true.

The most important parts of the PR are the following.

- The code that previously lived in `src/cli/commands/test/iac-local-execution` has been moved to `src/cli/commands/test/iac/local-execution`.
- The module at `src/cli/commands/test/iac/index.ts` has a full command that takes care of the whole execution of `snyk iac test`.
- The only overlap with the original test command at `src/cli/commands/test/index.ts` is an `if` statement that will delegate to the command mentioned above if needed.

Everything else is just about adjusting imports and references to the old directory.

This PR doesn't add or remove any functionality. Everything should work as before.

In the recent past, the IaC team asked for many reviews to the Hammer and Snyk Open Source teams. The reason for it is too much overlap between the code behind the `snyk test` and `snyk iac test` commands. This PR aims at minimizing this overlap, which will generate less extra work for the Hammer and Snyk Open Source teams


jahed-snyk (approved):
Looks good from Hammer's POV. 🎉 

I'm not sure if duplicating the `testEcosystem` flow is the way Project Envelope wants to go here (vs. only extracting bits where IaC diverges). But I'm assuming whoever owns that flow will review that.

robcresswell (approved):

PR #3103, opened 2022-04-06
[CFG-1660] new ui for `snyk iac describe`

moadibfr:

#### What does this PR do?
This PR update the `snyk iac describe` output with new ui by the design team.
(https://miro.com/app/board/uXjVOVYR5zg=/?share_link_id=306344151216)
This is a first iteration for the ui.

Sorry about that but I made some refactoring to not have everything inside a single big typescript file, I hope this won't make the reviewing too painfull.
The pr introduce some change in drift.ts for the logic about driftctl first run output handling (mainly to use driftctl fmt only for html output).
Main addition will then be inside drift/output.ts.

https://snyksec.atlassian.net/browse/CFG-1660

#### Screenshots

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "env-paths": "^2.0.0",
     "glob": "^7.1.7",
     "global-agent": "^2.1.12",
+    "jsondiffpatch": "^0.4.1",
     "lodash.assign": "^4.2.0",
     "lodash.camelcase": "^4.3.0",
     "lodash.clonedeep": "^4.5.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "env-paths": "^2.0.0",
         "glob": "^7.1.7",
         "global-agent": "^2.1.12",
+        "jsondiffpatch": "^0.4.1",
         "lodash.assign": "^4.2.0",
         "lodash.camelcase": "^4.3.0",
         "lodash.clonedeep": "^4.5.0",
@@ -6368,6 +6369,11 @@
         "node": ">=0.3.1"
       }
     },
+    "node_modules/diff-match-patch": {
+      "version": "1.0.5",
+      "resolved": "https://registry.npmjs.org/diff-match-patch/-/diff-match-patch-1.0.5.tgz",
+      "integrity": "sha512-IayShXAgj/QMXgB0IWmKx+rOPuGMhqm5w6jvFxmVenXKIzRqTAAsbBPT3kWQeGANj3jGgvcvv4yK6SxqYmikgw=="
+    },
     "node_modules/diff-sequences": {
       "version": "27.4.0",
       "resolved": "https://registry.npmjs.org/diff-sequences/-/diff-sequences-27.4.0.tgz",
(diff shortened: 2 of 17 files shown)


output
```diff
@@ -0,0 +1,335 @@
+{
```
sundowndev-snyk, line 1: shouldn't this file named `output.json` ?

craigfurman (approved):
Full disclosure: this isn't the deepest review I've ever given. It's not a rubber stamp either - basically, all I've done is test this out, and then scan through the overall code structure and tests.

This was a bit tricky to review due to the combined refactor (a good refactor though), and functional changes. Ideally these would be in different commits. But I know 2 different people worked on it at different times, so no big deal.

Overall, LGTM! But perhaps get one more set of eyeballs on this, it's a large patch.

Non-blocking suggestion that's probably better for a follow-on: instead of having lib/iac/drift.ts and lib/iac/drift/ separately, we could have a single directory lib/iac/drift/, with a thin public interface in `index.ts`. I'm no TS expert, but AFAIK index.ts is a special case that can be imported by the name of its parent directory.

wbeuil:
> Non-blocking suggestion that's probably better for a follow-on: instead of having lib/iac/drift.ts and lib/iac/drift/ separately, we could have a single directory lib/iac/drift/, with a thin public interface in index.ts. I'm no TS expert, but AFAIK index.ts is a special case that can be imported by the name of its parent directory.

That's a pretty good observation here Craig, I didn't think of this one and I 100% agree with you.

package.json
```diff
@@ -84,6 +84,7 @@
     "env-paths": "^2.0.0",
     "glob": "^7.1.7",
     "global-agent": "^2.1.12",
+    "jsondiffpatch": "^0.4.1",
```
jahed-snyk, line 87: Looks like this package is not maintained (2 years). Is there anything more maintained?

https://snyk.io/advisor/npm-package/jsondiffpatch

microdiff seems like an option, no dependencies too, though it's not as popular.

https://snyk.io/advisor/npm-package/microdiff

wbeuil: It's not doing exactly what we want `microdiff`. First, it does not have a format function to create a JSON like string. We could create one for the result of microdiff but the problem is a little more complex. Imagine those 2 JSON files:

```json
{
  "Statement": [
    {
      "Action": [
        "sqs:ListQueues",
        "sqs:ListQueueTags",
        "sqs:GetQueueAttributes",
        "sns:ListTopics"
      ],
      "Effect": "Allow",
      "Resource": "*",
      "Sid": ""
    }
  ],
  "Version": "2012-10-17"
}
```

```json
{
  "Statement": [
    {
      "Action": [
        "sqs:ListQueueTags",
        "sqs:GetQueueAttributes",
        "sns:ListTopics"
      ],
      "Effect": "Allow",
      "Resource": "*",
      "Sid": ""
    }
  ],
  "Version": "2012-10-17"
}
```

The results are quite different:

- jsondiffpatch understood that I removed one key (1 change)
- microdiff understood that I change each key by the one below.. (4 changes)

So i'm not convinced by it's usefulness here even though it's still maintained.

moadibfr: json-diff was better but it has a dependency with a strange licence that prevent us from using it. this was the best fit I found...

jahed-snyk (approved):
Approving as there's not much choice around jsondiffpatch. Please be aware though that if jsondiffpatch or its dependencies have vulns, IaC will need to find an alternative or fork the project to resolve it as an incident. It might be worth IaC forking it now, publishing a fork, and using it here to avoid any potential malicious releases in the future.

PR #3108, opened 2022-04-07
update smoke test for iac help

moadibfr:

#### What does this PR do?

Update smoke test to match the new main iac help page.


jahed-snyk (approved):
Not so sure about these tests. Seems very brittle. But that's something we'll need to fix outside of this PR so approved.

jahed-snyk (changes requested):
Blocking as we need to fix two tests and also run smoke tests suite.

PR #3119, opened 2022-04-10
feat: remove-deeproxy-url-configuration

shirlupo:
[NEBULA-443](https://snyksec.atlassian.net/browse/NEBULA-443)

We wish that the customer of ST will not need to add this configuration and only configure 'API' value.

Please refer to Jira ticket for more information.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -39,12 +39,14 @@ test/smoke/spec/iac/ @snyk/group-infrastructure-as-code
 test/smoke/spec/snyk_code_spec.sh @snyk/zenith
 test/smoke/spec/snyk_basic_spec.sh @snyk/hammer
 test/smoke/.iac-data/ @snyk/group-infrastructure-as-code
+test/jest/unit/lib/endpoint-config-test.spec.ts @snyk/nebula
 test/jest/unit/lib/formatters/iac-output/ @snyk/group-infrastructure-as-code
 test/jest/unit/lib/formatters/test/format-test-results.spec.ts @snyk/hammer @snyk/snyk-open-source @snyk/magma
 test/jest/unit/iac/ @snyk/group-infrastructure-as-code
 test/jest/unit/lib/iac/ @snyk/group-infrastructure-as-code
 test/jest/acceptance/iac/ @snyk/group-infrastructure-as-code
 test/jest/acceptance/snyk-apps @snyk/moose
+src/lib/code-config.ts @snyk/nebula
 src/lib/errors/describe-required-argument-error.ts @snyk/group-infrastructure-as-code
 src/lib/errors/describe-exclusive-argument-error.ts @snyk/group-infrastructure-as-code
 src/lib/errors/no-supported-sast-files-found.ts @snyk/zenith
--- a/config.default.json
+++ b/config.default.json
@@ -4,7 +4,6 @@
   "devDeps": false,
   "PRUNE_DEPS_THRESHOLD": 40000,
   "MAX_PATH_COUNT": 1500000,
-  "CODE_CLIENT_PROXY_URL": "https://deeproxy.snyk.io",
   "NPM_TREE_SIZE_LIMIT": 6.0e6,
   "YARN_TREE_SIZE_LIMIT": 6.0e6
 }
--- a/test/jest/unit/lib/endpoint-config-test.spec.ts
+++ b/test/jest/unit/lib/endpoint-config-test.spec.ts
@@ -1,0 +1,29 @@
+import * as codeConfig from '../../../../src/lib/code-config';
+import config from '../../../../src/lib/config';
+
+describe('Testing deeproxy URL', () => {
+  const configApiDefault = config.API;
+  const codeClientUrlDefault = config.CODE_CLIENT_PROXY_URL;
+
+  afterEach(() => {
+    config.API = configApiDefault;
+    config.CODE_CLIENT_PROXY_URL = codeClientUrlDefault;
+  });
+
+  test('uses api URL to determine deeproxy URL when not provided without app. prefix', () => {
+    config.API = 'https://snyk.io/api/v1';
+    expect(codeConfig.getCodeClientProxyUrl()).toBe('https://deeproxy.snyk.io');
+  });
+
+  test('uses api URL to determine deeproxy URL when provided with app. prefix', () => {
+    config.API = 'https://app.snyk.io/api/v1';
+    expect(codeConfig.getCodeClientProxyUrl()).toBe('https://deeproxy.snyk.io');
+  });
+
+  test('uses a custom deeproxy endpoint when provided by SNYK_CODE_CLIENT_PROXY_URL environment', () => {
+    config.CODE_CLIENT_PROXY_URL = 'https://deeproxy.custom.url.snyk.io';
+    expect(codeConfig.getCodeClientProxyUrl()).toBe(
+      'https://deeproxy.custom.url.snyk.io',
+    );
+  });
+});
--- a/src/lib/code-config.ts
+++ b/src/lib/code-config.ts
@@ -1,0 +1,11 @@
+import config from './config';
+
+export function getCodeClientProxyUrl() {
+  return (
+    config.CODE_CLIENT_PROXY_URL ||
+    config.API.replace(/snyk\.io.*/, 'snyk.io').replace(
+      /\/\/(app\.)?/,
+      '//deeproxy.',
+    )
+  );
+}
--- a/src/lib/plugins/sast/analysis.ts
+++ b/src/lib/plugins/sast/analysis.ts
@@ -19,6 +19,7 @@ import { getProxyForUrl } from 'proxy-from-env';
 import { bootstrap } from 'global-agent';
 import chalk from 'chalk';
 import * as debugLib from 'debug';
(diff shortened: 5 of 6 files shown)


src/lib/config.ts
```diff
@@ -45,6 +45,17 @@ if (endpoint && endpoint !== config.API) {
   config.API = endpoint;
 }
 
+// getting deeproxy url if is not specified
+if (
+  !config.CODE_CLIENT_PROXY_URL ||
+  config.CODE_CLIENT_PROXY_URL.length === 0
+) {
+  config.CODE_CLIENT_PROXY_URL = config.API.replace(
+    /snyk.io.*/,
```
snyk-nebula-test, line 54: Regex feedback:

`.` is a special character in regex. Meaning if you don't escape it with `\.`, the regex will do a special operation with the `.` rather than considering it part of a string.

So `snyk.io` needs to be replaced with `snyk\.io`

neonnoon: Welcome to the team @snyk-nebula-test

src/lib/config.ts
```diff
@@ -45,6 +45,17 @@ if (endpoint && endpoint !== config.API) {
...
+    'snyk.io',
+  ).replace(/\/\/(app.)?/, '//deeproxy.');
+}
+
```
metju90, line 58: What did you is cool, and perhaps it's the cleanest. Deeproxy will be deprecated at some point. So with that in mind, what we can do:

A.1) leave your code as and if people don't understand how this works they can always Slack us
A.1) leave your code as is and document it the default behavior and how it can be overwritten

B) Leave the `config.default.json` unchanged and edit your logic to only change `config.CODE_CLIENT_PROXY_UR` when `config.APP` contains `https://app`. 

With B, when people look at the `config.default.json` they will see the default value of `CODE_CLIENT_PROXY_URL. This improves the DX (developer experience)

src/lib/config.ts
```diff
@@ -45,6 +45,17 @@ if (endpoint && endpoint !== config.API) {
   config.API = endpoint;
 }
 
+// getting deeproxy url if is not specified
+if (
+  !config.CODE_CLIENT_PROXY_URL ||
+  config.CODE_CLIENT_PROXY_URL.length === 0
```
metju90, line 51: nitpick:  We dont need the second condition. if `config.CODE_CLIENT_PROXY_URL` is  `""`, the first condition is sufficient.

src/lib/config.ts
shirlupo: @metju90 

one more possibility is to leave this variable as a comment in the config file
 OR remove it and add it to the README.

src/lib/config.ts
shirlupo: @snyk-nebula-test  - This is fixed, thanks

src/lib/config.ts
ronenl3: I'm with adding it to the README, and remove it from the config

pkey:
@shirlupo provided this is a bit tricky logic, I suggest to write a few unit tests for this.

src/lib/config.ts
pkey: It is better to be explicit because developer would expect that the actual value is the one that is in `config.default.json` or `config.local.json`. If we then fiddle with it somewhere, it will cause confusion. So removing the value and then providing instructions how to set the development environment up, I think, is the way to go here.

From customer's POV I think none of this matters, everything is an improvement.

test/tap/endpoint-config.test.ts
```diff
@@ -117,3 +117,37 @@ test('throws an error when endpoint option is not a valid URL', (t) => {
   );
   t.end();
 });
+
+test('uses api URL to determine deeproxy URL when not provided', (t) => {
```
pkey, line 121: @shirlupo nice! I would also suggest adding a test that checks whether the replacement logic (`app` -> `deeproxy`) works well.

config.default.json
```diff
@@ -4,7 +4,6 @@
   "devDeps": false,
   "PRUNE_DEPS_THRESHOLD": 40000,
   "MAX_PATH_COUNT": 1500000,
-  "CODE_CLIENT_PROXY_URL": "https://deeproxy.snyk.io",
```
pkey, line 7: question: so if this one is removed and I am developing locally, how do I know that I will have to provide this value? Definitely makes sense to update README with this information, IMO.

test/tap/endpoint-config.test.ts
metju90: Awesome 🎉 

I would have 2 tests related to using API URL to determine deeproxy URL.

1. 
```
test('uses api URL to determine deeproxy URL when not provided, when API does not have "app." prefix', ...)
```

2.
```
test('uses api URL to determine deeproxy URL when not provided, when API contains "app." prefix', ...)
```

With these tests you'd be testing your regex for both cases -- when API=http://snyk.io and when API=http://app.snyk.io

What do you think?

shirlupo: @metju90 @pkey  pushed more tests

test/tap/endpoint-config.test.ts
```diff
@@ -117,3 +117,58 @@ test('throws an error when endpoint option is not a valid URL', (t) => {
...
+        },
+      },
+    },
+  });
+  t.equal(config.CODE_CLIENT_PROXY_URL, deeproxyDefaultEndpoint);
+  t.end();
+});
+
+test('uses api URL to determine deeproxy URL when provided with app. prefix', (t) => {
+  console.log(
```
pkey, line 138: :p

test/tap/endpoint-config.test.ts
```diff
@@ -117,3 +117,58 @@ test('throws an error when endpoint option is not a valid URL', (t) => {
   );
   t.end();
 });
+
+test('uses api URL to determine deeproxy URL when not provided without app. prefix', (t) => {
+  const deeproxyDefaultEndpoint = 'https://deeproxy.snyk.io';
+  const { default: config } = proxyquire('../../src/lib/config', {
```
metju90, line 123: Can you make  process.env.SNYK_API = 'https://snyk.io'; here? So we are 100% that we are testing `SNYK_API` without `app.` prefix

shirlupo: @metju90 at the beginning of the file it is defined as default, so no need in this.

metju90 (approved):

test/tap/endpoint-config.test.ts
```diff
@@ -117,3 +117,58 @@ test('throws an error when endpoint option is not a valid URL', (t) => {
...
+      },
+    },
+  });
+  t.equal(config.CODE_CLIENT_PROXY_URL, deeproxyDefaultEndpoint);
+  t.end();
+});
+
+test('uses api URL to determine deeproxy URL when provided with app. prefix', (t) => {
+  console.log(
+    'hey!! PROOXY API PROCESS ENV' + process.env.SNYK_CODE_CLIENT_PROXY_URL,
```
metju90, line 139: hey!! @shirlupo , don't forget this 😝

admons (approved):

test/jest/unit/lib/endpoint-config-test.spec.ts
```diff
@@ -0,0 +1,37 @@
+describe('Testing deeproxy URL', () => {
+  const OLD_ENV = process.env;
+
+  beforeEach(() => {
+    jest.resetModules();
+    process.env = { ...OLD_ENV };
```
jahed-snyk, line 6: This will affect other tests as process.env is a shared global.

The logic around deciding what `CODE_CLIENT_PROXY_URL` should be can be a function in its own module and unit tested without using global state. For example, we can have a function:

```js
function getCodeClientProxyUrl(config = configModule) {
  return config.CODE_CLIENT_PROXY_URL || config.API.replace(
      /snyk\.io.*/,
      'snyk.io',
    ).replace(/\/\/(app\.)?/, '//deeproxy.')
}
```

And in tests, we can pass in custom `config` values to test against instead of modifying global state.

So now places like `analysis.ts` should use `getCodeClientProxyUrl()` instead of the raw `config.CODE_CLIENT_PROXY_URL`

jahed-snyk (approved):
Approving to avoid blocking. Please do refactor out the global state mutation in a follow-up PR. 🙏 

The main selling point for the refactor is so your team can take ownership of code-specific config modules and can avoid needing to ask Hammer for reviews.

.github/CODEOWNERS
```diff
@@ -39,6 +39,7 @@ test/smoke/spec/iac/ @snyk/group-infrastructure-as-code
 test/smoke/spec/snyk_code_spec.sh @snyk/zenith
 test/smoke/spec/snyk_basic_spec.sh @snyk/hammer
 test/smoke/.iac-data/ @snyk/group-infrastructure-as-code
+test/jest/unit/lib/endpoint-config-test.spec.ts @snyk/hammer
```
jahed-snyk, line 42: Should this be assigned to Nebula?

You can also add `src/lib/code-config.ts` to this list with Nebula as codeowners.

jahed-snyk (approved):

robcresswell (approved):

patricia-v (approved):

magdziarek (approved):

PR #3122, opened 2022-04-12
test: Updated FF usage in TF vars deref test

ofekatr:

#### What does this PR do?

- Updates the way the `iacTerraformVarSupport` FF is used in the `test/jest/acceptance/iac/test-terraform-var-deref.spec.ts` acceptance tests to match the common practices applied in recent IaC acceptance tests, by:
    - Removing the special FF configuration in the fake server that sets the value for all feature flags to be true when the org name is `tf-lang-support`.
    - Using the `setFeatureFlag` utility function for controlling feature flag values.

- `test/jest/acceptance/iac/test-terraform-var-deref.spec.ts`

- Run the test file with Jest.

### DIFF
--- a/test/acceptance/fake-server.ts
+++ b/test/acceptance/fake-server.ts
@@ -337,13 +337,6 @@ export const fakeServer = (basePath: string, snykToken: string): FakeServer => {
       return;
     }
 
-    if (org === 'tf-lang-support') {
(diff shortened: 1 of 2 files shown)


ipapast (approved):
nice one 🥇

PR #3127, opened 2022-04-12
chore: add CLI_V2 to TrackedIntegration enum

PeterSchafer:
Signed-off-by: Peter Schäfer <101886095+PeterSchafer@users.noreply.github.com>

#### What does this PR do?
Introduce additional tracked integration name

Run `SNYK_INTEGRATION_NAME=CLI_V2 snyk --debug` and see the IntegrationName in the Analytics output


jahed-snyk (approved):
thought(non-blocking): I'm wondering for the longer term if we want to call it something more generic like `CLI`. That fact that it's v2 is in the INTEGRATION_VERSION and if we have a v3 or later, we'll have to create a new INTEGRATION_NAME for it to make sense.

PR #3132, opened 2022-04-13
feat: improve ci detection

ghost:
ci-info seems to the be the most popular library to use for CI detection. We can rely on it to detect both current and future CI services.

https://www.npmjs.com/package/ci-info

https://snyk.io/advisor/npm-package/ci-info


darscan:
Closing as stale

PR #3135, opened 2022-04-13
Feat: Implement IaC output failures list section [CFG-1572]

ofekatr:

#### What does this PR do?

- Implements the new failures list section for the Snyk IaC test output.
- Applies the new failures list section in the test flow, wrapped in the `iacCliOutput` feature flag.

#### Notes for the reviewer

- The relatively high number of files is mainly due to the added fixture files for the tests.
- As in the original output, for flows in which the output only consists of a single error message, the case is the same for this implementation of the new output as well as of now. Other output design options are currently being explored for this single error message output ([see here](https://snyk.slack.com/archives/C030UA3U1RB/p1649271803177019))

- `src/cli/commands/test/iac/index.ts`

- Execute the `snyk iac test` command without the `iacCliOutput` feature flag, and ensure the output is the same as the original.
- Execute the `snyk iac test` command with the `iacCliOutput` feature flag, ensure the test failures are displayed in the new section, and that it matches [the design](https://www.figma.com/file/buNFZqr1Uhf4SaifHw4WXE/CLI-IaC-Output-Improvements?node-id=0%3A1).

- As part of the new design for the Snyk IaC test output , we’d like to replace the current per-project failure reasons output, with an aggregated failures summary at the end of the test output. This can be viewed in [the following design](https://www.figma.com/file/buNFZqr1Uhf4SaifHw4WXE/CLI-IaC-Output-Improvements?node-id=0%3A1).

- [CFG-1572](https://snyksec.atlassian.net/browse/CFG-1572)

#### Screenshots

- Before
    image

- After
    image


src/lib/formatters/iac-output/v2/failures-list.ts
```diff
@@ -0,0 +1,64 @@
+import chalk from 'chalk';
+import { EOL } from 'os';
+
+import { IacFileInDirectory } from '../../../types';
+import { colors } from './color-utils';
+
+export function formatIacTestFailures(testFailures: IacFileInDirectory[]) {
```
YairZ101, line 7: Do you think that `formatOutput...` will make more sense for those kind of functions?
We had the chance to talk about this topic [before](https://github.com/snyk/cli/pull/3101#discussion_r846753103) and both of us agreed that `get..` sounds to general but I think `format...` might be confusing cause we are using this prefix in other use cases.

test/jest/unit/lib/formatters/iac-output/failures-list.spec.ts
```diff
@@ -0,0 +1,73 @@
+import chalk from 'chalk';
+import * as fs from 'fs';
+import * as pathLib from 'path';
+
+import { formatIacTestFailures } from '../../../../../../src/lib/formatters/iac-output';
+import { colors } from '../../../../../../src/lib/formatters/iac-output/v2/color-utils';
+import { IacFileInDirectory } from '../../../../../../src/lib/types';
+
+const testFialureFixtures: IacFileInDirectory[] = JSON.parse(
```
YairZ101, line 9: typo - `Fialure` => `Failure`

test/jest/acceptance/iac/iac-output.spec.ts
```diff
@@ -47,138 +49,273 @@ describe('iac test output', () => {
...
+      // Assert
+      expect(stdout).not.toContain('Invalid Files');
     });
 
-    describe('when providing a path to a valid directory', () => {
-      it('should not display the file meta sections', async () => {
+    describe.each(['--json', '--sarif'])(
+      'when providing the %s flag',
+      (dataFormatFlag) => {
+        it('should not show an initial message for JSON output', async () => {
```
YairZ101, line 114: Shouldn't the name of the test be `should not show an initial message for ${dataFormatFlag} output`?

src/lib/formatters/iac-output/v2/failures-list.ts
ofekatr: Since all of the functionality for generating the test was exists in the `formatters` directory, naming the functions with the `format` prefix aligns the behavior we're adding to with the existing terminology in the CLI.

Some of the functions in which we use the `format` prefix are technically mappers, these include:
1. `formatAndShareResults`
2. `formatScanResults`
3. `formatShareResults`

These originated from the `FormattedResult` type, found in `src/cli/commands/test/iac/local-execution/types.ts`, which later was the inspiration for the naming of the `FormattedIssue` type in `FormattedIssue`.

In most languages, the formatter is defined as the tool for formatting and outputting data, based on how it should be displayed, usually in text, files, pages, or other similar forms of visual outputs ([see definition for Java](https://www.developer.com/design/what-is-a-java-formatter-in-java-programming/))

The `get` prefix feels a bit too abstract, and will always require and `output` suffix. The same goes for `generate`, `create`, etc..
I've considered `display`, but that's also incorrect, as these functions only generate output without printing it.

I'm happy to bring `get` back on the table and discuss it further

YairZ101: > Some of the functions in which we use the format prefix are technically mappers, these include:

Then maybe a more suitable action will be change those functions from `format` to `map`?

> In most languages, the formatter is defined as the tool for formatting and outputting data, based on how it should be displayed, usually in text, files, pages, or other similar forms of visual outputs ([see definition for Java](https://www.developer.com/design/what-is-a-java-formatter-in-java-programming/))

Thanks for sharing this info, in this case I'm convinced that we should stay with the `format` prefix.

ofekatr: > Then maybe a more suitable action will be to change those functions from format to map?

Good call, I 100% agree on this. 
I can do so in a separate refactor PR 💯

test/jest/acceptance/iac/iac-output.spec.ts
ofekatr: Nice catch! Fixed in [0cec22e8].(https://github.com/snyk/cli/pull/3135/commits/0cec22e87162e9dccc9627d8c9849f4a92c5e2a6)

test/jest/unit/lib/formatters/iac-output/failures-list.spec.ts
ofekatr: Great catch, fixed in [0cec22e8](https://github.com/snyk/cli/pull/3135/commits/0cec22e87162e9dccc9627d8c9849f4a92c5e2a6).

YairZ101 (approved):
LGTM 🚀

PR #3144, opened 2022-04-17
fix: set rule index result by security rules mapping

ShaharAvigezer:

#### What does this PR do?
Set the `ruleIndex` of a result to match the index of the rule in the security rules map

`src/lib/plugins/sast/analysis.ts`

run `code test` on a repo, in each result, look for rule index and make sure it matches in the rules array

When we run a code test, we first alter the rules array to use only security rules. The source of the problem lies when we don't address the `ruleIndex` that is now outdate since the rules indexes have changed.

[Jira Ticket](https://snyksec.atlassian.net/browse/ZEN-23)

#### Screenshots
N/A

#### Additional comments
I've also fixed the test fixture to match the desired result

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,6 +22,19 @@
       "console": "integratedTerminal",
       "program": "${workspaceFolder}/node_modules/.bin/jest",
       "args": ["${relativeFile}"]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Jest Current File",
+      "console": "integratedTerminal",
+      "program": "${workspaceFolder}/node_modules/.bin/jest",
+      "args": ["${relativeFile}"],
+      "runtimeArgs": [
+        "--inspect-brk",
+        "${workspaceRoot}/node_modules/.bin/jest",
+        "--runInBand"
+      ]
     }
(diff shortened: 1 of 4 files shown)


Spoor2709 (changes requested):
Required: Looks good to me can you put the debug change into a different commit please 😃

.vscode/launch.json
```diff
@@ -22,6 +22,19 @@
       "console": "integratedTerminal",
       "program": "${workspaceFolder}/node_modules/.bin/jest",
       "args": ["${relativeFile}"]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Jest Current File",
```
JackuB, line 29: Why is this needed? 🤔 I think code breakpoints/debugger works with the Jest launcher above as well?

And do you need `runInBand` (serial run) when you are running only a single file?

ShaharAvigezer: The existing launcher didn't work for me so I added this enhancement. It's fine by me to drop it if you don't believe it's necessary. WDYT?

Spoor2709 (approved):
Looks good to me, Great Job!

PR #3154, opened 2022-04-19
chore: fix help text documentation

Todai88:

#### What does this PR do?

Fixes incorrect example in help-text.

Read the help-text associated with the example fix-up.

Likely no reason to. It should just be released.

The support-team recently got a ticket from Yahoo! Japan suggesting our documentation provides
an incorrect example based on associated help text.
What's unintentionally confusing is this section:
>    Default: 4 (the current working directory and 3 sub-directories).

   Example: Limit search to the specified directory (or the current directory if no <PATH> is

  specified) plus three levels of subdirectories.

   --detection-depth=3

The assumption, based on the helptext, is that the example should be `--detection-depth=4`.
It's either that, or something else happening in our code that needs to be taken in account in the help text.

[Ticket in reference](https://snyk.zendesk.com/agent/tickets/22644)

* [Ticket](https://snyk.zendesk.com/agent/tickets/22644)

### DIFF
--- a/help/cli-commands/iac-test.md
+++ b/help/cli-commands/iac-test.md
@@ -37,7 +37,7 @@ Use to indicate how many subdirectories to search. `DEPTH` must be a number, 0 o
 
 Default: no limit.
 
-Example: `--detection-depth=3` limits search to the specified directory (or the current directory if no `<PATH>` is specified) plus three levels of subdirectories.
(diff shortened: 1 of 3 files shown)


craigfurman (changes requested):
I tested this out (for IaC at least), and the docs actually seem right - a detection depth of N will go N directories deep _not_ including the root of the scan. IMO this is actually a bit surprising, and because we don't support 0, we don't support "only scanning the root", which seems arbitrary.

PR #3157, opened 2022-04-20
fix: Skip unsupported IaC files

ipapast:

#### What does this PR do?

This commit will fix an issue where unsupported IaC files were printed in the scan output for users without the iacTerraformVarSupport flag.
This was due to a refactoring, where a check for the valid IaC files was skipped. 
It is also not considered part of the failedFiles, as it was not tested at all. (I checked the previous behaviour with an older version as well, e.g 1.840.0)

This is now added (hardcoded) for the flow that the flag is off. This flow is planned to be deprecated next week so I chose to do a quick fix for the customers now.

- Make sure the `iacTerraformVarSupport` flag is off.
- Run `snyk-dev iac test test/fixtures/iac/terraform/var_deref` (which contains non supported IaC files such as auto.tfvars) and see that the results do not contain `unsupported file extension anymore`

The refactored/deleted code was doing the check here before https://github.com/snyk/cli/blob/1c7f002e39d36a805e1b145cbb5b681d7cd6b2f5/src/cli/commands/test/iac-local-execution/file-loader.ts#L59-L60

#### Screenshots
Before:
image

After:
image

### DIFF
--- a/src/cli/commands/test/iac/local-execution/file-parser.ts
+++ b/src/cli/commands/test/iac/local-execution/file-parser.ts
@@ -38,7 +38,9 @@ export async function parseFiles(
   let nonTfFileData: IacFileData[] = [];
 
   if (!isTFVarSupportEnabled) {
-    nonTfFileData = filesData;
+    nonTfFileData = filesData.filter((fileData) =>
+      ['tf', 'json', 'yaml', 'yml'].includes(fileData.fileType),
+    );
   } else {
     tfFileData = filesData.filter((fileData) =>
       VALID_TERRAFORM_FILE_TYPES.includes(fileData.fileType),
--- a/test/jest/acceptance/iac/test-directory.spec.ts
+++ b/test/jest/acceptance/iac/test-directory.spec.ts
@@ -45,7 +45,7 @@ describe('Directory scan', () => {
     expect(stdout).toContain('Failed to parse YAML file');
     expect(stdout).toContain('Failed to parse JSON file');
     expect(stdout).toContain(
-      '28 projects, 20 contained issues. Failed to test 8 projects.',
+      '28 projects, 20 contained issues. Failed to test 5 projects.',
     );
     expect(exitCode).toBe(1);
   });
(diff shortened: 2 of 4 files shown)


src/cli/commands/test/iac/local-execution/file-parser.ts
```diff
@@ -38,7 +38,9 @@ export async function parseFiles(
   let nonTfFileData: IacFileData[] = [];
 
   if (!isTFVarSupportEnabled) {
-    nonTfFileData = filesData;
+    nonTfFileData = filesData.filter((fileData) =>
```
teodora-sandu, line 41: I'm not sure what we'll do with the CLI soon so maybe this is a bit extra but I'm not keen on hardcoding these here. 

Do you think it's worth doing something like:
```
export const VALID_TERRAFORM_FILE_TYPES: string[] = [
  'tf',
  'tfvars',
];
const VALID_NON_TERRAFORM_FILE_TYPES: string[] = [
  'json',
  'yaml',
  'yml',
];
export const VALID_FILE_TYPES: string[] = [
    ...VALID_TERRAFORM_FILE_TYPES,
    ...VALID_NON_TERRAFORM_FILE_TYPES,
];
export const OLD_VALID_FILE_TYPES: string[] = [
  'tf',
  ...VALID_NON_TERRAFORM_FILE_TYPES,
];
```

And using `OLD_VALID_FILE_TYPES` in this branch and `VALID_FILE_TYPES` in the other?

Just a suggestion, leaving it up to you

ipapast: > This is now added (hardcoded) for the flow that the flag is off. This flow is planned to be deprecated next week so I chose to do a quick fix for the customers now.

yeah I was trying to saying in the description that we are going to delete it in apprixamately a week that's why I added them hardcoded. Deleting the hardcoded values looked easier for me than deleting more code for the new type as well. Not that it would make much difference tbh 🤷

teodora-sandu (approved):
🎖️

PR #3174, opened 2022-04-27
chore(ci): integrate v2 into release pipeline

jahed-snyk:
- Move v2's pipeline into v1's pipeline after v1 binaries are built.
- Use locally built v1 binaries for building v2 binaries instead of downloading the latest v1 release.
- Fix v2's versioning to be consistent. Version files no longer contain "v".
    - Previous implementations removed the "v" in "dev" versions.
    - Use `.` instead of `+` for v2 versioning as `+` is whitespace in URLs.
- Run v2 builds and tests on `master`.
- Prefix jobs with "v2 /" to easily differentiate from v1 jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -529,6 +529,8 @@ jobs:
     working_directory: /home/circleci/snyk
     steps:
       - checkout
+      - attach_workspace:
+          at: .
       - go/install:
           version: << pipeline.parameters.go_version >>
       - restore_cache:
@@ -539,6 +541,7 @@ jobs:
           environment:
             GOOS: linux
             GOARCH: amd64
+            CLI_V1_LOCATION: ../binary-releases
           command: make build build-test install prefix=. -e
       - run:
           name: Build darwin/amd64
@@ -546,6 +549,7 @@ jobs:
           environment:
             GOOS: darwin
             GOARCH: amd64
+            CLI_V1_LOCATION: ../binary-releases
           command: make build build-test install prefix=. -e
       - run:
           name: Build windows/amd64
@@ -553,6 +557,7 @@ jobs:
           environment:
             GOOS: windows
             GOARCH: amd64
+            CLI_V1_LOCATION: ../binary-releases
           command: make build build-test install prefix=. -e
       - save_cache:
           key: go-mod-{{ arch }}-{{ checksum "cliv2/go.sum" }}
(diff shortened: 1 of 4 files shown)


cliv2/Makefile
```diff
@@ -50,11 +50,9 @@ endif
 _V1_DOWNLOAD = true
 ifneq ($(CLI_V1_LOCATION), $(_EMPTY)) 
 	ifeq ($(CLI_V1_VERSION_TAG), $(_EMPTY))
-$(error ERROR! CLI_V1_VERSION_TAG needs to be specified)
-		#_V1_DOWNLOAD = true
-	else
-		_V1_DOWNLOAD = false
+		CLI_V1_VERSION_TAG = $(shell cat "$(CLI_V1_LOCATION)/version")
```
PeterSchafer, line 53: praise: I like this solution a lot! 

question: The build log shows something like
`-- Using cliv1 version ( 1.913.0-dev.6d47f74652ab955c48229f1eb639943f4acbd299 )`
This string will be part of the cliv2 version. Is this correct? will the version number change in case of a release?

PeterSchafer (commented):
Looks very good to me, maybe two things we need to clarify before merging are:
* the question about the version number, see comment.
* how to handle/merge changes in cliv2 folder

cliv2/Makefile
jahed-snyk: >This string will be part of the cliv2 version. Is this correct?

Yes.

>will the version number change in case of a release?

No, once the version has been set in master during build, it will be released with that version. The version seen in PR branches (dev pre-release) are not the same as the one used in master (minor bump).

jahed-snyk: Noticed, `-dev` won't be valid semver when combined with v2's version. I'll fix that by making it into a `.` for v2.

.circleci/config.yml
```diff
@@ -539,20 +541,23 @@ jobs:
           environment:
             GOOS: linux
             GOARCH: amd64
+            CLI_V1_LOCATION: ../binary-releases
           command: make build build-test install prefix=. -e
```
PeterSchafer, line 545: question: not sure whether or not this might problems so just the question to be safe, should we use `make clean` before each build?

PeterSchafer (approved):
Just left one maybe more or less optional comment.

.circleci/config.yml
jahed-snyk: Shouldn't need to. cliv2's directory will be clean at this point since all the previous steps are related to v1 and only the things we choose to persist to workspace are carried over between jobs.

PR #3162, opened 2022-04-22
feat: support base64 encoding

soniqua:
- [x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/CONTRIBUTING.md) rules

#### What does this PR do?
- Adds support for base64 encoded payloads when executing `snyk code test`.
- Depends on `code-client` `4.11.0` (https://github.com/snyk/code-client/pull/144)

#### Where should the reviewer start?
- Review the above PR (https://github.com/snyk/code-client/pull/144). Changes made here are to support an additional boolean (`base64Encoding`).

#### How should this be manually tested?
- Ensure `code-client` `4.11.0` is available (either via `npm` after the above PR is merged, or by pulling the linked PR above and importing locally)
- A version of `deeproxy` with base64 encoding support is needed (https://github.com/snyk/deeproxy/pull/133)
- Set the `SNYK_USE_BASE64_ENCODING` environment variable to `true`
- Execute `snyk code test`

#### Any background context you want to provide?
This change is currently required for single tenant code analysis only.

Single tenant `deeproxy` instances are protected by Akamai WAF. When executing `snyk code test` the payload (source code transmitted in the clear) creates false positives and is blocked. Encoding to base64 should mask this content without need for more costly encryption activities.

#### What are the relevant tickets?
https://snyksec.atlassian.net/browse/NEBULA-520

#### Screenshots
-

#### Additional questions
-

PR #3183, opened 2022-04-29
feat: support base64 encoding

soniqua:

#### What does this PR do?
- Adds support for base64 encoded payloads when executing `snyk code test`
- Targets single-tenant deployments of Snyk only

- Changes made to `src/lib/config.ts` to handle the new config key, and `src/lib/plugins/sast/analysis.ts` to handle setting the new `base64Encoding` variable.

- A version of `deeproxy` with base64 encoding support is required (supported from https://github.com/snyk/deeproxy/pull/135 onwards)
- Either:
  - set the `SNYK_CFG_USE_BASE64_ENCODING` environment variable to `true`, or
  - set the `USE_BASE64_ENCODING` config value to `true`
- Execute `snyk code test -d` - the debug entry `snyk-code Base 64 Encoding enabled: true +0ms` should be present
- Validate the request to the `/bundle` endpoint is base64 encoded

- This is targeted only for single tenant deployments (hence if not set, base64 encoding is disabled as the `deeproxy` service will not accept base64 encoded payloads in multi-tenant).

- See https://github.com/snyk/cli/pull/3162 for a previous iteration of this PR

https://snyksec.atlassian.net/browse/NEBULA-520

#### Screenshots
`-`

#### Additional questions
`-`

### DIFF
--- a/src/lib/code-config.ts
+++ b/src/lib/code-config.ts
@@ -1,4 +1,5 @@
 import config from './config';
+import { config as userConfig } from './user-config';
 
 export function getCodeClientProxyUrl() {
   const url = new URL(config.API);
@@ -7,4 +8,13 @@ export function getCodeClientProxyUrl() {
     config.CODE_CLIENT_PROXY_URL ||
     domain.replace(/\/\/(app\.)?/, '//deeproxy.')
   );
+}
+
+export function getBase64Encoding(
+  enabled = userConfig.get('use-base64-encoding'),
+): boolean {
+  if (enabled) {
+    return enabled.toLowerCase() === 'true';
+  }
+  return false;
 }
--- a/src/lib/plugins/sast/analysis.ts
+++ b/src/lib/plugins/sast/analysis.ts
@@ -7,6 +7,7 @@ import { ReportingDescriptor, Result } from 'sarif';
 import { SEVERITY } from '../../snyk-test/legacy';
 import { api } from '../../api-token';
 import config from '../../config';
+import { getBase64Encoding } from '../../code-config';
 import { spinner } from '../../spinner';
 import { Options } from '../../types';
 import { SastSettings, Log } from './types';
@@ -57,6 +58,9 @@ async function getCodeAnalysis(
     ? sastSettings.localCodeEngine.url
     : getCodeClientProxyUrl();
 
+  const base64Encoding = getBase64Encoding();
+  debug(`base64 encoding enabled: ${base64Encoding}`);
+
   // TODO(james) This mirrors the implementation in request.ts and we need to use this for deeproxy calls
   // This ensures we support lowercase http(s)_proxy values as well
   // The weird IF around it ensures we don't create an envvar with
@@ -82,7 +86,7 @@ async function getCodeAnalysis(
     ? severityToAnalysisSeverity(options.severityThreshold)
     : AnalysisSeverity.info;
   const result = await analyzeFolders({
-    connection: { baseURL, sessionToken, source, requestId },
+    connection: { baseURL, sessionToken, source, requestId, base64Encoding },
     analysisOptions: { severity },
     fileOptions: { paths: [root] },
     analysisContext: {
--- a/test/jest/acceptance/snyk-code-config/snyk-code-config.spec.ts
+++ b/test/jest/acceptance/snyk-code-config/snyk-code-config.spec.ts
@@ -1,0 +1,13 @@
+import { runSnykCLI } from '../../util/runSnykCLI';
+
+jest.setTimeout(1000 * 60);
+
+test('it does not enable base64 encoding by default', async () => {
+  const { code, stdout } = await runSnykCLI('config get use-base64-encoding', {
+    env: {
+      ...process.env,
+    },
+  });
+  expect(code).toEqual(0);
+  expect(stdout).toEqual('');
+});
--- a/test/jest/unit/snyk-code/snyk-code-config.spec.ts
+++ b/test/jest/unit/snyk-code/snyk-code-config.spec.ts
@@ -1,0 +1,15 @@
+import { getBase64Encoding } from '../../../../src/lib/code-config';
+
+describe('test base64 setting', () => {
+  test('if not defined, base64 encoding is disabled', () => {
+    expect(getBase64Encoding()).toBe(false);
+  });
+
+  test('if set to true, base64 encoding is enabled', () => {
+    expect(getBase64Encoding('true')).toBe(true);
+  });
+
+  test('if set to false, base64 encoding is disabled', () => {
+    expect(getBase64Encoding('false')).toBe(false);
+  });
+});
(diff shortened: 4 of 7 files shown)


jahed-snyk (commented):
suggestion: Instead of defining base64Encoding twice (once in config.ts and again in analysis.ts), you can take a similar strategy to https://github.com/snyk/cli/pull/3119 with a `getBase64Encoding` function and unit test it.

pkey:
@soniqua any idea why we have a different behaviour on multi-tenant when it comes to accepting base64? To me it sounds like it could be an issue as we want our environment to be identical.

soniqua:
> @soniqua any idea why we have a different behaviour on multi-tenant when it comes to accepting base64? To me it sounds like it could be an issue as we want our environment to be identical.

The entrypoint for single tenant is currently behind the Akamai WAF. The WAF has rules associated with the deeproxy endpoints which has flagged source code uploaded in the clear as a false positive (and blocked the request). Mutli-tenant is not currently behind this set of WAF rules, but *should* be and likely will be in the future.

test/jest/unit/snyk-code/snyk-code-config.spec.ts
```diff
@@ -0,0 +1,25 @@
+import config from '../../../../src/lib/config';
```
pkey, line 1: suggestion: I was debating in my head whether it makes more sense to separate concerns and test just the fact the we are passing the right argument (your approach) or actually not treat code client as a black box and have an integration test. I would go with the latter but provided we have tests in `code-client`, this shouldn't make a huge difference.

src/lib/config.ts
```diff
@@ -46,6 +47,12 @@ if (endpoint && endpoint !== config.API) {
   config.API = endpoint;
 }
 
+// allow toggling of base64 payload encoding
+const useBase64Encoding = userConfig.get('use-base64-encoding');
+if (useBase64Encoding && useBase64Encoding.toLowerCase() === 'true') {
```
metju90, line 52: Question: how is useBase64Encoding a string?

pkey (approved):
left a few comments, let me know what you think. Otherwise looks good.

src/lib/plugins/sast/analysis.ts
```diff
@@ -57,6 +58,9 @@ async function getCodeAnalysis(
     ? sastSettings.localCodeEngine.url
     : getCodeClientProxyUrl();
 
+  const base64Encoding = getBase64Encoding();
+  debug(`base64 encoding enabled: ${base64Encoding}`);
```
metju90, line 62: Do we need to log this?

src/lib/config.ts
soniqua: `userConfig` will return whatever `process.env.SNYK_CFG_USE_BASE64_ENCODING` var is set to - which will always be a string.

src/lib/plugins/sast/analysis.ts
pkey: I think this might be useful. Also this would only log once on start, I believe, so no effect on quotas.

soniqua: We have a similar log in https://github.com/snyk/cli/commit/2d1abfbab for the use of a custom endpoint. Not necessary, but will at least show if/when customers use this that they're setting the CLI up correctly to use this feature.

metju90 (commented):
@soniqua  Do we plan to set`USE_BASE64_ENCODING`  true by default after we test? or do we plan to leave things as they are and only enable the flag when needed?

soniqua:
> @soniqua Do we plan to set`USE_BASE64_ENCODING` true by default after we test? or do we plan to leave things as they are and only enable the flag when needed?

The latter for now - the capability to handle base64 encoded payloads is feature-flag driven on `deeproxy` at the moment, so we'd want to ensure that was set to be enabled by default before doing so at the `cli` level.

src/lib/config.ts
```diff
@@ -46,6 +47,12 @@ if (endpoint && endpoint !== config.API) {
...
+  config.USE_BASE64_ENCODING = true;
+}
```
jahed-snyk, line 54: Should all of this be moved to code-config.ts? Then the flag's behaviour can all be tested in one place. Also avoids exposing a `config.USE_BASE64_ENCODING` global.

`userConfig.get('use-base64-encoding')` can be a default value for `getBase64Encoding`'s input parameter and in tests you can pass through other possible values without needing to modify and reset global modules in the test.

If you go with @pkey's suggestion, then we'll only need 1 acceptance test to make sure the default value is correctly wired up (rather than permutations of possible values, which is faster as unit tests).

src/lib/code-config.ts
```diff
@@ -8,3 +9,10 @@ export function getCodeClientProxyUrl() {
     domain.replace(/\/\/(app\.)?/, '//deeproxy.')
   );
 }
+
+export function getBase64Encoding(enabled=userConfig.get('use-base64-encoding')): boolean {
+  if (enabled){
+    return (enabled.toLowerCase() === 'true')
+  }
+  return false
+}
```
metju90, line 18: Can you run the code formatter pls? (pipeline will fail)

test/jest/acceptance/snyk-code-config/snyk-code-config.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+
+test('it does not enable base64 encoding by default', async () => {
+
+  const { code, stdout } = await runSnykCLI('config get use-base64-encoding', {
+    env: {
+      ...process.env,
+    },
+  });
+  expect(code).toEqual(0);
+  expect(stdout).toEqual("");
```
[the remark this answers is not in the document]
soniqua: That (by default) base64 encoding is disabled

test/jest/acceptance/snyk-code-config/snyk-code-config.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+jest.setTimeout(1000 * 60);
+
+test('it does not enable base64 encoding by default', async () => {
+  const { code, stdout } = await runSnykCLI('config get use-base64-encoding', {
+    env: {
+      ...process.env,
+    },
+  });
+  expect(code).toEqual(0);
+  expect(stdout).toEqual('');
```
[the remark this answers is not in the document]
soniqua: Nope, it'll return `true` to `stdout` - the exit code will be `0` either way

jahed-snyk (approved):

test/jest/acceptance/snyk-code-config/snyk-code-config.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+});
```
jahed-snyk, line 13: This test isn't doing what we think. The default value is defined by `getBase64Encoding` whereas `snyk config get use-base64-encoding` is only checking the value in the user's configstore (which always defaults to nothing / unset rather than default). So it's just testing `snyk config` behaviour.

To test the actual wiring in `analysis.ts > analyzeFolders` we'll want to use `fake-server` (a mock API server) and check if the API request sent the expected value given a code scan command using a `SNYK_CONF_USE_BASE64_ENCODING=true` env var.

Check out `test/jest/acceptance/oauth-token.spec.ts` as an example.

novalex (approved):

PR #3190, opened 2022-05-02
feat: deprecation notice when user run `snyk iac report`

YairZ101:

#### What does this PR do?
We would like to remove the command `snyk iac report` in the least disruptive way, this PR adds a deprecation notice / warning when a user runs the command `snyk iac report` that let them know when are we planning to remove the command and which command and flag are replacing it.

1. Run `snyk iac report`
2. Check that the warning is being printed at the end of output
3. Run `snyk iac test --report`
4. Check that a warning is not being printed 

https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-1794

#### Screenshots
image

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -53,7 +53,7 @@ export default async function test(
   const { options: originalOptions, paths } = processCommandArgs(...args);
 
   if (originalOptions.iac) {
-    return await iacTestCommand(...args);
+    return await iacTestCommand(false, ...args);
   }
 
   const options = setDefaultTestOptions(originalOptions);
--- a/src/cli/commands/report/index.ts
+++ b/src/cli/commands/report/index.ts
@@ -1,7 +1,7 @@
 import { MethodArgs } from '../../args';
 import { IaCTestFlags } from '../test/iac/local-execution/types';
 import { TestCommandResult } from '../types';
-import test from '../test';
+import test from '../test/iac/';
 import { hasFeatureFlag } from '../../../lib/feature-flags';
 import { Options } from '../../../lib/types';
 import { UnsupportedFeatureFlagError } from '../../../lib/errors';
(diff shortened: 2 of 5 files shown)


src/cli/commands/test/index.ts
```diff
@@ -53,7 +53,7 @@ export default async function test(
   const { options: originalOptions, paths } = processCommandArgs(...args);
 
   if (originalOptions.iac) {
-    return await iacTestCommand(...args);
+    return await iacTestCommand(false, ...args);
```
ipapast, line 56: just to confirm, this is the only way to differentiate because the report is an alias of the --report flag, right?
at first I thought, why don't we just check the args array for the options.report value? But then I remembered the alias situation, so I guess when we check the args array `report: true` will be returned for both cases? Is that correct?

YairZ101: > this is the only way to differentiate because the report is an alias of the --report flag, right?

Correct, as you mentioned `snyk iac report` is simply an alias for `snyk iac test --report` so we don't have a good way to differentiate between the two.
Another option I thought about was to manually add to the `options` a field that indicates that the user used `snyk iac report` (something like `options.command = 'report'`) and using this field as the condition to print the warning message but it felt to me like abusing the `options` for something that it was never meant to be used for.

ipapast: yeah... from one hand I agree with what you are saying about the options. from the other hand this would not require reviews from the other teams.

the other option is to create our own "options" object to pass to the iacTestCommand so that we don't end up on this again in the future. But I'm not sure if and how often we will need something like this.

In general this is something to be deleted in a month - so I'm happy to get it out this way. Maybe the best thing we can do is add a comment on top of it saying this hardcoded value is temporary and we will remove it in June?

Then, if we have a similar use case again after its deletion, we tackle it different in the future?

ipapast (approved):

src/cli/commands/test/index.ts
YairZ101: > Maybe the best thing we can do is add a comment on top of it saying this hardcoded value is temporary and we will remove it in June?

sounds like a good idea, I'll add it.

robcresswell (approved):

PR #3191, opened 2022-05-02
chore: remove-path-from-url

shirlupo:


pkey (approved):
this is a preparation for potentially adding another candidate string to  replace. In that case, looks good to me + this is also cleaner.

PR #3193, opened 2022-05-02
Replace value of x-snyk-cli-version with v2 version if set

maxjeffos:

#### What does this PR do?

Updates the CLIv2 proxy so that, if the `x-snyk-cli-version` header is present on a request, we replace it with the v2 version.

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -70,7 +70,8 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 	}
 
 	// init proxy object
-	wrapperProxy, err := proxy.NewWrapperProxy(envVariables.UpstreamProxy, envVariables.CacheDirectory, debugLogger)
+
+	wrapperProxy, err := proxy.NewWrapperProxy(envVariables.UpstreamProxy, envVariables.CacheDirectory, cli.GetFullVersion(), debugLogger)
 	if err != nil {
 		fmt.Println("Failed to create proxy")
 		fmt.Println(err)
--- a/cliv2/internal/proxy/proxy_test.go
+++ b/cliv2/internal/proxy/proxy_test.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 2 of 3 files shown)


cliv2/internal/proxy/proxy_test.go
```diff
@@ -75,3 +76,62 @@ func Test_canGoThroughProxy(t *testing.T) {
 	_, err = os.Stat(wp.CertificateLocation)
 	assert.NotNil(t, err) // this means the file is gone
 }
+
+func Test_xSnykCliVersionHeaderIsReplaced(t *testing.T) {
+	debugLogger := log.New(os.Stderr, "", log.Ldate|log.Ltime|log.Lmicroseconds|log.Lshortfile)
+
+	wp, err := proxy.NewWrapperProxy("", []string{"static.snyk.io"}, "", "the-v2-version", debugLogger)
```
PeterSchafer, line 83: minor suggestion: putting expected values into explicit variables improves maintainability of tests since the value needs to be defined only in one location. This could apply to "the-v2-version" and statuscode 200.

cliv2/cmd/cliv2/main.go
```diff
@@ -82,7 +82,7 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 	}
 
 	// init proxy object
-	wrapperProxy, err := proxy.NewWrapperProxy(envVariables.UpstreamProxy, envVariables.SnykDNSNames, envVariables.CacheDirectory, debugLogger)
+	wrapperProxy, err := proxy.NewWrapperProxy(envVariables.UpstreamProxy, envVariables.SnykDNSNames, envVariables.CacheDirectory, cli.V2Version, debugLogger)
```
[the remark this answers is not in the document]
maxjeffos: Thanks. Done!

cliv2/internal/proxy/proxy_test.go
maxjeffos: Thanks. Done. I kept the `200` as a literal since that one seems so obvious.

PeterSchafer (approved):

PR #3198, opened 2022-05-03
Fix proxy behavior test to work with cliv2

maxjeffos:

#### What does this PR do?

Makes the `test/jest/acceptance/analytics.spec.ts` - `sends analytics for `snyk test` with no vulns found` test work with CLIv2. The error message is different in CLIv2 since it goes through the Go proxy.

### DIFF
--- a/test/jest/acceptance/proxy-behavior.spec.ts
+++ b/test/jest/acceptance/proxy-behavior.spec.ts
@@ -9,16 +9,9 @@ const FAKE_HTTP_PROXY = `http://localhost:${fakeServerPort}`;
 jest.setTimeout(1000 * 60 * 1);
 
 describe('Proxy configuration behavior', () => {
-  if (isCLIV2()) {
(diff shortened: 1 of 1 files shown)


test/jest/acceptance/proxy-behavior.spec.ts
```diff
@@ -29,9 +22,21 @@ describe('Proxy configuration behavior', () => {
       expect(code).toBe(0);
```
PeterSchafer, line 22: question: As far as I understand this test, it is testing an error case which results in not executing the request cli command. Naively I would have thought the this should cause an error code (>0). Do we have any guideline on when to set which return code?

PeterSchafer (approved):

test/jest/acceptance/proxy-behavior.spec.ts
PeterSchafer: Thanks!

PR #3201, opened 2022-05-04
fix: Don't create duplicated IaC projects when sharing results

francescomari:

#### What does this PR do?

This PR collects IaC projects into project groups depending on the repository they belong to. Because every IaC project corresponds to a file, and each project is associated to the corresponding repository, this PR also avoids creating duplicated projects whenever a file is scanned more than once using variadic arguments on the command line.

The bulk of the changes is in `src/cli/commands/test/iac/local-execution/process-results/share-results-formatter.ts`. Other changes fix the caller of the modified function and the tests that broke.

Try to scan the same file by using different paths. This will create just one project group, with only one project in it.

```
snyk iac test --report ./test/fixtures/iac/terraform ./test/fixtures/iac/terraform/sg_open_ssh.tf
```

image

Try to scan multiple files from different Git repositories. This will create one project group per Git repository, and every IaC project will be assigned to the right project group depending on the repository they are in. In the following example, `cli` and `terraform-goof` are two repositories.

```
snyk iac test --report cli/test/fixtures/iac/terraform terraform-goof
```

image

[CFG-1788](https://snyksec.atlassian.net/browse/CFG-1788)


src/cli/commands/test/iac/local-execution/process-results/share-results-formatter.ts
```diff
@@ -45,3 +38,54 @@ function groupByFilePath(scanResults: IacFileScanResult[]) {
 
   return Object.values(groupedByFilePath);
 }
+
+function computePaths(
```
ofekatr, line 42: In this PR, the original implementation for the `computePaths` function is no longer important, and a new implementation was added here. 
It seems like the original implementation is still being used in `src/cli/commands/test/iac/local-execution/process-results/results-formatter.ts`. 
Would you mind providing a brief description of the difference between the 2, and why we need the original implementation when formatting test results?

francescomari: The original implementation is still used when generating SARIF data. I didn't change the previous implementation of `computePaths` in order to not upset the SARIF use case.

The main difference between the two implementations is in how the parent directory of the scanned paths is computed. For this, the original implementation needs two pieces of information:

- the path to the current file being scanned;
- the path that was passed on the command line, and that eventually led to the scanned file being discovered.

For example, if the file being scanned is `/foo/bar/baz/main.tf`, and it was discovered because `/foo/bar/baz/main.tf` was passed on the command line as is, then the project name will be `baz` and the target file will be the path to the input file, relative to the current working directory. On the other hand, if the file being scanned is `/foo/bar/baz/main.tf`, and it was discovered because `/foo/bar` was passed on the command line, the project name will be `bar` and the target file will be the path to the file relative to the `bar` directory.

This is not good enough to deduplicate the scanned paths, as different arguments on the command line may lead to different target file paths pointing to the same actual file. So, the solution is to identify the Git repository the file belongs to, and make each path relative to that. This way, even if the same file is scanned more than once, it will result in only one path, and it will belong to only one project group. As a side effect, this strategy doesn't need to rely on the path that was passed on the command line.

francescomari: > It concerns me a bit that we'd diverge into 2 different ways of computing paths for these use-cases. I think it's worth bringing this up in the team channel so as to inquire whether or not it's possible to update the path computation for the SARIF format to the new implementation as well for consistency.

It doesn't concern me at all. Sharing results, and SARIF are different presentation strategies that might have different requirements when computing paths. The `computePaths` in this file is an implementation detail of this module, and is specifically implemented for this presentation strategy. I will bring it up to the team and ask for feedback about this point, though.

> If not, I think it's worth making it clearer in the code when each computation is relevant, either by making the function names more specific, changing the `utils.ts` file name which only has the `computePaths` function to be SARIF-centered, like `sarif-utils.ts`, or any other solution that comes to mind.

If the path computation for SARIF should use the same strategy, it's trivial to pull `computePaths` out in a separate module and make it available to the SARIF code, too. At the moment, as described in the issue, we don't want to affect how SARIF computes paths for the scanned files.

> Another question I had in mind is the expected behavior when files are not inside git repositories. Would the output for the original and the new implementations be the same?

When no Git repository can be found, `computePaths` falls back to the current working directory. This is a sensible default because, most of the times, you are going to run `snyk` from the same directory over and over (think about running `snyk` as part of a CI pipeline). When no Git repository can be detected, the paths will be relative to the current working directory. This is different from the original implementation of `computePaths` described above, where the current working directory is used only in some specific cases, depending on the command line arguments passed to the command.

src/cli/commands/test/iac/local-execution/process-results/share-results-formatter.ts
```diff
@@ -45,3 +38,54 @@ function groupByFilePath(scanResults: IacFileScanResult[]) {
...
+  const relativeFilePath = path.relative(currentDirectory, absoluteFilePath);
+
+  return {
+    targetFilePath: absoluteFilePath,
+    projectName: currentDirectoryName,
+    targetFile: relativeFilePath,
+  };
+}
+
+function getGitRootOrCwd(currentPath: string): string {
```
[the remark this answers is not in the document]
francescomari: Why now? Do we plan to reuse this? Isn't this premature reuse?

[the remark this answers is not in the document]
ofekatr: You raise a good point. It's more for the sake of separation of the logic here, which could also help to test these components in a more contained fashion. 
That being said, if it's not being to be used elsewhere at the moment, I'm happy to keep it as is.

src/cli/commands/test/iac/local-execution/process-results/share-results-formatter.ts
ipapast: I read the comments and I don't see any concerns for splitting the implementation of two paths. 
I like @francescomari 's explanation that we are talking about two different presenterviews that prepare the results in a different format.


>If not, I think it's worth making it clearer in the code when each computation is relevant, either by making the function names more specific, changing the utils.ts file name which only has the computePaths function to be SARIF-centered, like sarif-utils.ts, or any other solution that comes to mind.

Regarding the confusion about which `computePath` function is used where, the one inline the file is pretty self explanatory, right?

The second one that is extracted to `utils.ts` can either get a comment to say this is used only for the results-formatter.ts etc, or, even get moved back closer to its implementation (it's exported solely for that function). 
We sometimes exported functions in seperate files to unit test them easier with mocks. This is used only as a call in `results-formatter.ts` so if we move it back closer to its implementation it should be also self explanatory. 
What do you think?

ofekatr: @ipapast Makes sense, happy to go with this suggested approach.

francescomari:
@ofekatr I think we reached an agreement on the SARIF use case. Can we move this PR forward?

ofekatr:
It was my understanding that we agreed to go with the scoping suggestion from @ipapast to ensure the original implementation for the `computePaths` function is more coupled with its current usage in the code.
Happy to approve as it's not blocking, but I think it's best to add this refactoring bit before merging.

ofekatr (approved):

PR #3204, opened 2022-05-05
chore: Added an invalid files count to the test summary section [CFG-1787]

ofekatr:

#### What does this PR do?

- Adds the invalid files count to the test summary section, to increase its visibility. This change is based on this design.

    src/lib/formatters/iac-output/v2/test-summary.ts

- Test a directory with any number of invalid files (including none), and ensure that the correct number of invalid files appears in the test summary section.

In [the new test output design](https://www.figma.com/file/buNFZqr1Uhf4SaifHw4WXE/CLI-IaC-Output-Improvements?node-id=0%3A1), we are currently not displaying the number of files that failed to be tested in the test summary section, but it's included in the title for the test failures section.
We would like to add it to the test summary section as well, to increase visibility.

- [CFG-1787](https://snyksec.atlassian.net/browse/CFG-1787)

#### Screenshots

image

### DIFF
--- a/test/jest/acceptance/iac/iac-output.spec.ts
+++ b/test/jest/acceptance/iac/iac-output.spec.ts
@@ -83,9 +83,11 @@ describe('iac test output', () => {
           EOL +
           '✗ Files with issues: 3' +
           EOL +
+          '  Invalid files: 1' +
+          EOL +
           '  Ignored issues: 8' +
           EOL +
-          '  Total issues: 28 [ 0 critical, 4 high, 8 medium, 16 low ]',
+          '  Total issues: ',
       );
     });
 
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -319,6 +319,7 @@ export default async function(
     const iacTestSummary = `${formatIacTestSummary(
       {
         results,
+        failures: iacScanFailures,
         ignoreCount: iacIgnoredIssuesCount,
       },
       iacOutputMeta,
--- a/src/lib/formatters/iac-output/v2/test-summary.ts
+++ b/src/lib/formatters/iac-output/v2/test-summary.ts
@@ -60,6 +60,12 @@ function formatCountsSection(testData: IacTestData): string {
   );
 
   countsSectionProperties.push(
+    `${INDENT}Invalid files: ${colors.info.bold(
+      `${testData.failures?.length ?? 0}`,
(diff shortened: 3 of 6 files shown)


test/jest/acceptance/iac/iac-output.spec.ts
```diff
@@ -83,6 +83,8 @@ describe('iac test output', () => {
           EOL +
           '✗ Files with issues: 3' +
           EOL +
+          '  Invalid files: 1' +
+          EOL +
           '  Ignored issues: 8' +
           EOL +
           '  Total issues: 28 [ 0 critical, 4 high, 8 medium, 16 low ]',
```
ipapast, line 90: line 90: 
can we not assert on number of issues, please? this causes flakiness from time to time when we update policies, etc. 
It's also not the goal of this test I believe.
```suggestion
          '  Total issues:',
```

ipapast: (even though this change is not part of this PR, I didn't know where to comment on)

ofekatr: It's technically a part of what is being tested here. Apart from the format of the issues summary line, it tests the fact that the number of issues found for each severity adds up to the number of total issues found.

As a compromise, we could use a regular expression that simply ensures the text has numeric values in the relevant locations. However, a regular expression check for these cases would cost more in terms of resources.

What do you think?

ipapast: I think the best scenario would be to update all tests in all places, but that's a seperate discussion. That's why I suggest these tests to go out without numbers and not fail suddently at some other point in time

ofekatr: Looking at the rest of the tests in this suite, most would break in the event of any changes to our policies, as we also test the output of the issues list. I'm happy to make the requested change, but I'm not sure it'd be enough from preventing the output tests from breaking in the event of content changes.

ipapast: Yeap anything with numbers that relies on live data will break. I'm just saying let's not continue a pattern we know it's fragile just because we are used to doing it so far. does it make sense?

ofekatr: Will do 💯

ipapast (approved):
:shipit:

PR #3206, opened 2022-05-06
feat: update code client

JohnBatts418:
Updates the code client to the latest version.

This update means Code-client will now retry on 500 failures when making downstream calls. 

Code-client change [here](https://github.com/snyk/code-client/pull/145)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@open-policy-agent/opa-wasm": "^1.6.0",
         "@snyk/cli-interface": "2.11.0",
         "@snyk/cloud-config-parser": "^1.14.1",
-        "@snyk/code-client": "^4.11.0",
+        "@snyk/code-client": "^4.11.1",
         "@snyk/dep-graph": "^1.27.1",
         "@snyk/docker-registry-v2-client": "^2.6.1",
         "@snyk/fix": "file:packages/snyk-fix",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1966,7 +1966,7 @@
         "needle": "^3.0.0",
         "p-map": "^3.0.0",
         "uuid": "^8.3.2",
-        "yaml": "^1.10.2"
+        "yaml": "^2.0.1"
```
jahed-snyk, line 1969: yaml@2 drops Node 12 support. Snyk CLI supports Node 12 and above so we can't accept this until we drop Node 12 support (likely next year as usual) 🤔 Any reason we need to upgrade?

https://github.com/eemeli/yaml/releases/tag/v2.0.0

> Drop Node.js 12 support (end-of-life 2022-04-30)

jahed-snyk (changes requested):

package-lock.json
JohnBatts418: I was trying to use a slightly newer version of our code-client, https://www.npmjs.com/package/@snyk/code-client/v/4.12.0 but looks like it introduces some issues, I've informed the team to fix, and for this PR I've changed it to just be the minor update 4.11.1 which wont introduce this yaml update :)

jahed-snyk (approved):

PR #3207, opened 2022-05-06
chore: Added support for custom rules [CFG-1830]

ofekatr:

#### What does this PR do?

- Adds support for custom rules in the issue description for the new test output, by:
    - Rather than a docs URL, which custom rules don't have, we display the custom rule's public ID.

    src/lib/formatters/iac-output/v2/issues-list/issue.ts

1. Run the `snyk iac test` command with a custom rules bundles.
2. Ensure that all the expected issues generated from custom rules are displayed, and have the following format for the `Rule` property: `Rule: custom rule: <rule-public-ID>`

Currently, for issues generated by custom rules, we do not show the Rule property in the issue description, as it normally holds the documentation URL and custom rules do not have it. Since we use these documentation URLs to display security rule public IDs, we don’t include the custom rule public IDs in the output at all. 

We’d like to update the Rule property to specify the rule’s public ID, in the case of custom rules.

- [CFG-1830](https://snyksec.atlassian.net/browse/CFG-1830)

#### Screenshots

image

#### Additional notes

- The design used for the `Rule` property in this PR was approved in [the following Slack thread](https://snyk.slack.com/archives/C030UA3U1RB/p1650465503629659).


test/jest/unit/lib/formatters/iac-output/v2/issues-list/issue.spec.ts
```diff
@@ -0,0 +1,84 @@
+import * as pathLib from 'path';
+import * as fs from 'fs';
+import chalk from 'chalk';
+import { colors } from '../../../../../../../../src/lib/formatters/iac-output/v2/color-utils';
+import { formatIssue } from '../../../../../../../../src/lib/formatters/iac-output/v2/issues-list/issue';
+import { IacProjectType } from '../../../../../../../../src/lib/iac/constants';
+import { FormattedOutputResult } from '../../../../../../../../src/lib/formatters/iac-output/v2/issues-list/types';
+import { AnnotatedIacIssue } from '../../../../../../../../src/lib/snyk-test/iac-test-result';
+
+describe('formatIssue', () => {
```
teodora-sandu, line 10: I have a feeling for output related tests we usually resort to acceptance tests. See for example https://github.com/snyk/snyk/blob/d07b434ee83e033e9524bda76e818a8679b3635e/test/jest/acceptance/iac/custom-rules.spec.ts. 

This README says for example acceptance tests would be from a customer's perspective: https://github.com/snyk/cli/blob/d07b434ee83e033e9524bda76e818a8679b3635e/CONTRIBUTING.md#acceptance-tests

Is there a reason you chose to write these as unit tests?

ofekatr: @teodora-sandu At first I was uncertain as to which of these two would be more appropriate for this test, and thought acceptance tests shouldn't cover all the possible scenarios, but rather major ones, while covering these cases more thoroughly in unit tests.

Having read the contribution guidelines I believe you're right. I'll update the PR to cover this case via an acceptance test instead.

teodora-sandu (approved):

PR #3213, opened 2022-05-06
fix: SARIF output for IaC

teodora-sandu:

#### What does this PR do?
This PR fixes a problem in the GitHub action where the incorrect `PROJECTROOT` is computed (without a `/` at the end) when there is no `git` command.

The commits show how the code was refactored so it's easier to test, a test added to make sure we don't re-introduce this regression, and the actual fix.

It's hard to test manually because we all have `git` and we don't want to delete it just for this use case. Can see the failure here though: https://github.com/teodora-sandu/terraform-lambda-api-gateway/runs/6322494869?check_suite_focus=true

In https://github.com/snyk/cli/commit/30746efb0cc9b3dc5d466af5603af9abf76e59a2 we can see the regression. The reason we didn't catch it is because the GitHub action acceptance tests in https://github.com/snyk/snyk/blob/0447fcd47b6e2e6d3850a26a3c6b584a1851d510/test/jest/acceptance/iac/github-action/iac.spec.ts would use `git` and so never run into this path. 

Slack thread: https://snyk.slack.com/archives/CRV0PMFDH/p1651699191381929

### DIFF
--- a/src/lib/formatters/iac-output/v1/index.ts
+++ b/src/lib/formatters/iac-output/v1/index.ts
@@ -144,7 +144,7 @@ export function createSarifOutputForIac(
   try {
     repoRoot = getRepoRoot();
   } catch {
-    repoRoot = basePath;
+    repoRoot = pathLib.join(basePath, '/'); // the slash at the end is required, otherwise the artifactLocation.uri starts with a slash
   }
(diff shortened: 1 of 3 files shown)


ipapast (approved):
👌

PR #3214, opened 2022-05-06
fix: Ensure that IaC shared results paths use forward slashes

francescomari:

#### What does this PR do?

This PR ensures that the paths generated during an IaC scan and shared with the app always use forward slashes, independently of the platform where `snyk` runs.

Run `snyk iac test --report` on Windows and verify that the results are correctly shared with the app and use forward slashes.


ipapast (approved):
A run in windows VM also confirms that we don't create duplicate projects, and the '/' is fixed.

image
image

PR #3218, opened 2022-05-09
chore: Update test error message color [CFG-1385]

ofekatr:

#### What does this PR do?

- Adjusts the color for test failure messages to the new IaC output color scheme.

    src/lib/errors/legacy-errors.js

- Execute the `snyk iac test` command in a way that'd make it fail, such as testing a non-existing path, or scanning a path with no supported IaC files.

Currently in the new output, whenever a test flow fails, an output with only the error message is produced. This error message has the original shade of red used for errors/failures in the previous test output format. In the new test output format, we’ve defined a new shade of red which we would like to apply here as well.

- [CFG-1835](https://snyksec.atlassian.net/browse/CFG-1835)

#### Screenshots

Screen Shot 2022-05-09 at 18 15 23

### DIFF
--- a/src/cli/main.ts
+++ b/src/cli/main.ts
@@ -46,6 +46,7 @@ import { SarifFileOutputEmptyError } from '../lib/errors/empty-sarif-output-erro
 import { InvalidDetectionDepthValue } from '../lib/errors/invalid-detection-depth-value';
 import { obfuscateArgs } from '../lib/utils';
 import { EXIT_CODES } from './exit-codes';
+import { hasFeatureFlag } from '../lib/feature-flags';
 
 const debug = Debug('snyk');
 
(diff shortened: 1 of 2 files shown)


src/cli/main.ts
```diff
@@ -121,7 +122,11 @@ async function handleError(args, error) {
     console.log(output);
   } else {
     if (!args.options.quiet) {
-      const result = errors.message(error);
+      const shouldUseNewIacOutput =
+        args.options.iac &&
+        (await hasFeatureFlag('iacCliOutput', args.options));
```
jahed-snyk, line 127: issue: Can we avoid adding IaC-specific logic in `main`? This adds complexity into an already overloading module. As an example to avoid this: IaC can use extend `Error` with a `FormattedError` class which provides a `getFormattedMessage` for showing in CLI output with the correct colour. `legacy-error` can then use this function instead of it's available on the error object.

jahed-snyk (changes requested):
thought: Not too sure about using different colours for errors. Ideally our CLI colours for various context should be the same across products. I'll try bring relevant people into the Slack thread for thoughts. Maybe we can make IaC's red the default if it's aligned with the latest CLI branding decisions.

YairZ101 (approved):

ofekatr:
Closed based on [this discussion](https://snyk.slack.com/archives/C0127HWU0E7/p1652109744760259)

PR #3219, opened 2022-05-09
fix: git targets for variadic paths

YairZ101:

#### What does this PR do?
Currently when the user is running `snyk iac test --report {path1} {path2}` where `path1` and `path2` are git repo directories we group all the projects in the platform under one dropdown menu which is named after the current working directory, this is not the behaviour our users are expecting because from their point of view they are scanning 2 different directories so the projects should be grouped up into 2 different groups.
This PR fixes this issue by adding to the `git` commands we are executing the context of the path we are currently scanning.

Run `snyk iac test --report {path1} {path2}` before the change and after and check the difference.
In addition to that, `getInfo` is being used by other products in their `monitor` command so we should check that this change doesn't affect them.

https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-1756

### DIFF
--- a/src/lib/project-metadata/target-builders/git.ts
+++ b/src/lib/project-metadata/target-builders/git.ts
@@ -5,9 +5,13 @@ import { GitTarget } from '../types';
 // for scp-like syntax [user@]server:project.git
 const originRegex = /(.+@)?(.+):(.+$)/;
 
-export async function getInfo(
-  isFromContainer: boolean,
-): Promise<GitTarget | null> {
+export async function getInfo({
+  isFromContainer,
+  cwd,
+}: {
+  isFromContainer: boolean;
+  cwd?: string;
+}): Promise<GitTarget | null> {
   // safety check
   if (isFromContainer) {
     return null;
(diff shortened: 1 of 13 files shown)


src/lib/project-metadata/target-builders/git.ts
```diff
@@ -5,19 +5,24 @@ import { GitTarget } from '../types';
...
+  isFromContainer: boolean;
+  cwd?: string;
+}): Promise<GitTarget | null> {
   // safety check
   if (isFromContainer) {
     return null;
   }
 
   const target: GitTarget = {};
+  cwd = cwd || process.cwd();
```
francescomari, line 21: This is unnecessary. `cwd` defaults to `process.cwd()` if not specified.

francescomari (approved):

src/lib/project-metadata/target-builders/git.ts
YairZ101: Good to know :)
I've removed it and added `undefined` to the expected type [here](https://github.com/snyk/cli/blob/f210f1a55eb7fad023fa2c4431d856e3202a76a4/src/lib/sub-process.ts#L6).

Avishagp (approved):

PR #3225, opened 2022-05-11
fix: new CLI output styling [CFG-1841]

teodora-sandu:

#### What does this PR do?

This PR updates the styling used by IaC in the new CLI output, including:
- general colours 
- the custom rules informational message (colour and padding)
- the style of titles (need to be white and bold)
- the padding of contents underneath titles (2 pixels to the right)
- removing the spinner failure message
- the style of `Tip` messages (need to be white)

1. `npm run build`
2. `snyk-dev iac test test/fixtures/iac/terraform/var_deref` for success case
3. `snyk-dev iac test test/fixtures/iac/terraform/sg_open_ssh_invalid_hcl2.tf` for error case
4. `snyk-dev iac test ./test/fixtures/iac/terraform --rules=bundle.tar.gz` for custom rules case

In https://snyk.slack.com/archives/C0127HWU0E7/p1652109744760259 it was discovered that we were using the incorrect colours for formatting.

https://snyksec.atlassian.net/browse/CFG-1841

#### Screenshots
Before:
- custom rules edge case (notice two spinner messages, custom rules message on same line as spinner message, and never-ending spinner)
Screenshot 2022-05-12 at 10 34 44

After:
- success
Screenshot 2022-05-13 at 14 17 12

- success with failed files
Screenshot 2022-05-13 at 14 31 03

- error case
Screenshot 2022-05-13 at 14 40 06

- custom rules fixed edge case, colours, and added log before error message (new and old output)
Screenshot 2022-05-13 at 14 20 25

Screenshot 2022-05-13 at 11 03 22


src/cli/commands/test/iac/index.ts
```diff
@@ -149,10 +149,10 @@ export default async function(
   }
 
   try {
-    testSpinner?.start(spinnerMessage);
-
     const rulesOrigin = await initRules(iacOrgSettings, options);
 
+    testSpinner?.start(spinnerMessage);
```
teodora-sandu, line 154: This means the spinner will start after the rules are initialised. I think this is okay because we want to start showing the spinner when the testing actually starts.

src/cli/commands/test/iac/index.ts
```diff
@@ -213,6 +213,14 @@ export default async function(
         }
       }
     }
+  } catch (err) {
+    const res = formatTestError(err);
```
teodora-sandu, line 217: We should catch other thrown errors so we can render the not successsful log, i.e. `✖ Unable to complete the test.`

src/lib/formatters/iac-output/v2/color-utils.ts
```diff
@@ -14,12 +14,12 @@ type SeverityColor = {
...
 export const colors: IacOutputColors = {
   severities: {
-    critical: chalk.hex('#AB191A'),
-    high: chalk.hex('#CE501A'),
-    medium: chalk.hex('#D68000'),
-    low: chalk.hex('#88879E'),
+    critical: chalk.magenta,
+    high: chalk.red,
+    medium: chalk.yellow,
+    low: chalk.gray,
```
jahed-snyk, line 20: issue: For low severity, the colour is default (white). IIRC, when we tried grey, it was not very visible in many themes and blended with the terminal background.

These colours are in `lib/theme.ts`.

teodora-sandu: I agree with you that it's very hard to read. In [this thread](https://snyk.slack.com/archives/C0127HWU0E7/p1652254800431229?thread_ts=1652109744.760259&cid=C0127HWU0E7) @andygongea mentioned that gray should be used for low severity to match the Snyk colours in the UI. We could defer this conversation until later and go for white now but I'd like to know what Andy thinks

teodora-sandu: We agreed on white, thanks for flagging this up!

src/cli/commands/test/iac/index.ts
```diff
@@ -229,7 +229,7 @@ export default async function(
   const foundVulnerabilities = vulnerableResults.length > 0;
 
   if (notSuccess) {
-    testSpinner?.fail(spinnerFailureMessage + EOL);
+    testSpinner?.stop();
```
teodora-sandu, line 232: We don't want to print out anything but we want to stop the spinner

src/lib/formatters/iac-output/v2/issues-list/issue.ts
```diff
@@ -13,7 +13,13 @@ export function formatIssue(result: FormattedOutputResult): string {
 
   const propertiesOutput = formatProperties(result);
 
-  return titleOutput + EOL + propertiesOutput;
+  return (
```
ofekatr, line 16: nit: Refactor suggestion for the `formatIssue` function, for organizing the addition of the prefix padding to all output components of an issue.

```suggestion
export function formatIssue(result: FormattedOutputResult): string {
  const issueOutputComponents: string[] = [
    formatTitle(result.issue),
    ...formatProperties(result),
  ];

  return issueOutputComponents
    .map((issueOutputComponent) => contentPadding + issueOutputComponent)
    .join(EOL);
}
```

src/lib/formatters/iac-output/v2/user-messages.ts
```diff
@@ -19,10 +20,17 @@ export const spinnerMessage = colors.info(
 export const spinnerSuccessMessage = colors.info('Test completed.');
 
 /**
- * Displayed when a test fails.
+ * Message for using custom rules.
  */
-export const spinnerFailureMessage = colors.info(
-  'Unable to complete the test.',
+export const customRulesMessage = chalk.white(
```
ofekatr, line 25: Could be replaced with `colors.info`, so that we don't use `chalk` directly here.

src/lib/formatters/iac-output/v2/user-messages.ts
```diff
@@ -19,10 +20,17 @@ export const spinnerMessage = colors.info(
...
+  'Using custom rules to generate misconfigurations.',
+);
+
+/**
+ * Message for using custom rules.
+ */
+export const customRulesReportMessage = chalk.white(
```
ofekatr, line 32: Could be replaced with colors.info, so that we don't use `chalk` directly here.

src/cli/commands/test/iac/index.ts
teodora-sandu: Have a look at the ticekt please: https://snyksec.atlassian.net/browse/CFG-1841. Andy and I did a review of this PR over a call and he mentioned that we shouldn't print out  the extra spinner message `Unable to complete the test` at the end of each failure. I don't see that in the designs anywhere anyway 😄

src/lib/formatters/iac-output/v2/issues-list/issue.ts
teodora-sandu: Thanks for the suggestion. I find this format a bit harder to read personally. If you prefer this then feel free to open a refactoring PR after my PR

src/cli/commands/test/iac/local-execution/rules.ts
```diff
@@ -46,14 +49,13 @@ export async function initRules(
     (isOCIRegistryURLProvided || customRulesPath) &&
     !(options.sarif || options.json)
   ) {
-    let userMessage = 'Using custom rules to generate misconfigurations.';
+    let userMessage = `${customRulesMessage}\n`;
```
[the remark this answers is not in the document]
teodora-sandu: Re 2: I'm not changing the behaviour in this PR and while I agree with you, this change would require a bit more work. I am happy to think of a way to do this in a separate PR but I don't want to overcomplicate this PR

src/cli/commands/test/iac/index.ts
ofekatr: In which case, could we at least update the description for this PR to mention these changes as well?

teodora-sandu: Sure I'll copy them from the ticket if that's helpful

src/cli/commands/test/iac/local-execution/rules.ts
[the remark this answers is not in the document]
ofekatr: @teodora-sandu The change is the additional newline added for this user message, which I simply wanted to address in the standard way that was done in the new output format in other locations.
If you'd like, we could handle the newline itself in a separate PR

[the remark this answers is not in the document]
teodora-sandu: There's an additional newline yes but there's also an existing newline here that I'm updating. Both are used for formatting so I wouldn't really make any differentiation between them.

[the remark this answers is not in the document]
teodora-sandu: I've opened https://github.com/snyk/cli/pull/3245 but I still want to think about it a bit more. I'm curious what you think about the approach I have or if you had a better idea in mind

ofekatr (approved):
Phenomenal work @teodora-sandu. 
Thanks for driving this PR through oceans, volcanos, and terminals. LGTM :shipit: 

Alt Text

PR #3226, opened 2022-05-12
refactor: remove logic for --fail-on from iac test

francescomari:

#### What does this PR do?

This PR removes logic related to the `--fail-on` flag from the `iac test` command. The command does not support or use the `--fail-on` flag.

Only inactive code has been removed. The functionality should not be altered. Test this PR as you like.


ipapast (approved):
This is indeed in our list of unsupported flags, so approval removed.
It would throw an error anyway if we tried to pass it.

`Unsupported flag "--fail-on" provided. Run snyk iac test --help for supported flags`

PR #3227, opened 2022-05-12
refactor: remove inactive code from iac test

francescomari:
This PR removes some inactive code from the `iac test` command:

- Support for `snyk protect`, which has been deprecated in favour of `@snyk/protect` and is not relevant to the `iac test` command anyway.
- Support for detecting ecosystems, only relevant for the `code test` command.
- Argument validation for the `docker test` command, which is an alias for `container test`.


teodora-sandu:
I'm all up for removing unnecessary code. Do we want to also remove the following https://github.com/snyk/cli/blob/master/src/cli/commands/test/iac/index.ts#L360-L364.

francescomari:
Nice catch! I will remove it as part of this PR.

teodora-sandu (approved):
LGTM

PR #3229, opened 2022-05-12
chore: update go version to 1.18 for generics

maxjeffos:
Update Go version to 1.18 to support generics

### DIFF
--- a/.circleci/chocolatey.config
+++ b/.circleci/chocolatey.config
@@ -5,6 +5,6 @@
     <package id="sbt" version="1.5.5" />
     <package id="awscli" version="2.4.12" />
     <package id="nodejs" version="16.14.2" />
-    <package id="go" version="1.17.8" />
+    <package id="go" version="1.18.2" />
     <package id="mitmproxy" version="7.0.4" />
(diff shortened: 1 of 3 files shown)


PeterSchafer (approved):

PeterSchafer:
> issue: Also needs updating in:
> 
> * .circleci/chocolatey.config
> * .circleci/config.yml

question: Can we make it a single source of truth somehow? Having multiple places to change a version calls for trouble :)

jahed-snyk (approved):

maxjeffos:
> > issue: Also needs updating in:
> > 
> > * .circleci/chocolatey.config
> > * .circleci/config.yml
> 
> question: Can we make it a single source of truth somehow? Having multiple places to change a version calls for trouble :)

We could set an env var in CircleCI and use that in the CircleCI config. Then we'd have to somehow templateize the chocolatey.config and render it before running. Then somehow we'd have to pass that into the go.mod file which seems not ideomatic to templateize that. Seems like an excessive amount of yak shaving for a small thing. What do you think @PeterSchafer

PR #3231, opened 2022-05-12
fix: ignore errors on multiple paths if there were some results [CFG-1784]

ipapast:
This commit fixes a bug where if we had multiple paths in a scan, and one of them threw an error, the entire scan was failing.
Now we are going to still show success results if there are any, and not fail the entire scan. 
Before we were treating the success and failures as the only possible options, but now I extended it to also have a "partial" success, when both sucess and errors were found.

The bug affected both the legacy output and the new output.

For the legacy we reverted to the old behaviour, showing the error when we scanned it.
For the new output we construct these errors in the end, and there's another ticket in play that does that, so I thought I'll open a seperate PR for that ticket to add it. (CFG-1834, CFG-1836)

P.S. the reason that the error is not bold red is that we don't store this type of error as an `iacScanFailures` during the scan.  A failure is different than a "fatal" error.
https://github.com/snyk/cli/blob/6da59aff54754047270b680f9fc3efed0bfcae80/src/cli/commands/test/iac/index.ts#L133-L136

We catch it later and we don't store it in that array. 
Then, when we format all the failures we don't have this error in there: 
https://github.com/snyk/cli/blob/6da59aff54754047270b680f9fc3efed0bfcae80/src/cli/commands/test/iac/index.ts#L286

CFG-1784

You need to run these scenarios twice, one with the `iacCliOutput` FF on and one with the FF off.

1. Single directory that does not exist: - both outputs
`snyk-dev iac test non-existing-folder` 
image

2. Single file that does not exist: - both outputs
`snyk-dev iac test file-abc.tf`
image

3a. Multiple directories (both successful) - old output
`snyk-dev iac test test/fixtures/iac/kubernetes test/fixtures/iac/terraform/var_deref/nested_var_deref/`
image

3b Multiple directories (both successful) - new output
image

4a. Multiple directories (one fails, order of paths does not matter) - old output
` snyk-dev iac test non-existing-folder test/fixtures/iac/kubernetes`
image
image

4b. Multiple directories (one fails, order of paths does not matter) - new output
image
image

5a. one filepath fails - one directory succeeds - old output
image
image

5b new output - same as 4b. no error reported.

CFG-1784

### DIFF
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -11,7 +11,7 @@ import {
   TestOptions,
 } from '../../../../lib/types';
 import { MethodArgs } from '../../../args';
-import { TestCommandResult } from '../../../commands/types';
+import { TestCommandResult } from '../../types';
 import {
   LegacyVulnApiResult,
   TestResult,
@@ -24,7 +24,11 @@ import {
 } from '../../../../lib/formatters';
 import * as utils from '../utils';
 import {
+  failuresTipOutput,
   formatIacTestFailures,
+  formatIacTestSummary,
+  formatShareResultsOutput,
+  getIacDisplayedIssues,
   getIacDisplayErrorFileOutput,
   iacTestTitle,
   shouldLogUserMessages,
@@ -42,14 +46,11 @@ import { processCommandArgs } from '../../process-command-args';
 import { formatTestError } from '../format-test-error';
 import { displayResult } from '../../../../lib/formatters/test/display-result';
 
-import { isIacShareResultsOptions } from './local-execution/assert-iac-options-flag';
-import { assertIaCOptionsFlags } from './local-execution/assert-iac-options-flag';
-import { hasFeatureFlag } from '../../../../lib/feature-flags';
 import {
-  formatIacTestSummary,
-  formatShareResultsOutput,
-  getIacDisplayedIssues,
-} from '../../../../lib/formatters/iac-output';
+  assertIaCOptionsFlags,
+  isIacShareResultsOptions,
+} from './local-execution/assert-iac-options-flag';
+import { hasFeatureFlag } from '../../../../lib/feature-flags';
 import { initRules } from './local-execution/rules';
 import {
   cleanLocalCache,
@@ -57,7 +58,6 @@ import {
 } from './local-execution/measurable-methods';
 import config from '../../../../lib/config';
 import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-entitlement-error';
-import { failuresTipOutput } from '../../../../lib/formatters/iac-output';
 import * as ora from 'ora';
 
 const debug = Debug('snyk-test');
@@ -124,7 +124,6 @@ export default async function(
           iacOrgSettings,
           rulesOrigin,
         );
-
         iacOutputMeta = {
           orgName: results[0]?.org,
           projectName: results[0]?.projectName,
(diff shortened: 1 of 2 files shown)


src/cli/commands/test/iac/index.ts
```diff
@@ -137,19 +136,25 @@ export default async function(
       } catch (error) {
         // not throwing here but instead returning error response
         // for legacy flow reasons.
+        // TODO: Check if we really need this anymore.
+        //  All errors that we throw are of Error type, so we shouldn't be needing this check/conversion.
```
ipapast, line 140: talking to myself to remind of the investigation, not blocking

src/cli/commands/test/iac/index.ts
```diff
@@ -170,21 +175,33 @@ export default async function(
...
-      (res.result &&
-        res.result.cloudConfigResults &&
-        res.result.cloudConfigResults.length),
+  const [successResults, errorResults] = results.reduce(
+    ([successResults, errorResults], result) => {
+      return !(result instanceof Error)
+        ? [[...successResults, result], errorResults]
+        : [successResults, [...errorResults, result]];
+    },
+    [[], []],
```
ipapast, line 184: I split the array into two: successful results and error results to be able to use them later on.
if the reduce is not readable enough, I can convert it a for loop or use lodash.partition in a next PR (I was thinking to do it in a new PR to not trigger another team's review. wdyt?)

src/cli/commands/test/iac/index.ts
```diff
@@ -170,21 +174,33 @@ export default async function(
...
+        (res.vulnerabilities && res.vulnerabilities.length) ||
+        (res.result &&
+          res.result.cloudConfigResults &&
+          res.result.cloudConfigResults.length),
+    );
+  }
+  const hasErrors = errorResults.length;
+  const isPartialSuccess = errorResults.length && successResults.length;
+  const foundVulnerabilities = vulnerableResults.length;
+
```
ipapast, line 199: previously it was all black or white: a scan would either succeed or fail. I introduced the concept of partial success where you can both successes and (non-fatal) errors.  I think we can improve this more but taking it one step at a time

src/cli/commands/test/iac/index.ts
```diff
@@ -24,7 +24,11 @@ import {
 } from '../../../../lib/formatters';
 import * as utils from '../utils';
 import {
+  failuresTipOutput,
   formatIacTestFailures,
+  formatIacTestSummary,
+  formatShareResultsOutput,
+  getIacDisplayedIssues,
```
ipapast, line 31: sorry about this, I think it's my IDE formatting repositioning. it shouldn't be any new imports

src/cli/commands/test/iac/index.ts
teodora-sandu: I dislike reduces anyway, but as far as reduces go this isn't too bad so will leave it up to you to decide.

src/cli/commands/test/iac/index.ts
```diff
@@ -170,21 +174,33 @@ export default async function(
...
   );
-  const errorResults = results.filter((res) => res instanceof Error);
-  const notSuccess = errorResults.length > 0;
-  const foundVulnerabilities = vulnerableResults.length > 0;
 
-  if (notSuccess) {
-    testSpinner?.fail(spinnerFailureMessage + EOL);
-  } else {
+  let vulnerableResults;
+  if (successResults) {
```
teodora-sandu, line 187: Two questions:
1. Why are we still using `results` instead of `successResults` on line 188?
2. If we can use `successResults`, then do we need the if statement on line 187? I think `successResults` will either be an empty array or a populated array so we don't need to check

teodora-sandu:
> P.S. the reason that the error is not bold red is that we don't store this type of error as an iacScanFailures during the scan.

I might be fixing this in the following PR:
https://github.com/snyk/cli/pull/3225

src/cli/commands/test/iac/index.ts
teodora-sandu: We use `(!hasErrors || isPartialSuccess)` in a lot of places, would it be better to say here:
```
const isPartialSuccess = !hasErrors || successResults.length;
```
and use that directly everywhere else in the code? Just because there's a lot of if statements already and I don't want to make it even harder to read

teodora-sandu (approved):
I've tested it with the cases you mentioned and with `--report` and `--json` too and (like I expected) only the files with the issues are included. I've left some comments about how to refactor in some small places but approving this 🎉

src/cli/commands/test/iac/index.ts
ipapast: yeah, much better thank you!

PR #3233, opened 2022-05-13
fix: use correct auth header for api requests

jahed-snyk:
This is an immediate fix to ensure we use the correct header based on if SNYK_OAUTH_TOKEN is used.
We'll look into improving the auth experience in CLI in the near future.


gitphill (approved):

ipapast (approved):
looks good, thank you

novalex (approved):

PR #3239, opened 2022-05-13
feat: include thrown errors in IaC failures section [CFG-1834]

ipapast:

#### What does this PR do?

We fixed the bug where a multiple paths' scan would fail entirely, if one of the paths failed (https://github.com/snyk/cli/pull/3231).
As noted in that PR, the thrown errors were shown in the legacy output but not in the new one. 

So this PR adds all errors in the `Test Failures` summary, both failures from the scan that did not throw, or errors that were thrown and we caught.

This will also unlock the next piece of work which is to add a count of paths' error in the summary at the bottom.

Also tested with `--json`, `--report` flags.
`--sarif` flag fails, this was the case in many old version of the CLI as well. I will need to investigate more and open another PR for that seperately.

Code is open to refactoring as it gets bigger, I think I might to extract it when I try to introduce the failure counts next.

- Turn the `iacCliOutput` FF on.

1. Multiple directories (one fails, one succeeds order of paths does not matter):
` snyk-dev iac test non-existing-f test/fixtures/iac/kubernetes`
image

2. one filepath fails - one directory succeeds
image

3. Multiple paths - more than one errors, but some successful results
image

4. Multiple paths with the same error:
image

Other scenarios, although there are no changes:

6. Single directory that does not exist (no change):
`snyk-dev iac test non-existing-folder` 
image

7. Single file that does not exist (no change):
`snyk-dev iac test file-abc.tf`
image

8. Multiple directories (both successful - no change):
`snyk-dev iac test test/fixtures/iac/kubernetes test/fixtures/iac/terraform/var_deref/nested_var_deref/`
image

### DIFF
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -58,6 +58,7 @@ import {
 import config from '../../../../lib/config';
 import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-entitlement-error';
 import * as ora from 'ora';
+import { IaCTestFailure } from '../../../../lib/formatters/iac-output/v2/types';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
@@ -81,7 +82,7 @@ export default async function(
   const results = [] as any[];
 
   // Holds an array of scanned file metadata for output.
-  let iacScanFailures: IacFileInDirectory[] | undefined;
+  let iacScanFailures: IacFileInDirectory[] = [];
   let iacIgnoredIssuesCount = 0;
   let iacOutputMeta: IacOutputMeta | undefined;
 
@@ -130,20 +131,16 @@ export default async function(
         };
 
         res = results;
-        iacScanFailures = failures;
+        iacScanFailures = failures ? failures : [];
         iacIgnoredIssuesCount += ignoreCount;
       } catch (error) {
-        // TODO: Check if we really need this anymore.
-        //  All errors that we throw are of Error type, so we shouldn't be needing this check/conversion.
         res = formatTestError(error);
-        // we can store all errors here for printing the paths and errors later in the new output.
       }
 
       // Not all test results are arrays in order to be backwards compatible
       // with scripts that use a callback with test. Coerce results/errors to be arrays
       // and add the result options to each to be displayed
       // TODO: Similarly to above, do we actually need to convert this to an array?
-      // I think we do not need this in IaC
       const resArray: any[] = Array.isArray(res) ? res : [res];
 
       for (let i = 0; i < resArray.length; i++) {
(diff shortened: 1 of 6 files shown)


src/lib/formatters/iac-output/v2/types.ts
```diff
@@ -6,3 +6,8 @@ export interface IacTestData {
   results: IacTestResponse[];
   failures?: IacFileInDirectory[];
 }
+
+export type FormattedFailure = {
```
ipapast, line 10: suggestions for names welcome.
I thought of Failure, FailureInfo, FailureMeta, BasicFailure but left it to the "formatted" prefix

src/cli/commands/test/iac/index.ts
```diff
@@ -131,20 +132,16 @@ export default async function(
         };
 
         res = results;
-        iacScanFailures = failures;
+        iacScanFailures = failures ? failures : [];
         iacIgnoredIssuesCount += ignoreCount;
       } catch (error) {
-        // TODO: Check if we really need this anymore.
```
teodora-sandu, line 137: I see you removed these comments. Did you figure out why this would be needed?

src/lib/formatters/iac-output/v2/types.ts
teodora-sandu: Is it a bit confusing for it to have "formatted' when it goes into `formatIacTestFailures`, which is meant to format it? An alternative would be to change the name to `IacTestFailure` maybe. WDYT?

On the other hand, the `IacFileInDirectory` type we use for `iacFailures` is weird as well. I'm wondering if we can rename it (maybe to `IacTestFailure`) and also see if `failureReason` can be made required and `fileType` optional, so we don't need to do all of the mapping in `src/cli/commands/test/iac/index.ts`. As far as I can tell this type is only used for failures but the name is misleading and the type definition might be legacy. Just food for thought though.

teodora-sandu:
@ipapast Not related to your changes here because this is what we already do but do you think the `--json` output should contain the error?
Screenshot 2022-05-17 at 11 29 26

ipapast:
Putting this on draft as it touches CFG-1834 and I need to see again the definition of done.

src/cli/commands/test/iac/index.ts
ipapast: yes. 99% not needed as all our errors are of type CustomError. however I didn't intend to leave those comments in the code the first time I added them, it was note to myself. that's why I removed them

src/lib/formatters/iac-output/v2/types.ts
ipapast: I think all that this variable is, is a Scan/Test Faliure. I don't see why we need to prefix things with IaC when we are in files of our domain entirely. In theory, no one else should use these types in their files. But I'm not that fussed about it.

The IaCFileInDirectory is used in a few places and it's also extended by the IaCFileData which we also in a few places. 
https://github.com/snyk/cli/blob/990591069d5f8a9396fa8c6e140e2b1266c9f5ce/src/cli/commands/test/iac/local-execution/types.ts#L19
If we use it, and we make the `fileType` required, we will just have to use a `non null assertion` operator in a few places. I just wanted to avoid all of this tbh. I pushed a new commit to demo how it looks and decide what's better.

ipapast:
> @ipapast Not related to your changes here because this is what we already do but do you think the `--json` output should contain the error? Screenshot 2022-05-17 at 11 29 26

not sure about it. I think we use this json structure "string": "string" widespread across the product. we need to check where/how it is used and give it some thought

src/lib/formatters/iac-output/v2/types.ts
ipapast: I talked with @teodora-sandu and we decided going with the original approach of having a different type for failures as the optional parameters force us to use a lot of non null assertion operators in other places.

teodora-sandu (approved):

PR #3249, opened 2022-05-18
chore(ci): split versioning from build

jahed-snyk:
Build job needs to be split so that we can parallelise it, but we only want to version bump once.

This change introduces a Makefile for CLIv1 which will gradually grow to include more build steps.

### DIFF
--- a/release-scripts/make-binaries.sh
+++ b/release-scripts/make-binaries.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-mkdir binary-releases
+# Do not run this file locally. To build release artifacts, see CONTRIBUTING.
 
 mv "$(npm pack --workspace '@snyk/fix')" binary-releases/snyk-fix.tgz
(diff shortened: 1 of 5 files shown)


release-scripts/make-binaries.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-mkdir binary-releases
+# Do not run this file locally. To build release artifacts, see CONTRIBUTING.
```
maxjeffos, line 4: suggestion: might be nice to have a fail-safe mechanism so that this script fails if it is run locally. Maybe look for the presence of one of the standard CircleCI vars?

maxjeffos (approved):

release-scripts/make-binaries.sh
jahed-snyk: This script will soon be replaced as part of splitting builds so I'll leave it as it is for now to avoid too much reshuffling.

PR #3254, opened 2022-05-19
fix: unmanaged cancelled jobs

danlucian:

#### What does this PR do?

- throw an error when the unmanaged c++ jobs are cancelled or contain error

### DIFF
--- a/src/lib/polling/polling-test.ts
+++ b/src/lib/polling/polling-test.ts
@@ -8,7 +8,7 @@ import { getAuthHeader } from '../api-token';
 import { ScanResult } from '../ecosystems/types';
 
 import { ResolveAndTestFactsResponse } from './types';
-import { delayNextStep } from './common';
+import { delayNextStep, handleProcessingStatus } from './common';
 import { TestDependenciesResult } from '../snyk-test/legacy';
 
 export async function requestTestPollingToken(
@@ -53,6 +53,8 @@ export async function pollingTestWithTokenUntilDone(
   };
 
   const response = await makeRequest<ResolveAndTestFactsResponse>(payload);
+
+  handleProcessingStatus(response);
 
   if (response.result) {
     const {
--- a/test/jest/unit/lib/ecosystems/resolve-test-facts.spec.ts
+++ b/test/jest/unit/lib/ecosystems/resolve-test-facts.spec.ts
@@ -1,5 +1,6 @@
 import { Options } from '../../../../../src/lib/types';
 import * as pollingTest from '../../../../../src/lib/polling/polling-test';
+import * as promise from '../../../../../src/lib/request/promise';
 import { depGraphData, scanResults } from './fixtures/';
 import { resolveAndTestFacts } from '../../../../../src/lib/ecosystems/resolve-test-facts';
 import * as pluginAnalytics from '../../../../../src/lib/ecosystems/plugin-analytics';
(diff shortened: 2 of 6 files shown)


test/jest/unit/lib/ecosystems/resolve-test-facts.spec.ts
```diff
@@ -50,6 +51,42 @@ describe('resolve and test facts', () => {
...
+      });
+
+      const [testResults, errors] = await resolveAndTestFacts(
+        'cpp',
+        scanResults,
+        {} as Options,
+      );
+
+      expect(testResults).toEqual([]);
+      expect(errors[0]).toContain(expected);
```
dekelund, line 86: Don't know if it's needed in TS for unit tests in jests, but do we want to check the length first? or do we want to rely on the exception being catched by the test framework?

src/lib/polling/polling-test.ts
```diff
@@ -54,6 +55,12 @@ export async function pollingTestWithTokenUntilDone(
 
   const response = await makeRequest<ResolveAndTestFactsResponse>(payload);
 
+  if (response?.status === 'CANCELLED' || response?.status === 'ERROR') {
```
dekelund, line 58: I assume these are the only available statuses that we lacked coverage for?
Maybe we should have yet another test after the if-statement to check if it's ok (or not ok), and then fail with a message indicating that we received unexpected behaviour

dekelund (approved):
Some comments, but it doesn't block the PR itself.

src/lib/polling/polling-test.ts
danlucian: We lacked error handling only. This is just an if guard to short circuit the chain in case of a `CANCELLED` or `ERRORED` processing.

Once you have the response you don't get an 'OK' status code but the results directly. 
Having results, without an 'OK' status, is considered to be enough for the happy flow based on the [current implementation](https://github.com/snyk/cli/blob/master/src/lib/polling/polling-test.ts#L57).

dekelund: So either we increase the code and make it more complex to cover future changes, or we keep it as it is, but it's more brittle if something change.

test/jest/unit/lib/ecosystems/resolve-test-facts.spec.ts
dekelund: yea I understand what it does, I was more curious if something change, how we want it to behave. Let us say that the unit test will fail (due to unintended changes in the code), and the errors is empty. Do we want to cover those failures by catching "out of bounds"-exception or by checking length. But let's keep it like this, if that's how we have written the other tests.

src/lib/polling/polling-test.ts
jahed-snyk, line 58: suggestion: Could the API endpoint respond with a non-200 and a server-side error message instead?

https://github.com/snyk/cli/blob/a041d7d8588f85984e8add270c58f1a169a3ae96/src/lib/request/promise.ts#L10-L15

This will make it more flexible, especially for exiting CLI installations and won't need CLI releases and updates for future changes.

jahed-snyk (changes requested):

src/lib/polling/polling-test.ts
danlucian: Unfortunately not, we don't want to break the actual API definition. The implementation is doing just some heartbeating and the body of the response will tell us if a job was canceled or not (by an upstream service that we use).

src/lib/polling/polling-test.ts
```diff
@@ -54,6 +55,12 @@ export async function pollingTestWithTokenUntilDone(
...
+    throw new FailedToRunTestError(
+      'We were unable to run the test. Please run the command again with the `-d` flag and contact support@snyk.io with the contents of the output',
+    );
+  }
```
jahed-snyk, line 62: suggestion: Could we use a more descriptive error here? Is there one available in the response? IIRC API failures usually come with an error ID which can be used for tracing.

jahed-snyk (approved):

src/lib/polling/polling-test.ts
danlucian: That's a good point! I will try to embed the trace ID as well. Thanks!

PR #3257, opened 2022-05-20
chore(ci): migrate artifact builds to makefile

jahed-snyk:
We want to parallelise the "Build Artifacts" job. Moving each artifact's build to its own `make` target will let us easily run them individually while also keeping definitions explicit and discoverable through a single entrypoint: the Makefile.

I've split Makefile migrations into separate commits to make reviews easier.

More information: https://www.notion.so/snyk/CLI-Speed-Up-Time-to-Build-in-CI-0d7ce418ecd14fa0bd8fac1ac391eea1


Makefile
```diff
@@ -42,3 +55,58 @@ binary-releases/release.json: binary-releases/version $(wildcard binary-releases
...
+#   https://github.com/vercel/pkg#targets
+binary-releases/snyk-linux-arm64: prepack | binary-releases
+	$(PKG) -t node16-linux-arm64 -o binary-releases/snyk-linux-arm64 --no-bytecode
+	$(MAKE) binary-releases/snyk-linux-arm64.sha256
+
+binary-releases/snyk-macos: prepack | binary-releases
+	$(PKG) -t node16-macos-x64 -o binary-releases/snyk-macos
+	$(MAKE) binary-releases/snyk-macos.sha256
+
+binary-releases/snyk-win.exe: prepack | binary-releases
```
PeterSchafer, line 96: question: I'm wondering whether it would be reasonable to split out the signing as a separate target? 
I created 2 targets for cliv2 and the main reason for me was, that I wanted to enable developers to also be able to build an exe even if they don't have access to signing secrets. What do you think?

PeterSchafer, line 96: question: I'm wondering whether it would be reasonable to split out the signing as a separate target? 
I created 2 targets for cliv2 and the main reason for me was, that I wanted to enable developers to also be able to build an exe even if they don't have access to signing secrets. What do you think?

Makefile
```diff
@@ -5,7 +5,13 @@
 # Documentation: https://www.gnu.org/software/make/manual/make.html
 #
 
-# First target is default when running `make`.
+PKG := npx pkg ./
+# compression takes around 3 minutes, so only do it on master
+ifeq ($(CIRCLE_BRANCH), master)
```
PeterSchafer, line 10: suggestion: instead of relying on an implicit environment Variable, this could be done via an explicit parameter. A makefile should be worried about the CI system and its variable but just offer a features in its interface (the cmd args)

PeterSchafer, line 10: suggestion: instead of relying on an implicit environment Variable, this could be done via an explicit parameter. A makefile should be worried about the CI system and its variable but just offer a features in its interface (the cmd args)

Makefile
jahed-snyk: Agreed, in a future PR. I made a task in our cycle board to refactor signing steps to be similar to CLIv2 to avoid any divergence.

Makefile
jahed-snyk: Agreed. I'll rollback this commit to focus this PR and introduce another PR for this build time saving.

Makefile
jahed-snyk: Resolving as this comment is duplicate of https://github.com/snyk/cli/pull/3257#discussion_r879283380

Makefile
jahed-snyk: Resolving as this comment is a duplicate of https://github.com/snyk/cli/pull/3257#discussion_r879293279

PeterSchafer (approved):

PR #3259, opened 2022-05-22
chore: Add invalid paths count to test summary [CFG-1836]

ofekatr:

#### What does this PR do?

- Adds the `Invalid paths` count to the test summary section.

```
src/lib/formatters/iac-output/v2/test-summary.ts
```

- Run `snyk-dev iac test` with multiple paths, some of which should throw an error, some of which should include invalid files.
- Assert that the test summary section includes the correct values for the `Invalid paths` and `Invalid files` counts.

In [the new test output design](https://www.figma.com/file/buNFZqr1Uhf4SaifHw4WXE/CLI-IaC-Output-Improvements?node-id=0%3A1), we are currently displaying the number of files that failed to be tested in the test summary section, but we don’t display the number of paths that failed to be displayed.
We would like to add it to the test summary section as well.

- [CFG-1836](https://snyksec.atlassian.net/browse/CFG-1836)

#### Screenshots

image


YairZ101:
I have a few concerns:
1. It is not super clear what the `Invalid paths` counter is counting, for example: in the image you shared, which 2 paths are invalid?
2. The `Invalid files` counter shows 6 invalid files but in the errors list we can see 7 failed files.

I think our main focus should be to align on the terminology we are using and in addition to that which information is critical to show to the user.

PR #3260, opened 2022-05-23
chore: Implement new test failure message [CFG-1861]

ofekatr:

#### What does this PR do?

- Implements the new test failure message format, according to [this design](https://www.figma.com/file/buNFZqr1Uhf4SaifHw4WXE/Snyk-IaC-CLI?node-id=1684%3A15297).

```
src/cli/commands/test/iac/index.ts
```

#### Notes for the reviewer

- This review includes a new class called `FormattedCustomError`, which extends `CustomError`. This class can shift user message formatting from the CLI's error handling middleware, to the error class. The class adds a boolean flag called `isFormatted`, which is set to `true`. This indicates to the error handling middleware whether or not to format the user message.
- The change above was inspired by [this PR comment](https://github.com/snyk/cli/pull/3218#discussion_r868306574).
- For errors that are related to the flow, and not particular paths, such as initializing the local cache, fetching/initializing custom rules, fetching org settings, etc., the error format does not change.

1. Run `SNYK_IAC_OUTPUT_V2=true snyk iac test` with one or more paths that cannot complete the test successfully or have no files that were successfully scanned. 
2. Ensure the test failure message is aligned with the design and that the values are correct.
3. Run `SNYK_IAC_OUTPUT_V2=true snyk iac test` with one or more paths, some of which can be paths for which the test cannot be completed successfully.
4. Ensure the output was not changed for these flows.

In the context of this ticket, test failures are considered to be the cases in which there are no valid files which were successfully scanned, and the test only resulted in failures. This includes but isn’t limited to:

- Given a path to a file, failing to test this file.
- Given a path to a directory, failing to test all files in this directory.
- Given a path to a directory with no files to test.
- Failing to set up the environment for the test (e.g., downloading the security rules bundle, setting up the local cache, fetching the org settings, fetching remote custom rules, using invalid custom rules, etc.)
- Given multiple paths of files/dirs, with all of them resulting in failures, either path-based or file-based.

When the entire test flow fails, in the original output format, we’d only display the failure reason of the first failing path, with no further output. We’d like to modify that and adapt it to the design of the new output format.

- [CFG-1861](https://snyksec.atlassian.net/browse/CFG-1861)

#### Screenshots

image

### DIFF
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -26,6 +26,7 @@ import * as utils from '../utils';
 import {
   failuresTipOutput,
   formatIacTestFailures,
+  formatFailuresList,
   formatIacTestSummary,
   formatShareResultsOutput,
   getIacDisplayedIssues,
@@ -34,6 +35,7 @@ import {
   shouldLogUserMessages,
   spinnerMessage,
   spinnerSuccessMessage,
+  IaCTestFailure,
 } from '../../../../lib/formatters/iac-output';
 import { extractDataToSendFromResults } from '../../../../lib/formatters/test/format-test-results';
 
@@ -58,7 +60,7 @@ import {
 import config from '../../../../lib/config';
 import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-entitlement-error';
 import * as ora from 'ora';
-import { IaCTestFailure } from '../../../../lib/formatters/iac-output/v2/types';
+import { CustomError, FormattedCustomError } from '../../../../lib/errors';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
@@ -243,8 +245,10 @@ export default async function(
 
   let response = '';
 
-  if (isNewIacOutputSupported && isPartialSuccess) {
-    response += EOL + getIacDisplayedIssues(successResults, iacOutputMeta!);
+  if (isNewIacOutputSupported) {
+    if (isPartialSuccess) {
+      response += EOL + getIacDisplayedIssues(successResults, iacOutputMeta!);
+    }
   } else {
     response += results
       .map((result, i) => {
@@ -285,6 +289,27 @@ export default async function(
       }))
       .concat(thrownErrors);
 
+    if (hasErrors && !isPartialSuccess) {
+      response += chalk.bold.red(summaryMessage);
+
+      // take the code of the first problem to go through error
+      // translation
+      // HACK as there can be different errors, and we pass only the
+      // first one
+      const error: CustomError =
+        isNewIacOutputSupported && allTestFailures
+          ? new FormattedCustomError(
+              errorResults[0].message,
+              formatFailuresList(allTestFailures),
+            )
+          : new CustomError(response);
+      error.code = errorResults[0].code;
+      error.userMessage = errorResults[0].userMessage;
+      error.strCode = errorResults[0].strCode;
+
+      throw error;
+    }
+
     response += isNewIacOutputSupported
       ? EOL.repeat(2) + formatIacTestFailures(allTestFailures)
       : iacScanFailures
@@ -318,19 +343,6 @@ export default async function(
         summariseErrorResults(errorResultsLength) +
         '\n';
     }
-  }
-
-  if (hasErrors && !isPartialSuccess) {
-    response += chalk.bold.red(summaryMessage);
-    const error = new Error(response) as any;
-    // take the code of the first problem to go through error
-    // translation
-    // HACK as there can be different errors, and we pass only the
-    // first one
-    error.code = errorResults[0].code;
-    error.userMessage = errorResults[0].userMessage;
-    error.strCode = errorResults[0].strCode;
-    throw error;
   }
 
   if (foundVulnerabilities) {
--- a/src/lib/errors/legacy-errors.js
+++ b/src/lib/errors/legacy-errors.js
@@ -3,6 +3,7 @@ const chalk = require('chalk');
 const { SEVERITIES } = require('../snyk-test/common');
 const { errorMessageWithRetry } = require('./error-with-retry');
 const analytics = require('../analytics');
+const { FormattedCustomError } = require('./formatted-custom-error');
 
 const errors = {
   connect: 'Check your network connection, failed to connect to Snyk API',
@@ -86,8 +87,12 @@ module.exports.message = function(error) {
       codes[error.code || error.message] ||
       errors[error.code || error.message];
     if (message) {
-      message = message.replace(/(%s)/g, error.message).trim();
-      message = chalk.bold.red(message);
+      if (error instanceof FormattedCustomError) {
+        message = error.formattedUserMessage;
+      } else {
+        message = message.replace(/(%s)/g, error.message).trim();
+        message = chalk.bold.red(message);
+      }
     } else if (error.code) {
       // means it's a code error
       message =
--- a/src/lib/formatters/iac-output/v2/failures/list.ts
+++ b/src/lib/formatters/iac-output/v2/failures/list.ts
@@ -37,7 +37,7 @@ function groupTestFailuresByFailureReason(
   }, {});
 }
 
-function formatFailuresList(testFailures: IaCTestFailure[]) {
+export function formatFailuresList(testFailures: IaCTestFailure[]): string {
   const testFailuresByReason = groupTestFailuresByFailureReason(testFailures);
   return Object.entries(testFailuresByReason)
     .map(([failureReason, testFailures]) =>
--- a/src/lib/errors/formatted-custom-error.ts
+++ b/src/lib/errors/formatted-custom-error.ts
@@ -1,0 +1,16 @@
+import chalk from 'chalk';
+import { CustomError } from '.';
+
+export class FormattedCustomError extends CustomError {
+  public formattedUserMessage: string;
+
+  constructor(
+    message: string,
+    formattedUserMessage: string,
+    userMessage?: string,
+  ) {
+    super(message);
+    this.userMessage = userMessage || chalk.reset(formattedUserMessage);
+    this.formattedUserMessage = formattedUserMessage;
+  }
+}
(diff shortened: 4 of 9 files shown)


robcresswell (approved):

src/cli/commands/test/iac/index.ts
```diff
@@ -285,6 +291,23 @@ export default async function(
       }))
       .concat(thrownErrors);
 
+    if (hasErrors && !isPartialSuccess) {
+      if (isNewIacOutputSupported && allTestFailures) {
+        throw new IacTestFailureError(allTestFailures);
+      } else {
+        response += chalk.bold.red(summaryMessage);
```
jahed-snyk, line 298: thought: I'm not sure if we want the `error.message` itself to be formatted. Typically `error.message` is plain text, and it may be consumed by various error-consuming APIs which aren't aware of `isFormatted`.

Which is why I previously suggested introducing a separate `error.getFormattedMessage` or `error.formattedMessage` property. This way the concept of a formatted message is only visible to those that render errors to the terminal.

Not blocking on this as we can iterate on it in the future if problems arise.

jahed-snyk (approved):

src/cli/commands/test/iac/index.ts
ofekatr: @jahed-snyk I've attempted at updating the implementation to be more generic, and apply your suggestions.

In my solution:
- `FormattableCustomError` has a `formattedUserMessage` getter.
- `FormattableCustomError` has a private singleton member has a singleton member, `_formattedUserMessage` to which the formatted user message is assigned to on the first time the `formattedUserMessage` getter is called.
- `FormattableCustomError` has a protected `formatUserMessage` member, which is a function that defines the formatting logic. This can be overridden by the inheriting error classes, for them to define their formatting logic.

src/lib/errors/formattable-custom-error.ts
```diff
@@ -0,0 +1,27 @@
...
+  constructor(message: string) {
+    super(message);
+
+    this.formatUserMessage = () => {
+      let userMessage = this.userMessage || this.message;
+      userMessage = userMessage.replace(/(%s)/g, userMessage).trim();
+      userMessage = chalk.bold.red(userMessage);
+
+      return userMessage;
+    };
```
jahed-snyk, line 18: suggestion: Does anything use this default implementation? If not, should we remove it? 
We can introduce it later when it's needed.

If we remove it, we won't need the getter/setter and caching logic for `formattedUserMessage` which would reduce a lot of complexity here.

ofekatr: @jahed-snyk great question, I've added this default implementation to prevent people from creating instances for this parent class, and by doing so accidentally bypassing the formatting stage in the error handling middleware of the CLI.

That being said, if they don't define the `formattedUserMessage` than the CLI will fall back to the original formatting so it should be fine

ipapast (commented):
First of all, some context for anyone else seeing this in the future: I reached out to Ofek in DM and we clarified what's the expected functionality: 

- this adds functionality for one more scenario to the existing failure outputs. the new scenario is: user runs a scan with one or more paths and the entire scan fails. 

We will now display entire scan errors as a list (formatted). However, if these errors are "fatal" such as downloading rules from S3, we will keep the "old" generic error message when we fail and won't use the new formatted errors.

src/cli/commands/test/iac/index.ts
```diff
@@ -285,6 +291,23 @@ export default async function(
       }))
       .concat(thrownErrors);
 
+    if (hasErrors && !isPartialSuccess) {
+      if (isNewIacOutputSupported && allTestFailures) {
+        throw new IacTestFailureError(allTestFailures);
```
ipapast, line 296: the changes introduced in this PR mean that we will now wrap almost all our errors (except the fatal ones) in a `IacTestFailureError`, with its own code error and a message. 

- Have we thought/talked about it with product of this potentially breaking the existing analytics, and also making it difficult to track error codes from now on? Even though the message will be included in the general error message, it's still wrapped in a big string so it's more difficult to find it.

an example: 
image

```
"error-message": "Unable to complete the test. Failure reasons: [{\"filePath\":\"blahlah\",\"failureReason\":\"Could not find any valid IaC files\"}]",
    "error": "IacTestFailureError: Unable to complete the test. Failure reasons: [{\"filePath\":\"blahlah\",\"failureReason\":\"Could not find any valid IaC files\"}]\n    at Object.default_1 [as default] (/Users/iliannapapastefanou/snyk-repos/snyk/dist/cli/webpack:/snyk/src/cli/commands/test/iac/index.ts:296:15)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at test (/Users/iliannapapastefanou/snyk-repos/snyk/dist/cli/webpack:/snyk/src/cli/commands/test/index.ts:56:12)\n    at runCommand (/Users/iliannapapastefanou/snyk-repos/snyk/dist/cli/webpack:/snyk/src/cli/main.ts:53:25)\n    at main (/Users/iliannapapastefanou/snyk-repos/snyk/dist/cli/webpack:/snyk/src/cli/main.ts:317:11)\n    at /Users/iliannapapastefanou/snyk-repos/snyk/dist/cli/webpack:/snyk/src/cli/index.ts:11:3\n    at Object.callHandlingUnexpectedErrors (/Users/iliannapapastefanou/snyk-repos/snyk/dist/cli/webpack:/snyk/src/lib/unexpected-error.ts:28:5)",
    "error-code": 1230,
    "error-str-code": "IAC_TEST_FAILURE_ERROR",
```

src/lib/errors/legacy-errors.js
```diff
@@ -86,8 +86,12 @@ module.exports.message = function(error) {
       codes[error.code || error.message] ||
       errors[error.code || error.message];
     if (message) {
-      message = message.replace(/(%s)/g, error.message).trim();
-      message = chalk.bold.red(message);
+      if (error.formattedUserMessage) {
+        message = error.formattedUserMessage;
```
ipapast, line 90: instead of checking by the field existence, we could also check by the type of error here.

src/lib/formatters/iac-output/v2/failures/list.ts
```diff
@@ -37,7 +37,7 @@ function groupTestFailuresByFailureReason(
   }, {});
 }
 
-function formatFailuresList(testFailures: IaCTestFailure[]) {
+export function formatFailuresList(testFailures: IaCTestFailure[]) {
```
ipapast, line 40: add string as a return type

src/cli/commands/test/iac/index.ts
```diff
@@ -379,3 +389,18 @@ export default async function(
...
+        iacTestFailures,
+      )}`,
+    );
+    this.code = IaCErrorCodes.IacTestFailureError;
+    this.strCode = getErrorStringCode(this.code);
+
+    this.formattedUserMessage = formatFailuresList(iacTestFailures);
+    this.userMessage = chalk.reset(this.formattedUserMessage);
+  }
+}
```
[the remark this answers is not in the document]
ipapast: next alternative idea:

let's say that we don't format the error before we throw it. Then, in a similar logic to the above, we remove the need for an extra class just to add a `formattedUserMessage` field, but we define a function inline of this class , eg `getFormattedUserMessage()` and we add the implementation of the `formatFailuresList` inline this `IacTestFailureError` class.

from both, I'd mostly prefer the first as it sounds simpler to format the error outside of the error class, then throw it.

src/lib/formatters/iac-output/v2/failures/list.ts
ofekatr: Fixed in [7ba2beb8](https://github.com/snyk/cli/pull/3260/commits/7ba2beb80595dd6d27d0e377f16d6b67c72b7829)

src/lib/errors/legacy-errors.js
ofekatr: It's been specifically requested by @jahed-snyk from team Hammer, which own the logic in `src/lib/errors/legacy-errors.js`, and I agree with this approach, as it feels like a safer assertion than relying on the class type in this case. This assertion also checks the property has a non-empty string, which ensures a valid message is being printed.

src/cli/commands/test/iac/index.ts
ofekatr: Great observation @ipapast, and this breaking change wasn't fully considered yet, but rather was treated as an implementation detail. 
We could potentially revert the error codes logic here back to the original, omit the distinct error code for `IacTestFailureError`, and instead, we use the first error's code and message in the constructor. 
That way, we get the original behavior around the error codes.

What do you think?

src/lib/errors/legacy-errors.js
ipapast: I have seen the message from @jahed-snyk. I didn't read it as a request but as a suggestion. I'm pretty sure Jahed is open to ideas and suggestions.

The way I see it, you have an if condition;
Whether that condition checks for a `formattedUserMessage` field (which exists only in IaC), or a type of `IacTestFailureError` (which exists only in IaC), it's the same thing. The code is still iac agnostic in my opinion.

src/cli/commands/test/iac/index.ts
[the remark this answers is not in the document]
ipapast: I just suggested an alternative idea to reduce the amount of new errors. I don't see a need for those when we can keep it simpler.

src/cli/commands/test/iac/index.ts
ipapast: happy to keep the original behaviour. we can double check later how we want to improve this

ipapast (approved):

src/lib/errors/legacy-errors.js
jahed-snyk: >I'm pretty sure Jahed is open to ideas and suggestions.

Agreed, I'm not too fussed how we go about this (why I've approved and not blocked) and I don't have full context so I'm mostly providing short-term suggestions to consider either now or in future PRs as needs arise.

It may be useful to directly reference `FormattedCustomError` to show where this `formattedUserMessage` is coming from. Though, I see Ofek's point about empty strings too since `formattedUserMessage` is optional.

We could make `formattedUserMessage` non-optional (a constructor parameter), which would reduce complexity as we won't need the null/undefined check. I think it's safe to assume `formattedUserMessage` is a non-empty string as that point as whoever's using it will need to intentionally pass in an empty string.

ofekatr: @ipapast @jahed-snyk I've updated the implementation based on the suggestions, the `FormattedCustomError` constructor now requires the `formattedUserMessage`, and the condition has been changed to be `error instanceof FormattedCustomError`.

PR #3261, opened 2022-05-23
fix: (iac) last error does not override previous

ipapast:

#### What does this PR do?

Small regression that happened after we merged https://github.com/snyk/cli/pull/3239.
I had a refactor before merging and accidentally changed the line that stores all the failures, resulting in our scan showing just the last path's failures in the output.

I switched it back to store all failures in the for loop, so now we are going to see all of them in the output.
The test validates the behaviour.

for  `snyk-dev iac test test/fixtures/iac/terraform/ test/fixtures/iac/cloudformation/`, we will see all errors for both outputs:
image

image

### DIFF
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -132,7 +132,7 @@ export default async function(
         };
 
         res = results;
-        iacScanFailures = failures ? failures : [];
+        iacScanFailures = [...iacScanFailures, ...(failures || [])];
         iacIgnoredIssuesCount += ignoreCount;
       } catch (error) {
         res = formatTestError(error);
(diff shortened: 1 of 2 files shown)


src/cli/commands/test/iac/index.ts
```diff
@@ -132,7 +132,7 @@ export default async function(
         };
 
         res = results;
-        iacScanFailures = failures ? failures : [];
+        iacScanFailures = [...iacScanFailures, ...(failures || [])];
```
ipapast, line 135: let me know if there's a smarter way to do this, the failures can be of `IacFileInDirectory | undefined` so I needed to guard it somehow

francescomari: I don't think there is a better way to do that 🤷

francescomari (approved):

PR #3262, opened 2022-05-23
refactor: create only one project group per execution

francescomari:

#### What does this PR do?

This PR changes the behaviour of the `iac test` command when multiple paths are passed on the command line. This PR changes the behaviour of the `--report` flag, too. The implemented behaviour is the following:

- `iac test` recursively scans the current working directory by default.
- If one or more paths are passed on the command line, those paths must be relative to the current working directory.
- If one or more paths are passed on the command line, only those files or directories are scanned. If a path is a directory, it will be recursively scanned.
- If the `--report` flag is used, the `iac test` command creates one project group in the platform named after the name of the current working directory.
- If the current working directory is a Git repository, additional information will be extracted and attached to the projects and the project group.

This PR extracts the scan logic into [src/cli/commands/test/iac/scan.ts](./src/cli/commands/test/iac/scan.ts). You can start the review fro there.

Move anywhere on the file system and run `snyk iac test ..`. The command will fail with an error, because the only provided path is outside of the current working directory.

```
$ snyk iac test ..

Testing .....

Path ".." is outside the current working directory
```

Move in the root of this repository and run `snyk iac test --report`. This command creates only one project group in the platform, named after this repository. All the project names will be the paths of the IaC files in this repository, relative from the current working directory.

image

Move into the [test/fixtures/iac](./test/fixtures/iac) directory and run `snyk iac test --report`. The command creates only one project group named `iac`, because the current working directory is not a Git repository. All the project names are relative from the current working directory.

image

This behaviour has been discussed in [this document](https://docs.google.com/document/d/1ZzsYEKEtJsz79ZdzfywuBQ7sJt4LCcvqjehLdVL6ldY/edit?usp=sharing). The document contains additional links to interesting conversations that shaped the implementation in this PR.

[CFG-1840](https://snyksec.atlassian.net/browse/CFG-1840)

### DIFF
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -1,28 +1,16 @@
 import * as Debug from 'debug';
 import { EOL } from 'os';
-import * as cloneDeep from 'lodash.clonedeep';
-import * as assign from 'lodash.assign';
 import chalk from 'chalk';
 
-import {
-  IacFileInDirectory,
-  IacOutputMeta,
-  Options,
-  TestOptions,
-} from '../../../../lib/types';
 import { MethodArgs } from '../../../args';
 import { TestCommandResult } from '../../types';
-import {
-  LegacyVulnApiResult,
-  TestResult,
-} from '../../../../lib/snyk-test/legacy';
+import { LegacyVulnApiResult } from '../../../../lib/snyk-test/legacy';
 import { mapIacTestResult } from '../../../../lib/snyk-test/iac-test-result';
 
 import {
   summariseErrorResults,
   summariseVulnerableResults,
 } from '../../../../lib/formatters';
-import * as utils from '../utils';
 import {
   failuresTipOutput,
   formatIacTestFailures,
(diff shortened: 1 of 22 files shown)


src/cli/commands/test/iac/scan.ts
```diff
@@ -0,0 +1,127 @@
...
+import { IacOrgSettings } from './local-execution/types';
+import * as pathLib from 'path';
+
+export async function scan(
+  iacOrgSettings: IacOrgSettings,
+  options: any,
+  testSpinner: ora.Ora | undefined,
+  paths: string[],
+  orgPublicId: string,
+  projectRoot: string,
```
ipapast, line 31: something to think for refactoring later:

we could potentially combine some of these to reduce the number of params. I'm just thinking it's going to be easier to read than having 6 params.
eg: orgPublicId + iacOrgSettings are org meta, then projectRoot, paths, options are related to the scan arguments, etc

ipapast: sounds good, it was actually very readable when going through the changes.

ipapast (commented):
We had a chat with @francescomari  and Esra regarding the expected design and also some edge cases that fail, such as `snyk iac test . .. --report`. 
Francesco is making some changes and we are going to review again

francescomari:
@ipapast I added a new error class in `scan.ts`. That will need to be updated as soon as we have more details by the content team.

ipapast:
@francescomari I tested different scenarios we discussed on the call, including snyk iac . .. --report and they all look good now.

But I also tried out something more weird, like:
```
cwd=/Users/iliannapapastefanou/snyk-repos/snyk/test/fixtures/iac/terraform

Then run:

snyk-dev iac test . .. ../arm/ var_deref --report
```

You will see a project name appearing as 'var_deref" which is not true. When I check the platform everything is correctly grouped under "snyk"
image

francescomari:
@ipapast that was a really good catch. It should be fixed at 73d30c66, according to the unit tests. Let's see what the acceptance tests say.

src/cli/commands/test/iac/index.ts
```diff
@@ -102,71 +71,20 @@ export default async function(
...
+    iacIgnoredIssuesCount,
+    results,
+    resultOptions,
+  } = await scan(
+    iacOrgSettings,
+    options,
+    testSpinner,
+    paths,
+    orgPublicId,
+    process.cwd(),
```
ofekatr, line 86: suggestion: If we always pass the CWD here, is there a need for it to be included in the parameters?
If so, we could define the CWD as the default value?

francescomari: Done at 5502bf12.

ipapast:
> @ipapast that was a really good catch. It should be fixed at [73d30c6](https://github.com/snyk/cli/commit/73d30c666005f3bafce828191093fd73d7db8e0e), according to the unit tests. Let's see what the acceptance tests say.

they pass now and running it manually looked good 👍  I'll approve and let's see how the product discussion goes.

ipapast (approved):

francescomari:
Given the latest conversations about this feature, the scope of this PR changed. Instead of replacing the current implementation for processing results, the PR adds a new implementation (single project group) in `src/cli/commands/test/iac/local-execution/process-results/v2` and keeps the old one (multiple project groups) as-is in `src/cli/commands/test/iac/local-execution/process-results/v1`. 

The code in `src/cli/commands/test/iac/scan.ts` choses the implementation at runtime depending on the presence of a `projectRoot`. Because no `projectRoot` is currently passed to `scan()`, the new implementation is inactive and the old one is still in use, as before.

I could have been more granular when splitting the two implementations, but I decided not to at the cost of some code duplication. When the new implementation will become the default, removing the old one will be as easy as deleting the `v1` module, removing the class `MultipleGroupsResultsProcessor`, and adjusting `scan()`. I didn't want to introduce granular `if` statements buried deep in the implementation of the result processor. In this case, I believe that some code duplication makes the code easier to maintain.

Some tests are missing for `SingleGroupResultsProcessor`. I will add those in a separate PR.

PR #3282, opened 2022-05-26
fix: maven scan all unmanaged

gitphill:
See https://github.com/snyk/snyk-mvn-plugin/pull/115
Fix package manager name in resulting dep-graph.

Also includes minor dependency bumps:
* https://github.com/snyk/snyk-mvn-plugin/pull/112
* https://github.com/snyk/snyk-mvn-plugin/pull/117
* https://github.com/snyk/snyk-mvn-plugin/pull/113
* https://github.com/snyk/snyk-mvn-plugin/pull/118


package-lock.json
```diff
@@ -16681,23 +16681,41 @@
...
-        "@snyk/cli-interface": "2.11.0",
+        "@snyk/cli-interface": "2.11.3",
         "@snyk/dep-graph": "^1.23.1",
-        "@snyk/java-call-graph-builder": "1.23.3",
+        "@snyk/java-call-graph-builder": "1.23.6",
         "debug": "^4.1.1",
         "glob": "^7.1.6",
         "needle": "^2.5.0",
-        "tslib": "1.11.1"
+        "tmp": "^0.1.0",
```
jahed-snyk, line 16694: issue(blocking): Can we use the latest instead? `^0.2.1` like what java-call-graph-builder, go-plugin and gradle-plugin uses.

I think semver's `^` won't match it on `^0.1.0` as it's on a `0` major, so it causes a duplicate package.

package-lock.json
```diff
@@ -16681,23 +16681,41 @@
...
     "node_modules/snyk-mvn-plugin": {
-      "version": "2.29.1",
-      "resolved": "https://registry.npmjs.org/snyk-mvn-plugin/-/snyk-mvn-plugin-2.29.1.tgz",
-      "integrity": "sha512-wW5uT5ec1/xmfikROkGxv8XDr234m+JCgf5BqWSWo7OEv+xcJag/obOuvu+5LBGn8WfJnRvRAgQ6gndU+oQLHw==",
+      "version": "2.29.6",
+      "resolved": "https://registry.npmjs.org/snyk-mvn-plugin/-/snyk-mvn-plugin-2.29.6.tgz",
+      "integrity": "sha512-dtcfb88EqYR+hhpmhDXF/PDU/aa3CkYyPExY1QPhdA7Vcwk0Cotj+VgAFN0oFwYA8vd3VxmgKjFoPX97fGo+HA==",
       "dependencies": {
-        "@snyk/cli-interface": "2.11.0",
+        "@snyk/cli-interface": "2.11.3",
```
jahed-snyk, line 16688: suggestion: Should we make this a `^` to avoid duplicate interface packages? It seems Python, Maven and Gradle plugins are using fixed versions causing duplicates, the rest are using caret versions.

jahed-snyk (dismissed):
Noticed some duplicate nested dependencies being introduced. I've left some comments related to that.

package-lock.json
gitphill: removing tmp completely here https://github.com/snyk/snyk-mvn-plugin/pull/128

gitphill: ok updated, thanks @jahed-snyk

PR #3290, opened 2022-06-01
chore: add condition to diverge to new flow for iac [CFG-1816]

ipapast:

#### What does this PR do?

This PR adds a divergion from the existing flow to the new flow (TBC) for the UPE integration.

The code currently does nothing. this commit just prepares the directory structure. I added a dummy message for fun :) It can be deleted on the next commit.
I have named the directory v2 but it's open to renaming as we progress with building up this flow.

Also included a simple acceptance test to prove the point of the feature flag and command flag. We can refactor and use this later, when we have actual code. 

#### How to test

follow the same scenarios as the acceptance test.
- with FF on and --experimental
- with FF on , without --experimental
- with FF off, and --experimental
- with FF off, without --experimental
- with typos on the command flag

image

### DIFF
--- a/test/jest/acceptance/iac/experimental.spec.ts
+++ b/test/jest/acceptance/iac/experimental.spec.ts
@@ -1,0 +1,59 @@
+import { startMockServer } from './helpers';
+import { FakeServer } from '../../../acceptance/fake-server';
+
+jest.setTimeout(50_000);
+
+describe('iac --experimental', () => {
+  let server: FakeServer;
+  let run: (
+    cmd: string,
+  ) => Promise<{ stdout: string; stderr: string; exitCode: number }>;
+  let teardown: () => void;
+  const upeFeatureFlag = 'iacCliUnifiedEngine';
+
+  beforeAll(async () => {
+    ({ server, run, teardown } = await startMockServer());
+  });
+
+  afterEach(() => {
+    server.restore();
+  });
+
+  afterAll(async () => teardown());
+
+  describe(`"${upeFeatureFlag}" feature flag is ON`, () => {
+    beforeEach(() => {
+      server.setFeatureFlag(upeFeatureFlag, true);
+      server.setFeatureFlag('iacCliOutputRelease', true);
+    });
+
+    it('diverts to the new flow when --experimental flag is used', async () => {
+      const { exitCode, stdout } = await run(
+        `snyk iac test ./iac/arm/rule_test.json --experimental`,
+      );
+      expect(stdout).not.toContain('Snyk Infrastructure as Code');
+      expect(stdout).not.toContain('Issues');
+      expect(exitCode).toBe(0);
+    });
+
+    it('does not divert to the new flow without --experimental flag', async () => {
+      const { exitCode, stdout } = await run(
+        `snyk iac test ./iac/arm/rule_test.json`,
+      );
+      expect(stdout).toContain('Snyk Infrastructure as Code');
+      expect(stdout).toContain('Issues');
+      expect(exitCode).toBe(1);
+    });
+  });
+
+  describe(`"${upeFeatureFlag}" feature flag is OFF`, () => {
+    it('does not divert to the new flow with just --experimental flag', async () => {
+      server.setFeatureFlag(upeFeatureFlag, false);
+      const { exitCode, stdout } = await run(
+        `snyk iac test ./iac/arm/rule_test.json --experimental`,
+      );
+      expect(stdout).toContain('Unsupported flag "--experimental" provided');
+      expect(exitCode).toBe(2);
+    });
+  });
+});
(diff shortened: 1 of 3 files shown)


test/jest/acceptance/iac/experimental.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { startMockServer } from './helpers';
+import { FakeServer } from '../../../acceptance/fake-server';
+
+jest.setTimeout(50_000);
+
+describe('iac --experimental', () => {
```
YairZ101, line 6: should we split this describe into two sub-describes? one for FF enable and for disable?

test/jest/acceptance/iac/experimental.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    const { exitCode, stdout } = await run(
+      `snyk iac test ./iac/arm/rule_test.json --experimental`,
+    );
+    expect(stdout).toContain('Unsupported flag "--experimental" provided');
+    expect(exitCode).toBe(2);
+  });
+
+  it('does not divert to the new flow with FF on only', async () => {
+    const { exitCode, stdout } = await run(
+      `snyk iac test ./iac/arm/rule_test.json`,
```
YairZ101, line 46: you forget to add the `--experimental` flag to the CLI command

src/cli/commands/test/iac/index.ts
```diff
@@ -88,6 +89,20 @@ export default async function(
     throw new UnsupportedEntitlementError('infrastructureAsCode');
   }
 
+  if (
```
YairZ101, line 92: Sharing my thoughts, feel free to disagree :)
Do you think we should diverge to the new testing flow even earlier than here (maybe where we check for [`originalOptions.iac`](https://github.com/snyk/cli/blob/master/src/cli/commands/test/index.ts#L55))?

test/jest/acceptance/iac/experimental.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        `snyk iac test ./iac/arm/rule_test.json`,
+      );
+      expect(stdout).toContain('Infrastructure as code issues:');
+      expect(exitCode).toBe(1);
+    });
+  });
+
+  describe(`"${upeFeatureFlag}" feature flag is OFF`, () => {
+    it('does not divert to the new flow with just --experimental flag', async () => {
+      server.setFeatureFlag(upeFeatureFlag, false);
```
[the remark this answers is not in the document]
ipapast: there's no point to add extra lines right now for one test, in my opinion

src/cli/commands/test/iac/index.ts
ipapast: Note to ourselves: This type of conversation is another reason why this should not have been a single task/PR. chat about dead of lines of code that can be moved literally on any other pr

ipapast: > Do you think we should diverge to the new testing flow even earlier than here (maybe where we check for [originalOptions.iac](https://github.com/snyk/cli/blob/master/src/cli/commands/test/index.ts#L55))?

maybe? why were you thinking it? I was thinking that it makes to validate the arguments provided (experimental) and also the entitlement

YairZ101: > Note to ourselves: This type of conversation is another reason why this should not have been a single task/PR. chat about dead of lines of code that can be moved literally on any other pr

I agree, I'll make sure that the next tickets will hold more product value / a more sensible amount of work.\
Thx for the feedback :)

> maybe? why were you thinking it? I was thinking that it makes to validate the arguments provided (experimental) and also the entitlement, before we check for the upe flag and diverge.

Mostly because I believe that in the future we will create a new implementation for `getIacOrgSettings` that will follow what we've discussed [here](https://miro.com/app/board/uXjVO8qNv4s=/?moveToWidget=3458764524883973497&cot=14) and diverging into a new function that doesn't include any logic which is part of v1 (local-exec flow) will let us implement new functionality more easily.
WDYT?

ipapast: so, you are suggesting to diverge earlier and not check/validate anything, but instead add whatever validation is needed etc in the future?

YairZ101: > you are suggesting to diverge earlier and not check/validate anything

do you mean the command validation?

ipapast: basically, is your suggestion to move the new "test" function after this line?
```  
const options = setDefaultTestOptions(originalOptions);
if (...) {
  test ()
  return ...
 }
validateTestOptions(options);
validateCredentials(options);
....
```

YairZ101 (approved):
🔥 🔥 🔥

gitphill (approved):

PR #3304, opened 2022-06-08
chore: Locate UPE Executable [CFG-1846]

ofekatr:

#### What does this PR do?

- Adds logic for locating the UPE executable, which currently needs to be available locally. At a later stage, we will add a downloading step for when the executable is not cached locally.
- Refactors the pre-test step according to [this proposal](https://snyk.slack.com/archives/C02JMMTLUF9/p1655040737968789).

    src/lib/iac/test/v2/setup/policy-engine.ts

- Ensure the `iacCliUnifiedEngine` FF is activated.
- Prepare the UPE executable.
- Place it in your CLI's cache directory.
- Run `snyk-dev iac test --experimental -d`
- Ensure the debug logs include `Using cached UPE binary: <path-to-cached-upe>`
- Run `SNYK_IAC_UPE_PATH=<custom-upe-path> snyk-dev iac test --experimental -d`, with a customized path to a UPE executable.
- Ensure the debug logs include `Custom IaC UPE binary path detected in configuration: <path-to-cached-upe>`

One of the steps in [the pre-test phase](https://miro.com/app/board/uXjVO8qNv4s=/?moveToWidget=3458764524883855643&cot=14) is to locate the UPE binary locally and if we don’t find it to download it, the end result of this step should be a path to the UPE binary.

This ticket will focus on the 1st part of the process which is locating the UPE binary locally.

The implementation of this task will follow [the following diagram](https://miro.com/app/board/uXjVO8qNv4s=/?moveToWidget=3458764525145003511&cot=14).

- [CFG-1846](https://snyksec.atlassian.net/browse/CFG-1846)

### DIFF
--- a/src/cli/args.ts
+++ b/src/cli/args.ts
@@ -35,6 +35,7 @@ const DEBUG_DEFAULT_NAMESPACES = [
   'snyk-test',
   'snyk',
   'snyk-code',
+  'snyk-iac',
   'snyk:find-files',
   'snyk:run-test',
   'snyk:prune',
--- a/src/lib/iac/test/v2/index.ts
+++ b/src/lib/iac/test/v2/index.ts
@@ -1,0 +1,12 @@
+import { TestConfig } from './types';
+import { setup } from './setup';
+
+export { TestConfig } from './types';
+
+export async function test(testConfig: TestConfig) {
+  await setup(testConfig);
+
+  // TODO: Add the rest of the test steps
+
+  return;
+}
--- a/src/lib/iac/test/v2/setup/index.ts
+++ b/src/lib/iac/test/v2/setup/index.ts
@@ -1,0 +1,8 @@
+import { TestConfig } from '../types';
+import { initRules } from './rules';
+import { initPolicyEngine } from './policy-engine';
+
+export async function setup(testConfig: TestConfig) {
+  await initPolicyEngine(testConfig);
+  await initRules(testConfig);
+}
--- a/src/lib/iac/test/v2/setup/rules.ts
+++ b/src/lib/iac/test/v2/setup/rules.ts
@@ -1,8 +1,23 @@
 import * as fs from 'fs';
 import * as tar from 'tar';
-import { CustomError } from '../../../../../lib/errors';
-import { getErrorStringCode } from '../local-execution/error-utils';
-import { IaCErrorCodes } from '../local-execution/types';
+import { CustomError } from '../../../../errors';
+import { getErrorStringCode } from '../../../../../cli/commands/test/iac/local-execution/error-utils';
+import { IaCErrorCodes } from '../../../../../cli/commands/test/iac/local-execution/types';
+import { TestConfig } from '../types';
+
+export async function initRules(testConfig: TestConfig) {
+  const bundleLocator = new RulesBundleLocator(
+    testConfig.cachedBundlePath,
+    testConfig.userBundlePath,
+  );
+  const bundlePath = bundleLocator.locateBundle();
+
+  if (bundlePath) {
+    console.log(`found rules bundle at ${bundlePath}`);
+  } else {
+    console.log('no rules bundle found');
+  }
+}
 
 export class RulesBundleLocator {
   constructor(
--- a/src/lib/iac/test/v2/types.ts
+++ b/src/lib/iac/test/v2/types.ts
@@ -1,0 +1,6 @@
+export interface TestConfig {
+  cachedBundlePath: string;
+  cachedPolicyEnginePath: string;
+  userBundlePath?: string;
+  userPolicyEnginePath?: string;
+}
--- a/src/cli/commands/test/iac/local-execution/types.ts
+++ b/src/cli/commands/test/iac/local-execution/types.ts
@@ -351,6 +351,9 @@ export enum IaCErrorCodes {
 
   // Rules bundle errors.
   InvalidUserRulesBundleError = 1130,
+
+  // Unified Policy Engine executable errors.
+  InvalidUserPolicyEnginePathError = 1140,
 }
 
 export interface TestReturnValue {
--- a/src/cli/commands/test/iac/v2/index.ts
+++ b/src/cli/commands/test/iac/v2/index.ts
@@ -1,19 +1,15 @@
 import chalk from 'chalk';
-import { TestCommandResult } from '../../../types';
-import { RulesBundleLocator } from './rules';
-import config from '../../../../../lib/config';
 import envPaths from 'env-paths';
-import * as path from 'path';
+import * as pathLib from 'path';
+import * as testLib from '../../../../../lib/iac/test/v2';
+import { TestConfig } from '../../../../../lib/iac/test/v2';
+import config from '../../../../../lib/config';
+import { TestCommandResult } from '../../../types';
 
 export async function test(): Promise<TestCommandResult> {
-  const bundleLocator = createRulesBundleLocator();
-  const bundlePath = bundleLocator.locateBundle();
+  const testConfig = prepareTestConfig();
 
-  if (bundlePath) {
-    console.log(`found rules bundle at ${bundlePath}`);
-  } else {
-    console.log('no rules bundle found');
-  }
+  await testLib.test(testConfig);
 
   let response = '';
   response += chalk.bold.green('new flow for UPE integration - TBC...');
@@ -24,9 +20,14 @@ export async function test(): Promise<TestCommandResult> {
   );
 }
 
-function createRulesBundleLocator(): RulesBundleLocator {
+function prepareTestConfig(): TestConfig {
   const systemCachePath = config.CACHE_PATH ?? envPaths('snyk').cache;
-  const cachedBundlePath = path.join(systemCachePath, 'iac', 'bundle.tar.gz');
-  const userBundlePath = config.IAC_BUNDLE_PATH;
-  return new RulesBundleLocator(cachedBundlePath, userBundlePath);
+  const iacCachePath = pathLib.join(systemCachePath, 'iac');
+
+  return {
+    cachedBundlePath: pathLib.join(iacCachePath, 'bundle.tar.gz'),
+    userBundlePath: config.IAC_BUNDLE_PATH,
+    cachedPolicyEnginePath: pathLib.join(iacCachePath, 'snyk-iac-test'),
+    userPolicyEnginePath: config.IAC_POLICY_ENGINE_PATH,
+  };
 }
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -23,6 +23,7 @@ interface Config {
   DRIFTCTL_PATH?: string;
   DRIFTCTL_URL?: string;
   IAC_BUNDLE_PATH?: string;
+  IAC_POLICY_ENGINE_PATH?: string;
   IAC_OUTPUT_V2?: boolean;
 }
 
--- a/src/lib/iac/drift/driftctl.ts
+++ b/src/lib/iac/drift/driftctl.ts
@@ -23,6 +23,7 @@ import * as child_process from 'child_process';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as crypto from 'crypto';
+import { isExe } from '../file-utils';
 
 const debug = debugLib('driftctl');
 
@@ -422,18 +423,6 @@ function driftctlUrl(): string {
   }
 
   return `${dctlBaseUrl}/${driftctlVersion}/${driftctlFileName()}`;
-}
-
-function isExe(dctlPath: string): Promise<boolean> {
-  return new Promise<boolean>((resolve) => {
-    fs.access(dctlPath, fs.constants.X_OK, (err) => {
-      if (err) {
-        resolve(false);
-        return;
-      }
-      resolve(true);
-    });
-  });
 }
 
 function createIfNotExists(path: string) {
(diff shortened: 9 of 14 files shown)


src/cli/args.ts
```diff
@@ -35,6 +35,7 @@ const DEBUG_DEFAULT_NAMESPACES = [
   'snyk-test',
   'snyk',
   'snyk-code',
+  'snyk-iac',
```
ofekatr, line 38: See [thread](https://snyk.slack.com/archives/C0127HWU0E7/p1654690206531289)

Avishagp (approved):

src/lib/iac/test/v2/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface TestConfig {
+  cachedBundlePath: string;
+  cachedUpePath: string;
+  customBundlePath?: string;
```
teodora-sandu, line 4: question: Is it confusing to use "custom" here? Can we use something else (`userBundlePath` like it used to be, or `localBundlePath` if you didn't like that). Just because it currently makes me think of the custom rules bundle

src/lib/iac/test/v2/setup/index.ts
```diff
@@ -0,0 +1,19 @@
+import { TestConfig } from '../types';
+import { RulesBundleLocator } from './rules';
+import { initUpe } from './upe';
+
+export async function setup(testConfig: TestConfig) {
+  await initUpe(testConfig);
```
teodora-sandu, line 6: Let's call this `initPolicyEngine` because acronyms can be confusing. See also [this thread](https://snyk.slack.com/archives/C036BK48B4H/p1654782980058569) where it's discussed that we want to call this just "policy engine"

src/lib/iac/test/v2/setup/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const bundleLocator = new RulesBundleLocator(
+    testConfig.cachedBundlePath,
+    testConfig.customBundlePath,
+  );
+  const bundlePath = bundleLocator.locateBundle();
```
teodora-sandu, line 12: suggestion: Could we move this into an `initRulesBundle` or somethig like that inside the `rules.ts` file so that this setup function doesn't need to know about the RulesBundleLocator?

src/lib/iac/test/v2/setup/rules.ts
```diff
@@ -1,8 +1,8 @@
 import * as fs from 'fs';
 import * as tar from 'tar';
-import { CustomError } from '../../../../../lib/errors';
-import { getErrorStringCode } from '../local-execution/error-utils';
-import { IaCErrorCodes } from '../local-execution/types';
+import { CustomError } from '../../../../errors';
+import { getErrorStringCode } from '../../../../../cli/commands/test/iac/local-execution/error-utils';
+import { IaCErrorCodes } from '../../../../../cli/commands/test/iac/local-execution/types';
```
teodora-sandu, line 5: Note (not for this PR) to move error stuff to lib too

teodora-sandu, line 5: Note (not for this PR) to move error stuff to lib too

src/lib/iac/test/v2/setup/index.ts
teodora-sandu: Maybe even the `console.log`s, since for the`initUpe` function they are also included in there, up to you

src/lib/iac/test/v2/setup/upe.ts
```diff
@@ -0,0 +1,60 @@
+import * as createDebugLogger from 'debug';
+import { isExe } from '../../../file-utils';
+import { CustomError } from '../../../../errors';
+import { IaCErrorCodes } from '../../../../../cli/commands/test/iac/local-execution/types';
+import { getErrorStringCode } from '../../../../../cli/commands/test/iac/local-execution/error-utils';
+import { TestConfig } from '../types';
+
+const debugLogger = createDebugLogger('snyk-iac');
+
+export class InvalidCustomUpePathError extends CustomError {
```
teodora-sandu, line 10: What do you think of storing all errors in one file? Since we're starting this from scratch that way we can use that file as the documentation for all our errors

src/lib/iac/test/v2/setup/upe.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  // Lookup in cache.
+  else {
+    if (await isExe(cachedUpePath)) {
+      debugLogger('Using cached UPE binary: %s', cachedUpePath);
+      return cachedUpePath;
+    } else {
+      debugLogger('UPE binary was not cached: %s', cachedUpePath);
+    }
+  }
```
teodora-sandu, line 50: This `if else` logic is very similar to the one in the `RulesBundleLocator`. Do you think it's worth having a single "CacheOrUserPathLocator" class that does this logic?

src/lib/iac/test/v2/index.ts
```diff
@@ -0,0 +1,9 @@
+import { TestConfig } from './types';
```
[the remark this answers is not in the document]
ofekatr: Excellent question @teodora-sandu. 
From my understanding, the `cli` is meant for handling the I/O logic components of any flow in the CLI. Compared to a server, the command modules in the `cli` directory serve as controllers in which we process the request, collect inputs, flag values, relevant environment variables and CLI configuration and overall parse all the provided data to a request the internal services and libraries can process. 

The `lib` module contains business logic libraries, which are agnostic to the user interface they are used for. 
While the majority of the code for the original IaC local execution flow was in `cli`, I think it's the equivalent of having all of your business logic defined in your controller on a server.

src/lib/iac/test/v2/types.ts
ofekatr: Fixed in [`6c305438`](https://github.com/snyk/cli/pull/3304/commits/6c3054383401564c92112f7770e973499b0fee63)

test/jest/unit/lib/iac/test/v2/setup/upe.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  const defaultTestConfig = {
+    cachedUpePath: `iac/cache/path/snyk-iac-test`,
+  };
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  describe('when a custom path for a UPE executable is not provded', () => {
+    describe('when the UPE executable was cached', () => {
```
teodora-sandu, line 20: nit: Just a preference so feel free to ignore it but there's no need to have a describe block with a single it block inside it.

src/lib/iac/test/v2/index.ts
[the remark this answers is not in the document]
teodora-sandu: I agree with you, I think maybe we can discuss this in the weekly tech meeting so we can all get on the same page for this

teodora-sandu (commented):
I've left some comments, let me know if you want to discuss any! The nits are not blocking, overall a great job 🥇

src/lib/iac/test/v2/setup/rules.ts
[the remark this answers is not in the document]
ofekatr: I'm really glad this was brought up.
There are several things we could improve in our usage of errors in the CLI, to make them more generalized for Snyk IaC rather than defined for local execution, as we've added many errors that are only relevant for other commands/flows. 
This is planned to be brought up during the Technical Weekly meetings.

src/lib/iac/test/v2/setup/upe.ts
ofekatr: I'm actually not in favor of this approach in cases where the error is only relevant in a single module. 
I think they are more accessible when positioned at the top of the module in which they are relevant for. 

This way, rather than searching for a specific error in a module filled with errors of all different kinds and flows, you know precisely in which module to find your error, which is incidentally the module in which the error is being used.

For errors that are used in several modules and are associated with a particular module, I'd still have it defined at the top of that module. 

For errors that are used in several modules and not associated with a particular module, I'd create an `errors.ts` module in the right directory depending on the scope, and add it there, so that all module that need the error are in that directory.

src/lib/iac/test/v2/setup/upe.ts
ofekatr: Stellar way of thinking. 
Two comments:
- Generally unless presented with a strong need, I'd avoid using classes in typescript in general, as their disadvantages exceed their advantages. Instead, I'd advise taking advantage of TypeScript's modules architecture. Some links that elaborate further ([link 1](https://dev.to/vangware/you-don-t-need-classes-jk4), [link 2](https://www.reddit.com/r/typescript/comments/65zq36/is_the_idea_that_classes_are_bad_still_relevant/)
- In these two examples, the logic for ensuring a valid UPE executable is different from the validation logic for the rules bundle. While we could implement an abstract solution that allows us to inject different validation logics, I think that'd take more effort than just having both 😆

src/lib/iac/test/v2/setup/index.ts
teodora-sandu: I think since it's in v2 then it will be clear which one it is. Eventually this will become the singular policy engine, and like I said the `upe` acronym can be a bit confusing.

src/lib/iac/test/v2/setup/upe.ts
teodora-sandu: Agreed that generally speaking classes can be avoided. I expected the loading logic for the two to be quite similar and I'd like us to maintain a similar coding style if possible, especially so early on in the re-implementation of the CLI. I would like to avoid having different ways of doing the same thing if possible, but I'll give it a think outside of this code review 😆

src/lib/iac/test/v2/setup/upe.ts
teodora-sandu: I think what you're saying is it's useful to know what file an error is for? I'm not sure what that'd be good for, although of course we all have different development processes so maybe it's just that we work differently

We've talked previously as team about having a single place in code to store errors, so that for example it works as internal documentation for us and documentation for the content team for reviews. 

For me the fact we even have to think so deeply about where an error should go based on its usage shows that we should have them all in one place. But we can take the discussion outside of this PR if you want

ofekatr: > We've talked previously as team about having a single place in code to store errors
Is there a way to link the discussion? If not, would it be alright with you if we bring it up in the channel or in the upcoming Technical Weekly meeting?

src/lib/iac/test/v2/setup/index.ts
ofekatr: Fixed in [`6c305438`](https://github.com/snyk/cli/pull/3304/commits/6c3054383401564c92112f7770e973499b0fee63)

src/lib/iac/test/v2/setup/upe.ts
teodora-sandu: I don't think we discussed it on Slack - it was mostly in discussions. Let's discuss at the technical meeting?

ofekatr: Added to the Technical Weekly [here](https://www.notion.so/snyk/Weekly-Technical-Meeting-86d2e62818344e77af52bae44ac5e1bf?p=fac7fb068ee14f8aaeba97f41c011dfa).

teodora-sandu (approved):

src/lib/iac/test/v2/setup/upe.ts
francescomari: > Generally unless presented with a strong need, I'd avoid using classes in typescript in general, as their disadvantages exceed their advantages.

This statement hinges with your definition of "strong need". Classes and object oriented design are great tools to enforce stricter separation of concerns, clearer APIs, and proper hiding of implementation details.

> I'd advise taking advantage of TypeScript's modules architecture.

In general, you don't need language support for classes and interfaces to come up with an object-oriented design. In fact, C programmers have been following object oriented design practices since the 70s while using a language with very primitive features. Lisp programmers had language support for functions similar to JavaScript's, and they were aggressively using closures to implement what we would call private class fields today.

That said, I think that the module system in Node can only bring you so far. Effectively, a module is what in OO programming is called a singleton object. For more complicated designs, you need to reach beyond that level of abstraction. Currently, our code is littered with stuff like this:

```
type Params = {
  // A dozen fields...
}

function doSomething(params: Params) {
  // Do something with the params.
}
```

The future maintainers of `doSomething()` need to understand every single field in `Params` and how the function manipulates it. Even worse, a programmer who wants to call `doSomething()` will be confronted with the task of filling up a big object, and he needs to understand what every field means. Effectively, the author of `doSomething()` did a poor job of making clear:

- what piece of data is an implementation detail and what is part of the API,
- what field represents configuration and what represents an actual run-time parameter,
- what is the intention of the function from its signature.

Now, I'm not here to tell you to always use classes or to never use functions, that would be silly. Do what you think is right from a design perspective. I would suggest, though, to be wary of blanket statements dismissing 50+ years of software design literature because somebody didn't like how JavaScript implements classes.

YairZ101: > I expected the loading logic for the two to be quite similar and I'd like us to maintain a similar coding style if possible, especially so early on in the re-implementation of the CLI.

I 100% agree with @teodora-sandu here and this was also my expectation.
@ofekatr / @francescomari do you want to open this topic for async discussion in the team channel or do you want to discuss it in the tech huddle?

PR #3306, opened 2022-06-09
chore: locate rules bundle in iac test v2

francescomari:

#### What does this PR do?

This PR adds the logic to locate a rules bundle based on user input and falling back to the local cache if none is found. Please note that, once the bundle is located, the path is currently printed to the console. This is not going to be the final behaviour, but there is no consumer of the bundle path at the moment.

The bulk of the logic is in `src/cli/commands/test/iac/v2/rules.ts`.

Enable the `iacCliUnifiedEngine` and run `snyk` with different arguments.

Use the bundle provided by the user:

```
SNYK_IAC_BUNDLE_PATH=test/fixtures/iac/custom-rules/custom.tar.gz snyk iac test --experimental
```

Use the bundle from the cache (your system cache might be at a different path):

```
mkdir -p ~/Library/Caches/snyk-nodejs/iac
cp test/fixtures/iac/custom-rules/custom.tar.gz ~/Library/Caches/snyk-nodejs/iac/bundle.tar.gz
snyk iac test --experimental
```

Verify that when both a user bundle and a cached bundle exist, the user bundle is chosen:

```
mkdir -p ~/Library/Caches/snyk-nodejs/iac
cp test/fixtures/iac/custom-rules/custom.tar.gz ~/Library/Caches/snyk-nodejs/iac/bundle.tar.gz
SNYK_IAC_BUNDLE_PATH=test/fixtures/iac/custom-rules/custom.tar.gz snyk iac test --experimental
```

[CFG-1884](https://snyksec.atlassian.net/browse/CFG-1884)

### DIFF
--- a/src/cli/commands/test/iac/v2/index.ts
+++ b/src/cli/commands/test/iac/v2/index.ts
@@ -1,7 +1,20 @@
 import chalk from 'chalk';
 import { TestCommandResult } from '../../../types';
+import { RulesBundleLocator } from './rules';
+import config from '../../../../../lib/config';
+import envPaths from 'env-paths';
+import * as path from 'path';
 
 export async function test(): Promise<TestCommandResult> {
+  const bundleLocator = createRulesBundleLocator();
+  const bundlePath = bundleLocator.locateBundle();
+
+  if (bundlePath) {
+    console.log(`found rules bundle at ${bundlePath}`);
+  } else {
+    console.log('no rules bundle found');
+  }
+
   let response = '';
   response += chalk.bold.green('new flow for UPE integration - TBC...');
   return TestCommandResult.createHumanReadableTestCommandResult(
@@ -9,4 +22,11 @@ export async function test(): Promise<TestCommandResult> {
     '',
     '',
   );
+}
+
+function createRulesBundleLocator(): RulesBundleLocator {
+  const systemCachePath = config.CACHE_PATH ?? envPaths('snyk').cache;
+  const cachedBundlePath = path.join(systemCachePath, 'iac', 'bundle.tar.gz');
+  const userBundlePath = config.IAC_BUNDLE_PATH;
+  return new RulesBundleLocator(cachedBundlePath, userBundlePath);
 }
--- a/src/cli/commands/test/iac/v2/rules.ts
+++ b/src/cli/commands/test/iac/v2/rules.ts
@@ -1,0 +1,76 @@
+import * as fs from 'fs';
+import * as tar from 'tar';
+import { CustomError } from '../../../../../lib/errors';
+import { getErrorStringCode } from '../local-execution/error-utils';
+import { IaCErrorCodes } from '../local-execution/types';
+
+export class RulesBundleLocator {
+  constructor(
+    private cachedBundlePath: string,
+    private userBundlePath?: string,
+  ) {}
+
+  locateBundle(): string | null {
+    const locatedUserBundle = this.locateUserBundle();
+
+    if (locatedUserBundle) {
+      return locatedUserBundle;
+    }
+
+    return this.locateCachedBundle();
+  }
+
+  private locateUserBundle(): string | null {
+    if (!this.userBundlePath) {
+      return null;
+    }
+
+    if (!fs.existsSync(this.userBundlePath)) {
+      return null;
+    }
+
+    if (!fs.statSync(this.userBundlePath).isFile()) {
+      throw new InvalidUserRulesBundleError('user bundle is not a file');
+    }
+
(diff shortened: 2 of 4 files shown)


src/cli/commands/test/iac/v2/rules.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (!this.userBundlePath) {
+      return null;
+    }
+
+    if (!fs.existsSync(this.userBundlePath)) {
+      return null;
+    }
+
+    if (!fs.statSync(this.userBundlePath).isFile()) {
+      throw new Error('user bundle is not a file');
```
teodora-sandu, line 29: Will this error be seen by customers? Should we create a new error that extends the `CustomError` like this:
https://github.com/snyk/cli/blob/fd0a3d8c5172773fab56823c9ce246706eefe1ea/src/cli/commands/test/iac/local-execution/local-cache.ts#L206

src/cli/commands/test/iac/v2/rules.ts
```diff
@@ -0,0 +1,46 @@
...
+  private locateUserBundle(): string | null {
+    if (!this.userBundlePath) {
+      return null;
+    }
+
+    if (!fs.existsSync(this.userBundlePath)) {
+      return null;
+    }
+
+    if (!fs.statSync(this.userBundlePath).isFile()) {
```
teodora-sandu, line 28: Should we also check that it's a tar.gz file? The diagram in the ticket does that, not sure if it's helpful since we would throw en error further down the line anyway but what do you think?

src/cli/commands/test/iac/v2/rules.ts
```diff
@@ -0,0 +1,46 @@
...
+  private locateCachedBundle(): string | null {
+    if (!fs.existsSync(this.cachedBundlePath)) {
+      return null;
+    }
+
+    if (!fs.statSync(this.cachedBundlePath).isFile()) {
+      return null;
+    }
+
+    return this.cachedBundlePath;
```
teodora-sandu, line 44: nit: Could extract the validation logic to another function if it's the same for the user and cached bundle

src/cli/commands/test/iac/v2/index.ts
```diff
@@ -10,3 +23,10 @@ export async function test(): Promise<TestCommandResult> {
     '',
   );
 }
+
+function createRulesBundleLocator(): BundleLocator {
+  const systemCachePath = config.CACHE_PATH ?? envPaths('snyk').cache;
+  const cachedBundlePath = path.join(systemCachePath, 'iac', 'bundle.tar.gz');
+  const userBundlePath = config.IAC_BUNDLE_PATH;
```
teodora-sandu, line 30: Should we create a new environment variable for this? This environment variable is used by us to test new Snyk bundles before publishing them (https://github.com/snyk/cloud-config-opa-policies/blob/master/docs/create_release.md#updating-the-policy-bundle-used-by-the-cli) so if we start using it to test the new UPE bundles then we'll lose that functionality

src/cli/commands/test/iac/v2/rules.ts
francescomari: Done at 6870674c.

src/cli/commands/test/iac/v2/rules.ts
francescomari: The two functions are not the same. Some conditions result in an error for the user-provided bundle, but not for the cached bundle. The checks are similar, but the results are not.

src/cli/commands/test/iac/v2/index.ts
francescomari: Not sure about that. The environment variable is used exactly for that purpose in this PR. If a user specifies that variable, the command uses the rules from that bundle. This wouldn't even break the way we currently test the rules bundle. What do you think?

src/cli/commands/test/iac/v2/rules.ts
francescomari: While I added the new custom error, this creates a dependency from the "new" CLI code to the "old" one because of `IaCErrorCodes` and `getErrorStringCode`. I'm not really enthusiastic about that. Do you think we should have a new set of errors and error utilities in the new CLI stuff?

src/cli/commands/test/iac/v2/index.ts
teodora-sandu: Ah okay that error I was seeing was just me being stupid. I agree with you. When we use the command without `--experimental` it will continue working so yes let's just use this env var

teodora-sandu (approved):
Tested it and it works 👍

PR #3312, opened 2022-06-10
Fix/v2 http proxy

PeterSchafer:

#### What does this PR do?
In addition to the HTTPS_PROXY CLIv2 needs to specify the HTTP_PROXY environment variable. 

Due to the change it is necessary to disable two HTTP proxy related tests. One can be enabled when error handling is improved and reasonable asserting can be done, the other seems to test for a specific needle (npm package in CLIv1) behavior, which is necessary to reproduce in CLIv2.

### DIFF
--- a/test/jest/acceptance/proxy-behavior.spec.ts
+++ b/test/jest/acceptance/proxy-behavior.spec.ts
@@ -7,7 +7,6 @@ const SNYK_API_HTTP = 'http://snyk.io/api/v1';
 const FAKE_HTTP_PROXY = `http://localhost:${fakeServerPort}`;
 
 jest.setTimeout(1000 * 60 * 1);
-
 describe('Proxy configuration behavior', () => {
   describe('*_PROXY against HTTPS host', () => {
     it('tries to connect to the HTTPS_PROXY when HTTPS_PROXY is set', async () => {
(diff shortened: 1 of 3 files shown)


Avishagp (approved):

test/jest/acceptance/proxy-behavior.spec.ts
```diff
@@ -56,41 +57,51 @@ describe('Proxy configuration behavior', () => {
...
+          env: {
+            ...process.env,
+            HTTP_PROXY: FAKE_HTTP_PROXY,
+            SNYK_API: SNYK_API_HTTP,
+            SNYK_HTTP_PROTOCOL_UPGRADE: '0',
+          },
+        });
 
-      expect(code).toBe(0);
+        // eslint-disable-next-line jest/no-standalone-expect
```
Avishagp, line 72: Why did we need to add the eslint disable? I know we usually treat it as bad practice. Is there another way to test this without ignoring jest rules?

test/jest/acceptance/proxy-behavior.spec.ts
```diff
@@ -7,6 +7,7 @@ const SNYK_API_HTTP = 'http://snyk.io/api/v1';
 const FAKE_HTTP_PROXY = `http://localhost:${fakeServerPort}`;
 
 jest.setTimeout(1000 * 60 * 1);
+const itIfNotCLIV2 = () => (isCLIV2() ? it.skip : it);
```
Avishagp, line 10: Not sure about this. Can't we add a check in the describe to check if the test is being run for CLIV2? that method makes it a bit harder to understand

Avishagp (approved):

PR #3313, opened 2022-06-10
chore: print a warning when the user uses paths outside CWD

francescomari:

#### What does this PR do?

This PR lets the `snyk iac test` command print a warning when the user provides one or more paths to scan that lay outside the current working directory.

Run the command without paths and check that the warning message is not displayed:

```
snyk iac test
```

Run the command with no paths laying outside the current working directory and check that the warning message is not displayed:

```
snyk iac test foo.tf bar.tf
```

Run the command with at least one path laying outside the current working directory and check that the warning message is displayed:

```
snyk iac test foo.tf bar.tf ../baz.tf
```

[CFG-1967](https://snyksec.atlassian.net/browse/CFG-1967)

#### Screenshots

image

### DIFF
--- a/src/cli/commands/test/iac/index.ts
+++ b/src/cli/commands/test/iac/index.ts
@@ -41,6 +41,7 @@ import { UnsupportedEntitlementError } from '../../../../lib/errors/unsupported-
 import * as ora from 'ora';
 import { CustomError, FormattedCustomError } from '../../../../lib/errors';
 import { scan } from './scan';
+import * as path from 'path';
 
 const debug = Debug('snyk-test');
 const SEPARATOR = '\n-------------------------------------------------------\n';
@@ -73,6 +74,10 @@ export default async function(
 
   if (shouldLogUserMessages(options, isNewIacOutputSupported)) {
     console.log(EOL + iacTestTitle + EOL);
+
+    if (paths.some(isOutsideCurrentWorkingDirectory)) {
+      printCurrentWorkingDirectoryTraversalWarning();
+    }
 
     testSpinner = ora({ isSilent: options.quiet, stream: process.stdout });
   }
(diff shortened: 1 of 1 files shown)


src/cli/commands/test/iac/index.ts
```diff
@@ -74,6 +75,20 @@ export default async function(
...
 
+    if (paths.some(isOutsideCurrentWorkingDirectory)) {
+      console.log(
+        chalk.yellow(
+          'Warning: Scanning paths outside the current working directory is deprecated and',
+        ),
+      );
+      console.log(
+        chalk.yellow(
+          'will be removed in the future. Please see the documentation for further details.',
```
aron, line 86: Can we add a link to the documentation?

francescomari: Would something like this work?

image

src/cli/commands/test/iac/index.ts
```diff
@@ -74,6 +75,26 @@ export default async function(
   if (shouldLogUserMessages(options, isNewIacOutputSupported)) {
     console.log(EOL + iacTestTitle + EOL);
 
+    if (paths.some(isOutsideCurrentWorkingDirectory)) {
```
teodora-sandu, line 78: Do you think it would be nicer to have just one `console.log` for this message instead of multiple for the end-of-line? We can use `EOL` instead to represent the newline. I'm thinking it would make it more obvious what makes up a single user message.

francescomari: I will extract the message in a new constant in this file. It's probably not worth it to spread the code around for something that is going to go away in one month. Having everything here will just make it easier to delete.

francescomari: In 33894715 I created  a new function responsible for building the message and printing it. It's a bit clearer than before, and it could be even clearer if the formatter would stop adding newlines to the code in semi-random places.

teodora-sandu (approved):
LGTM 👍

PR #3317, opened 2022-06-13
fix: support HTTP(S) proxies in iac-test

craigfurman:

#### What does this PR do?

Configure HTTP clients to respect HTTP(S)_PROXY environment variables.
Ignore TLS identity verification errors (this is dangerous) when the
`--insecure` flag is passed, matching the behaviour of `snyk test`.

I manually tested this by building the CLI and running in the 3 following scenarios. First, I ran `mitmdump` (from the mitmproxy package) in the background, bound to tcp:8080.

1. `node ./dist/cli iac test a-cloudformation-file.yaml` works as expected and does not call the proxy
2. `https_proxy=http://localhost:8080 node ./dist/cli iac test a-cloudformation-file.yaml` exits with an error, because I did not configure my local TLS certificate store with a cert from mitmproxy.
3. `https_proxy=http://localhost:8080 node ./dist/cli iac test a-cloudformation-file.yaml --insecure` works, and calls the proxy

This is intended to be useful for users behind corporate proxies. Ideally no-one will use `--insecure`, and will instead configure a CA from (poorly behaved) HTTPS proxies that do act as MITM proxies, but this is a feature we support elsewhere so it should probably work consistently.

No ticket AFAIK, this came up informally as part of on-call work: https://snyk.slack.com/archives/CRV0PMFDH/p1654811904912649

### DIFF
--- a/src/cli/commands/test/iac/local-execution/local-cache.ts
+++ b/src/cli/commands/test/iac/local-execution/local-cache.ts
@@ -3,6 +3,8 @@ import * as fs from 'fs';
 import { EngineType, IaCErrorCodes } from './types';
 import * as needle from 'needle';
 import * as rimraf from 'rimraf';
+import { getProxyForUrl } from 'proxy-from-env';
+import { bootstrap } from 'global-agent';
 import { createIacDir, extractBundle, isValidBundle } from './file-utils';
 import * as Debug from 'debug';
 import { CustomError } from '../../../../../lib/errors';
@@ -10,8 +12,11 @@ import * as analytics from '../../../../../lib/analytics';
 import ReadableStream = NodeJS.ReadableStream;
 import { getErrorStringCode } from './error-utils';
 import config from '../../../../../lib/config';
+import { Global } from '../../../../args';
 
 const debug = Debug('iac-local-cache');
+
+declare const global: Global;
 
 export const LOCAL_POLICY_ENGINE_DIR = '.iac-data';
 
@@ -141,7 +146,15 @@ export async function initLocalCache({
   // always overwrite whatever might be there.
   try {
     const BUNDLE_URL = 'https://static.snyk.io/cli/wasm/bundle.tar.gz';
-    const response: ReadableStream = needle.get(BUNDLE_URL);
+    const proxyUri = getProxyForUrl(BUNDLE_URL);
+    if (proxyUri) {
+      bootstrap({
+        environmentVariableNamespace: '',
+      });
+    }
+    const response: ReadableStream = needle.get(BUNDLE_URL, {
+      rejectUnauthorized: !global.ignoreUnknownCA,
+    });
     await extractBundle(response);
   } catch (e) {
     throw new FailedToDownloadRulesError();
--- a/test/jest/unit/iac/local-cache.spec.ts
+++ b/test/jest/unit/iac/local-cache.spec.ts
@@ -30,6 +30,7 @@ describe('initLocalCache - downloads bundle successfully', () => {
 
     expect(needle.get).toHaveBeenCalledWith(
       expect.stringContaining('bundle.tar.gz'),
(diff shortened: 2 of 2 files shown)


src/cli/commands/test/iac/local-execution/local-cache.ts
```diff
@@ -141,7 +146,15 @@ export async function initLocalCache({
   // always overwrite whatever might be there.
   try {
     const BUNDLE_URL = 'https://static.snyk.io/cli/wasm/bundle.tar.gz';
-    const response: ReadableStream = needle.get(BUNDLE_URL);
+    const proxyUri = getProxyForUrl(BUNDLE_URL);
+    if (proxyUri) {
+      bootstrap({
```
craigfurman, line 151: this copies some of the logic in our needle wrapper, https://github.com/snyk/cli/blob/master/src/lib/request/request.ts.

I had trouble coaxing a streamable response out of that wrapper, and it seemed simpler to just call the same 2 methods here instead rather than rework the interface in our wrapper.

test/jest/unit/iac/local-cache.spec.ts
```diff
@@ -17,21 +18,27 @@ describe('initLocalCache - downloads bundle successfully', () => {
...
-    expect(needle.get).toHaveBeenCalledWith(
-      expect.stringContaining('bundle.tar.gz'),
+    expect(request.makeRequest).toHaveBeenCalledWith(
+      expect.objectContaining({
+        method: 'get',
+        url: expect.stringContaining('bundle.tar.gz'),
+      }),
     );
-    expect(spy).toHaveBeenCalledWith(mockReadable);
+    expect(spy).toHaveBeenCalledWith(expect.any(Readable));
```
craigfurman, line 41: yes, this test is now less precise, but I couldn't figure out how to meaningfully compare the streams. I played around with reading them both into buffers and comparing that, but had some issues. These tests are overly coupled to the implementation anyway, and IMO are not particularly helpful for development.

Highlighting it in case anyone wants to push back, of course

src/cli/commands/test/iac/local-execution/local-cache.ts
francescomari: AFAIK `needle` supports downloading the body directly to the file system via the `output` option, but it needs to be verified whether the version we use supports that. Maybe that was the reason why `needle.get()` was using directly. `needle.request()`, which is used by `makeRequest()` under the hood, doesn't return a stream.

craigfurman: We could use `output`, it appears to be present in our version of needle. We're streaming from the socket right into untar today, which is kind of nice. However, we then unpack that tar right onto the filesystem anyway I think, so perhaps being consistent in that way is not the worst thing in the world.

maxjeffos (approved):

src/cli/commands/test/iac/local-execution/local-cache.ts
craigfurman: I've actually now reverted all the way back to how this PR started, duping calls to `bootstrap` to set up global proxy state from the makeRequest wrapper. Mutating global state from a command handler is not ideal, but we already do this every time makeRequest is called (if a proxy is requested).

This way, we keep piping the stream directly into tar, which is more optimal for large custom rule bundles. I hadn't initially noticed this code was also used for custom rules.

craigfurman:
@francescomari amazing, do you want to PR that and I'll close this?

francescomari:
@craigfurman I have no preference. Whatever works best for you.

craigfurman:
Well if you PR it, I can review it 😁

francescomari:
Alright. Let's close this, and I will open a new PR based on the latest changes.

francescomari:
Closing in favour of #3348.

PR #3324, opened 2022-06-15
fix: cwd error

YairZ101:


src/cli/commands/test/iac/local-execution/types.ts
```diff
@@ -299,6 +299,7 @@ export enum IaCErrorCodes {
   // file-loader errors
   NoFilesToScanError = 1010,
   FailedToLoadFileError = 1011,
+  CurrentWorkingDirectoryTraversalError = 1012,
```
YairZ101, line 302: This might not be the correct section for this error but I also don't think we should create a new section just for this specific case.

PR #3327, opened 2022-06-19
chore: Implement UPE executable downloading

ofekatr:

#### What does this PR do?

- Adds abstract logic for downloading cache resources for the IaC test flow.
- Adds logic for downloading the Policy Engine executable when it is not cached locally.

- Ensure the binary is not cached locally.
- Ensure you have the `iacCliUnifiedEngine` FF activated.
- Run `snyk iac test --experimental`
- Ensure the flow finishes successfully without errors.

- The step after achieving [CFG-1890: UPE CLI Integration - Milestone #0](https://snyksec.atlassian.net/browse/CFG-1890) is to stop relying on a UPE binary that will be available locally and instead download it, this will enable us to achieve  [CFG-1891: Milestone #1](https://snyksec.atlassian.net/browse/CFG-1891).

- [CFG-1885](https://snyksec.atlassian.net/browse/CFG-1885)

### DIFF
--- a/src/lib/iac/file-utils.ts
+++ b/src/lib/iac/file-utils.ts
@@ -44,3 +44,11 @@ export async function isArchive(path: string): Promise<boolean> {
     return false;
   }
 }
+
+export async function saveFile(
+  dataBuffer: Buffer,
+  savePath: string,
+): Promise<void> {
+  await fsPromises.writeFile(savePath, dataBuffer);
+  await fsPromises.chmod(savePath, 0o744);
+}
--- a/src/lib/iac/test/v2/setup/local-cache/policy-engine/constants/index.ts
+++ b/src/lib/iac/test/v2/setup/local-cache/policy-engine/constants/index.ts
@@ -3,9 +3,11 @@ import { formatPolicyEngineFileName } from './utils';
 /**
  * The Policy Engine release version associated with this Snyk CLI version.
  */
-export const releaseVersion = '0.1.0';
+export const policyEngineReleaseVersion = '0.1.0';
 
 /**
  * The Policy Engine executable's file name.
  */
-export const policyEngineFileName = formatPolicyEngineFileName(releaseVersion);
+export const policyEngineFileName = formatPolicyEngineFileName(
+  policyEngineReleaseVersion,
+);
--- a/src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
+++ b/src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
@@ -1,0 +1,126 @@
+import * as pathLib from 'path';
+import * as crypto from 'crypto';
+import * as createDebugLogger from 'debug';
+
+import { getErrorStringCode } from '../../../../../../../cli/commands/test/iac/local-execution/error-utils';
+import { IaCErrorCodes } from '../../../../../../../cli/commands/test/iac/local-execution/types';
+import { CustomError } from '../../../../../../errors';
+import { TimerMetricInstance } from '../../../../../../metrics';
+import { TestConfig } from '../../../types';
+import { fetchCacheResource } from '../utils';
+import { policyEngineFileName, policyEngineReleaseVersion } from './constants';
+import { saveFile } from '../../../../../file-utils';
+
+const debugLog = createDebugLogger('snyk-iac');
+
+export async function downloadPolicyEngine(
+  testConfig: TestConfig,
+): Promise<string> {
+  let downloadDurationSeconds = 0;
+
+  const timer = new TimerMetricInstance('iac_policy_engine_download');
+  timer.start();
+
+  const dataBuffer = await fetch();
+
+  assertValidChecksum(dataBuffer);
+
+  const cachedPolicyEnginePath = await cache(
+    dataBuffer,
+    testConfig.iacCachePath,
+  );
+
+  timer.stop();
+  downloadDurationSeconds = Math.round((timer.getValue() as number) / 1000);
+
+  debugLog(
+    `Downladed Policy Engine successfully in ${downloadDurationSeconds} seconds`,
+  );
+
+  return cachedPolicyEnginePath;
+}
+
+async function fetch(): Promise<Buffer> {
+  debugLog(`Fetching Policy Engine executable from ${policyEngineUrl}`);
+
+  let policyEngineDataBuffer: Buffer;
+  try {
+    policyEngineDataBuffer = await fetchCacheResource(policyEngineUrl);
+  } catch (err) {
+    throw new FailedToDownloadPolicyEngineError();
+  }
+  debugLog('Policy Engine executable was fetched successfully');
+
+  return policyEngineDataBuffer;
+}
+
+export const policyEngineUrl = `https://static.snyk.io/cli/iac/test/v${policyEngineReleaseVersion}/${policyEngineFileName}`;
+
+export class FailedToDownloadPolicyEngineError extends CustomError {
+  constructor() {
+    super(`Failed to download cache resource from ${policyEngineUrl}`);
+    this.code = IaCErrorCodes.FailedToDownloadPolicyEngineError;
+    this.strCode = getErrorStringCode(this.code);
+    this.userMessage =
+      `Could not fetch cache resource from: ${policyEngineUrl}` +
+      '\nEnsure valid network connection.';
+  }
+}
+
+function assertValidChecksum(dataBuffer: Buffer): void {
+  const computedChecksum = crypto
+    .createHash('sha256')
+    .update(dataBuffer)
+    .digest('hex');
+
+  if (computedChecksum !== policyEngineChecksum) {
+    throw new FailedToDownloadPolicyEngineError();
+  }
+
+  debugLog('Fetched Policy Engine executable has valid checksum');
+}
+
+export const policyEngineChecksum = {
+  'snyk-iac-test_0.1.0_Linux_x86_64':
+    '0b0d846cd74bf42676f79875ab30f20dc08529aedb94f2f6dd31a67c302b78e4',
+  'snyk-iac-test_0.1.0_Darwin_x86_64':
+    '896960a09b6adf699185f443428cda25ccb30123440bc8735e1a05df1e9cbc12',
+  'snyk-iac-test_0.1.0_Windows_x86_64.exe':
+    'a0b0b5781218f42d121cd1c6bca2d5ea27fbe76b1e9817e37745586c276cabda',
+  'snyk-iac-test_0.1.0_Linux_arm64':
+    'b0c0bd9a06cc3d556b526b868e5a4ac4c5a3938899c4e312607f6828626debe9',
+  'snyk-iac-test_0.1.0_Darwin_arm64':
+    'cf7a327378983810ea043774a333bdd724e3866a815555d59ec5cd8aa25ea5ec',
+  'snyk-iac-test_0.1.0_Windows_arm64.exe':
+    'd88b24c611c8d37c2df9382e6e1b39933b926ee2ed438f1fdb69c51da5086fc5',
+}[policyEngineFileName]!;
+
+async function cache(
+  dataBuffer: Buffer,
+  iacCachePath: string,
+): Promise<string> {
+  const savePath = pathLib.join(iacCachePath, policyEngineFileName);
+
+  debugLog(`Caching Policy Engine executable to ${savePath}`);
+
+  try {
+    await saveFile(dataBuffer, savePath);
+  } catch (err) {
+    throw new FailedToCachePolicyEngineError(savePath);
+  }
+
+  debugLog(`Policy Engine executable was successfully cached`);
+
+  return savePath;
+}
+
+export class FailedToCachePolicyEngineError extends CustomError {
+  constructor(savePath: string) {
+    super(`Failed to cache Policy Engine executable to ${savePath}`);
+    this.code = IaCErrorCodes.FailedToCachePolicyEngineError;
+    this.strCode = getErrorStringCode(this.code);
+    this.userMessage =
+      `Could not write the downloaded cache resource to: ${savePath}` +
+      '\nEnsure the cache directory is writable.';
+  }
+}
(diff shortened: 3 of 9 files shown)


src/lib/config.ts
```diff
@@ -24,6 +24,7 @@ interface Config {
   DRIFTCTL_URL?: string;
   IAC_BUNDLE_PATH?: string;
   IAC_POLICY_ENGINE_PATH?: string;
+  IAC_POLICY_ENGINE_URL?: string;
```
YairZ101, line 27: For visibility, this env variable purpose is to enable local debugging of the policy engine by providing a specific URL to download the binary from.
Love this addition @ofekatr !

src/lib/iac/file-utils.ts
```diff
@@ -9,3 +9,17 @@ export async function isExe(path: string): Promise<boolean> {
     return false;
   }
 }
+
+export async function createDirIfNotExists(path: string): Promise<void> {
+  let pathExists = true;
+
+  try {
+    await fsPromises.stat(path);
```
YairZ101, line 17: why did you change the implementation of this function?

src/lib/iac/test/v2/setup/policy-engine/index.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function initPolicyEngine(testConfig: TestConfig) {
+  let policyEnginePath = await lookupLocal(testConfig);
+
+  const isPolicyEngineCached = !!policyEnginePath;
+
+  if (!isPolicyEngineCached) {
+    policyEnginePath = await download(testConfig);
+  }
+
+  analytics.add('iac-policy-engine-binary-is-cached', isPolicyEngineCached);
```
YairZ101, line 15: I think we might want to tackle `analytics` in another PR/ticket while working together with product about the relevant information that we would like to save.
WDYT?

src/lib/iac/file-utils.ts
ofekatr: Since the `fs.exists` method was deprecated, and since it's generally advised to use async methods rather than sync methods like `existsSync`, I've followed the advice in [the `fs` docs](https://iojs.org/api/fs.html#fs_fs_exists_path_callback), and used the async function `fs.promises.stat`

src/lib/iac/test/v2/setup/policy-engine/index.ts
ofekatr: These analytics use the same properties as DriftCTL, so if the latter was approved it's likely these properties will be used for the UPE as well. 
That being said, I agree. It'd be better to tackle these separately.

src/lib/iac/test/v2/setup/policy-engine/download.ts
```diff
@@ -0,0 +1,146 @@
...
+    `Policy Engine executable fetched successfully from ${policyEngineUrl}`,
+  );
+
+  return {
+    policyEngineBuffer,
+    isUserProvidedUrl: !!testConfig.userPolicyEngineUrl,
+  };
+}
+
+function assertValidChecksum(policyEngineBuffer: Buffer): void {
```
YairZ101, line 76: nit: WDYT about moving all the checksum logic to the `util.ts` file?

ofekatr: Checksum validation is only relevant when downloading the executable, so I prefer to keep it in the downloading logic's scope in the code.

src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
```diff
@@ -0,0 +1,29 @@
+import { TestConfig } from '../../../types';
+import { downloadResource } from '../utils';
```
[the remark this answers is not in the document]
ofekatr: The short answer is encapsulation. Since there was no need for consumers of this module to be aware of the deeper levels in the files hierarchy of `utils`, I've surfaced the relevant exports to the `index.ts` file. It's similar to public class methods, and it makes the consumption of `utils` easier.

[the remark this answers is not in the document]
aron: Sorry for a drive-by review @ofekatr, I just wanted to share a counter point here based on my past experience. I think your logic makes perfect sense for external dependencies or where you want clear encapsulation between two distinct libraries with different owning teams.

The drawback of the index (import/re-export) approach within the same library/codebase is that it provides two entry-points to import the same function. This makes it very easy to create circular dependencies and very hard to then debug & untangle them. I've worked on several projects in the past that used this pattern that eventually had to revert to a policy of one function = one import path. It makes things much easier to reason about long term.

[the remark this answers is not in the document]
ofekatr: You raised a great point @aron, I've removed this `index.ts` file, and updated the import to be from `utils/download-resource`

src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
```diff
@@ -0,0 +1,124 @@
...
+    policyEngineDataBuffer = await fetchCacheResource(policyEngineUrl);
+  } catch (err) {
+    throw new FailedToDownloadPolicyEngineError();
+  }
+  debugLog('Policy Engine executable was fetched successfully');
+
+  return policyEngineDataBuffer;
+}
+
+export const policyEngineUrl = `https://static.snyk.io/cli/iac/test/v${policyEngineReleaseVersion}/${policyEngineFileName}`;
```
YairZ101, line 55: move to `constants.ts`?

src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
```diff
@@ -0,0 +1,124 @@
...
+    super(`Failed to download cache resource from ${policyEngineUrl}`);
+    this.code = IaCErrorCodes.FailedToDownloadPolicyEngineError;
+    this.strCode = getErrorStringCode(this.code);
+    this.userMessage =
+      `Could not fetch cache resource from: ${policyEngineUrl}` +
+      '\nEnsure valid network connection.';
+  }
+}
+
+function assertValidChecksum(dataBuffer: Buffer): void {
```
YairZ101, line 68: can you make this a more generic function that we could potentially reuse in the future?

test/jest/unit/lib/iac/test/v2/setup/local-cache/policy-engine/download.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  it('caches the fetched cache resource', async () => {
+    // Arrange
+    const testTestConfig: TestConfig = cloneDeep(defaultTestTestConfig);
+    const testDataBuffer = Buffer.from('test-data-buffer');
+
+    jest
+      .spyOn(localCacheUtils, 'fetchCacheResource')
+      .mockImplementation(async () => testDataBuffer);
+    jest.spyOn(crypto, 'createHash').mockImplementation(
```
YairZ101, line 72: you used in multiple places `mockImplementation` to return a single value without making any specific calculations, you can use `mockResolvedValue` or `mockReturnValue` to achieve the same thing but in a much simpler way.

src/lib/iac/test/v2/setup/local-cache/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { makeRequest } from '../../../../../request';
+import { CustomError } from '../../../../../errors';
+
+export async function fetchCacheResource(url: string): Promise<Buffer> {
+  const { res, body: cacheResourceBuffer } = await makeRequest({
```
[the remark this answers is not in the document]
ofekatr: What would the benefit be in your opinion? 
A catch block here will re-throw an error. Replacing the original error with a custom one will make us lose information about what made `fetchCacheResource` fail, making it harder to troubleshoot. 
I think re-throwing errors should require a valuable reason, such as replacing an internal error for a user-facing one. Since this utility function is only meant to throw internal errors, custom or not, there wouldn't be much value in catching an internal error and throwing an internal non-custom error.

src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
ofekatr: Should we need the function in the future, we can extract it and generalize it. 
I don't think it'd be worth prematurely refactoring it to be general.

test/jest/unit/lib/iac/test/v2/setup/local-cache/policy-engine/download.spec.ts
ofekatr: I prefer preventing the execution of the actual function, to prevent any possible side-effects and to make the run time a bit faster in the case of time consuming operations, such as hash calculations.

src/lib/iac/test/v2/setup/local-cache/policy-engine/constants/index.ts
```diff
@@ -3,9 +3,11 @@ import { formatPolicyEngineFileName } from './utils';
 /**
  * The Policy Engine release version associated with this Snyk CLI version.
  */
-export const releaseVersion = '0.1.0';
+export const policyEngineReleaseVersion = '0.1.0';
```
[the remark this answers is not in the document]
ofekatr: I wanted to create a separation between the `formatPolicyEngineFileName` utility function used for generating the `policyEngineFileName` constant, and `policyEngineFileName` itself. 
Having both of these exportable from the save level will create confusion about which should be imported and used.

test/jest/unit/lib/iac/test/v2/setup/local-cache/policy-engine/download.spec.ts
YairZ101: In both scenarios you are preventing the execution of the actual function, the difference is that `mockResolvedValue` and `mockReturnValue` are more readable.

src/lib/iac/test/v2/setup/local-cache/utils.ts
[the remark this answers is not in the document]
YairZ101: > Replacing the original error with a custom one will make us lose information about what made fetchCacheResource fail, making it harder to troubleshoot.

good point.

src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
YairZ101: > Generally, I think resources should sit on the lowest level at which they are being used.

sounds good to me :)

src/lib/iac/test/v2/setup/local-cache/utils.ts
jahed-snyk, line 5: suggestion: If these downloads are large, consider using `streamRequest` instead which @craigfurman recently introduced in https://github.com/snyk/cli/pull/3317

craigfurman: @francescomari actually gets credit for that, but yes it's probably a good idea to stream unless we have a good reason not to

jahed-snyk: My bad 😅 Got confused for a bit as I could've sworn I reviewed it. It's actually this commit: https://github.com/snyk/cli/pull/3348

YairZ101 (approved):
🕍

src/lib/iac/test/v2/setup/local-cache/utils.ts
ofekatr: Thank you for the suggestion @jahed-snyk. 
Our downloading flow consists of the following steps:
1. Fetching the executable.
2. Asserting the validity of its checksum.
3. Saving the executable to the files system.

Since step #2 requires the full content of the executable, it cannot begin before the executable is fully fetched. Do you think fully aggregating streamed response before proceeding to step #2 still has a benefit in this case?

Happy to discuss this further and change the implementation if need be 🙂

jahed-snyk: There's value if there are intermediate steps which can take advantage of streams. IIRC, #3348 benefits from streaming extraction. Checksum computation can also be streamed and so can writing to the filesystem.

NodeJS docs has some examples of streaming with crypto lib: https://nodejs.org/api/crypto.html#class-cipher

jahed-snyk: On benefits of streaming, this should cover the main ideas: https://nodejs.dev/learn/nodejs-streams To summarise, by processing large files in smaller pieces asynchronously, we can reduce memory usage and processing time.

Whether it's worth pursuing depends on the size of your data and how slow each step is. Benefits aren't obvious on smaller downloads since either way it'd be pretty fast and low on memory usage.

YairZ101: > Checksum computation can also be streamed

In our use case, we need to compute the checksum for the whole file we downloaded because we check if it equals the expected checksum and only if it matches the expected value we save it locally.
I've never used streams so I might be wrong here but if understand it correctly we don't have a use case for it in our case because we need to wait for the whole file to get streamed to compute the checksum for it (which is similar to `makeRequest`).
Please correct me if I misunderstood anything I would love to fully understand the way we should use streams.

YairZ101: > meaning that the HTTP response body has been copied to the file

To make sure I fully understand, does it mean we need to save the HTTP response body to a local file?

francescomari: Yes, this is how it would work.

ofekatr: Are there any security risks with this approach we should be wary of? 
The file we will save to the user's file system will not be authenticated at first.

YairZ101: > Yes, this is how it would work.

In this case, I don't think we should use this approach because we will save something that we didn't validate before.

ofekatr: [CFG-2020](https://snyksec.atlassian.net/browse/CFG-2020)  was created for updating the downloading implementation to use streaming

PR #3330, opened 2022-06-20
refactor: IaC test v2 setup

ofekatr:

#### What does this PR do?

- Fixes the locally cached Policy Engine executable name.
- Adds a local cache initialization step to the test setup which:
    - Wraps the logic for initializing the rules bundle and the Policy Engine executable.
    - Ensures the IaC local cache directory exists.
    - Replaces the specific cache paths for the Policy Engine executable and the rules bundle with an IaC cache path, as it's enough for the former two can be calculated with it.

- The only logic piece this PR should change, is the name of the locally cached Policy Engine executable the CLI will look for.

- Ensure the IaC test v2 tests still pass.
- Ensure the test files are structured correctly following the refactor.
- Ensure the IaC test v2 flow still behaves the same as before.

- The refactor in this PR was extracted from #3327, where we add the logic for downloading the Policy Engine executable when it is not found locally.

- [CFG-1885](https://snyksec.atlassian.net/browse/CFG-1885)

### DIFF
--- a/src/lib/iac/test/v2/setup/local-cache/index.ts
+++ b/src/lib/iac/test/v2/setup/local-cache/index.ts
@@ -1,0 +1,19 @@
+import { TestConfig } from '../../types';
+import { initRules } from './rules';
+import { initPolicyEngine } from './policy-engine';
+import { createDirIfNotExists } from '../../../../file-utils';
+import { CustomError } from '../../../../../errors';
+import { FailedToInitLocalCacheError } from '../../../../../../cli/commands/test/iac/local-execution/local-cache';
+
+export async function initLocalCache(testConfig: TestConfig) {
+  try {
+    await createDirIfNotExists(testConfig.iacCachePath);
+
+    const policyEnginePath = await initPolicyEngine(testConfig);
+    const rulesBundlePath = await initRules(testConfig);
+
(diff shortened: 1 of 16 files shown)


src/lib/iac/test/v2/setup/local-cache/index.ts
```diff
@@ -0,0 +1,19 @@
+import { TestConfig } from '../../types';
+import { initRules } from './rules';
+import { initPolicyEngine } from './policy-engine';
+import { createDirIfNotExists } from '../../../../file-utils';
+import { CustomError } from '../../../../../errors';
+import { FailedToInitLocalCacheError } from '../../../../../../cli/commands/test/iac/local-execution/local-cache';
+
+export async function initLocalCache(testConfig: TestConfig) {
+  try {
+    await createDirIfNotExists(testConfig.iacCachePath);
```
YairZ101, line 10: should we let the `download-engine` and `download-rules` functions to handle the creation of the directory?

YairZ101 (approved):
image

src/lib/iac/test/v2/setup/local-cache/index.ts
YairZ101: okay sounds good to me

YairZ101 (approved):

PR #3331, opened 2022-06-20
chore: scan IaC files with the Policy Engine

francescomari:

#### What does this PR do?

This PR updates the experimental `iac test` command to execute `snyk-iac-test`, passing to it the rules bundle and the paths to be scanned. The experimental command collects and parses the standard output of the executable, and collects its standard error to be printed to a debug logger.

Follow the flow from `src/cli/commands/test/iac/v2/index.ts`. The execution of `snyk-iac-test` is in `src/lib/iac/test/v2/scan.ts`.

Set the environment variables to point to the engine and the bundle:

```
export SNYK_IAC_POLICY_ENGINE_PATH=...
export SNYK_IAC_BUNDLE_PATH=...
```

Make sure you have the `iacCliUnifiedEngine` feature flag enabled, and invoke the command like this:

```
snyk iac test --experimental
```

[CFG-1896](https://snyksec.atlassian.net/browse/CFG-1896)

### DIFF
--- a/src/cli/commands/test/index.ts
+++ b/src/cli/commands/test/index.ts
@@ -61,7 +61,7 @@ export default async function test(
       (await hasFeatureFlag('iacCliUnifiedEngine', options)) &&
       options.experimental
     ) {
-      return await iacTestCommandV2.test();
+      return await iacTestCommandV2.test(...args);
     } else {
       return await iacTestCommand(...args);
     }
--- a/src/cli/commands/test/iac/local-execution/types.ts
+++ b/src/cli/commands/test/iac/local-execution/types.ts
@@ -349,9 +349,13 @@ export enum IaCErrorCodes {
 
   // Rules bundle errors.
   InvalidUserRulesBundleError = 1130,
+  BundleNotFoundError = 1131,
 
   // Unified Policy Engine executable errors.
   InvalidUserPolicyEnginePathError = 1140,
+
+  // Scan errors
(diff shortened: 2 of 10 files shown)


src/cli/commands/test/index.ts
```diff
@@ -61,7 +61,7 @@ export default async function test(
       (await hasFeatureFlag('iacCliUnifiedEngine', options)) &&
       options.experimental
     ) {
-      return await iacTestCommandV2.test();
+      return await iacTestCommandV2.test(...args);
```
YairZ101, line 64: the function `processCommandArgs` above takes care of processing the command and returning the paths and the options, we can use it instead of creating our own implementation

francescomari: I decided to follow the same approach of the current version of the `iac test` command, where we do our option parsing independently, I thought we wanted to keep the same external interface for the current and the experimental version of the command.

YairZ101: in the current implementation, we are not really doing our own option parsing, we are still using `processCommandArgs` simply in a different place.

robcresswell (approved):

YairZ101 (approved):

src/cli/commands/test/index.ts
YairZ101: WDYT about changing both of the functions to get the `options` and `paths` as arguments instead of reusing `processCommandArgs` in each one of them?

PR #3403, opened 2022-06-27
chore: bump `snyk-iac-test` version

YairZ101:

### DIFF
--- a/src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
+++ b/src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
@@ -81,18 +81,18 @@ function assertValidChecksum(dataBuffer: Buffer): void {
 }
 
 export const policyEngineChecksum = {
-  'snyk-iac-test_0.1.0_Linux_x86_64':
-    '0b0d846cd74bf42676f79875ab30f20dc08529aedb94f2f6dd31a67c302b78e4',
-  'snyk-iac-test_0.1.0_Darwin_x86_64':
-    '896960a09b6adf699185f443428cda25ccb30123440bc8735e1a05df1e9cbc12',
-  'snyk-iac-test_0.1.0_Windows_x86_64.exe':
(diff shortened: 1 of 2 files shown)


src/lib/iac/test/v2/setup/local-cache/policy-engine/download.ts
```diff
@@ -81,18 +81,18 @@ function assertValidChecksum(dataBuffer: Buffer): void {
...
-    '896960a09b6adf699185f443428cda25ccb30123440bc8735e1a05df1e9cbc12',
-  'snyk-iac-test_0.1.0_Windows_x86_64.exe':
-    'a0b0b5781218f42d121cd1c6bca2d5ea27fbe76b1e9817e37745586c276cabda',
-  'snyk-iac-test_0.1.0_Linux_arm64':
-    'b0c0bd9a06cc3d556b526b868e5a4ac4c5a3938899c4e312607f6828626debe9',
-  'snyk-iac-test_0.1.0_Darwin_arm64':
-    'cf7a327378983810ea043774a333bdd724e3866a815555d59ec5cd8aa25ea5ec',
-  'snyk-iac-test_0.1.0_Windows_arm64.exe':
-    'd88b24c611c8d37c2df9382e6e1b39933b926ee2ed438f1fdb69c51da5086fc5',
+  'snyk-iac-test_0.2.0_Linux_arm64':
```
ofekatr, line 84: nit: Do you think we should use the inline version from `policyEngineReleaseVersion` so that we don't need to update each release name every time we bump the version manually?

ofekatr: We would still need to update each checksum, so it might not be worth it

YairZ101: > We would still need to update each checksum, so it might not be worth it

+1
the big pain point here is updating the checksum manually, updating the names is the easy part.
we can sit one day and try to think of a better way to do this checksum validation but I think currently this solution fits us because its quick and dirty and we can progress faster with the integration

ofekatr (approved):

PR #3406, opened 2022-06-27
fix: move checkPaths() function out of main()

aron:

#### What does this PR do?

Moves the `checkPaths()` function outside of the `main()` function. This is only used to verify that a path to a package manifest has not been provided without the `--file` flag. Other commands like `snyk iac test` do require filepaths to be passed as args and this check causes issues where we cannot look for misconfigurations in a file called package.json for example.

The fix is to move the implementation outside of the core flow into `test` and `monitor`. I've also excluded Container by checking for `options.docker`.

Run the following:

    snyk test package.json # should error with --file message

Then test:

    snyk iac test package.json # should not error

image


src/cli/commands/monitor/index.ts
```diff
@@ -79,6 +80,10 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
     );
   }
 
+  if (!options.docker) {
+    checkOSSPaths(options);
```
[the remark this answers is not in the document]
jahed-snyk: In terms of args processing flow, beyond `processCommandArgs`, `options._` currently is never used AFAIK, `paths` is instead. Since `paths` is what further steps use I think we should check that rather than `options._`.

The existence of `option._` might be unintentional or required for some backwards compatibility. 🤔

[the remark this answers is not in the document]
aron: Okay, checked this out, `paths` works, so if it's more future proof then let's go with that.

aron:
> Should IaC add a test to avoid regressions?

Done

test/jest/acceptance/iac/check-path-regression.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  beforeAll(async () => {
+    const result = await startMockServer();
+    run = result.run;
+    teardown = result.teardown;
+  });
+
+  afterAll(async () => teardown());
+
+  it('supports scanning a project matching an OSS manifest name', async () => {
+    const { stdout, exitCode } = await run(`snyk iac test ../../package.json`);
```
ofekatr, line 20: suggestion: [CFG-1973](https://snyksec.atlassian.net/browse/CFG-1973) which is planned to be picked up after July 11th will render input paths outside the current working directory invalid. 
The changes for this ticket will likely break this test.

While we could revisit this test and update the input path to be within the CWD when doing this ticket, maybe we could do it in this PR?

test/jest/acceptance/iac/check-path-regression.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { startMockServer } from './helpers';
+
+jest.setTimeout(50000);
+
+describe('checkPath() regression test snyk/cli#3406', () => {
```
ofekatr, line 5: nit: The paths in the original function's name was in plural.
```suggestion
describe('checkPaths() regression test snyk/cli#3406', () => {
```

src/lib/check-paths.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  MissingTargetFileError,
+  UnsupportedOptionCombinationError,
+} from './errors';
+import { isPathToPackageFile } from './detect';
```
ofekatr, line 5: suggestion: This function is only being called in the `checkOSSPaths` function. 
Rather than importing it, can it be placed in `src/lib/check-paths.ts` instead?

ofekatr (approved):
Commented small suggestions and nits, none should block :shipit:
Feel free to review these with no obligation to respond to any of them.

src/lib/check-paths.ts
aron: I'd rather not, as I'd then need to export all the constants used by isPathToPackageFile

orsagie (approved):
Looks good to me

PR #3415, opened 2022-06-27
Improve API URL configuration behavior

JackuB:
Allright, over 300 lines, but the change is actually not that big.

**Goal is to allow users to set a single configuration option pointing to an API endpoint and let CLI figure out the rest.**

With the REST API being the main (and only) API path used in the near future, I think it make sense to settle on the **default as `https://api.snyk.io/rest/` as per https://apidocs.snyk.io**

For example:

```console
SNYK_API=https://api.custom-domain.dev snyk test
```

Additional benefits introduced in this PR:
- making a hierarchy of envvar vs. config option a bit clearer and communicated to the user
- better error AND debug messages for all options (currently we don't show these!)
- no need to configure any other envvar/option by default. But still offer an override for local development and specific use cases.

There are still a few weakpoints, like manipulating the api to app subdomain strings. But I'd like to sort those out in the CLI v2 configurations and then just update CLI v1 to consume those.

Merging this PR is slightly risky, because it needs to take account specific use cases in the wild. 
I'd like to extend the test coverage if we'll find any other domains/use cases in the wild.

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -1,6 +1,4 @@
 {
-  "API": "https://snyk.io/api/v1",
-  "API_REST_URL": "https://api.snyk.io/rest",
   "devDeps": false,
   "PRUNE_DEPS_THRESHOLD": 40000,
   "MAX_PATH_COUNT": 1500000,
--- a/src/cli/commands/auth/index.ts
+++ b/src/cli/commands/auth/index.ts
@@ -1,4 +1,3 @@
-import * as url from 'url';
 import * as open from 'open';
 import { v4 as uuidv4 } from 'uuid';
 import * as Debug from 'debug';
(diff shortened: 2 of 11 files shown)


src/lib/code-config.ts
```diff
@@ -6,7 +6,7 @@ export function getCodeClientProxyUrl() {
   const domain = url.origin;
   return (
     config.CODE_CLIENT_PROXY_URL ||
-    domain.replace(/\/\/(app\.)?/, '//deeproxy.')
+    domain.replace(/\/\/(ap[pi]\.)?/, '//deeproxy.')
```
PeterSchafer, line 9: question: Are there tests that cover these changes?

src/cli/commands/auth/index.ts
```diff
@@ -29,7 +31,9 @@ function resetAttempts() {
 async function webAuth() {
   const token = uuidv4(); // generate a random key
 
-  let urlStr = authUrl + '/login?token=' + token;
+  apiUrl.pathname = '/login';
+  apiUrl.searchParams.append('token', token);
+  let urlStr = apiUrl.toString();
```
PeterSchafer, line 36: question: Are there tests that cover these changes?

JackuB: This is currently covered by one tap test and a Smoke test

PeterSchafer (approved):

src/lib/code-config.ts
metju90, line 9: Would appreciate if you can save me some time by explaining what was changed in this line?

config.default.json
```diff
@@ -1,6 +1,4 @@
 {
-  "API": "https://snyk.io/api/v1",
```
metju90, line 2: Why are we removing this?

What will be the default value if we run CLI from our local machines?

gitphill (approved):

metju90 (approved):

PR #3417, opened 2022-06-28
feat: container json response with app vulns

yaronschwimmer:

#### What does this PR do?

We want to support scanning app vulns by default. Before we can do that,
we have to support JSON response without the `experimental` flag, and
without creating any breaking changes for existing users. The solution
is to add another field `applications` to the response object, which
will contain all of the app vulns results.

https://snyksec.atlassian.net/browse/MYC-101


mladkau (approved):
Good luck with the new format!!! Great that we finally have a go ...

mladkau (approved):

PeterSchafer (approved):

PR #3418, opened 2022-06-28
chore: Rename policy-engine to test-engine

ofekatr:

#### What does this PR do?

- Renames policy-engine to test-engine

We’re currently referring to the `snyk-iac-test` executable as the Policy Engine executable in the code, [example](https://github.com/snyk/cli/blob/master/src/lib/iac/test/v2/setup/local-cache/policy-engine/index.ts).
This is misleading, as we’re not actually downloading and using the UPE directly, but rather the `snyk-iac-test`, which is meant to perform multiple steps in the test flow.

#### Additional information

- [Slack thread](https://snyk.slack.com/archives/C02JMMTLUF9/p1656346281440879)


francescomari (approved):

aron:
Suggestion: what are CLI using to refer to the "plugins" in the CLI v2? Would be good to align on the naming here? test-engine is a good name but would snyk-iac-test-plugin be more forward looking?

PeterSchafer (approved):

ofekatr:
> Suggestion: what are CLI using to refer to the "plugins" in the CLI v2? Would be good to align on the naming here? test-engine is a good name but would snyk-iac-test-plugin be more forward looking?

@PeterSchafer Could help us with this one?

From team Config's side we're still not 100% sure what the scope of the `snyk-iac-test` executable would be. 
Our current viewpoint is that it would not cover all of the CLI IaC steps, but rather it serves as the engine holding most of the IaC test execution's business logic for the scanning and result potentially the processing step (currently in discussion). [This diagram](https://miro.com/app/board/uXjVOt3RXo0=/) should make this clearer. 

Since:
- The logic for the test's setup and post-test steps isn't currently being planned to get included in this executable.
- The logic for scanning and result processing could be shared with Snyk Cloud, while the setup and teardown could not.
- Plugins are meant to be registered and applied a bit differently, and will mostly be referred in the Golang implementation of CLI v2. I'm not sure whether the plugin would be per IaC command or be an IaC plugin as a whole.

francescomari:
I don't think we should focus too much on naming right now. There are too many unknowns re. the CLIv2, the scope of `snyk-iac-test` and the overall CLI flow we are implementing to make a naming choice that is both accurate and forward-looking. Since the name of `snyk-iac-test` is only referred in two places (in the `snyk-iac-test` repository itself and in a constant in the `snyk` CLI), and it's not a user-facing name, I think it's alright to leave naming decision for later.

ofekatr:
> I think it's alright to leave naming decision for later.

Agreed, closing the PR for now

PR #3424, opened 2022-06-29
chore: Error handling PoC

ofekatr:

#### What does this PR do?

- A PoC for improving error handling for chained errors, including:
    - Chaining error stacks in debug logs.
    - Ensuring the original user message is displayed.

```
src/lib/errors/nested-custom-error.ts
```
### Notes for the reviewer

- The initial idea was to use the optional `cause` property for the `Error` constructor, but following [this discussion](https://snyk.slack.com/archives/C0127HWU0E7/p1656493212716839), it was agreed that since it was only introduced in the latest `Node.js` versions, we cannot apply it for this implementation. Instead, we made our own implementation to mimic the behavior.
- To prevent any unexpected behaviors with the existing errors, I've introduced a new error class, called `NestedCustomError`.
- The new properties introduces with `NestedCustomError` are:
    - `nestedName`: Gets the original error name from the deepest chained error.
    - `nestedMessage`: Chains the error's message with all the messages of its nested errors.
    - `nestedCode`: Gets the original error code from the deepest chained error.
    - `nestedStrCode`: Gets the original error string code from the deepest chained error.
    - `nestedUserMessage`: Gets the original user message from the deepest chained error.
    - `nestedStack`: Chains the error's stack with all the stacks of its nested errors.
 

- throw a `NestedCustomError` error with a chain of internal errors.
- Ensure the error stacks are chained correctly in the debug logs.
- Ensure the original user message is displayed to the users.

#### What are the relevant threads?

- #ask-team-hammer:
    - https://snyk.slack.com/archives/C0127HWU0E7/p1655816867354399
    - https://snyk.slack.com/archives/C0127HWU0E7/p1656493212716839
-  #team-iac-config:
    - https://snyk.slack.com/archives/C02JMMTLUF9/p1655816525274439

#### Screenshots

image

image

### DIFF
--- a/src/cli/main.ts
+++ b/src/cli/main.ts
@@ -84,6 +84,15 @@ async function runCommand(args: Args) {
 
 async function handleError(args, error) {
   spinner.clearAll();
+
+  if (typeof error === 'object') {
+    error.stack = error.nestedStack || error.stack;
+    error.userMessage = error.nestedUserMessage || error.userMessage;
+    error.code = error.nestedCode || error.code;
+    error.strCode = error.nestedStrCode || error.strCode;
+    error.userMessage = error.nestedUserMessage || error.userMessage;
+  }
+
   let command = 'bad-command';
   let exitCode = EXIT_CODES.ERROR;
 
--- a/src/lib/errors/nested-custom-error.ts
+++ b/src/lib/errors/nested-custom-error.ts
@@ -1,0 +1,67 @@
+import { CustomError } from './custom-error';
+
+export class NestedCustomError extends CustomError {
+  constructor(message: string, innerError?: any) {
+    super(message);
+    this.innerError = innerError;
+  }
+
+  public get nestedName(): string {
+    return this.innerError?.nestedName || this.innerError?.name || this.name;
+  }
+
+  public get nestedStack(): string | undefined {
+    const stackComponents: string[] = [];
+    if (this.stack) {
+      stackComponents.push(this.stack);
+    }
+
+    const nestedStack = this.innerError?.nestedStack || this.innerError?.stack;
+    if (nestedStack) {
+      stackComponents.push(nestedStack);
+    }
+
+    return stackComponents.join('\n' + ' '.repeat(2) + '[cause]: ');
+  }
+
+  public get nestedMessage(): string {
+    let message = this.message;
+
+    const innerErrorMessage =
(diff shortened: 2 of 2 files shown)


src/cli/main.ts
```diff
@@ -84,6 +84,15 @@ async function runCommand(args: Args) {
 
 async function handleError(args, error) {
   spinner.clearAll();
+
+  if (typeof error === 'object') {
```
PeterSchafer, line 88: Question: Is there a reason why this is not part of NestedCustomError? From a coupling point of view, a consumer of an error should not need to be required to handle different cases and know about the specific error implementations.

ofekatr: Great input, @PeterSchafer. 
This here is my workaround after not accomplishing this behavior in the `NestedCustomError` class. 
My initial approach was creating getters for these properties which exist as data members in `CustomError`. 
However, the code consistently used the data from `CustomError` rather than the getters with the same names in `NestedCustomError`. 

I would much prefer an implementation that contains this logic in the scope of the `NestedCustomError` class, so any suggestions would absolutely be on the table.

PeterSchafer: suggestion: Assuming that an instance of `CustomError` or `NestedCustomError` is "immutable" in the sense that it gets created and not changed afterwards, I know the members are public, then I think it should work to move the logic into the constructor of `NestedCustomError`. Because at the point where you create a `NestedCustomError`, you already have the Error you want to nest and thereby know how the final error with all its nested errors looks like.
In this sense, I think it would also be possible to merge the functionality into the constructor of `CustomError` and not have `NestedCustomError` as an additional class at all.

I hope this makes sense?!

ofekatr: > Assuming that an instance of CustomError or NestedCustomError is "immutable"
Unfortunately, that's not the case, and there are several instances where an error instance is modified after it's been created.

PeterSchafer (approved):

src/cli/main.ts
ofekatr: I'm concerned this sort of encapsulation may end up adding confusion for the consumers. 
The original intention was to ensure this nested error still holds both its own properties and its nested properties so that both are still usable. 
If we add an additional method like `getNestedProperties` or a getter named `nestedProperties` or a method that returns an error object with the nested properties as the error's own properties, this is likely to complicate the usage more than needed.

PR #3436, opened 2022-07-04
chore: cleanup iac custom rules analytics [CFG-1144]

teodora-sandu:

#### What does this PR do?

Removes an unused metric: https://snyksec.atlassian.net/browse/CFG-1144

PR to remove it from `registry`: https://github.com/snyk/registry/pull/28021
BI ticket that removed it from Looker: https://snyksec.atlassian.net/browse/B1-1341


YairZ101 (approved):
Quick question: does it mean that the col that held this metric in table will also get removed? and if so, does it happen automatically or someone needs to do that manually?

YairZ101:
> Do you think we should?

I don't think it is super important and if there are other unused cols there then I think ours can stay.
Thx for clarifying this :)

PR #3448, opened 2022-07-11
Feat/update vuln urls to point to security.snyk.io

e102:
#### What does this PR do?
Updates URL links for vulns in the CLI output to point to security.snyk.io, the new snyk vulnerability db which generally has better UX and more info than the legacy site the links are currently pointing to.

### DIFF
--- a/src/lib/spotlight-vuln-notification.ts
+++ b/src/lib/spotlight-vuln-notification.ts
@@ -41,7 +41,7 @@ export function notificationForSpotlightVulns(
           );
 
           for (const vulnId of foundSpotlightVulnsIds) {
-            message += `  - ${vulnId} (See https://snyk.io/vuln/${vulnId})`;
+            message += `  - ${vulnId} (See https://security.snyk.io/vuln/${vulnId})`;
           }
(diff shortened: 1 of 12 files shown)


src/lib/spotlight-vuln-notification.ts
```diff
@@ -41,7 +41,7 @@ export function notificationForSpotlightVulns(
           );
 
           for (const vulnId of foundSpotlightVulnsIds) {
-            message += `  - ${vulnId} (See https://snyk.io/vuln/${vulnId})`;
+            message += `  - ${vulnId} (See https://security.snyk.io/vuln/${vulnId})`;
```
thisislawatts, line 44: **issue(blocking):** Whilst we're in this area can we switch to using the configuration value rather than a hard coded string?

For example:

```suggestion
            message += `  - ${vulnId} (See ${config.PUBLIC_VULN_DB_URL}/vuln/${vulnId})`;
```

Avishagp (approved):

orsagie (approved):

PR #3459, opened 2022-07-14
fix: fix broken test after security URL change

Avishagp:
Fix broken test following security URL change


maxjeffos:
@Avishagp I'd suggest making your commit message a `test:` rather than a `fix:`

PR #3467, opened 2022-07-15
test: kerberos/jest setup

PeterSchafer:

#### What does this PR do?
In this PR we introduce acceptance tests for proxy authentication implemented in CLIv2. The proxy authentication currently supports [SPNEGO based authentication](https://www.ietf.org/rfc/rfc4559.txt) and thereby mainly kerberos (all platforms) and NTLM (on Windows machines). 
The PR introduces a Docker based test environment spinning up a linux container hosting a Squid Proxy and a Kerberos KDC, with which the executable under test interacts.

Unfortunately due to limitations on our CircleCi Runners, the tests currently only run for linux builds.

The test suite can be found here cliv2/test/acceptance/proxy_authentication.spec.ts

### DIFF
--- a/cliv2/Makefile
+++ b/cliv2/Makefile
@@ -169,8 +169,12 @@ whiteboxtest:
 	@echo "$(LOG_PREFIX) Running $@"
 	@$(GOCMD) test -cover ./...
 
+.PHONY: acceptancetest 
+acceptancetest: build
+	TEST_SNYK_COMMAND="$(BUILD_DIR)/$(V2_EXECUTABLE_NAME)" npx jest
+
 .PHONY: test
-test: whiteboxtest blackboxtest
+test: whiteboxtest blackboxtest acceptancetest
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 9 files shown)


cliv2/Makefile
```diff
@@ -169,8 +169,12 @@ whiteboxtest:
 	@echo "$(LOG_PREFIX) Running $@"
 	@$(GOCMD) test -cover ./...
 
+.PHONY: acceptancetest 
+acceptancetest: build
+	TEST_SNYK_COMMAND="$(BUILD_DIR)/$(V2_EXECUTABLE_NAME)" npx jest
+
 .PHONY: test
-test: whiteboxtest blackboxtest
+test: whiteboxtest blackboxtest acceptancetest
```
maxjeffos, line 177: I'm not seeing where `acceptancetest` (in this context) is every invoked and it isn't in the Makefile help.
I suggest either getting rid of it, or changing renaming `blackboxtest` to `acceptancetest` (since acceptance test is usually the term we on Hammer use to refer to blackbox tests).

maxjeffos (approved):

cliv2/Makefile
PeterSchafer: In a Makefile the syntax is the target name, in this case "test" and the targets it depends on "whiteboxtest blackboxtest acceptancetest". This means that acceptancetest is being checked and if necessary run when test is checked.

PR #3469, opened 2022-07-18
fix: container app vulns json with experimental flag

yaronschwimmer:

#### What does this PR do?
In a recent [PR](https://github.com/snyk/cli/pull/3417) we released a new JSON structure for container tests with app vulns, eliminating the need for the `experimental` flag when doing so. Since some customers are still using the `experimental` flag and expect the output in the older format, we are reverting the JSON output for container app vulns to how it was when using the experimental flag.

https://snyksec.atlassian.net/browse/MYC-139


JackuB:
You could also probably add `test/jest/acceptance/snyk-test/app-vuln-container-project.spec.ts` as owned by Container in the CODEOWNERS

PR #3471, opened 2022-07-19
Chore/add rechable deprecation message

e102:
Add's deprecation warning for `--reachable` flag

### DIFF
--- a/help/cli-commands/monitor.md
+++ b/help/cli-commands/monitor.md
@@ -180,13 +180,13 @@ Auto-detect maven jars, aars, and wars in given directory. To monitor individual
 
 ### `--reachable`
 
-WARNING: Will be deprecated on the 1st of September 2022
+WARNING: Will be deprecated on the 15th of August 2022
 
 Analyze your source code to find which vulnerable functions and packages are called. Cannot be used with `--all-projects`
 
 ### `--reachable-timeout=<TIMEOUT>`
 
-WARNING: Will be deprecated on the 1st of September 2022
+WARNING: Will be deprecated on the 15th of August 2022
 
 Specify the amount of time (in seconds) to wait for Snyk to gather reachability data. If it takes longer than `<TIMEOUT>`, reachable vulnerabilities are not reported. This does not affect regular test or monitor output.
 
--- a/help/cli-commands/monitor.md
+++ b/help/cli-commands/monitor.md
@@ -180,9 +180,13 @@ Auto-detect maven jars, aars, and wars in given directory. To monitor individual
 
 ### `--reachable`
 
+WARNING: Will be deprecated on the 1st of September 2022
+
 Analyze your source code to find which vulnerable functions and packages are called. Cannot be used with `--all-projects`
 
 ### `--reachable-timeout=<TIMEOUT>`
+
+WARNING: Will be deprecated on the 1st of September 2022
 
 Specify the amount of time (in seconds) to wait for Snyk to gather reachability data. If it takes longer than `<TIMEOUT>`, reachable vulnerabilities are not reported. This does not affect regular test or monitor output.
 
--- a/help/cli-commands/monitor.md
+++ b/help/cli-commands/monitor.md
@@ -180,13 +180,13 @@ Auto-detect maven jars, aars, and wars in given directory. To monitor individual
 
 ### `--reachable`
 
-WARNING: Will be deprecated on the 15th of August 2022
+WARNING: Will be removed on 15 August 2022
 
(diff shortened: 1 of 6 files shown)


help/cli-commands/monitor.md
```diff
@@ -180,10 +180,14 @@ Auto-detect maven jars, aars, and wars in given directory. To monitor individual
 
 ### `--reachable`
 
+WARNING: Will be deprecated on the 1st of September 2022
```
JackuB, line 183: Will you add these to the User docs as well?

src/lib/reachable-vulns.ts
```diff
@@ -65,3 +66,7 @@ export async function validatePayload(
   }
   return true;
 }
+
+export const reachableVulnsRemovalMessage = theme.color.status.warn(
+  `${theme.icon.WARNING} Starting on August 15th 2022, the reachable flag (--reachable) will not be supported.We are currently developing new and improved capabilities for prioritizing vulnerabilities.Please follow https://updates.snyk.io for news and updates, and we are sorry for the temporary inconvenience.\n`,
```
JackuB, line 71: You mention 1st of September 2022 above, but August 15th here. Should these be the same?

Maybe you should also check with @awileysnyk on what kind of date format we want to use in communications.

help/cli-commands/test.md
```diff
@@ -185,10 +185,14 @@ Auto-detect maven jars, aars, and wars in given directory. To test individually
 
 ### `--reachable`
 
+WARNING: Will be deprecated on the 1st of September 2022
```
cfereday, line 188: **discussion:** why are these different dates for test and monitor?

src/lib/options-validator.ts
```diff
@@ -10,6 +10,16 @@ export async function validateOptions(
   options: (Options & TestOptions) | (Options & MonitorOptions),
   packageManager?: SupportedPackageManagers,
 ): Promise<void> {
+  if (options.reachableVulns) {
```
cfereday, line 13: **discussion:** where are these warning alerts being channelled? Should they be alerting in #apollo-monitoring?

src/lib/reachable-vulns.ts
```diff
@@ -65,3 +66,7 @@ export async function validatePayload(
   }
   return true;
 }
+
+export const reachableVulnsRemovalMessage = theme.color.status.warn(
```
cfereday, line 70: **discussion:** will this appear for both `test` and `monitor`? If so this could be confusing as there are different dates

cfereday (commented):
Left a few comments, final question is: has this been tested locally to validate behaviour works as expected?

src/lib/reachable-vulns.ts
e102: Good catch. Fixed.

The dates and messaging were decided by @taldromi who's the product owner for reachability. Do you think it's also neccessary to get sign off from awiley?

help/cli-commands/monitor.md
e102: I'm not sure what the user docs are. Do you mean the gitbook site? (If so, I thought editing here would update the gitbook automatically)

help/cli-commands/test.md
e102: fixed

src/lib/options-validator.ts
e102: They're not alerts in the alerting and monitoring sense. They're alerts in the "yellow warning in the CLI" sense

src/lib/reachable-vulns.ts
e102: Oh. The different dates were my bad. The dates are meant to be the same. Should be fixed  now.

src/lib/options-validator.ts
cfereday: Ah ok

cfereday (approved):

help/cli-commands/test.md
```diff
@@ -185,10 +185,14 @@ Auto-detect maven jars, aars, and wars in given directory. To test individually
 
 ### `--reachable`
 
+WARNING: Will be deprecated on the 15th of August 2022
```
JackuB, line 188: Suggestion:

```suggestion
WARNING: Will be removed on 15 August 2022
```

My understanding was that it is already deprecated today and on 15th it will be removed?
Also the date format should be the same in all messages, and I'd suggest sticking to the UK format we are using in other places as well.

JackuB (approved):
In general, approving, but left a suggestion on wording and date format

PR #3484, opened 2022-07-25
feat: pass configuration to snyk-iac-test

francescomari:

#### What does this PR do?

This PR adds support for passing configuration to the `snyk-iac-test` executable. This PR includes in the configuration some values (org name, API endpoint and authorization header) that will be used in future changes to share results in `snyk-iac-test`.

The logic for this change is contained in `src/lib/iac/test/v2/scan/index.ts`.

This test is transparent to the user. Running `snyk iac test --experimental` should work as before. If you run that with the `-d` option, you should see a message from `snyk-iac-test` about finding the configuration:

```
2022/07/25 12:15:12 using config: /var/folders/1d/b93fm2nj567fn0bm8f5mlg6h0000gn/T/snyk-16nA3H/config.json
```

[CFG-2044](https://snyksec.atlassian.net/browse/CFG-2044)

### DIFF
--- a/src/lib/iac/test/v2/scan/index.ts
+++ b/src/lib/iac/test/v2/scan/index.ts
@@ -5,6 +5,12 @@ import { IaCErrorCodes } from '../../../../../cli/commands/test/iac/local-execut
 import { getErrorStringCode } from '../../../../../cli/commands/test/iac/local-execution/error-utils';
 import * as newDebug from 'debug';
 import { SnykIacTestOutput } from './results';
+import * as os from 'os';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as rimraf from 'rimraf';
+import config from '../../../../config';
+import { getAuthHeader } from '../../../../api-token';
 
 const debug = newDebug('snyk-iac');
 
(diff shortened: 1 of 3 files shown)


src/lib/iac/test/v2/scan/index.ts
```diff
@@ -47,6 +72,33 @@ export function scan(
   return output;
 }
 
+function createConfig(options: TestConfig): string {
+  try {
+    const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'snyk-'));
```
YairZ101, line 77: why you didn't use our cache dir?

francescomari: The cache dir is usually used for artifacts that you want to create (or download) and reuse multiple times. I don't want to use the config multiple times, I want it scoped to a single execution of `snyk-iac-test` and throw it away. Some of those values might change between one execution and the other anyway, e.g. if the user authenticates the CLI with a different token.

YairZ101: makes sense, thanks for the explanation :)

YairZ101 (approved):
rocket

PR #3501, opened 2022-07-27
chore: Fix IaC test exit code when issues are found

ofekatr:

#### What does this PR do?

- Fixes the exit code for the new IaC test flow to be `1` instead of `0`, as is expected by the Snyk CLI ([source](https://docs.snyk.io/snyk-cli/cli-reference#subcommands-of-cli-commands)).

```
src/lib/iac/test/v2/output.ts
```

- Run `snyk iac test` with a directory/file with issues, in all three formats: text, JSON, SARIF.
- Ensure the exist code is 1.
- Run `snyk iac test --experimental` with a directory/file with issues, in all three formats: text, JSON, SARIF.
- Ensure the exist code is 1.
- Run `snyk iac test` with a directory/file without issues, in all three formats: text, JSON, SARIF.
- Ensure the exist code is 0.
- Run `snyk iac test --experimental` with a directory/file without issues, in all three formats: text, JSON, SARIF. 
- Ensure the exist code is 0.

- At the moment when the new test flow returns issues, the exit code is `0`, but based on the status codes list in the CLI’s help doc:
```
Exit codes
  Possible exit codes and their meaning:

  0: success, no vulnerabilities found
  1: action_needed, vulnerabilities found
  2: failure, try to re-run command
  3: failure, no supported projects detected
```

We should receive `1` when issues were found.
Running `snyk iac test` in all output formats confirms it, as it currently returns exist code `1`.

#### Additional information

- [Slack thread](https://snyk.slack.com/archives/C02JMMTLUF9/p1658908472627959)
- [Snyk CLI docs](https://docs.snyk.io/snyk-cli/cli-reference#subcommands-of-cli-commands)

### DIFF
--- a/src/lib/iac/test/v2/output.ts
+++ b/src/lib/iac/test/v2/output.ts
@@ -16,6 +16,7 @@ import { jsonStringifyLargeObject } from '../../../json';
 import { IacOrgSettings } from '../../../../cli/commands/test/iac/local-execution/types';
 import { SnykIacTestError } from './errors';
 import { convertEngineToSarifResults } from './sarif';
+import { CustomError } from '../../../errors';
 
 export function buildOutput({
   scanResult,
@@ -91,6 +92,11 @@ function buildTestCommandResultData({
     responseData = buildTextOutput({ scanResult, projectName, orgSettings });
   }
 
+  const isFoundIssues = !!scanResult.results?.vulnerabilities?.length;
+  if (isFoundIssues) {
+    throw new FoundIssuesError({ responseData, jsonData, sarifData });
+  }
+
   return { responseData, jsonData, sarifData };
 }
 
@@ -140,4 +146,24 @@ function buildTextOutput({
   response += EOL;
 
   return response;
+}
+
+interface FoundIssuesErrorProps {
+  responseData: string;
+  jsonData: string;
+  sarifData: string;
+}
+
+export class FoundIssuesError extends CustomError {
+  public jsonStringifiedResults: string;
+  public sarifStringifiedResults: string;
+
+  constructor(props: FoundIssuesErrorProps) {
+    super(props.responseData);
+    this.code = 'VULNS' as any;
+    this.strCode = 'VULNS';
+    this.userMessage = props.responseData;
+    this.jsonStringifiedResults = props.jsonData;
+    this.sarifStringifiedResults = props.sarifData;
+  }
 }
--- a/test/jest/unit/cli/commands/test/iac/v2/index.spec.ts
+++ b/test/jest/unit/cli/commands/test/iac/v2/index.spec.ts
@@ -10,6 +10,7 @@ import { test } from '../../../../../../../../src/cli/commands/test/iac/v2/index
 import { Options, TestOptions } from '../../../../../../../../src/lib/types';
 import { isValidJSONString } from '../../../../../../acceptance/iac/helpers';
 import { IacOrgSettings } from '../../../../../../../../src/cli/commands/test/iac/local-execution/types';
+import { FoundIssuesError } from '../../../../../../../../src/lib/iac/test/v2/output';
 
 jest.setTimeout(1000 * 10);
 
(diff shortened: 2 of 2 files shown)


YairZ101 (approved):
LGTM
the rest of the exit codes are working as expected?

src/cli/commands/test/iac/scan.ts
```diff
@@ -82,6 +82,11 @@ export async function scan(
         assertIaCOptionsFlags(process.argv);
 
         if (pathLib.relative(projectRoot, path).includes('..')) {
+          console.log('\n', {
```
YairZ101, line 85: console.log

ofekatr:
> the rest of the exit codes are working as expected?
The exit code for failures is `0`, which is incorrect. 
IIUC, we intend to have additional work on failures, after which we ensure the correct exit code, which is `2`

src/lib/iac/test/v2/output.ts
```diff
@@ -91,6 +92,11 @@ function buildTestCommandResultData({
     responseData = buildTextOutput({ scanResult, projectName, orgSettings });
   }
 
+  const isFoundIssues = !!scanResult.results?.vulnerabilities?.length;
+  if (isFoundIssues) {
+    throw new FoundIssuesError({ responseData, jsonData, sarifData })
```
YairZ101, line 97: so to make sure that we exit withe code `1` we need to throw an error? is it what the other products are doing?

YairZ101:
> IIUC, we intend to have additional work on failures, after which we ensure the correct exit code, which is 2

What stops us from fixing the other 2 error codes now as well?

src/lib/iac/test/v2/output.ts
ofekatr: Correct, in the local-execution flow, we did so to align with the other products. 
The way this error is handled can be viewed in `src/cli/main.ts`.
Team Hammer also confirmed this recently [here](https://snyk.slack.com/archives/C0127HWU0E7/p1658936937633039?thread_ts=1658930596.107169&cid=C0127HWU0E7).

ofekatr:
- When there are no issues, we get `0` as expected.
- With this PR, when there are issues, we would get `1` as expected.
- When the CLI fails, we automatically get `2` as expected. However, in case all of the tested files failed, we should likely return `2` like in local-execution, but we should discuss with the team about how this should be first.
- When only scanning unsupported files, we need to return `3`, which we currently don't.

In general I would separate these to their own PRs to ensure we review these cautiously and decouple the changes.

src/lib/iac/test/v2/output.ts
YairZ101: interesting, I don't remember doing this in the text output, do you know why we have different results processing between the two?

ofekatr: We don't, it happens for the text output as well ([source](https://github.com/snyk/cli/blob/99b8ee02372417d351c19970606631caef36bf5b/src/cli/commands/test/iac/output.ts#L299-L310))

YairZ101:
>In general I would separate these to their own PRs 

okay, please make sure to open a ticket for this bug so we could track it.

src/lib/iac/test/v2/output.ts
YairZ101: can you please point me to exact line where we do it?

YairZ101: ah I was talking about the new output flow for the `--experimental`.
does it mean we have the same bug there?

ofekatr: The bug is only there, for all three output formats

YairZ101: okay, now I understand.
This is very very weird design but I guess we need to deal with it :)

ofekatr:
> In general I would separate these to their own PRs

- For exit code `2`: [CFG-2076](https://snyksec.atlassian.net/browse/CFG-2076)
- For exit code `3`: [CFG-2077](https://snyksec.atlassian.net/browse/CFG-2077)

PR #3511, opened 2022-08-01
feat: pass projectTags arg to snyk-iac-test [CFG-2051]

ipapast:

#### What does this PR do?

This PR passes the values of the `--project-tags` flag to the snyk-iac-test binary. You can check it in combination with this branch: https://github.com/snyk/snyk-iac-test/pull/54 where the command accepts the new argument.

I used the existing `parseTags` function in the CLI which handles the validation and errors of the `--project-tags`.  There is a second check on the snyk-iac-test side as well, TBC if we need both.

- Go to this branch: https://github.com/snyk/snyk-iac-test/tree/feat/get-project-tags-arg
- Run locally `go build -o snyk-iac-test .`
- Now on the CLI side, point to the binary you just built (and a bundle) and try it out: 

```
SNYK_IAC_POLICY_ENGINE_PATH=/Users/iliannapapastefanou/snyk-repos/snyk-iac-test/snyk-iac-test \
SNYK_IAC_BUNDLE_PATH=/Users/iliannapapastefanou/snyk-repos/opa-rules/dist.tar.gz \
snyk-dev iac test test/fixtures/iac/terraform/sg_open_ssh.tf --experimental --report --project-tags=key1=value1,key2=value2

```

TODO: update the snyk-iac-test version when the branch is merged.

Handling the values of this command such as formatting them and adding them to the endpoint will be handled in a new PR.

CFG-2051

### DIFF
--- a/src/lib/iac/test/v2/scan/index.ts
+++ b/src/lib/iac/test/v2/scan/index.ts
@@ -98,6 +98,15 @@ function processFlags(
     flags.push('-project-lifecycle', options.attributes.lifecycle.join(','));
   }
 
+  if (options.projectTags) {
+    const stringifiedTags = options.projectTags
+      .map((tag) => {
+        return `${tag.key}=${tag.value}`;
+      })
+      .join(',');
+    flags.push('-project-tags', stringifiedTags);
+  }
+
   if (options.report) {
     flags.push('-report');
   }
--- a/src/cli/commands/test/iac/local-execution/index.ts
+++ b/src/cli/commands/test/iac/local-execution/index.ts
@@ -153,7 +153,7 @@ export function removeFileContent({
   };
 }
 
-function parseTags(options: IaCTestFlags) {
+export function parseTags(options: IaCTestFlags) {
   if (options.report) {
     return generateTags(options);
   }
--- a/src/cli/commands/test/iac/v2/index.ts
+++ b/src/cli/commands/test/iac/v2/index.ts
@@ -10,6 +10,7 @@ import { buildOutput } from '../../../../../lib/iac/test/v2/output';
 import { getIacOrgSettings } from '../local-execution/org-settings/get-iac-org-settings';
 import { Options, TestOptions } from '../../../../../lib/types';
 import { generateProjectAttributes } from '../../../monitor';
+import { parseTags } from '../local-execution';
 
 export async function test(
   paths: string[],
@@ -51,6 +52,7 @@ async function prepareTestConfig(
 
   const org = (options.org as string) || config.org;
   const orgSettings = await getIacOrgSettings(org);
+  const projectTags = parseTags(options);
 
   const attributes = parseAttributes(options);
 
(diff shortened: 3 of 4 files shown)


YairZ101 (approved):
LGTM

src/lib/iac/test/v2/scan/index.ts
```diff
@@ -40,10 +40,12 @@ function scanWithConfig(
   configPath: string,
 ): SnykIacTestOutput {
   const args = ['-bundle', rulesBundlePath, '-config', configPath];
-
   if (options.severityThreshold) {
     args.push('-severity-threshold', options.severityThreshold);
   }
+  if (options.projectTags) {
+    args.push('-project-tags', JSON.stringify(options.projectTags));
```
francescomari, line 47: `JSON.stringify()` will not generate an argument value in the same way as described in the corresponding PR in `snyk-iac-test`. Is this a leftover?

ipapast: yes. just changed to `options.projectTags.toString()` instead

francescomari: Does the `toString()` of the `Tag` do the right thing? A `Tag` seems to be a plain object.

ipapast: Sorry, that was too fast push again 😓 
snyk-iac-test expects a string as a flag, I updated it like this:

```
if (options.projectTags) {
    const stringifiedTags = options.projectTags.map((tag) => {
      return Object.values(tag).join(':');
    });
    args.push('-project-tags', stringifiedTags.toString());
  }
```

stringifiedTags is an array of the key value pairs, eg if I run `--project-tags=key1=value1,another2=val2`, I'll get:

`[ 'key1:value1', 'another2:val2' ]`

and then `stringifiedTags.toString` will pass the  `key1:value1,another2:val2` string to the executable.

src/lib/iac/test/v2/scan/index.ts
```diff
@@ -39,7 +39,32 @@ function scanWithConfig(
...
+    args.push('-project-environment', options.attributes.environment.join(','));
+  }
+
+  if (options.attributes?.lifecycle) {
+    args.push('-project-lifecycle', options.attributes.lifecycle.join(','));
+  }
+
+  if (options.projectTags) {
+    const stringifiedTags = options.projectTags.map((tag) => {
+      return Object.values(tag).join(':');
```
francescomari, line 64: `Object.values()` return the values of properties in an object by iterating over each property according to the same semantics of `for..in`

> The traversal order, as of modern ECMAScript specification, is well-defined and consistent across implementations. Within each component of the prototype chain, all non-negative integer keys (those that can be array indices) will be traversed first in ascending order by value, **then other string keys in ascending chronological order of property creation**.

This code is fragile. First, if the tag objects are created differently, the order of the key and the value properties would change. Second, if another property is added to the tag object (`Tag` is just an interface, after all), you are going to join more than two components. I would prefer if you built the tuples by directly using the key and the value properties.

Moreover, looking at the corresponding PR in snyk-iac-test, the separator between keys and value is supposed to be a `=`. I'm fine with using `:` instead, since we don't process the tags in snyk-iac-test yet.

Sorry for being a nuisance!

francescomari (changes requested):

src/lib/iac/test/v2/scan/index.ts
ipapast: we chatted on slack, pushed new commit

francescomari (approved):

PR #3521, opened 2022-08-02
extensions init

maxjeffos:


cliv2/internal/cliv2/arg_parser.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+var v1Commands = []string{
+	"test",
+	"monitor",
+	"code",
+	"iac",
+	"container",
+	"auth",
+	// TODO: what else?
```
Avishagp, line 37: fix, log4shell, config, policy, ignore

cliv2/internal/cliv2/cliv2.go
```diff
@@ -53,8 +62,7 @@ const (
 //go:embed cliv2.version
 var SNYK_CLIV2_VERSION_PART string
 
-func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) *CLI {
-
+func NewCLIv2(cacheDirectory string, extensions []*extension.Extension, argParserRootCmd *cobra.Command, debugMode bool, debugLogger *log.Logger) *CLI {
```
Avishagp, line 65: Should we rename this function?

cliv2/internal/cliv2/arg_parser.go
PeterSchafer: I think this part should eventually not be hard coded but we should try to extract this from cliv1

cliv2/internal/cliv2/arg_parser.go
```diff
@@ -37,22 +37,60 @@ func MakeArgParserConfig(extensions []*extension.Extension, config *CliConfigura
...
 
-var v1Commands = []string{
-	"test",
-	"monitor",
-	"code",
-	"iac",
-	"container",
-	"auth",
-	// TODO: what else?
+type Command struct {
```
maxjeffos, line 40: @Avishagp maybe use a more descriptive name here - like `NodeCLICommandMeta` or something just because the `Command` is starting to get heavily over-loaded. For example, cobra.Command, extension.Command (from the extension lib).

PR #3557, opened 2022-08-08
chore: Moved snyk-iac-test error mapping to be after scanning

ofekatr:

#### What does this PR do?

- Adds mapping for the `snyk-iac-test` output (`SnykIacTestOutput`) to the internal test output model (`TestOutput`). This includes:
    - Mapping `snyk-iac-test` scan errors to Snyk CLI errors.
    - Omitting `snyk-iac-test`'s raw results property, as we currently do not use it in the test flow. 
    - Adds the `path` and `fields` properties to `SnykIacTestError`.

#### Background context

- We currently do the mapping of `snyk-iac-test` errors to Snyk CLI errors only for the text output, when this mapping is relevant for all test output formats in the CLI.
- We currently do not use the `rawResults` property received from `snyk-iac-test`.
- We currently don't hold a top-level `path` property in test errors, but the JSON output formatted expects it when adding errors.
- We currently don't add the `fields` property from `snyk-iac-test` errors to the `SnykIacTestError` instances we create.

### DIFF
--- a/src/lib/iac/test/v2/errors.ts
+++ b/src/lib/iac/test/v2/errors.ts
@@ -45,10 +45,26 @@ export function getErrorUserMessage(code: number): string {
 }
 
 export class SnykIacTestError extends CustomError {
-  constructor(error: ScanError) {
-    super(error.message);
-    this.code = error.code;
+  public fields: { path: string; [key: string]: string };
+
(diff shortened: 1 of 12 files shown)


francescomari (approved):
Well done!

PR #3565, opened 2022-08-10
refactor: Create Issue interface for text output

ofekatr:

#### What does this PR do?

- Creates an `Issue` interface for the text output, to define which properties are relevant for the output. This also makes the mapping of `snyk-iac-test` results to issues used by the text output formatter easier, as it highlights which properties should be mapped and populated.


YairZ101 (approved):
great small change!

PR #3569, opened 2022-08-11
refactor: derive the Policy Engine version from the checksums

francescomari:

#### What does this PR do?

Implements a function to derive the version of the `snyk-iac-test`from the checksum. This allows a developer to upgrade the dependency to `snyk-iac-test` by just pasting a new checksum, which is a change that must be done in one place only.


ipapast (approved):
YAS

PR #3570, opened 2022-08-11
chore: change project name source

YairZ101:

#### What does this PR do?
Removes the project name calculation that we are doing in the CLI and instead starts using the project name that we are returning from `snyk-iac-test`

### DIFF
--- a/src/lib/iac/test/v2/output.ts
+++ b/src/lib/iac/test/v2/output.ts
@@ -18,17 +18,16 @@ import { convertEngineToSarifResults } from './sarif';
 import { CustomError, FormattedCustomError } from '../../../errors';
 import { SnykIacTestError } from './errors';
 import stripAnsi from 'strip-ansi';
+import * as path from 'path';
 
 export function buildOutput({
   scanResult,
   testSpinner,
-  projectName,
   orgSettings,
   options,
 }: {
   scanResult: TestOutput;
   testSpinner?: Ora;
-  projectName: string;
   orgSettings: IacOrgSettings;
   options: any;
 }): TestCommandResult {
@@ -40,7 +39,6 @@ export function buildOutput({
 
   const { responseData, jsonData, sarifData } = buildTestCommandResultData({
     scanResult,
-    projectName,
     orgSettings,
     options,
   });
@@ -62,15 +60,16 @@ export function buildOutput({
 
 function buildTestCommandResultData({
   scanResult,
-  projectName,
   orgSettings,
   options,
 }: {
   scanResult: TestOutput;
-  projectName: string;
   orgSettings: IacOrgSettings;
   options: any;
 }) {
+  const projectName =
+    scanResult.results?.metadata?.projectName ?? path.basename(process.cwd());
+
   const jsonData = jsonStringifyLargeObject(
     convertEngineToJsonResults({
       results: scanResult,
--- a/src/lib/iac/test/v2/scan/results.ts
+++ b/src/lib/iac/test/v2/scan/results.ts
@@ -32,6 +32,11 @@ export interface SnykIacTestOutput {
 export interface Results {
   resources?: Resource[];
   vulnerabilities?: Vulnerability[];
+  metadata?: Metadata;
+}
+
+export interface Metadata {
+  projectName: string;
 }
 
 export interface Vulnerability {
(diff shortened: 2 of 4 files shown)


src/lib/iac/test/v2/scan/results.ts
```diff
@@ -32,6 +32,11 @@ export interface SnykIacTestOutput {
 export interface Results {
   resources?: Resource[];
   vulnerabilities?: Vulnerability[];
+  metadata?: Metadata;
```
ofekatr, line 35: Thought: Does the generation of the metadata depends on the existence of valid results? 
If there are no results would there not be metadata? 
Unless there's a direct coupling between the existence of `results` and `metadata` it makes more sense to me that `metadata` would be a top-level property of `TestOutput`. 
If the project name generation is conditional, would it make sense to have the `metadata` property as required, and `projectName` within it as conditional?

src/lib/iac/test/v2/output.ts
```diff
@@ -62,15 +60,16 @@ export function buildOutput({
...
-  projectName,
   orgSettings,
   options,
 }: {
   scanResult: TestOutput;
-  projectName: string;
   orgSettings: IacOrgSettings;
   options: any;
 }) {
+  const projectName =
```
ofekatr, line 70: Suggestion: I think this default value can cause more confusion than good. 
If we considered `projectName` to be required and assert it with this default value, wouldn't it make more sense to make it required when receiving it from `snyk-iac-test`? 
In what scenarios does `snyk-iac-test` not return a project name and we require this default value?

src/lib/iac/test/v2/scan/results.ts
YairZ101: >  Does the generation of the metadata depends on the existence of valid results?

Currently, it does because we are generating the metadata in the results processing stage which we enter only when there are valid results.
AFAIK when we don't generate results we are only displaying an error without any information about the tests so there is no reason to generate the metadata.

> If the project name generation is conditional, would it make sense to have the metadata property as required, and projectName within it as conditional?

That's an implementation detail that we can always change in the future if necessary.

src/lib/iac/test/v2/output.ts
YairZ101: > If we considered projectName to be required and assert it with this default value, wouldn't it make more sense to make it required when receiving it from snyk-iac-test?

I agree with you that it would make sense if would consider it required but currently, we do not.
as I said above it is an implementation detail so if in the future we see it as problematic we can always change it.

> In what scenarios does snyk-iac-test not return a project name and we require this default value?

When we failed to generate results.

src/lib/iac/test/v2/scan/results.ts
ofekatr: I don't think it's an implementation detail, as it actually defines the contract we have with `snyk-iac-test` in the different flows and user stories. 
On the other hand, the fact that we calculate the project name as part of the results processing step is an implementation detail, as it can be generated in another scope which then allows us to return it in all the desired flows.

src/lib/iac/test/v2/output.ts
ofekatr: In that case at the very least if `results` exists that means `metadata` should exist as well, meaning there'd be no need to have it as optional if it's there

ofekatr (approved):
As mentioned in the comments it'd make more sense to me personally to have `metadata` as a top-level required property but I don't consider it a blocker and we can always re-iterate on this in the future. 

Apart from that, LGTM ✅

src/lib/iac/test/v2/output.ts
YairZ101: The reason that `metadata` is optional is the same reason that `resources` is optional.
I agree with your logic here but I rather align with the current `results` type structure.

src/lib/iac/test/v2/scan/results.ts
YairZ101: > as it actually defines the contract we have with snyk-iac-test in the different flows and user stories

But the current flows and user stories do not display the project name if the test didn't produce results, this is why I think it is an implementation detail (similar to what you said in your 2nd half of the message).

YairZ101: In addition to that, personally, I believe we should stop displaying the project name in the test output because it is misleading because it doesn't match the platform definition of a "project".

PR #3571, opened 2022-08-11
chore: regenerate lockfile

Avishagp:
Looks like our lockfile got corrupted somehow. This PR fixes it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -136,37 +136,6 @@
         "node": ">=12"
       }
     },
-    "../cloud-config-parser": {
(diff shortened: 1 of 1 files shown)


JackuB:
Screen Shot 2022-08-12 at 10 31 44

I'm not a big fan that this PR is removing the resolved field for all dependencies 🤔 
I'll try to regenerate it myself

JackuB:
I've ran dozens of useless npm commands, but seems like this is the safest and smallest changeset we can do.

JackuB (approved):

PR #3574, opened 2022-08-11
Add `--additional-extension-path` option

maxjeffos:


cliv2/cmd/cliv2/main.go
```diff
@@ -60,9 +60,25 @@ func MainWithErrorCode(cliConfig *cliv2.CliConfiguration, args []string) int {
...
-		CacheDirectory: cliConfig.CacheDirectory,
-		Logger:         cliConfig.DebugLogger,
+	argParser := cliv2.BaseArgParser(cliConfig.DebugLogger, args)
+	persistentFlagValues, err := argParser.PeristentFlagValues(args)
+	if persistentFlagValues == nil {
+		fmt.Println("persistentFlagValues is nil")
+		return exit_codes.SNYK_EXIT_CODE_ERROR
+	}
+
+	cliConfig.Debug = persistentFlagValues.Debug
```
PeterSchafer, line 70: suggestion/question: this assignment of individual flags seems unnecessary code. The previous version used a single configuration structure that was being filled by the parser. What was the reason for changing this?

cliv2/cmd/cliv2/main.go
```diff
@@ -60,9 +60,25 @@ func MainWithErrorCode(cliConfig *cliv2.CliConfiguration, args []string) int {
...
+	cliConfig.Insecure = persistentFlagValues.Insecure
+	cliConfig.ProxyAddr = persistentFlagValues.ProxyAddr
+	cliConfig.AdditionalExtensionDirectoryPath = persistentFlagValues.AdditionalExtensionDirectoryPath
```
PeterSchafer, line 73: question: does the cli instance really need to know about this additional extensions path? to me it seems that this is a parameter for the ExtensionManager, whose responsibility it is to load the extension objects and the cli just uses these objects.

cliv2/cmd/cliv2/main.go
```diff
@@ -60,9 +60,25 @@ func MainWithErrorCode(cliConfig *cliv2.CliConfiguration, args []string) int {
 		}
 	}
 
-	extMngr := extension.New(&extension.Configuration{
-		CacheDirectory: cliConfig.CacheDirectory,
-		Logger:         cliConfig.DebugLogger,
+	argParser := cliv2.BaseArgParser(cliConfig.DebugLogger, args)
```
PeterSchafer, line 63: suggestion: we should see the Command line Parser as an instance that fills available configuration structures. these structures are defined by the instances that consume them like the CLIv2 instance or the ExtensionManager. By this we enable a design in which it doesn't matter who fills the configuration structures, which enables to add additional sources for configuration easier and more structured.

PR #3596, opened 2022-08-17
feat: stop using cached rules bundle

YairZ101:

#### What does this PR do?
This PR removes the ability to use the cached rules bundle and instead it will redownload the bundle each time.
The reason behind this is the switch to the legacy rules bundle which is not using versioning so the latest "version" is being uploaded to the same S3 bucket each time.


ofekatr (approved):

ofekatr:
I was expecting this to break the tests in https://github.com/snyk/cli/blob/master/test/jest/unit/lib/iac/test/v2/setup/local-cache/index.spec.ts, was the caching not covered in these tests?

PR #3600, opened 2022-08-17
fix: increase buffer size

YairZ101:

#### What does this PR do?
The PR fixes the problem where in some cases when scanning a directory with a high number of files, the scan failed due to an [ENOBUFS](https://www.encyclo.co.uk/meaning-of-ENOBUFS) error.
The fix is rather simple, it simply increases the max buffer.

### DIFF
--- a/src/lib/iac/test/v2/scan/index.ts
+++ b/src/lib/iac/test/v2/scan/index.ts
@@ -53,6 +53,7 @@ function scanWithConfig(
   const process = childProcess.spawnSync(policyEnginePath, args, {
     encoding: 'utf-8',
     stdio: 'pipe',
+    maxBuffer: 1024 * 1024 * 10, // The default value is 1024 * 1024, if we see in the future that multiplying it by 10 is not enough we can increase it further.
   });
 
   debug('policy engine standard error:\n%s', '\n' + process.stderr);


src/lib/iac/test/v2/scan/index.ts
```diff
@@ -53,6 +53,7 @@ function scanWithConfig(
   const process = childProcess.spawnSync(policyEnginePath, args, {
     encoding: 'utf-8',
     stdio: 'pipe',
+    maxBuffer: 1024 * 1024 * 10,
```
ofekatr, line 56: Do you know what value was originally being used? 
Do we know how bigger the buffer's cap just became? 
Why was this value chosen in particular?

YairZ101, line 56: The default value is `1024 * 1024`, if we see in the future that multiplying it by 10 is not enough we can increase it further.

ofekatr (approved):

ofekatr:
@YairZ101 Can you please share this solution with #cli for visibility?

YairZ101:
@ofekatr why do you think we should post this in #cli? this PR doesn't affect the CLI as a whole.
Do you have any concerns?

PR #3601, opened 2022-08-17
refactor: use IacTestFlags type instead of generic Options

ipapast:

#### What does this PR do?

I'm suggesting this change for using only the subset of types we actually need. this will also help the next PR which needs the Policy Options to be used at a later point in code. https://github.com/snyk/cli/pull/3602

### DIFF
--- a/src/cli/commands/test/iac/v2/index.ts
+++ b/src/cli/commands/test/iac/v2/index.ts
@@ -7,15 +7,15 @@ import { buildSpinner, printHeader } from '../output';
 import { spinnerMessage } from '../../../../../lib/formatters/iac-output';
 import { buildOutput } from '../../../../../lib/iac/test/v2/output';
 import { getIacOrgSettings } from '../local-execution/org-settings/get-iac-org-settings';
-import { Options, TestOptions } from '../../../../../lib/types';
 import { generateProjectAttributes } from '../../../monitor';
 import { parseTags } from '../local-execution';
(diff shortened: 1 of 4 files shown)


src/cli/commands/test/iac/v2/index.ts
```diff
@@ -7,15 +7,15 @@ import { buildSpinner, printHeader } from '../output';
...
 import { generateProjectAttributes } from '../../../monitor';
 import { parseTags } from '../local-execution';
 import { systemCachePath } from '../../../../../lib/iac/test/v2/scan';
 import { getFlag } from '../index';
+import { IaCTestFlags } from '../local-execution/types';
 
 export async function test(
   paths: string[],
-  options: Options & TestOptions,
+  options: IaCTestFlags,
```
ipapast, line 18: do you see any issue with using our own IaCTestFlags (subset of Options & TestOptions) here?

ipapast: I need PolicyOptions on my next PR, so I did `IaCTestFlags` instead of `Options & TestOptions & PolicyOptions`

YairZ101: I don't see a problem, I think sooner or later we would need to make the switch :)

YairZ101 (approved):

ofekatr (approved):

PR #3603, opened 2022-08-18
feat: add unmanaged service test call ff

dagrest:

#### What does this PR do?

- Implements Feature Flag to switch to a new polling function that uses a new Unmanaged Service APIs
- Implements a new polling function that uses a new Unmanaged Service APIs

Implementation begins in the following function: resolveAndTestFacts

Use the following options: test --unmanaged, do not forget to enable feature flag, current name is 'snykNewUnmanagedTest'

- [Implement CLI feature flag for new service: TUN-299](https://snyksec.atlassian.net/browse/TUN-299)
- [Update CLI polling functions to point at the unmanaged service: TUN-317](https://snyksec.atlassian.net/browse/TUN-317)

### DIFF
--- a/src/lib/ecosystems/resolve-test-facts.ts
+++ b/src/lib/ecosystems/resolve-test-facts.ts
@@ -1,16 +1,226 @@
 import { Options, PolicyOptions } from '../types';
 import { spinner } from '../../lib/spinner';
-import { Ecosystem, ScanResult, TestResult } from './types';
+import {
+  Ecosystem,
+  ScanResult,
+  TestResult,
+  FileSignaturesDetails,
+} from './types';
+import {
+  CreateDepGraphResponse,
+  GetIssuesResponse,
+  FileHashes,
+  Attributes,
+} from './unmanaged/types';
 import {
   requestTestPollingToken,
   pollingTestWithTokenUntilDone,
+  createDepGraph,
+  getDepGraph,
+  getIssues,
 } from '../polling/polling-test';
 import { extractAndApplyPluginAnalytics } from './plugin-analytics';
 import { findAndLoadPolicy } from '../policy';
 import { filterIgnoredIssues } from './policy';
 import { IssueData, Issue } from '../snyk-test/legacy';
+import { hasFeatureFlag } from '../feature-flags';
+import { delayNextStep } from '../polling/common';
+import {
+  convertDepGraph,
+  convertMapCasing,
+  convertToCamelCase,
+  getSelf,
+} from './unmanaged/utils';
 
 export async function resolveAndTestFacts(
+  ecosystem: Ecosystem,
+  scans: {
+    [dir: string]: ScanResult[];
+  },
+  options: Options & PolicyOptions,
+): Promise<[TestResult[], string[]]> {
+  const unmanagedDepsOverride = process.env.USE_UNMANAGED_DEPS;
+
+  const featureFlagEnabled = await hasFeatureFlag(
+    'snykNewUnmanagedTest',
+    options,
+  );
+
+  return featureFlagEnabled || unmanagedDepsOverride
+    ? resolveAndTestFactsUnmanagedDeps(scans, options)
+    : resolveAndTestFactsRegistry(ecosystem, scans, options);
+}
+
+async function submitHashes(
+  hashes: FileHashes,
+  orgId: string,
+): Promise<string> {
+  const response: CreateDepGraphResponse = await createDepGraph(hashes, orgId);
+
+  return response.data.id;
+}
+
+async function pollDepGraph(id: string, orgId: string): Promise<Attributes> {
+  let attempts = 0;
+  const maxAttempts = 50;
+  while (attempts < maxAttempts) {
+    try {
+      const response = await getDepGraph(id, orgId);
+      return response.data.attributes;
+    } catch (e) {
+      await delayNextStep(attempts, maxAttempts, 1000);
+      attempts++;
+    }
+  }
+
+  return Promise.reject('Failed to get DepGraph');
+}
+
+async function fetchIssues(
+  start_time,
+  dep_graph_data,
+  component_details,
+  orgId: string,
+) {
+  const response: GetIssuesResponse = await getIssues(
+    {
+      dep_graph: dep_graph_data,
+      start_time,
+      component_details,
+    },
+    orgId,
+  );
+
+  const issues = response.data.result.issues.map((issue) => {
+    const converted = convertToCamelCase<Issue>(issue);
+    converted.fixInfo = convertToCamelCase(converted.fixInfo);
+    return converted;
+  });
+
+  const issuesData = convertMapCasing<{
+    [issueId: string]: IssueData;
+  }>(response.data.result.issues_data);
+
+  const depGraphData = convertDepGraph(response.data.result.dep_graph);
+
+  const dependencyCount = response.data.result.dep_graph.graph.nodes.find(
+    (graphNode) => {
+      return graphNode.node_id === 'root-node';
+    },
+  )?.deps?.length;
+
+  const depsFilePaths = response.data.result.deps_file_paths;
+
+  const fileSignaturesDetails = convertMapCasing<FileSignaturesDetails>(
+    response.data.result.file_signatures_details,
+  );
+
+  return {
+    issues,
+    issuesData,
+    depGraphData,
+    dependencyCount,
+    depsFilePaths,
+    fileSignaturesDetails,
+  };
+}
+
+export async function resolveAndTestFactsUnmanagedDeps(
+  scans: {
+    [dir: string]: ScanResult[];
+  },
+  options: Options & PolicyOptions,
+): Promise<[TestResult[], string[]]> {
+  const results: any[] = [];
+  const errors: string[] = [];
+  const packageManager = 'Unmanaged (C/C++)';
+
+  let orgId = options.org || '';
+
+  if (orgId === '') {
+    const self = await getSelf();
+    orgId = self.default_org_context;
+  }
+
+  for (const [path, scanResults] of Object.entries(scans)) {
+    await spinner(`Resolving and Testing fileSignatures in ${path}`);
+    for (const scanResult of scanResults) {
+      try {
+        const id = await submitHashes(
+          { hashes: scanResult?.facts[0]?.data },
+          orgId,
+        );
+
+        const {
+          start_time,
+          dep_graph_data,
+          component_details,
+        } = await pollDepGraph(id, orgId);
+
+        const {
+          issues,
+          issuesData,
+          depGraphData,
+          dependencyCount,
+          depsFilePaths,
+          fileSignaturesDetails,
+        } = await fetchIssues(
+          start_time,
+          dep_graph_data,
+          component_details,
+          orgId,
+        );
+
+        const issuesMap: Map<string, Issue> = new Map();
+        issues.forEach((i) => {
+          issuesMap[i.issueId] = i;
+        });
+
+        const vulnerabilities: IssueData[] = [];
+        for (const issuesDataKey in issuesData) {
+          const issueData = issuesData[issuesDataKey];
+          const pkgCoordinate = `${issuesMap[issuesDataKey]?.pkgName}@${issuesMap[issuesDataKey]?.pkgVersion}`;
+          issueData.from = [pkgCoordinate];
+          issueData.name = pkgCoordinate;
+          issueData.packageManager = packageManager;
+          vulnerabilities.push(issueData);
+        }
+
+        const policy = await findAndLoadPolicy(path, 'cpp', options);
+
+        const [issuesFiltered, issuesDataFiltered] = filterIgnoredIssues(
+          issues,
+          issuesData,
+          policy,
+        );
+
+        results.push({
+          issues: issuesFiltered,
+          issuesData: issuesDataFiltered,
+          depGraphData,
+          depsFilePaths,
+          fileSignaturesDetails,
+          vulnerabilities,
+          path,
+          dependencyCount,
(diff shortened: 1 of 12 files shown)


src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,393 @@
...
+  for (let i = 0; i < attributes?.dep_graph_data?.pkgs.length; i++) {
+    pkgs.push(attributes?.dep_graph_data?.pkgs[i]);
+  }
+
+  const graph = {} as IssuesRequestGraphOpenApi;
+  const nodes = [] as IssuesRequestNodeOpenApi[];
+  graph.root_node_id = attributes?.dep_graph_data?.graph?.root_node_id;
+  for (let i = 0; i < attributes?.dep_graph_data?.graph?.nodes?.length; i++) {
+    const node = {} as IssuesRequestNodeOpenApi;
+    node.node_id = attributes?.dep_graph_data?.graph?.nodes[i]?.node_id;
```
dekelund, line 96: Can't you extract the content from `attributes?.dep_graph_data?.graph?.nodes[i]?` once and reuse it for both extractions of node and pkg id?

dekelund: or set it directly when you create the node object
const source = attributes?.dep_graph_data?.graph?.nodes[i]
```
const node = {
node_id: source?.node_id,
pkg_id: source?.pkg_id,
} as IssuesRequestNodeOpenApi
```

dekelund: and it would also reduce the extraction of nodes[i] below

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,393 @@
...
+  };
+
+  const componentDetails = {} as IssuesRequestComponentDetails;
+  for (const key in attributes?.component_details) {
+    const details = attributes?.component_details[key];
+    const filePaths: string[] = [];
+    for (const fileIndex in details?.file_paths) {
+      filePaths.push(details?.file_paths[fileIndex]);
+    }
+    const isuueDetails: IssuesRequestDetails = {
```
dekelund, line 127: spelling "isuueDetails"

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,393 @@
...
+      version: details?.version,
+      author: details?.author,
+      path: details?.path,
+      id: details?.id,
+      url: details?.url,
+      score: details?.score,
+      file_paths: filePaths,
+    };
+    componentDetails[key] = isuueDetails;
+  }
```
dekelund, line 138: Do details contains anything you don't want to copy? otherwise you could use this feature from TS/JS

```ts
const source = {
    a: 42,
    b: "hello",
    c: "te be replaced"
}

const dst = {...source, c: "replaced"};

console.dir(source);
console.dir(dst);
```

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,393 @@
...
+    hashesArray.push(hash);
+  }
+  return hashesArray;
+}
+
+function convertIssues(issuesOpenApi: IssueOpenApi[]) {
+  const issues: Issue[] = [];
+
+  for (let i = 0; i < issuesOpenApi.length; i++) {
+    const fInfo: FixInfo = {
```
dekelund, line 172: I would rename fInfo to info, since we only have one kind of info here

src/lib/ecosystems/types.ts
```diff
@@ -24,6 +30,291 @@ export interface UnknownTarget {
...
+  nodeId: string;
+}
+
+export interface Node {
+  nodeId: string;
+  pkgId: string;
+  deps: Dep[];
+}
+
+// interface Graph {
```
dekelund, line 58: Unused code, should not be merged, should it?

test/jest/unit/lib/ecosystems/fixtures/create-dep-graph-response.ts
```diff
@@ -0,0 +1,9 @@
+import {LocationResponse} from "../../../../../../src/lib/ecosystems/types";
+
+export const createDepgraphResponse = {
+    //id: '53b2fa4d-3e2a-491e-a73e-c4dafcb0ec6f',
```
dekelund, line 4: remove unused uuid

test/jest/unit/lib/ecosystems/fixtures/create-dep-graph-response.ts
```diff
@@ -0,0 +1,9 @@
...
+    id: '1234-1234',
+    location:
+        '/rest/unmanaged_ecosystem/depgraphs/53b2fa4d-3e2a-491e-a73e-c4dafcb0ec6f?version=2022-05-23~experimental',
```
dekelund, line 7: location contains the old uuid, is that intentional?

test/jest/unit/lib/ecosystems/fixtures/get-dep-graph-response.ts
```diff
@@ -0,0 +1,16 @@
+import {Attributes, ComponentDetailsOpenApi, Data} from "../../../../../../src/lib/ecosystems/types";
+import {depGraphDataOpenAPI} from "./dep-graph-open-api";
+
+const componentDetailsOpenApi: ComponentDetailsOpenApi = {};
+
+const attributes: Attributes = {
+    start_time: 1660137910316,
+    dep_graph_data: depGraphDataOpenAPI, // getDepGraphOpenApi(),// TODO: delete  depGraphDataOpenAPI,
```
dekelund, line 8: unused code

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,393 @@
...
+
+  const pkgs = [] as Pkg[];
+  for (let i = 0; i < attributes?.dep_graph_data?.pkgs.length; i++) {
+    pkgs.push(attributes?.dep_graph_data?.pkgs[i]);
+  }
+
+  const graph = {} as IssuesRequestGraphOpenApi;
+  const nodes = [] as IssuesRequestNodeOpenApi[];
+  graph.root_node_id = attributes?.dep_graph_data?.graph?.root_node_id;
+  for (let i = 0; i < attributes?.dep_graph_data?.graph?.nodes?.length; i++) {
```
dekelund, line 94: maybe replace for-loop with `attributes?.dep_graph_data?.graph?.nodes.map(node => somelambda)`

hmm what happens with length if nodes is undefined. I guess it's safe when you use questionmark to handle the extraction. I'm just curious

dekelund: otherwise you could extract it with

`const source = attributes?.dep_graph_data?.graph?.nodes || []`

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,393 @@
...
+    fileSignaturesDetails[fileSignatureDetailOpenApiKey] = fileSignaturesDetail;
+  }
+
+  return fileSignaturesDetails;
+}
+
+function convertDepGraph(depGraphOpenApi: DepGraphDataOpenAPI) {
+  const depGraph = {} as DepGraphData;
+
+  depGraph.pkgManager = depGraphOpenApi.pkg_manager;
```
dekelund, line 255: I would inline the attributes in one statement, something like this

```ts
depGraph = {
  pkgManager: depGraphOpenApi.pkg_manager,
  schemaVersion: depGraphOpenApi.schema_version,
  pkgs: depGraphOpenApi.pkgs,

  graph = {
    rootNodeId: depGraphOpenApi.graph.root_node_id,
    nodes: [] as Node[],
  } as Graph;
} as DepGraphData
```

src/lib/ecosystems/resolve-test-facts.ts
danielroymoore: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

test/jest/unit/lib/ecosystems/fixtures/get-dep-graph-response.ts
dekelund: we shouldn't keep TODO's in our code base

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -1,21 +1,231 @@
...
+  const results: any[] = [];
+  const errors: string[] = [];
+  const packageManager = 'Unmanaged (C/C++)';
+
+  let orgId = options.org || '';
+
+  if (orgId === '') {
+    const self = await getSelf();
+    orgId = self.default_org_context;
+  }
```
dekelund, line 143: I nearly want to replace 138-143 with

```ts
const orgId = options.org || (await getSelf()).default_org_context;
```

but it looks a bit ugly with the await inside the parentheses, on the other hand we are able to make orgId `const`which is a bit safer, and maybe easier to reason about.

So maybe it should be the following, even if it's a bit expensive in some cases.

```ts
const self = await getSelf();
const orgId = options.org || self.default_org_context;
```

src/lib/ecosystems/unmanaged/types.ts
```diff
@@ -0,0 +1,264 @@
...
+  title: string;
+  severity: SEVERITY;
+  fixed_in: string[];
+  legal_instructions?: string;
+  package_manager?: SupportedProjectTypes;
+  from?: string[];
+  name?: string;
+  publication_time?: string;
+  creation_time?: string;
+  cvsSv3?: string;
```
dekelund, line 262: uppercase "S" ?

src/lib/polling/polling-test.ts
```diff
@@ -6,10 +6,92 @@ import { Options } from '../types';
...
+      break;
+    } catch (e) {
+      if (waitingTimeMs < maxWaitnigTimeMs) {
+        waitingTimeMs += pollIntervalMs;
+      } else {
+        throw e;
+      }
+    }
+  }
+  return JSON.parse(result.toString());
```
dekelund, line 74: can't we move this return into the try, and reduce number or variables, alternative use a const in the try-block.

src/lib/polling/polling-test.ts
```diff
@@ -6,10 +6,92 @@ import { Options } from '../types';
...
+      'x-is-ci': isCI(),
+      authorization: getAuthHeader(),
+    },
+  };
+
+  const maxWaitnigTimeMs = 30000;
+  const pollIntervalMs = 5000;
+  let waitingTimeMs = pollIntervalMs;
+  let result = {} as GetDepGraphResponse;
+  while (waitingTimeMs <= maxWaitnigTimeMs) {
```
dekelund, line 61: misspelled word "waitnig" should be "waiting"

src/lib/polling/polling-test.ts
```diff
@@ -6,10 +6,92 @@ import { Options } from '../types';
...
+    try {
+      await sleep(waitingTimeMs);
+      result = await makeRequest<GetDepGraphResponse>(payload);
+      break;
+    } catch (e) {
+      if (waitingTimeMs < maxWaitnigTimeMs) {
```
dekelund, line 67: remove if and else with `waitingTimeMs += pollIntervalMs;`, then after the while-loop do a throw.

src/lib/polling/polling-test.ts
```diff
@@ -6,10 +6,92 @@ import { Options } from '../types';
...
+      authorization: getAuthHeader(),
+    },
+    body: issuesRequestAttributes,
+  };
+
+  const result = await makeRequest<GetIssuesResponse>(payload);
+  return JSON.parse(result.toString());
+}
+
+export async function getDepGraph(
```
dekelund, line 42: Based on the comments in this function it could look something like:

```ts
export async function getDepGraph(
  id: string,
  orgId: string,
): Promise<GetDepGraphResponse> {
  const payload = {
    method: 'GET',
    url: `${config.API_REST_URL}/org/${orgId}/unmanaged_ecosystem/depgraphs/${id}?version=2022-05-23~experimental`,
    json: true,
    headers: {
      'Content-Type': 'application/vnd.api+json',
      'x-is-ci': isCI(),
      authorization: getAuthHeader(),
    },
  };

  const maxWaitingTimeMs = 30000;
  const pollIntervalMs = 5000;
  let waitingTimeMs = pollIntervalMs;
  while (waitingTimeMs <= maxWaitingTimeMs) {
    try {
      await sleep(waitingTimeMs);
      const result: GetDepGraphResponse = await makeRequest<GetDepGraphResponse>(payload);
      return JSON.parse(result.toString());
    } catch (e) {
      waitingTimeMs += pollIntervalMs;
    }
  }

  throw <something>;  // I would create my own timeout exception here
}
```

dekelund: the problem I see though, is that JSON.parse might throw an exception. maybe we shouldn't do it as I said 😆

dekelund: on the other hand, the result instant in current implementation is misleading since it has been cast to a GetDepGraphResponse, but in reality it's a string.

src/lib/ecosystems/resolve-test-facts.ts
danielroymoore: The reason I didn't fetch self first was because I wanted to avoid a network call when it may not be needed.

src/lib/polling/polling-test.ts
danielroymoore: Resolving and raising as improvement ticket

dekelund (approved):

Avishagp (approved):

PR #3606, opened 2022-08-18
fix: none custom policies severity issues should be filtered out before sending them to registry

wbeuil:

#### What does this PR do?

Related to this [Tier 3 Zendesk ticket](https://snyk.zendesk.com/agent/tickets/30091) where if you tried to send to `/iac-share-results` endpoint issues with custom policies set to `NONE` in the Snyk UI, you were getting an error using the `--report` in the CLI. Whereas, without the `--report` flag, everything works fine.

Behind the user experience, the problem is that we got a weird bug where registry send back to the CLI a 400 introduced in [this PR](https://github.com/snyk/registry/pull/27508) and we're now throwing an error back to the client due to a change of errors handling introduced [here](https://github.com/snyk/cli/pull/3454).

This PR is simply filtering `NONE` custom policies severity issues prior to sending them through the `/iac-share-results` endpoint.

Create a simple terraform file like this one:

```
resource "aws_s3_bucket" "this" {
  force_destroy = true
}
```

And then run this first command to check that it works fine without any custom policies:

```
$ snyk iac test <YOUR_TERRAFORM_FILE>
```

Go into Snyk UI and change one policies to NONE:

image

And rerun the above command, it should send only the other issues and not the one you changed in the UI.

[Zendesk ticket](https://snyk.zendesk.com/agent/tickets/30091)

### DIFF
--- a/src/cli/commands/test/iac/local-execution/process-results/share-results-formatter.ts
+++ b/src/cli/commands/test/iac/local-execution/process-results/share-results-formatter.ts
@@ -22,7 +22,9 @@ export function formatShareResults(
       filePath: result.filePath,
       fileType: result.fileType,
       projectType: result.projectType,
-      violatedPolicies: result.violatedPolicies,
+      violatedPolicies: result.violatedPolicies.filter(
+        (violatedPolicy) => violatedPolicy.severity !== 'none',
+      ),
     };
   });
 }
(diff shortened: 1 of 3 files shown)


ipapast (approved):
looks good, thank you!

PeterSchafer (approved):

PR #3612, opened 2022-08-21
feat: add report summary

YairZ101:

#### What does this PR do?
This PR adds to the text output 2 new sections:
1. When using `--report` it adds a short summary of the generated projects and where to find them
2. When not using `--report` it adds a tip to let users know about the `--report` flag

1. Run `snyk-dev iac test --experimental --report` inside a dir that is a root of a repo and inside a dir which is not and see the correct project name with a link to the relevant projects page.
2. Run `snyk-dev iac test --experimental` and check for the correct tip at the bottom of the text output.

https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-2102

#### Screenshots
Report summary
image

Report tip
image

### DIFF
--- a/src/cli/commands/test/iac/scan.ts
+++ b/src/cli/commands/test/iac/scan.ts
@@ -38,7 +38,7 @@ export async function scan(
   remoteRepoUrl?: string,
   targetName?: string,
 ): Promise<{
-  iacOutputMeta: IacOutputMeta | undefined;
+  iacOutputMeta: IacOutputMeta;
   iacScanFailures: IacFileInDirectory[];
   iacIgnoredIssuesCount: number;
   results: any[];
--- a/src/lib/formatters/iac-output/v2/share-results.ts
+++ b/src/lib/formatters/iac-output/v2/share-results.ts
@@ -1,20 +1,19 @@
-import { IacOutputMeta } from '../../../types';
 import config from '../../../config';
 import { EOL } from 'os';
 import { colors, contentPadding } from './utils';
 
-export function formatShareResultsOutput(iacOutputMeta: IacOutputMeta) {
+export function formatShareResultsOutput(orgName: string, projectName: string) {
   return (
     colors.title('Report Complete') +
     EOL +
     EOL +
     contentPadding +
     'Your test results are available at: ' +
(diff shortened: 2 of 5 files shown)


src/cli/commands/test/iac/scan.ts
```diff
@@ -38,7 +38,7 @@ export async function scan(
   remoteRepoUrl?: string,
   targetName?: string,
 ): Promise<{
-  iacOutputMeta: IacOutputMeta | undefined;
```
YairZ101, line 41: I'm not sure why we have `undefined` here as an optional type, it looks like we will always return an `iacOutputMeta` because of the logic [here](https://github.com/snyk/cli/blob/d8e4ea8d47587524e6e82e372173a10da2c07c2e/src/cli/commands/test/iac/scan.ts#L51-L57)
reviewer, please double check this just to make sure I'm not missing something :)

ipapast: I think your change is good. we return a promise of IacOutputMeta. Even though a property of the object can be undefined (eg `gitRemoteUrl`,) the object itself will never be undefined. So I think it's safe to remove it.

src/lib/formatters/iac-output/v2/share-results.ts
```diff
@@ -1,20 +1,19 @@
-import { IacOutputMeta } from '../../../types';
 import config from '../../../config';
 import { EOL } from 'os';
 import { colors, contentPadding } from './utils';
 
-export function formatShareResultsOutput(iacOutputMeta: IacOutputMeta) {
+export function formatShareResultsOutput(orgName: string, projectName: string) {
```
ipapast, line 5: I'm curious, do you prefer 2 parameters over one from the same object? or is it something different I'm not seeing?

YairZ101: A few reasons:
1. The main reason, we stopped using the type `IacOutputMeta` in the new output logic and I didn't want to create a variable of it just for this specific function so instead I split the argument of the function to 2.
2. I think it makes it more clear what are the necessary arguments for this function, giving it an object makes it a bit vague in my opinion, I think going forward I'll try to stop giving functions a whole object and instead I'll try to specify the exact arguments that are needed.

PR #3614, opened 2022-08-22
chore: use docs url returned from `snyk-iac-test`

YairZ101:

#### What does this PR do?
This PR is a follow-up to https://github.com/snyk/snyk-iac-test/pull/94 where we started to add a link to the docs site to scan results.


src/lib/iac/test/v2/scan/results.ts
```diff
@@ -55,6 +55,7 @@ export interface Rule {
   references?: string;
   labels?: string[];
   category?: string;
+  documentation: string;
```
ofekatr, line 58: This will likely be an optional field since rules can also be generated from custom rules, which won't include a documentation link.

ofekatr (approved):
LGTM :shipit: 
Left a minor comment regarding the property added to the `Rule` interface.

src/lib/iac/test/v2/scan/results.ts
ofekatr: In that case, can you please add a `TODO` comment for it?

PR #3618, opened 2022-08-23
fix: `--target-name` bug

YairZ101:

#### What does this PR do?
IaC has recently introduced a new flag in the CLI called `--target-name` that as the name suggest allows the user to overwrite the target name in the platform.
We currently have a bug where when scanning a directory that is a root of a repo the target name is not getting overridden with the one that the user set by using the flag.
This PR adds the ability to add the name field to the target together with git information, in registry we can use the `name` to overwrite the `target` name.

1. Run registry using the fix PR linked below
2. cd to a directory that is the root of a repo
3. Run `snyk-dev iac test --report --target-name=test`
4. Open the UI and check for the correct target name

The fix in this PR is dependent on the fix in https://github.com/snyk/registry/pull/28707.

https://snyksec.atlassian.net/jira/software/c/projects/CFG/boards/301?modal=detail&selectedIssue=CFG-2101

### DIFF
--- a/src/lib/ecosystems/types.ts
+++ b/src/lib/ecosystems/types.ts
@@ -20,7 +20,7 @@ export interface ContainerTarget {
   image: string;
 }
 
-export interface UnknownTarget {
+export interface NamedTarget extends GitTarget {
   name: string; // Should be equal to the project name
 }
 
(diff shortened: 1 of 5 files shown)


src/lib/ecosystems/types.ts
```diff
@@ -20,7 +20,7 @@ export interface ContainerTarget {
   image: string;
 }
 
-export interface UnknownTarget {
+export interface NamedTarget extends GitTarget {
```
YairZ101, line 23: I'm not 100% sure if this is the best approach here, maybe adding the name as an optional field to `GitTarget`  is a better option?

orsagie (approved):
LGTM

francescomari (approved):

PR #3620, opened 2022-08-24
chore: handle GOOS alpine in Makefile

PeterSchafer:
#### What does this PR do?
Add handling of alpine builds to the Makefile. It introduces the possibility to run

`make build GOOS=alpine GOARCH=amd64`

#### How does it work?
While for the golang cli this will actually result in a linux build, the bundled TS cli will have to be build explicitly for alpine.

### DIFF
--- a/test/jest/acceptance/iac/file-output.spec.ts
+++ b/test/jest/acceptance/iac/file-output.spec.ts
@@ -85,28 +85,36 @@ describe('iac test --sarif-file-output', () => {
     expect(startLine).not.toEqual(-1);
   });
 
-  it('returns the correct paths for provided path', async () => {
-    const sarifOutputFilename = path.join(__dirname, `${uuidv4()}.sarif`);
-    const { stdout } = await run(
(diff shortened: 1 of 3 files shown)


test/jest/acceptance/iac/file-output.spec.ts
```diff
@@ -85,28 +85,36 @@ describe('iac test --sarif-file-output', () => {
...
-    expect(actualProjectRoot).toEqual(
-      pathToFileURL(path.join(path.resolve(''), '/')).href,
-    );
-  });
+  if (
+    process.env.TEST_SNYK_COMMAND &&
+    process.env.TEST_SNYK_COMMAND.includes('alpine')
+  ) {
+    // This test is temporarily skipped on Alpine Linux.
+    it.skip('returns the correct paths for provided path', async () => {});
```
Avishagp, line 93: Just to make it easier to understand, could you insert the condition to within the test? and then skip it inside like we do with isCLIv2?

Avishagp (approved):

PR #3640, opened 2022-08-30
feat: create temp filepath for iac engine to write results [CFG-2114]

ipapast:
#### What does this PR do?

This commit creates a temporary filepath and passes it through the `-output` flag to the snyk-iac-test executable.
The snyk-iac-test will then write all the results to this file, which the CLI will have to read and present to the user.

Notes: 
- the flag is not visible to the user, this will all happen in the background.
- the file is created with empty content on the CLI side, and then gets overwritten by the snyk-iac-test with the results. This comment explains also reasoning about runnings scans in parallel: https://snyksec.atlassian.net/browse/CFG-2114?focusedCommentId=509517
- the file is deleted when the results are read on the CLI side.
- if errors happen during the file creation they will get caught on the CLI side too.
- if errors happen during the file writing we will handle them on the snyk-iac-test side and return with exit code 1 to the CLI.

Use this in combination with https://github.com/snyk/snyk-iac-test/pull/106 to test.

- Go to the branch of snyk-iac-test above and build locally: `go build -o snyk-iac-test .`
- Create a bundle in `opa-rules` repository if you don't have one: `make dist`
- Then run a scan as usual to see results:
`SNYK_IAC_POLICY_ENGINE_PATH=~/snyk-iac-test/snyk-iac-test SNYK_IAC_BUNDLE_PATH=~/opa-rules/dist.tar.gz snyk-dev iac test test/fixtures/iac/terraform/sg_open_ssh.tf --experimental`

﻿﻿
image

### DIFF
--- a/src/lib/iac/test/v2/scan/index.ts
+++ b/src/lib/iac/test/v2/scan/index.ts
@@ -4,11 +4,7 @@ import { CustomError } from '../../../../errors';
 import { IaCErrorCodes } from '../../../../../cli/commands/test/iac/local-execution/types';
 import { getErrorStringCode } from '../../../../../cli/commands/test/iac/local-execution/error-utils';
 import * as newDebug from 'debug';
-import {
-  mapSnykIacTestOutputToTestOutput,
-  SnykIacTestOutput,
-  TestOutput,
-} from './results';
+import { mapSnykIacTestOutputToTestOutput, TestOutput } from './results';
 import * as os from 'os';
 import * as fs from 'fs';
 import * as path from 'path';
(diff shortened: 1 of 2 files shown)


src/lib/iac/test/v2/scan/index.ts
```diff
@@ -159,16 +167,18 @@ function createConfig(options: TestConfig): string {
     });
 
     fs.writeFileSync(tempConfig, configData);
+    fs.writeFileSync(tempOutput, '');
```
francescomari, line 170: If `snyk-iac-test` creates a file if missing, there is no need to write anything here. Just return the path where you want the output file to be. See my comment in the corresponding PR.

src/lib/iac/test/v2/scan/index.ts
```diff
@@ -159,16 +167,18 @@ function createConfig(options: TestConfig): string {
...
-    throw new ScanError(`unable to create config file: ${e}`);
+    throw new ScanError(`unable to create config/output file: ${e}`);
   }
 }
 
-function deleteConfig(configPath) {
+function deleteTemporaryFiles(configPath: string, outputPath: string) {
   try {
     rimraf.sync(path.dirname(configPath));
+    rimraf.sync(path.dirname(outputPath));
```
francescomari, line 181: Why not removing the temporary directory at once?

francescomari (approved):

src/lib/iac/test/v2/scan/index.ts
ipapast: updated in https://github.com/snyk/cli/pull/3640/commits/0a0e00e006a45bfc864cfa7700a57cec1ee4bfb4 and also updated the snyk-iac-test to create the file.

francescomari (approved):

PR #3817, opened 2022-09-05
chore: implement --about

PeterSchafer:

#### What does this PR do?
This PR add golang based licenses to the --about dialog. It adds retrieving the licenses on the fly to the build process, bundles them into the binary and prints them to the end of the typescript based license content.

* Retrieve licenses: scripts/prepare_licenses.sh
* Bundle licenses: internal/embedded/file.go
* Print licenses: internal/cliv2/cliv2.go

Run the cli with `--about`, pipe the output into a file and compare the output with the same output for the typescript cli

### DIFF
--- a/cliv2/.gitignore
+++ b/cliv2/.gitignore
@@ -10,3 +10,4 @@ _cache
 .dccache
 .vscode
 bin
+internal/embedded/_data
--- a/cliv2/internal/cliv2/cliv2.go
+++ b/cliv2/internal/cliv2/cliv2.go
@@ -48,6 +48,7 @@ const SNYK_CA_CERTIFICATE_LOCATION_ENV = "NODE_EXTRA_CA_CERTS"
 const (
 	V1_DEFAULT Handler = iota
 	V2_VERSION Handler = iota
+	V2_ABOUT   Handler = iota
 )
 
 //go:embed cliv2.version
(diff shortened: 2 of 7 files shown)


cliv2/.gitignore
```diff
@@ -10,3 +10,5 @@ _cache
 .dccache
 .vscode
 bin
+internal/embedded/_data
+**/MADE_UP_NAME
```
Avishagp, line 14: What is that? 👀 
also, could you please add a new line to the end of the file?

PeterSchafer: Q: What is that? 👀
A: Ignoring two directories, if this is was the question.

Avishagp: I mean the  MADE_UP_NAME part

PeterSchafer: Removed `**/MADE_UP_NAME` from gitignore instead cleaning up folder after test

cliv2/internal/cliv2/cliv2.go
```diff
@@ -134,13 +135,47 @@ func (c *CLI) commandVersion(passthroughArgs []string) int {
...
+	}
+
+	separator := "\n+-+-+-+-+-+-+\n\n"
+
+	allEmbeddedFiles := embedded.ListFiles()
+	for i := range allEmbeddedFiles {
+		f := &allEmbeddedFiles[i]
+		path := f.Path()
+
+		//fmt.Println(path)
```
Avishagp, line 152: think this was supposed to be deleted?

cliv2/scripts/prepare_licenses.sh
```diff
@@ -0,0 +1,26 @@
...
+# try to find all licenses via the go.mod file
+export GOBIN=$(pwd)/_cache
+go install github.com/google/go-licenses@latest
+PATH="$PATH:$GOBIN" go-licenses save ./... --save_path=./internal/embedded/_data/licenses --force --ignore github.com/snyk/cli/cliv2/
+
+manualLicenseDownload "https://raw.githubusercontent.com/davecgh/go-spew/master/LICENSE" github.com/davecgh/go-spew
+manualLicenseDownload "https://raw.githubusercontent.com/alexbrainman/sspi/master/LICENSE" github.com/alexbrainman/sspi
+manualLicenseDownload "https://raw.githubusercontent.com/pmezard/go-difflib/master/LICENSE" github.com/pmezard/go-difflib
+manualLicenseDownload "https://raw.githubusercontent.com/go-yaml/yaml/v3.0.1/LICENSE" gopkg.in/yaml.v3
+manualLicenseDownload "https://go.dev/LICENSE?m=text" go.dev
```
Avishagp, line 19: So if we add a new package we're gonna have to add the license manually here?

cliv2/internal/embedded/file.go
```diff
@@ -0,0 +1,88 @@
+package embedded
```
Avishagp, line 1: Why is this file named `file.go`?

cliv2/scripts/prepare_licenses.sh
PeterSchafer: It depends, we first run go-licenses, which should if working normally find the license, if it doesn't find the license, it needs to be added here.

cliv2/internal/cliv2/cliv2.go
PeterSchafer: removed!

cliv2/scripts/prepare_licenses.sh
Avishagp: Ok cool, so this is fine for now but we should think of a better, dynamic way for the future?

PeterSchafer: Correct, I created the follow up task HMMR-535.

Avishagp (approved):

PR #3825, opened 2022-09-06
chore: install git on alpine and re-enable test

PeterSchafer:

#### What does this PR do?
Install git on the alpine test container and re-enable the previously skipped test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -383,7 +383,7 @@ jobs:
       - run:
           name: Installing Node.js + other test dependencies
           command: |
-            apk add --update nodejs npm bash maven
+            apk add --update nodejs npm bash maven git
       - setup_npm:
           npm_cache_directory: /home/circleci/.npm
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -383,7 +383,7 @@ jobs:
       - run:
           name: Installing Node.js + other test dependencies
           command: |
-            apk add --update nodejs npm bash maven
+            apk add --update nodejs npm bash maven git
```
ofekatr, line 386: Question: What's the difference between the Alpine image used for testing and a standard Alpine image? 
Would Alpine images usually include `git` out of the box?

PeterSchafer: For testing we use a standard Alpine image. Alpine is a hardened linux distribution which comes with almost nothing and not even glibc.

ofekatr (approved):
⭐

Avishagp (approved):

PR #3835, opened 2022-09-09
chore: Introduce analytics in Extensible CLI

PeterSchafer:

#### What does this PR do?
This PR adds a basic implementation of the analytics module in golang. Therefore introduces a global configuration package to access configuration from different sources in a unified way. 
And some minor refactorings to reduce coupling and avoid cyclic dependencies.

For the integration of the feature the best location is `cliv2/cmd/cliv2/main.go`.

Run any command and use for example logz.io to see the events arrive. Commands that execute in the legacy CLI will create two analytic entries.

HMMR-528


cliv2/cmd/cliv2/main.go
```diff
@@ -75,39 +115,47 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
...
 	if err != nil {
 		fmt.Println("Failed to create proxy")
 		fmt.Println(err)
-		return cliv2.SNYK_EXIT_CODE_ERROR
+		cliAnalytics.AddError(err)
+		return constants.SNYK_EXIT_CODE_ERROR
 	}
 
 	wrapperProxy.SetUpstreamProxyAuthentication(envVariables.ProxyAuthenticationMechanism)
+	http.DefaultTransport = wrapperProxy.Transport()
```
Avishagp, line 141: Why was this added?

PeterSchafer: This ensures that the global DefaultTransport uses the Kerberos Proxy Authentication. Since Analytics traffic in this case doesn't go through the internal proxy. Which would technically work but would add unreasonable extra computations.

PR #3838, opened 2022-09-12
chore: Create smoke tests for experimental IaC tests

ofekatr:

#### What does this PR do?

- Creates smoke tests for the experimental IaC test.
- Creates a new Github Action to run these smoke tests every hour and for every PR with IaC files.

#### Notes for the reviewer

- As of writing this, the `IAC_SMOKE_TESTS_SNYK_TOKEN` secret with the authentication token was yet to be added by team Hammer, so the tests still fail on the authentication step of the `snyk iac test --experimental` command.

After [CFG-2061: Spike options for testing/monitoring resources availability for the CLI](https://snyksec.atlassian.net/browse/CFG-2061) , we ended up on agreeing that a smoke test which downloads the artefacts and runs an IaC scan will give us enough confidence that the reyling artefacts are available and scans are working.

- Language of choice:  Jest

- CI of choice: Github Action

- Note: Jest vs shellspec was a tie in the original spike, I reached out to Hammer to see if they have seen any benefits of using shellspec, and I think they just transferred them to shellspec for historic reasons. they are actually thinking of keeping one language and leaning to migrate them to jest for v2.

- [CFG-2106](https://snyksec.atlassian.net/browse/CFG-2106)
- [Slack thread](https://snyk.slack.com/archives/C0127HWU0E7/p1662981253519489)

### DIFF
--- a/.github/workflows/iac-smoke-tests.yml
+++ b/.github/workflows/iac-smoke-tests.yml
@@ -1,0 +1,67 @@
+name: Infrastructure as Code Smoke Tests
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+  release:
+    types: [published]
+  workflow_call:
+
+jobs:
+  run_iac_e2e_tests:
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu, macos, windows]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 15
+
+      - name: Install jq on macOS
+        if: ${{ matrix.os == 'macos' }}
+        run: |
+          brew install jq
+
+      - name: Install jq on Windows
+        if: ${{ matrix.os == 'windows'}}
+        run: |
+          iwr -useb get.scoop.sh -outfile 'install-scoop.ps1'
+          .\install-scoop.ps1 -RunAsAdmin
+          scoop install jq
+
+      - name: Install jq on Ubuntu
+        if: ${{ matrix.os == 'ubuntu' }}
+        run: |
+          sudo apt-get install jq
+
+      - name: Install dependencies
+        run: |
+          npm install
+
+      - name: Build Snyk CLI
+        run: |
+          npm run build
+
+      - name: Run IaC smoke tests - non-Windows
+        if: ${{ matrix.os != 'windows' }}
+        env:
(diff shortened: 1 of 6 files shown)


.github/workflows/iac-smoke-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: Build Snyk CLI
+        run: |
+          npm run build
+
+      - name: Run IaC smoke tests - non-Windows
+        if: ${{ matrix.os != 'windows' }}
+        env:
+          SNYK_TOKEN: ${{ secrets.IAC_SMOKE_TESTS_SNYK_TOKEN }}
+          TEST_SNYK_COMMAND: ${{ format('node {0}/dist/cli/index.js', github.workspace) }}
```
ofekatr, line 55: It was also considered to download and use the binary executable for the CLI, but this was option was dismissed for 2 reasons:
- The test is executed via a `package.json` script, which means the code checkout and dependencies installation are already required regardless, so not much overhead is added when building the CLI locally.
- This decreases the chances for any flakiness results from the binaries not being available for downloading for any of the operating systems. Additionally, this lets us isolate the inspection of the execution of IaC commands in the tests, without adding any additional steps.
- The binary executables are already tests in other smoke tests, so there's no need to include these here as well if not directly needed as part of what's being tested.

test/smoke/iac/test.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  it('resolves with a non-error exit code', async () => {
+    // Arrange
+    const filePath = 'iac/depth_detection/root.tf';
+
+    // Act
+    const { exitCode } = await run(`snyk iac test --experimental ${filePath}`);
+
+    // Assert
+    expect(exitCode).toBeLessThan(2);
+  });
```
ipapast, line 29: is there a reason that this is a new test case? can't we add the exitcode assertion on the first test suite so that we don't run the command twice?

ipapast (commented):
Can you add a README.md in the `test/smoke/iac` directory with a couple of sentences describing what the GH action does, and how you could run the files without the Github Action (npm run ...) ?

.github/workflows/iac-smoke-tests.yml
Avishagp: Hey, I don't really see why use a locally built version of the CLI instead of the CDN? CDN was never flakey so I can't really relate to the second point. I'm worried about using different methods of installation in different smoke tests. The solution here works but is quick and dirty, so I suggest using the CDN as a better, reliable solution instead

ofekatr: The idea is to test the CLI that includes the changes introduced in the PR, which will not be available with binary executables downloaded from the CDN

Avishagp (approved):

ipapast (approved):

PR #3841, opened 2022-09-13
feat: share results with the Cloud API

francescomari:

#### What does this PR do?

This PR upgrades the version of `snyk-iac-test` to enable sharing results with the Snyk Cloud API. Moreover, it enables users to use the `--report`, `--target-name`, and `--remote-repo-url` flags, which can be used to influence how the information about the shared results are computed.

**Note** The `--target-name` flag only works the first time the environment is created. This is because of a bug in the implementation of the Cloud API that is being fixed right now. Once the fix is in place, the user will be able to change the environment's name when sharing results even if the environment already exists.

Share results with:

```
snyk iac test --experimental --report path/to/scan
```

[CFG-2142](https://snyksec.atlassian.net/browse/CFG-2142)


YairZ101 (approved):
🎉

PR #3851, opened 2022-09-15
fix: unmanaged scan unknown archives

gitphill:

#### What does this PR do?
See https://github.com/snyk/snyk-mvn-plugin/pull/137

Adding warning message to indicate when projects have a dep-graph that contains packages with unknown versions (like in the Snyk UI).

#### Screenshots
Screenshot 2022-09-28 at 14 53 32

### DIFF
--- a/src/lib/formatters/test/display-result.ts
+++ b/src/lib/formatters/test/display-result.ts
@@ -22,6 +22,7 @@ import {
   getDisplayedOutput,
 } from '../../../lib/formatters/test/format-test-results';
 import { showMultiScanTip } from '../show-multi-scan-tip';
+import * as theme from '../../theme';
 
 export function displayResult(
   res: TestResult,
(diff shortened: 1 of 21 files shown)


src/lib/formatters/test/display-result.ts
```diff
@@ -66,6 +66,12 @@ export function displayResult(
   );
   const multiProjAdvice = multiProjectTip ? `\n\n${multiProjectTip}` : '';
 
+  const hasUnknownVersions = res.hasUnknownVersions
+    ? `\n\n${chalk.yellow(
```
JackuB, line 70: I'd recommend using the theme file instead of the chalk library directly. There is already a `warn` helper https://github.com/snyk/cli/blob/2831b80c9f39ff881a35e2d30640a5fcd3dd91fc/src/lib/theme.ts#L14

JackuB: DItto for errors a file lines below in this PR

JackuB (approved):
Approving for the hammer owned files. But left a suggestion

src/lib/formatters/test/display-result.ts
gitphill: ah yea good idea, we will do that thanks

PR #3859, opened 2022-09-22
Redact static strings from analytics

maxjeffos:

### DIFF
--- a/cliv2/internal/analytics/analytics_test.go
+++ b/cliv2/internal/analytics/analytics_test.go
@@ -113,3 +113,74 @@ func Test_SanitizeValuesByKey(t *testing.T) {
 	assert.Equal(t, expectedNumberOfRedacted, secretsCountBefore)
 	assert.Equal(t, 0, secretsCountAfter)
 }
+
+func Test_SanitizeUsername(t *testing.T) {
+	type sanTest struct {
+		ErrorLog string
+		Other    string
+	}
+
+	rawUserName := "someuser"
+	simpleUsername := "someuser"
+	replacement := "REDACTED"
+
+	inputStruct := sanTest{
+		ErrorLog: "/Users/" + rawUserName + "/some/path",
+		Other:    fmt.Sprintf("some other value where %s is contained", rawUserName),
+	}
+
+	input, _ := json.Marshal(inputStruct)
+	fmt.Println("Before: " + string(input))
+
+	// invoke method under test
+	output, err := SanitizeUsername(rawUserName, replacement, input)
+
+	fmt.Println("After: " + string(output))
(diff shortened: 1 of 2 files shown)


cliv2/internal/analytics/analytics_test.go
```diff
@@ -113,3 +114,65 @@ func Test_SanitizeValuesByKey(t *testing.T) {
...
+	numUsernameInstances := strings.Count(string(output), username)
+	assert.Equal(t, 0, numUsernameInstances)
+
+	// check that it works for Windows usernames
+	username = "somedomain\\someuser"
+	fmt.Println("username: " + username)
+	needles = []string{username}
+
+	inputStruct = sanTest{
+		ErrorLog: fmt.Sprintf("C:\\Users\\%s\\some\\path", username),
```
PeterSchafer, line 159: chore: The issue we want to fix, is that analytics data might contain go error messages have the user name included, for example if they contain a path, like the home dir. The path doesn't contain the domain, the test would better reflect the issue, if it would use for example user.HomeDir instead of constructing a path. When using HomeDir, the test actually fails on Windows.

The reason is that the path on Windows looks like this
`C:\Users\UserName`

But the username retrieved here looks like this
`DOMAIN\UserName`

One solution could be to strip away the domain, if available, and use the bare user name to search and replace.

cliv2/internal/analytics/analytics_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/user"
```
PeterSchafer, line 9: issue: imported but not used

cliv2/internal/analytics/analytics_test.go
```diff
@@ -113,3 +114,74 @@ func Test_SanitizeValuesByKey(t *testing.T) {
...
+	assert.Equal(t, "/Users/REDACTED/some/path", outputStruct.ErrorLog)
+	assert.Equal(t, "some other value where REDACTED is contained", outputStruct.Other)
+
+	// Check with Windows style domain\username
+	rawUserName = "somedomain\\someuser"
+	simpleUsername = "someuser"
+	replacement = "REDACTED"
+
+	inputStruct = sanTest{
+		ErrorLog: fmt.Sprintf("C:\\Users\\%s\\some\\path", rawUserName),
```
PeterSchafer, line 162: nitpick: here it should be "simpleUsername" but since the expectation also contains the domain name it matches.

maxjeffos: make sense, thanks

PeterSchafer (approved):

PR #3883, opened 2022-09-29
Snyk Protect: embed patches

JackuB:
This PR is for `@snyk/protect` package.

This change will make Snyk Protect work without calling Snyk API.
This is achieved by running (ideally) an one-off script to download all patches and metadata needed for patching. Result is a JSON file and a folder all the patches.

This affects package size, but not dramatically.
This change was made with minimal changes to the Protect package code itself, updating only the methods fetching patches to use the local JSON file and local patches.

Screen Shot 2022-09-29 at 15 35 07

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20089,11 +20089,45 @@
       },
       "devDependencies": {
         "cross-spawn": "^6.0.5",
-        "fs-extra": "^9.1.0"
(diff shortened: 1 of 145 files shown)


thisislawatts:
🧹 Closing as stale. No activity in more than 1 year.

PR #4105, opened 2022-10-05
feat: disable container app scan with feature flag

tommyknows:
#### What does this PR do?

This commit adds a check for the `containerCliAppVulnsEnabled` feature flag. If neither `--exclude-app-vulns` nor `--app-vulns` are set, the feature flag will be used to determine whether app vulnerabilities will be scanned or not.

Part of the work to enable app vulnerability scanning in containers by default. Build upon @yaronschwimmer's [change](https://github.com/snyk/cli/commit/d2aac4ba96cef4c93e75cbdf800fc2f7f3125192), but that needed some adjustments to ensure the ordering and precedence of flags.

https://snyksec.atlassian.net/browse/MYC-177

### DIFF
--- a/scripts/apply-patches.sh
+++ b/scripts/apply-patches.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+
+echo "Applying patches from vendor directory"
+
+for filename in ./vendor/*.patch; do
+    if ! git apply --reverse --check --quiet "$filename"; then
+        echo "Applying $filename"
+        git apply "$filename"
+    fi;
+done
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -48,6 +48,7 @@ import { isMultiProjectScan } from '../../../lib/is-multi-project-scan';
 import { getEcosystem, monitorEcosystem } from '../../../lib/ecosystems';
 import { getFormattedMonitorOutput } from '../../../lib/ecosystems/monitor';
 import { processCommandArgs } from '../process-command-args';
+import { hasFeatureFlag } from '../../../lib/feature-flags';
 
 const SEPARATOR = '\n-------------------------------------------------------\n';
 const debug = Debug('snyk');
@@ -96,14 +97,31 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
   // TODO remove 'app-vulns' options and warning message once
   // https://github.com/snyk/cli/pull/3433 is merged
   if (options.docker) {
-    if (!options['app-vulns'] || options['exclude-app-vulns']) {
+    // order is important here, we want:
+    // 1) exclude-app-vulns set -> no app vulns
+    // 2) app-vulns set -> app-vulns
+    // 3) neither set -> containerAppVulnsEnabled
+    if (options['exclude-app-vulns']) {
       options['exclude-app-vulns'] = true;
-    }
+    } else if (options['app-vulns']) {
+      options['exclude-app-vulns'] = false;
+    } else {
+      options['exclude-app-vulns'] = !(await hasFeatureFlag(
+        'containerCliAppVulnsEnabled',
+        options,
+      ));
 
-    // we can't print the warning message with JSON output as that would make
-    // the JSON output invalid.
-    if (!options['app-vulns'] && !options['json']) {
-      console.log(theme.color.status.warn(appVulnsReleaseWarningMsg));
+      // we can't print the warning message with JSON output as that would make
+      // the JSON output invalid.
+      // We also only want to print the message if the user did not overwrite
+      // the default with one of the flags.
+      if (
+        options['exclude-app-vulns'] &&
+        !options['json'] &&
+        !options['sarif']
+      ) {
+        console.log(theme.color.status.warn(appVulnsReleaseWarningMsg));
+      }
     }
   }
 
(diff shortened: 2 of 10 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -96,14 +97,27 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
...
+      options['exclude-app-vulns'] = !(await hasFeatureFlag(
+        'containerCliAppVulnsEnabled',
+        options,
+      ));
+
+      // we can't print the warning message with JSON output as that would make
+      // the JSON output invalid.
+      // We also only want to print the message if the user did not overwrite
+      // the default with one of the flags.
+      if (options['exclude-app-vulns'] && !options['json']) {
```
noaleibo1, line 118: It might be my basic TS knowledge, but don't we want here `options['exclude-app-vulns'] == null && !options['json']`?

src/cli/commands/test/index.ts
```diff
@@ -98,14 +98,27 @@ export default async function test(
   // TODO remove 'app-vulns' options and warning message once
   // https://github.com/snyk/cli/pull/3433 is merged
   if (options.docker) {
-    if (!options['app-vulns'] || options['exclude-app-vulns']) {
+    // order is important here, we want:
```
noaleibo1, line 101: I guess there isn't a nice way to reuse the code we have for monitor..?

test/acceptance/fake-server.ts
```diff
@@ -10,6 +10,7 @@ const featureFlagDefaults = (): Map<string, boolean> => {
   return new Map([
     ['cliFailFast', false],
     ['iacIntegratedExperience', false],
+    ['containerCliAppVulnsEnabled', false],
```
noaleibo1, line 13: Is this the only testing we have? Makes me feel a little uncomfortable 😅

src/cli/commands/monitor/index.ts
tommyknows: I don't think so, though I agree that having a negated boolean [is confusing](https://stackoverflow.com/questions/2526886/naming-booleans) (`exclude-app-vulns` vs. `app-vulns`). 

Either way, we want to print the message if `exclude-app-vulns` is `true`. Also, this condition is within the larger if-else chain of "has the user overwritten this through the CLI". We only get to this expression if the user *has not* done so, and relies on the setting of the feature flag on the API.

The reasoning behind this is that if the user explicitly sets `--app-vulns` or `--exclude-app-vulns`, he already knows about this & does not need a warning message :)

src/cli/commands/test/index.ts
tommyknows: I guess there is, but that would probably be for a separate ticket. It would be a good idea though so that we also don't need reviews from other teams when changing these parts....

test/acceptance/fake-server.ts
tommyknows: Not directly - I actually had a couple of tests fail before I set that, as it returned `true` as the default value? By setting this here we make sure the fake API returns false, which makes our tests pass :) 

I can add some more tests to check if everything behaves correctly if the API returns `true` though.

src/cli/commands/test/index.ts
tommyknows: (I'd also like to split out the API feature flag names into a separate package with constants, so that we don't need to copy that string...)

noaleibo1: 👍 for both

test/acceptance/fake-server.ts
noaleibo1: Great!

noaleibo1 (approved):

tommyknows:
Added a second commit to fix the test failures.

See the commit message for a proper description. What I want to add is why this just came up with this PR, and has not been an issue before.

`global-agent` has two different initialisation routines, either by `import 'global-agent/bootstrap';` or by `import {bootstrap} from 'global-agent'; bootstrap()`.

The second way means that `bootstrap` needs to be explicitly called, and before this PR it was only called *after* we were done talking to the docker socket.

By adding the `hasFeatureFlag` call, which in turn calls `bootstrap`, the agent has now been replaced *before* we talked to the docker socket, thus triggering the issue.

PeterSchafer (approved):

danlucian (approved):

scripts/apply-patches.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
danlucian: Okey dokey, sounds good to me.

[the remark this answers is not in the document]
tony-snoop: Hi @tommyknows - you mentioned in the above comment to let you know if it's not good enough.

Just a heads up about this issue that started happening today:
https://github.com/snyk/cli/issues/4183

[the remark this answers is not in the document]
danlucian: Really sorry for this, @tony-snoop! And thank you a lot for reporting it!
We are rollbacking the changes right now: https://github.com/snyk/cli/pull/4184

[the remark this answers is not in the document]
tony-snoop: It's all good - thanks for jumping onto the fix so quickly @danlucian and for rolling back the changes.
Really REALLY appreciated :)

Quick question - do you know when the new version will become publicly available on https://www.npmjs.com/package/snyk ?

It's currently on version `1.1045.0`. I'm refreshing the page regularly waiting for `1.1045.0` to appear :)

PR #4109, opened 2022-10-06
feat: use in_progress to decide if unmanage-deps task has finished

dekelund:

#### What does this PR do?

Make use of InProgress-field to decide if we are going to retry our unmanaged tasks.

### DIFF
--- a/src/lib/ecosystems/resolve-test-facts.ts
+++ b/src/lib/ecosystems/resolve-test-facts.ts
@@ -24,13 +24,13 @@ import { findAndLoadPolicy } from '../policy';
 import { filterIgnoredIssues } from './policy';
 import { IssueData, Issue } from '../snyk-test/legacy';
 import { hasFeatureFlag } from '../feature-flags';
-import { delayNextStep } from '../polling/common';
 import {
   convertDepGraph,
   convertMapCasing,
   convertToCamelCase,
   getSelf,
 } from './unmanaged/utils';
+import { sleep } from '../common';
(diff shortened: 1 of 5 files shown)


src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -206,11 +200,16 @@ export async function resolveAndTestFactsUnmanagedDeps(
           packageManager,
         });
       } catch (error) {
+        // TODO(dekelund): Add handling of max retries
```
dekelund, line 203: We need to implement this

src/lib/polling/polling-test.ts
```diff
@@ -54,24 +53,30 @@ export async function getDepGraph(
...
-    } catch (e) {
-      if (waitingTimeMs < maxWaitingTimeMs) {
-        waitingTimeMs += pollIntervalMs;
-      } else {
-        throw e;
-      }
+  const maxIntervalMs = 60000;
+  const minIntervalMs = 5000;
+
+  const maxAttempts = 31; // Corresponds to 25.5 minutes
```
dekelund, line 59: we could of course just use something like now - starttime as well, similar to davids old code. But instead of calculating it with sum, just subst. it

Avishagp (approved):

Avishagp (dismissed):
Tests are failing, agreed on waiting longer with approving

PeterSchafer (approved):

PR #4126, opened 2022-10-12
Use basic auth for internal CLI proxy

maxjeffos:
Adds basic authentication to the internal CLI golang proxy and then invokes TS CLI with proxy credentials.

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -140,7 +140,7 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 	wrapperProxy.SetUpstreamProxyAuthentication(envVariables.ProxyAuthenticationMechanism)
 	http.DefaultTransport = wrapperProxy.Transport()
 
-	port, err := wrapperProxy.Start()
+	err = wrapperProxy.Start()
 	if err != nil {
 		fmt.Println("Failed to start the proxy")
 		fmt.Println(err)
@@ -149,7 +149,8 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 	}
 
 	// run the cli
-	err = cli.Execute(port, wrapperProxy.CertificateLocation, args)
+	proxyInfo := wrapperProxy.ProxyInfo()
+	err = cli.Execute(proxyInfo, wrapperProxy.CertificateLocation, args)
 	if err != nil {
 		cliAnalytics.AddError(err)
 	}
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -150,7 +150,7 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 
 	// run the cli
 	proxyInfo := wrapperProxy.ProxyInfo()
-	err = cli.Execute(proxyInfo, wrapperProxy.CertificateLocation, args)
+	err = cli.Execute(proxyInfo, args)
 	if err != nil {
 		cliAnalytics.AddError(err)
 	}
--- a/cliv2/internal/cliv2/cliv2.go
+++ b/cliv2/internal/cliv2/cliv2.go
@@ -124,9 +124,9 @@ func (c *CLI) commandVersion(passthroughArgs []string) error {
 	}
 }
 
-func (c *CLI) commandAbout(proxyInfo *proxy.ProxyInfo, fullPathToCert string, passthroughArgs []string) error {
+func (c *CLI) commandAbout(proxyInfo *proxy.ProxyInfo, passthroughArgs []string) error {
 
-	err := c.executeV1Default(proxyInfo, fullPathToCert, passthroughArgs)
+	err := c.executeV1Default(proxyInfo, passthroughArgs)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


cliv2/cmd/cliv2/main.go
```diff
@@ -140,7 +140,7 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 	wrapperProxy.SetUpstreamProxyAuthentication(envVariables.ProxyAuthenticationMechanism)
 	http.DefaultTransport = wrapperProxy.Transport()
 
-	port, err := wrapperProxy.Start()
+	_, err = wrapperProxy.Start()
```
PeterSchafer, line 143: Suggestion: how about removing the port here, since the information is now available via ProxyInfo.

cliv2/cmd/cliv2/main.go
```diff
@@ -149,7 +149,8 @@ func MainWithErrorCode(envVariables EnvironmentVariables, args []string) int {
 	}
 
 	// run the cli
-	err = cli.Execute(port, wrapperProxy.CertificateLocation, args)
+	proxyInfo := wrapperProxy.ProxyInfo()
+	err = cli.Execute(proxyInfo, wrapperProxy.CertificateLocation, args)
```
PeterSchafer, line 153: Suggestion: How about adding the CertificateLocation to ProxyInfo and thereby reducing the interface here.

cliv2/internal/proxy/proxy.go
```diff
@@ -90,13 +115,27 @@ func (p *WrapperProxy) replaceVersionHandler(r *http.Request, ctx *goproxy.Proxy
...
+	return user == p.proxyUsername && p.proxyPassword == password
+}
+
+func (p *WrapperProxy) HandleConnect(req string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
+	basic := auth.BasicConnect("", p.checkBasicCredentials)
+	action, str := basic.HandleConnect(req, ctx)
+
+	if action == goproxy.OkConnect {
+		action, str = goproxy.AlwaysMitm.HandleConnect(req, ctx)
+	}
```
PeterSchafer, line 128: thought: it could make sense to add a logging here? especially if the authentication fails.

cliv2/internal/proxy/proxy_test.go
```diff
@@ -109,16 +145,23 @@ func Test_xSnykCliVersionHeaderIsReplaced(t *testing.T) {
 	}))
 	defer ts.Close()
 
-	// request without the header set
-	res, err := proxiedClient.Get(ts.URL)
+	req, err := http.NewRequest("GET", ts.URL, nil)
+	req.Header.Set("Proxy-Authorization", fmt.Sprintf("Basic %s", basicAuthValue(proxy.PROXY_USERNAME, wp.ProxyInfo().Password)))
```
PeterSchafer, line 149: looks like we forgot to clean this up yesterday :)

cliv2/internal/proxy/proxy_test.go
```diff
@@ -109,16 +145,23 @@ func Test_xSnykCliVersionHeaderIsReplaced(t *testing.T) {
...
 	if err != nil {
 		t.Fatal(err)
 	}
 	assert.Equal(t, 200, res.StatusCode)
 	assert.Equal(t, "", capturedVersion)
 
 	// request with the header set
-	req, _ := http.NewRequest("GET", ts.URL, nil)
+	req, _ = http.NewRequest("GET", ts.URL, nil)
+	req.Header.Set("Proxy-Authorization", fmt.Sprintf("Basic %s", basicAuthValue(proxy.PROXY_USERNAME, wp.ProxyInfo().Password)))
```
PeterSchafer, line 164: looks like we forgot to clean this up yesterday :)

cliv2/internal/proxy/proxy_test.go
maxjeffos: 👍

cliv2/internal/proxy/proxy_test.go
maxjeffos: 👍

cliv2/internal/cliv2/cliv2.go
```diff
@@ -214,28 +215,26 @@ func PrepareV1EnvironmentVariables(input []string, integrationName string, integ
 
 }
 
-func PrepareV1Command(cmd string, args []string, proxyPort int, caCertLocation string, integrationName string, integrationVersion string) (snykCmd *exec.Cmd, err error) {
-
-	proxyAddress := fmt.Sprintf("http://127.0.0.1:%d", proxyPort)
+func PrepareV1Command(cmd string, args []string, proxyInfo *proxy.ProxyInfo, integrationName string, integrationVersion string) (snykCmd *exec.Cmd, err error) {
+	proxyAddress := fmt.Sprintf("http://%s:%s@127.0.0.1:%d", proxy.PROXY_USERNAME, proxyInfo.Password, proxyInfo.Port)
```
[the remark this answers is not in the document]
maxjeffos: It's a constant, though, so adding it to the struct seems weird to me.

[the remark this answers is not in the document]
PeterSchafer: It is a constant because you implemented it that way. In general it is a parameter to a module which should be handled as a parameter.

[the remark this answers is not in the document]
maxjeffos: I just don't see the point in making it a variable and dealing with that complexity when we can always just use the same value.

PR #4173, opened 2022-10-26
fix: Errors from snyk-iac-test should not be swallowed

francescomari:

#### What does this PR do?

This PR fixes a bug where errors returned by `snyk-iac-test` are silently swallowed instead of being thrown and displayed to the user.


moadibfr (approved):
Tested it for cloud context specific error so LGTM

PR #4189, opened 2022-11-02
fix: apps create command

love-bhardwaj:
* Update API version to 2022-03-11~experimental
* Update formatting out app create output
* Update field name for data fetched from the API
* Upate make request rest to handle 400 and greater than 400 errors separately
* Add no compression option to request payload

### DIFF
--- a/src/lib/request/promise.ts
+++ b/src/lib/request/promise.ts
@@ -35,10 +35,14 @@ export async function makeRequestRest<T>(payload: any): Promise<T> {
       if (error) {
         return reject(error);
       }
-      if (res.statusCode >= 400) {
+      if (res.statusCode === 400) {
         return reject({
           code: res.statusCode,
           body: JSON.parse(body as any),
+        });
+      } else if (res.statusCode >= 401) {
+        return reject({
+          code: res.statusCode,
         });
       }
       resolve(JSON.parse(body as any) as T);
--- a/src/lib/request/request.ts
+++ b/src/lib/request/request.ts
@@ -30,7 +30,7 @@ function setupRequest(payload: Payload) {
 
   const versionNumber = getVersion();
   const body = payload.body;
-  let data;
+  let data = body;
 
   delete payload.body;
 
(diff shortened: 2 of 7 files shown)


dicosnyk (approved):

src/lib/request/promise.ts
```diff
@@ -35,11 +35,15 @@ export async function makeRequestRest<T>(payload: any): Promise<T> {
...
       }
-      if (res.statusCode >= 400) {
+      if (res.statusCode === 400) {
         return reject({
           code: res.statusCode,
           body: JSON.parse(body as any),
         });
+      } else if (res.statusCode >= 401) {
+        return reject({
+          code: res.statusCode,
```
Avishagp, line 45: Why is `body` removed for any code other than 400?

src/lib/request/request.ts
```diff
@@ -40,7 +40,10 @@ function setupRequest(payload: Payload) {
 
   payload.headers['x-snyk-cli-version'] = versionNumber;
 
-  if (body) {
+  const noCompression = payload.noCompression;
+
+  if (body && !noCompression) {
```
Avishagp, line 45: Should we include another scenario for when noCompression is true? isn't that change affecting the logic?

src/lib/request/promise.ts
love-bhardwaj: Ideally, the body of all error responses would look like this:

```json
{
  "errors": [
    {
      "detail": "Permission denied for this resource",
      "status": "403"
    }
  ],
  "jsonapi": {
    "version": "1.0"
  }
}
```

Which can be used to create a useful message for the user. But unfortunately, Cerberus's responses like 401, and 403 are not JSON but a single string (also causing JSON parsing errors). For those cases, we are not using the response anyways. As we can just map the status code to an error message directly.

For 400 bad requests, it is still relevant as we want to process the response to show the user inputs can be wrong.

Hope that makes sense 😅

src/lib/request/request.ts
love-bhardwaj: The previous behaviour was every request body was being compressed. Now with `noCompression` you have an option to not compress the request body if you don't want to like in our case.

So by default `noCompression` will always be false and the request body will always be compression. Unless you explicitly tell it to not compress. I which case the body of the request is used as is.

Avishagp (approved):

PR #4201, opened 2022-11-09
fix: handle self being undefined

candrews:

#### What does this PR do?
`getSelf()` can return undefined so that case must be handled,  otherwise the following error will occur:
```
TypeError: Cannot read properties of undefined (reading 'default_org_context')
```

Take a look at the defintion of the `getSelf()`: https://github.com/snyk/cli/blob/e4d0eaf1bbdcec185890894350d67a446452f164/src/lib/ecosystems/unmanaged/utils.ts#L50

### DIFF
--- a/src/lib/ecosystems/resolve-test-facts.ts
+++ b/src/lib/ecosystems/resolve-test-facts.ts
@@ -52,6 +52,10 @@ export async function resolveAndTestFacts(
     : resolveAndTestFactsRegistry(ecosystem, scans, options);
 }
 
+async function getOrgDefaultContext(): Promise<string> {
+  return (await getSelf())?.data.attributes.default_org_context;
+}
+
 async function submitHashes(
   hashes: FileHashes,
   orgId: string,
@@ -154,12 +158,12 @@ export async function resolveAndTestFactsUnmanagedDeps(
   const packageManager = 'Unmanaged (C/C++)';
   const displayTargetFile = '';
 
-  let orgId = options.org || '';
+  const orgId = options.org || (await getOrgDefaultContext()) || '';
   const target_severity: SEVERITY = options.severityThreshold || SEVERITY.LOW;
 
   if (orgId === '') {
-    const self = await getSelf();
-    orgId = self.data.attributes.default_org_context;
+    errors.push('organisation-id missing');
+    return [results, errors];
   }
 
   for (const [path, scanResults] of Object.entries(scans)) {


candrews:
@dagrest @dekelund @danielroymoore tagging you since you seem to be working in this area

candrews:
@ipapast @ahmed-agabani-snyk @gitphill (or anyone else)
Can you please take a look at this PR so developers don't experience these false test failures?

dekelund:
Hi @candrews sorry for the late response. We will look into your suggestion and come back to you.

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -151,7 +151,7 @@ export async function resolveAndTestFactsUnmanagedDeps(
 
   if (orgId === '') {
     const self = await getSelf();
-    orgId = self.default_org_context;
+    orgId = self?.default_org_context;
   }
```
dekelund, line 156: Thanks for the feedback and suggestion, appreciated it.
We are not able to proceed if we lack the ID since it's used by HTTP calls in submitHashes, pollDepGraphAttribute, and fetchIssues.

Therefore I recommend we exit the execution early. To shutdown gracefully, we might want to handle it like this:

```typescript
function getOrgDefaultContext(): string {
  return (await getSelf())?.default_org_context;
}
```

```suggestion
let orgId = options.org || getDefaultContext() || '';

if (orgId === '') {
  errors.push('organisation-id missing');
  return [results, errors];
}
```

Which means it would fallback to a default value, and return it as an error to the user.
[Playground](https://www.typescriptlang.org/play?#code/PQKgUABCEEIE4HsDuBnApnCXs4gZQAc40BDAEwhQE8A7AFxIA9cIBjBG1kuiAWzToALBGUjQAtJKnSZsufIUyxEAMKDE-CAHYArC2wA2AEwGAdABYAzLxT6cRyztMBGaymUAxAJbEAZgmYDAA47AE4jIIsABhs7bCCjZ1NLGPcofBJfEjgvCGcjOOcDJ3NzWLi8nSNTKLK06ABRMgBzNDyQiuctIPNTLSNyuy0oywtnG2UAeQIMEghjCp1QqJdUiohE0fMdCfTFfYPD6TAQYDAwEmpOCF8AV046Lw4IVro8NAAbXwAKAEoALggAAUNF50AAeEg0KgAPggAG9IBBiHRbnAaBB7mQ0L4vDQ0GQANxI4DAFFojHw7FZW4fOgAfQQcGa9PY9DQjDogIARNSSLSeEzmuI2XQOXRuQBfSXEyXnD4CCAIAiPDi2AC8CKRQsBAHJdQAaMByi5XVg3e6sVUYuD3P5arAKwXMgCSFE1yutKFMQogAB8-RBviQkCQvDxXu8vn9fgB+Ux8gWM5msjhizn+wO6rIfD4AIxIrAA1rriVg2SgEArTB8EM1vkK3b9ZedbTQ-oSgA)

dekelund: A note, the "suggestion" feature in GitHub did not allow me to select lines before row 151, so it's not possible to add suggestion with the buttons above.

candrews: Fantastic, I've implemented your suggestion and updated this PR accordingly.

My test, which is to run:
`npm ci && NODE_OPTIONS="--openssl-legacy-provider" npm run build:prod && XDG_CONFIG_HOME="$(mktemp -d)" SNYK_TOKEN="not-used" npm run test:unit`
still works, so this looks great to me :+1: 

Thank you!

candrews:
Rebased and addressed all conversations. I believe this MR to be ready for review and (hopefully) merging.

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -52,6 +52,10 @@ export async function resolveAndTestFacts(
     : resolveAndTestFactsRegistry(ecosystem, scans, options);
 }
 
+async function getOrgDefaultContext(): Promise<string> {
+  return (await getSelf())?.data.attributes.default_org_context;
```
dekelund, line 56: One last thing, was it intentional to add `.data.attributes`, isn't that already taken care of by `getSelf`?

```suggestion
  return (await getSelf())?.default_org_context;
```

src/lib/ecosystems/resolve-test-facts.ts
dekelund: Thanks @candrews , I will resolve the conversion.
FYI I have added a suggestion and comment to the latest version as well.

src/lib/ecosystems/resolve-test-facts.ts
candrews: `getSelf()`  returns a [`SelfResponse`](https://github.com/snyk/cli/blob/7a6c44cfbd1a66b09efbd6001848179f54190420/src/lib/ecosystems/unmanaged/utils.ts#L48) which, as of commit https://github.com/snyk/cli/commit/ff897719447e12990881541b62e44e003122eda0#diff-50795afceb61f2719e051435cf21b7dc96eebb618447a6e1f8fc3df515e0d198, doesn't have a `default_org_context` property (before that commit, the `data.attributes` were moved up, but after that commit, they no longer are); that information is accessed by `.data.attributes.default_org_context`.

dekelund: You are right, I looked at the wrong branch.

dekelund (approved):

candrews:
Is there anything that I can do to help this along?

I'd love to see it merged 🤞

dekelund:
@candrews I will come back to you, I'm trying to get check called "Danger" to get triggered so that we are able to merge it.

dekelund:
I was enforced to create a new PR to get the permissions for the "Danger"-result reported back to this PR. Now I'm able to merge it @candrews.

PR #4203, opened 2022-11-10
Chore/consecutive iac smoke tests failures workflow

ofekatr:

#### What does this PR do?
- Removes the previous IaC smoke tests alerting job, which triggered Slack alerts for every single IaC smoke tests run.
- Creates a new hourly scheduled workflow that triggers Slack alerts only if the 2 latest consecutive IaC smoke tests workflow runs failed.

- The implementation for this workflow leverages the implementation for the `cli-alert` workflow, which achieves a similar goal.
- The current test runs are rather flakey and produce false-positives at least 3-4 times every week.

#### More information

- [Slack thread](https://snyk.slack.com/archives/C02J9KR30JE/p1668010540723339)

### DIFF
--- a/.github/workflows/iac-cli-alert.yml
+++ b/.github/workflows/iac-cli-alert.yml
@@ -1,0 +1,23 @@
+name: Infrastructure as Code CLI alert
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+
+jobs:
+  check_tests:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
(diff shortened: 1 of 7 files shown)


Avishagp (changes requested):
`package-lock.json` should not be changed in this PR

ofekatr:
> `package-lock.json` should not be changed in this PR
@Avishagp It should, as it lists the dependencies for each of the workspaces, including `@snyk/iac-cli-alert`

Avishagp (approved):

PR #4222, opened 2022-11-30
feat: yarn lock v2 support plus improvements

JamesPatrickGill:
This adds support for yarn lock v2 to the CLI aswell as a heavier use of depgraphs. 

It also carries some slight Yarn v1 changes through due to using an outdated nodejs parser in the CLI.


src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -18,19 +20,40 @@ export async function inspect(
...
     plugin: {
       name: 'snyk-nodejs-lockfile-parser',
       runtime: process.version,
     },
-    scannedProjects: [{ depTree }],
+    scannedProjects,
   };
 }
+
+function isResDepGraph(depRes: PkgTree | DepGraph): boolean {
```
admons, line 57: Use the following pattern and you won't have to cast after in your if block
```typescript
function isDepGraph(item: PkgTree | DepGraph): item is DepGraph {
  return 'rootPkg' in item;
}

File: src/lib/plugins/nodejs-plugin/npm-lock-parser.ts [hunk 2]
```diff
@@ -5,14 +5,17 @@ import { spinner } from '../../spinner';
 import * as analytics from '../../analytics';
 import * as fs from 'fs';
 import * as lockFileParser from 'snyk-nodejs-lockfile-parser';
-import { PkgTree } from 'snyk-nodejs-lockfile-parser';
+import { PkgTree, InvalidUserInputError } from 'snyk-nodejs-lockfile-parser';
 import { Options } from '../types';
+import { DepGraph } from '@snyk/dep-graph';
+import { NodeLockfileVersion } from 'snyk-nodejs-lockfile-parser/dist/utils';
```
admons, line 11: This should be imported on line 8

src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
```diff
@@ -5,14 +5,17 @@ import { spinner } from '../../spinner';
...
+import { ProjectParseOptions } from 'snyk-nodejs-lockfile-parser/dist/dep-graph-builders/types';
```
admons, line 12: If this is being used outside of the library, then it should be exported on it's `index.ts` file and imported here the same way on line 8

src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
```diff
@@ -61,3 +84,45 @@ export async function parse(
...
+
+  switch (lockfileVersion) {
+    case NodeLockfileVersion.YarnLockV1:
+      return await lockFileParser.parseYarnLockV1Project(
+        manifestFileContents,
+        lockFileContents,
+        options,
+      );
+    case NodeLockfileVersion.YarnLockV2:
+      return lockFileParser.parseYarnLockV2Project(
```
admons, line 121: Weird that this is not an async method (in the library, I checked), maybe it should?
anyway if you're not changing it then ignore this comment

admons (changes requested):

src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
JamesPatrickGill: We would force the function to be async, wonder where we could do this

src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -18,19 +20,38 @@ export async function inspect(
...
     : await modulesParser.parse(root, targetFile, options);
 
-  if (depTree?.meta?.lockfileVersion) {
-    analytics.add('lockfileVersion', depTree.meta.lockfileVersion);
+  let scannedProjects: any[] = [];
+  if (isResDepGraph(depRes)) {
+    if (depRes.pkgManager.version) {
+      analytics.add(
+        'lockfileVersion',
+        (depRes as DepGraph)?.pkgManager.version,
```
admons, line 32: ```depRes.pkgManager.version```

src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -18,19 +20,38 @@ export async function inspect(
...
+      );
+    }
+    scannedProjects = [{ depGraph: depRes }];
+  } else {
+    if (depRes.meta?.lockfileVersion) {
+      analytics.add(
+        'lockfileVersion',
+        (depRes as PkgTree)?.meta?.lockfileVersion,
```
admons, line 40: ```depRes.meta?.lockfileVersion```

src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
```diff
@@ -47,9 +53,29 @@ export async function parse(
   });
   const resolveModuleSpinnerLabel = `Analyzing npm dependencies for ${lockFileFullPath}`;
   debug(resolveModuleSpinnerLabel);
+
+  const strictOutOfSync = options.strictOutOfSync !== false;
```
admons, line 57: What if this is null/undefined?

admons (changes requested):

src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -18,19 +20,32 @@ export async function inspect(
...
+  const depRes: PkgTree | DepGraph = getLockFileDeps
     ? await lockParser.parse(root, targetFile, options)
     : await modulesParser.parse(root, targetFile, options);
 
-  if (depTree?.meta?.lockfileVersion) {
-    analytics.add('lockfileVersion', depTree.meta.lockfileVersion);
+  let scannedProjects: any[] = [];
+  if (isResDepGraph(depRes)) {
+    if (depRes.pkgManager.version) {
+      analytics.add('lockfileVersion', depRes?.pkgManager.version);
```
admons, line 30: Remove question mark

src/lib/plugins/nodejs-plugin/index.ts
```diff
@@ -18,19 +20,32 @@ export async function inspect(
...
+    }
+    scannedProjects = [{ depGraph: depRes }];
+  } else {
+    if (depRes.meta?.lockfileVersion) {
+      analytics.add('lockfileVersion', depRes?.meta?.lockfileVersion);
```
admons, line 35: Remove question marks

admons (approved):

PeterSchafer (commented):
Suggestion: As you have added new logic and control flows, it would be important to add new tests for it. From the description, I would assume to find at least one additional acceptance test that uses a yarn v2 file. Unit tests would be great too.

PeterSchafer (approved):
As discussed with @JamesPatrickGill, adding additional tests will be done in a separate task. Approving to unblock.

PR #4226, opened 2022-12-01
feat: update unmanaged endpoints

danielroymoore:
This PR does two things:

1. Switches the unmanaged dep calls to the hidden endpoint
2. Fixes a bug when polling that was causing it to wait for 60 seconds each attempt.

### DIFF
--- a/src/lib/ecosystems/resolve-test-facts.ts
+++ b/src/lib/ecosystems/resolve-test-facts.ts
@@ -64,20 +64,24 @@ async function pollDepGraphAttributes(
   id: string,
   orgId: string,
 ): Promise<Attributes> {
-  const maxIntervalMs = 60000;
   const minIntervalMs = 5000;
+  const maxIntervalMs = 30000;
 
-  const maxAttempts = 31; // Corresponds to 25.5 minutes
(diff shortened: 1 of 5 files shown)


src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -64,20 +64,24 @@ async function pollDepGraphAttributes(
...
 
-  const maxAttempts = 31; // Corresponds to 25.5 minutes
+  let totalElaspedTime = 0;
+  let attempts = 1;
+  const maxElapsedTime = 1800000; // 30 mins in ms
 
   // Loop until we receive a response that is not in progress,
   // or we receive something else than http status code 200.
-  for (let i = 1; i <= maxAttempts; i++) {
+  while (totalElaspedTime <= maxElapsedTime) {
```
dekelund, line 76: for me it's a bit harder to reason about if we both have a counter and a time to think of. That's why we removed the time measurements from the beginning. But it's not a big deal to me. The most important thing is that you found the issue with max vs min.

dekelund (approved):

PR #4227, opened 2022-12-01
fix: upgrade http client

gitphill:

#### What does this PR do?
Upgrading needle from 2.6.0 to 3.0.0.

This version now supports automatically parsing JSON when HTTP responses
send Content-Type application/vnd.api+json header, which the latest Snyk
APIs use.

Explicitly opting out of this automatic parsing in makeRequestRest
because this currently parses a response using Content-Type
application/text.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
         "lodash.values": "^4.3.0",
         "marked": "^4.0.1",
         "micromatch": "4.0.2",
-        "needle": "2.6.0",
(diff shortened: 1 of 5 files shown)


michelkaporin (approved):
LGTM 👍 

Here's [the link](https://snyk.slack.com/archives/C0127HWU0E7/p1669828695975799) to the discussion on why we didn't upgrade to the most latest, for historical reference.

PR #4247, opened 2022-12-15
chore: clean old cliv1 versions in cache directory

Avishagp:
[Jira task](https://snyksec.atlassian.net/browse/HMMR-594)

### DIFF
--- a/cliv2/internal/cliv2/cliv2.go
+++ b/cliv2/internal/cliv2/cliv2.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
 
 	"github.com/snyk/cli/cliv2/internal/constants"
@@ -60,6 +61,8 @@ func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
 		stderr:           os.Stderr,
 	}
 
+	cli.ClearCache()
+
 	err = cli.ExtractV1Binary()
 	if err != nil {
 		fmt.Println(err)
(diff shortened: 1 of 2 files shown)


cliv2/internal/cliv2/cliv2.go
```diff
@@ -69,6 +72,32 @@ func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
 	return &cli, nil
 }
 
+func ClearCache(cli *CLI) error {
```
PeterSchafer, line 75: Suggestion: since ClearCache() requires fields from the current cli instance, why not use the CLI as receiver instead of a parameter.

cliv2/internal/cliv2/cliv2.go
```diff
@@ -69,6 +72,32 @@ func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
...
+        currentPath := path.Join(cli.CacheDirectory, file.Name())
+		if currentPath != v1BinaryPath {
+			err = os.RemoveAll(currentPath)
+			if err != nil {
+				fmt.Println("Error deleting an old version directory")
+			}
+		}
+		// Stop the loop after 5 deletions to not create too much overhead
+		if i == 5 {
+			return nil
```
PeterSchafer, line 94: Suggestion: Use a break to end the loop. Otherwise somebody might add code below the loop that is only reached in some cases. And maybe create a constant for "5"

cliv2/internal/cliv2/cliv2_test.go
```diff
@@ -222,3 +223,72 @@ func Test_executeUnknownCommand(t *testing.T) {
...
+	os.Mkdir(dir1, 0755)
+	os.Mkdir(dir2, 0755)
+	os.Mkdir(dir3, 0755)
+	os.Mkdir(dir4, 0755)
+	os.Mkdir(dir5, 0755)
+	os.Mkdir(dir6, 0755)
+
+	// clear cache
+	err := cli.ClearCache()
+	if err != nil {
```
PeterSchafer, line 281: FYI: you can also use `assert.Nil(t, err)`

cliv2/internal/cliv2/cliv2.go
```diff
@@ -69,6 +72,33 @@ func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
...
+
+	// Get current version binary's path
+	v1BinaryPath := path.Dir(c.v1BinaryLocation)
+	var maxVersionToDelete = 5
+    for i, file := range fileInfo {
+        currentPath := path.Join(c.CacheDirectory, file.Name())
+		if currentPath != v1BinaryPath {
+			err = os.RemoveAll(currentPath)
+			if err != nil {
+				fmt.Println("Error deleting an old version directory")
```
PeterSchafer, line 90: Suggestion: replace fmt.Println() by c.DebugLogger.Println() and add the currentPath

PeterSchafer (approved):

PR #4249, opened 2022-12-15
Integrate support of NODE_EXTRA_CA_CERTS 

PeterSchafer:

#### What does this PR do?
If a user specifies additional root CAs via the environment variable NODE_EXTRA_CA_CERTS, they get added to the internal Certificate Pool and forwarded to child processes.

Also Golang is updated to 1.19.4 for building and throughout the CI pipeline.

* main.go 
* proxy.go

Run and look at "test/jest/acceptance/extra-certs.spec.ts"

The Golang CLI uses NODE_EXTRA_CA_CERTS itself to set temporary certificates for child processes. If there are user defined certificates they need to be merged.


cliv2/cmd/cliv2/main.go
```diff
@@ -89,10 +89,17 @@ func sendAnalytics(analytics analytics.Analytics, debugLogger *log.Logger) {
 
 	res, err := analytics.Send()
 	errorCodeReceived := res != nil && 200 <= res.StatusCode && res.StatusCode < 300
-	if err == nil && !errorCodeReceived {
+	if err == nil && errorCodeReceived {
```
Avishagp, line 92: This logic was changed but the logging didn't, is this right?

Avishagp (approved):

cliv2/cmd/cliv2/main.go
PeterSchafer: Yes, I stumbled across this while debugging. The logic was incorrect and it resulted in an error message when actually being successful. I could actually have changed the variable name too. shall I do this?

Avishagp: Yes, I think we should! will make it more clear

PeterSchafer: done :)

PR #4252, opened 2022-12-17
chore: update os-name and adapt import

PeterSchafer:

#### What does this PR do?
* Update npm package os-name
* Since os-name changed to a pure ESM package, the usage of require() was replaced or the usage of the package was avoided


src/lib/analytics/getStandardData.ts
```diff
@@ -41,7 +40,8 @@ export async function getStandardData(
 
   let osNameString;
   try {
-    osNameString = osName(os.platform(), os.release());
+    const osName = await import('os-name');
```
Avishagp, line 43: Not sure I understand why is the import inside the try catch?

PeterSchafer: Since it is a dynamic import it could potentially fail. And since we are talking about analytics, I'd rather catch anything and handle an error gracefully than affecting the whole application.

Avishagp: Why should this be a dynamic import?

PeterSchafer: The package author made this a pure ESM package, so the require syntax doesn't work anymore. See https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c for details.

Avishagp: Yeah, I mean why not import like you did in the other files with `import * as os from 'os';`?

PeterSchafer: Because it doesn't work. I guess it is because of the mixture of ESM, Typescript and Javascript.

Avishagp (approved):

strassl-snyk (approved):
LGTM for @snyk/zenith

magdziarek (approved):

PR #4255, opened 2022-12-21
chore: release golang cli on linux

Avishagp:
Release golang CLI on linux.
[JIRA ticket](https://snyksec.atlassian.net/browse/HMMR-567)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -563,6 +563,12 @@ jobs:
             cp cliv2/bin/snyk-macos binary-releases/snyk-macos
             cp cliv2/bin/snyk-macos.sha256 binary-releases/snyk-macos.sha256
       - run:
+          name: Copy linux cliv2 binaries to binary-releases staging area
+          command: |
+            ls -la cliv2/bin
+            cp cliv2/bin/snyk-linux binary-releases/snyk-linux
+            cp cliv2/bin/snyk-linux.sha256 binary-releases/snyk-linux.sha256
+      - run:
           name: Signing shasums
           command: make binary-releases/sha256sums.txt.asc
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1054,6 +1088,10 @@ workflows:
             - Build (snyk-linux-arm64)
           go_os: linux
           go_arch: arm64
+      - v2-rename-linux-artifact:
+          name: v2 / Rename linux/amd64 artifact
+          requires:
+            - v2 / Sign (linux/amd64)
```
PeterSchafer, line 1094: Question: I think this should be "Build (linux/amd64)"

.circleci/config.yml
```diff
@@ -1140,8 +1178,8 @@ workflows:
           name: v2 / Regression Tests (linux/amd64)
           context: nodejs-install
           requires:
-            - v2 / Build (linux/amd64)
-          test_snyk_command: /mnt/ramdisk/snyk/cliv2/bin/snyk_linux_amd64
+            - v2 / Rename macos/amd64 artifact
```
PeterSchafer, line 1181: Question: why is this requiring "v2 / Rename macos/amd64 artifact"? Shouldn't it be linux?

PeterSchafer (changes requested):
Unfortunately the build doesn't run, please see my comments on how to fix it.

.circleci/config.yml
Avishagp: My bad!

.circleci/config.yml
Avishagp: I used the same logic from `v2-rename-windows` and `v2-rename-macos`, both require signing before renaming. The logic is sign -> rename -> build. Does that make sense?

PeterSchafer: In general that would be right, but linux artifacts don't get signed right now. 

If you look at the build log https://app.circleci.com/pipelines/github/snyk/cli/16176, you can see that this unknown job is the problem of your build.

You can also look at your changes here: https://github.com/snyk/cli/pull/4255/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L1087 the prerequisite for the tests was "build" not "sign". So this should also be the prerequisite for the "rename".

I hope this helps.

Avishagp: I'm sorry! thanks for explaining that

PeterSchafer:
Please also take rename/replace "snyk_linux_arm64"

PeterSchafer (approved):

PR #4257, opened 2022-12-22
fix: align tmp and cache dir

PeterSchafer:

#### What does this PR do?
Avoid file system conflicts on linux and align where the CLI stores files on the local system.

**Background**
On linux, temp dir resolves to /tmp and we extended it to /tmp/snyk. If any user chose to use create a folder or file with the same name, there is a conflict.


Avishagp (approved):

PeterSchafer:
Pair Reviewed with @asaf92 after doing additional changes.

PR #4265, opened 2022-12-29
Fix/revert golang cli release linux

Avishagp:
Revert https://github.com/snyk/cli/pull/4264


PeterSchafer (approved):

PeterSchafer:
This PR is just here, so that we can undo the release of the golang based cli to linux if necessary.

PR #4266, opened 2022-12-29
Chore/hmmr 612 extcli in ts

PeterSchafer:

#### What does this PR do?
Introduce a Typescript based bootstrapper and wrapper to distribute and run the Golang based Extensible CLI.


ts-binary-wrapper/src/bootstrap.ts
```diff
@@ -0,0 +1,20 @@
+import * as common from './common';
+import * as process from 'process';
+
+const config = common.getCurrentConfiguration();
+export const executable = config.getLocalLocation();
+
+if (process.argv.includes('exec')) {
+  (async function() {
```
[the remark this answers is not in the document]
PeterSchafer: Probably not. It is just limited ts/js knowledge.

ts-binary-wrapper/src/common.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  return exitCode;
+}
+
+export async function downloadExecutable(
+  downloadUrl: string,
+  filename: string,
+  filenameShasum: string,
+): Promise<number> {
+  return new Promise<number>(function(resolve) {
```
Avishagp, line 190: Should we use `await` instead?

ts-binary-wrapper/src/bootstrap.ts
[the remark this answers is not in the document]
PeterSchafer: applied discussed changes

ts-binary-wrapper/src/common.ts
```diff
@@ -0,0 +1,254 @@
...
+    const options = new URL(downloadUrl);
+    const tempFilename = path.join(__dirname, randomInt(100000).toString());
```
Avishagp, line 192: Is this the right way to do this (with `randomInt`)? Should we maybe call it `temp` and delete it later? would that be possible?

ts-binary-wrapper/src/common.ts
```diff
@@ -0,0 +1,254 @@
...
+    const fileStream = fs.createWriteStream(tempFilename);
+
+    const cleanupAfterError = (exitCode: number) => {
+      try {
+        fs.unlinkSync(tempFilename);
+      } catch (e) {
+        //
```
Avishagp, line 199: Should logging happen here?

ts-binary-wrapper/src/common.ts
PeterSchafer: I general I think it is totally fine to use promises. Or do you see any reason why we should await here?

ts-binary-wrapper/src/common.ts
```diff
@@ -0,0 +1,254 @@
...
+
+        // compare shasums
+        const actualShasum = shasum.digest('hex');
+        console.debug(
+          'Shasums:\n- actual:   ' +
+            actualShasum +
+            '\n- expected: ' +
+            filenameShasum,
+        );
+        if (filenameShasum && actualShasum != filenameShasum) {
```
Avishagp, line 222: Should we add parenthesis here?

ts-binary-wrapper/src/common.ts
Avishagp: I think it's more confusing to use promises and was taught it was a bit of a bad practice nowadays

ts-binary-wrapper/src/common.ts
PeterSchafer: what should we call temp?

Avishagp: The tempFilename, I'm a bit worried about `randomInt(100000)`

ts-binary-wrapper/src/common.ts
PeterSchafer: I don't think we need logging here. The context is that some error occurred and we want to cleanup the temp file. if this fails we don't gain any value of a logging message.

ts-binary-wrapper/src/common.ts
PeterSchafer: do we have any team policy or code guide around the usage of promises?

ts-binary-wrapper/src/common.ts
PeterSchafer: I'm still not sure, do you mean this?
`const temp = path.join(__dirname, randomInt(100000).toString());`

ts-binary-wrapper/src/common.ts
PeterSchafer: Actually, I'm not sure how to avoid the promise considering that the code inside is event driven.

ts-binary-wrapper/src/common.ts
Avishagp: Exactly!

ts-binary-wrapper/src/common.ts
Avishagp: Creating a separate function and calling it with `await`, but not sure if it's worth it...

ts-binary-wrapper/src/common.ts
PeterSchafer: sure I can do this.

PeterSchafer: done!

ts-binary-wrapper/src/common.ts
PeterSchafer: if you don't add try{} the script would exit ungracefully, which I wanted to avoid. Unfortunately it seems required to add the catch. I can add logging, to avoid having the `//` in there.

ts-binary-wrapper/src/common.ts
Avishagp: I still see the filename being decided by `randomInt`, is something wrong on my end?

ts-binary-wrapper/src/common.ts
Avishagp: I think that'd be a good idea

PeterSchafer: done!

ts-binary-wrapper/src/common.ts
PeterSchafer: The formatter removes the parenthesis. So I guess it has to stay the same?!

ts-binary-wrapper/src/common.ts
PeterSchafer: For the documentation reason:
We generate a random file name for a temporary file, that we use while downloading. The file name is generated using randomInt() to be simple and compatible with node 12. The temporary file is after a successful download renamed to the given file name or in case of errors deleted.

@Avishagp and I synced on this and decided to resolve this comment.

ts-binary-wrapper/src/common.ts
PeterSchafer: This would move the usage of promises just one level down the callgraph. I struggle seeing an actual benefit. But if you feel that this is important, I am ok to just do it and resolve this conversation.

Avishagp (approved):

PR #4274, opened 2023-01-06
fix: run snyk code test with the default org

strassl-snyk:
#### What does this PR do?

Removes the explicitly specified org from the Snyk Code smoke tests. As part of [#incident-2023-01-05-circelci-security-alert](https://snyk.slack.com/archives/C04HNK0F144) the `SMOKE_TESTS_SNYK_TOKEN` was rotated and the new service user no longer has access to the org.


bastiandoetsch (approved):
LGTM, you could also explicitly specify the snyk org, but not necessary, I guess.

novalex (approved):

PR #4318, opened 2023-01-11
chore: ignore .build folder for swift

DoronElir:

### DIFF
--- a/src/lib/find-files.ts
+++ b/src/lib/find-files.ts
@@ -66,8 +66,8 @@ export async function find(
   const found: string[] = [];
   const foundAll: string[] = [];
 
-  // ensure we ignore find against node_modules path.
-  if (path.endsWith('node_modules')) {
+  // ensure we ignore find against node_modules path and .build folder for swift.
+  if (path.endsWith('node_modules') || path.endsWith('.build')) {
     return { files: found, allFilesFound: foundAll };
   }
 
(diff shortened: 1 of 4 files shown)


src/lib/find-files.ts
```diff
@@ -62,14 +66,18 @@ export async function find(
   const found: string[] = [];
   const foundAll: string[] = [];
 
-  // ensure we ignore find against node_modules path.
-  if (path.endsWith('node_modules')) {
+  // ensure we ignore find against node_modules path and .build folder for swift.
+  if (path.endsWith('node_modules') || path.endsWith('.build')) {
```
orsagie, line 70: Maybe what we need to do here is to parse the path and look that the `.build` is not part of something like yyy.xxx.build?

test/tap/find-files.test.ts
```diff
@@ -108,6 +108,19 @@ test('find package.json file in test fixture ignoring node_modules', async (t) =
   t.same(result.sort(), expected.sort(), 'should return expected file');
 });
 
+test('find Package.swift file in test fixture ignoring .build folder', async (t) => {
```
orsagie, line 111: Make a test for the case of a/b/c.build

orsagie (approved):
One comment, but LGTM

Greg-Smulko:
This change broke our workflow! 😡 😉 https://xkcd.com/1172/

We have `paket.dependencies` and `paket.lock` files in a `.build` folder, together with other build-related stuff. As a result of this change, these paket files are ignored by Snyk. 😞

PR #4365, opened 2023-01-23
fix: avoid filesystem concurrency during initialization

PeterSchafer:

#### What does this PR do?
* Fix a race condition when multiple CLIs are executed at the same time. Therefore it introduces the usage of a filelock using https://github.com/gofrs/flock. 
* Refactor code and introduce Init() to have all code that alters the filesystem together and the synchronization is easy to apply.

* Fixed a bug in ClearCache() which didn't actually count deleted contents to end the loop.

* Improved debug logging
which caused failing tests main_integration_tests.go, instead of adapting these not sustainable tests, I cleaned up most of them, since we have either unit tests or jest acceptance tests covering the functionality. Removing this test suite completely requires more cleanup which should better be done in HEAD-60

* cliv2/internal/cliv2/cliv2.go

Run a script like the one below to spawn multiple processes almost at the same time. Doing this on the previous version of the code will show multiple attempts to extract the cliv1 binary due to failed shasum checks. The version on the branch should only extract the cliv1 once and other executions should find a proper binary to use without the need to extract.

> for i in {1..2}
do
    binary-releases/snyk-macos woof -d &
done

### DIFF
--- a/cliv2/internal/cliv2/cliv2_test.go
+++ b/cliv2/internal/cliv2/cliv2_test.go
@@ -4,16 +4,26 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"path"
 	"sort"
 	"testing"
+	"time"
 
 	"github.com/snyk/cli/cliv2/internal/cliv2"
 	"github.com/snyk/cli/cliv2/internal/constants"
 	"github.com/snyk/cli/cliv2/internal/proxy"
+	"github.com/snyk/cli/cliv2/internal/utils"
 
 	"github.com/stretchr/testify/assert"
 )
+
+func getCacheDir() string {
+	cacheDir := path.Join(os.TempDir(), "snyk")
+	os.RemoveAll(cacheDir)
+	os.MkdirAll(cacheDir, 0755)
+	return cacheDir
+}
 
 func Test_PrepareV1EnvironmentVariables_Fill_and_Filter(t *testing.T) {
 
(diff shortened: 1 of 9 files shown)


cliv2/internal/cliv2/cliv2_test.go
```diff
@@ -4,17 +4,27 @@ import (
...
 	"github.com/snyk/cli/cliv2/internal/cliv2"
 	"github.com/snyk/cli/cliv2/internal/constants"
 	"github.com/snyk/cli/cliv2/internal/proxy"
+	"github.com/snyk/cli/cliv2/internal/utils"
 
 	"github.com/stretchr/testify/assert"
 )
 
+func getCacheDir() string {
+	cacheDir := path.Join(os.TempDir(), "snyk")
```
bastiandoetsch, line 22: Shouldn't this be `SNYK_CACHE_DIR` environment variable?

cliv2/internal/cliv2/cliv2_test.go
```diff
@@ -150,21 +160,27 @@ func Test_prepareV1Command(t *testing.T) {
...
 
 	// run once
+	assert.Nil(t, cli.Init())
 	actualReturnCode := cliv2.DeriveExitCode(cli.Execute(getProxyInfoForTest(), []string{"--help"}))
 	assert.Equal(t, expectedReturnCode, actualReturnCode)
 	assert.FileExists(t, cli.GetBinaryLocation())
 	fileInfo1, _ := os.Stat(cli.GetBinaryLocation())
 
+	// sleep shortly to ensure that everything is applied
+	time.Sleep(500 * time.Millisecond)
```
bastiandoetsch, line 180: How about using `assert.Eventually` instead of sleeping?

cliv2/internal/cliv2/cliv2_test.go
```diff
@@ -176,16 +192,61 @@ func Test_executeRunV1(t *testing.T) {
 	os.RemoveAll(cacheDir)
```
bastiandoetsch, line 192: this could be a `deferred` call so that it is executed even if a panic/error occurs

cliv2/internal/cliv2/cliv2_test.go
```diff
@@ -176,16 +192,61 @@ func Test_executeRunV1(t *testing.T) {
...
+	// prove that we now can execute the invalid binary
+	_, binError = exec.Command(cli.GetBinaryLocation(), "--help").Output()
+	assert.Nil(t, binError)
+
+	fileInfo2, _ := os.Stat(cli.GetBinaryLocation())
+
+	assert.NotEqual(t, fileInfo1.ModTime(), fileInfo2.ModTime())
+
+	// cleanup
+	os.RemoveAll(cacheDir)
```
bastiandoetsch, line 234: see above - why not cleaning up with defer?

cliv2/internal/cliv2/cliv2_test.go
```diff
@@ -176,16 +192,61 @@ func Test_executeRunV1(t *testing.T) {
...
+	os.MkdirAll(tmpDir, 0755)
+	os.WriteFile(cli.GetBinaryLocation(), []byte("Writing some strings"), 0755)
+	fileInfo1, _ := os.Stat(cli.GetBinaryLocation())
+
+	// prove that we can't execute the invalid binary
+	_, binError := exec.Command(cli.GetBinaryLocation(), "--help").Output()
+	assert.NotNil(t, binError)
+
+	// sleep shortly to ensure that everything is applied
+	time.Sleep(500 * time.Millisecond)
```
bastiandoetsch, line 216: see above :)

cliv2/internal/cliv2/cliv2_test.go
bastiandoetsch: The cachedir could even clean itself up, if you pass T into the function and then register the cleanup as cleanup function.

cliv2/internal/cliv2/cliv2_test.go
PeterSchafer: done!

cliv2/internal/cliv2/cliv2_test.go
PeterSchafer: done! used defer

cliv2/internal/cliv2/cliv2_test.go
PeterSchafer: Interesting point, to be honest I'm not completely sure about the root cause of why sleep is necessary. I just had the build failing on a circle runner while working locally and on other runners. 
My assumption is that the final comparison of the modification time might not be the strongest check. Looking at time resolution etc, I was assuming that the sleep helps to ensure that modification times would be definitely different.
Looking at the comment it seems misleading. I'm definitely changing the comment.

cliv2/internal/cliv2/cliv2_test.go
PeterSchafer: see above ;)

cliv2/internal/cliv2/cliv2_test.go
PeterSchafer: done!

PR #4367, opened 2023-01-24
fix: check if cachedir is empty

PeterSchafer:

#### What does this PR do?
* Check if the specified cache directory is an empty string before trying to create it.
* Moved some logging to main for better debugging

### DIFF
--- a/cliv2/internal/cliv2/cliv2.go
+++ b/cliv2/internal/cliv2/cliv2.go
@@ -70,11 +70,13 @@ func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
 func (c *CLI) Init() (err error) {
 	c.DebugLogger.Println("Init start")
 
-	// ensure the specified base cache directory exists, this needs to be done even before acquiring the lock
-	if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
-		err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
-		if err != nil {
-			return err
+	if len(c.CacheDirectory) > 0 {
+		// ensure the specified base cache directory exists, this needs to be done even before acquiring the lock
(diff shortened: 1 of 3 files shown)


Avishagp (approved):

cliv2/internal/cliv2/cliv2.go
```diff
@@ -70,11 +70,13 @@ func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
...
 
-	// ensure the specified base cache directory exists, this needs to be done even before acquiring the lock
-	if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
-		err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
-		if err != nil {
-			return err
+	if len(c.CacheDirectory) > 0 {
+		// ensure the specified base cache directory exists, this needs to be done even before acquiring the lock
+		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
+			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
```
bastiandoetsch, line 76: maybe use `mkdirall`?

PeterSchafer: I would, but we are having a bit of a hen-egg situation. We need the cache directory to create the lock file in, but at the same time we want protect the filesystem operations with the lock file. Originally I didn't even want to do `mkdir` here, but acceptance tests showed that this was necessary to remain backwards compatible. Sure it could be that multiple folders would need to be created, but again this would cause issues since `mkdirall` is not an atomic operation.

PeterSchafer: Solution could be to create the lock file outside of the cache directory, but this would increase our foot print in the file system and could potentially cause issues during deployment.

PR #4373, opened 2023-01-26
chore: improve cache directory

Avishagp:
We can’t use MkdirAll since it’ll result in concurrency issues. Log information if user is trying to use a cache directory path with sub-directories. Also update go-application-framework to use version that returns “snyk” instead of an empty string in case of an error creating a cache directory

[JIRA ticket](https://snyksec.atlassian.net/browse/HEAD-84)

### DIFF
--- a/cliv2/internal/cliv2/cliv2.go
+++ b/cliv2/internal/cliv2/cliv2.go
@@ -15,7 +15,6 @@ import (
 	"strings"
 
 	"github.com/gofrs/flock"
-	"github.com/pkg/errors"
 	"github.com/snyk/cli/cliv2/internal/constants"
 	"github.com/snyk/cli/cliv2/internal/embedded"
 	"github.com/snyk/cli/cliv2/internal/embedded/cliv1"
@@ -76,11 +75,7 @@ func (c *CLI) Init() (err error) {
 		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
 			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
 			if err != nil {
-				// If the path given contains sub directories, ask user to create them (as using MkdirAll will result in concurrency issues)
-				if strings.Contains(c.CacheDirectory, "/") {
-					errors.Wrap("Cache directory path contains sub-directories. Please create all sub-directories in the path")
-				}
-				return err
+				return fmt.Errorf("Cache directory path is invalid: %w", err)
 			}
 		}
 	}
--- a/cliv2/internal/cliv2/cliv2.go
+++ b/cliv2/internal/cliv2/cliv2.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/gofrs/flock"
+	"github.com/pkg/errors"
 	"github.com/snyk/cli/cliv2/internal/constants"
 	"github.com/snyk/cli/cliv2/internal/embedded"
 	"github.com/snyk/cli/cliv2/internal/embedded/cliv1"
@@ -75,6 +76,10 @@ func (c *CLI) Init() (err error) {
 		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
 			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
 			if err != nil {
+				// If the path given contains sub directories, ask user to create them (as using MkdirAll will result in concurrency issues)
+				if strings.Contains(c.CacheDirectory, "/") {
+					errors.Wrap("Cache directory path contains sub-directories. Please create all sub-directories in the path")
+				}
 				return err
 			}
 		}
--- a/cliv2/go.mod
+++ b/cliv2/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/pkg/errors v0.9.1
 	github.com/snyk/cli-extension-sbom v0.0.0-20221212093410-6b474ed1a42a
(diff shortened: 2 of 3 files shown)


cliv2/internal/cliv2/cliv2.go
```diff
@@ -75,6 +75,10 @@ func (c *CLI) Init() (err error) {
 		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
 			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
 			if err != nil {
+				// If the path given contains sub directories, ask user to create them (as using MkdirAll will result in concurrency issues)
+				if strings.Contains(c.CacheDirectory, "/") {
+					fmt.Println("Cache directory path contains sub-directories. Please create all sub-directories in the path")
```
PeterSchafer, line 80: Please don't just use fmt and print to the console. Please use errors.Wrap(). Printing to the appropriate output will be handled in a view.

cliv2/internal/cliv2/cliv2.go
```diff
@@ -75,6 +76,10 @@ func (c *CLI) Init() (err error) {
 		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
 			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
 			if err != nil {
+				// If the path given contains sub directories, ask user to create them (as using MkdirAll will result in concurrency issues)
+				if strings.Contains(c.CacheDirectory, "/") {
+					errors.Wrap("Cache directory path contains sub-directories. Please create all sub-directories in the path")
```
bastiandoetsch, line 81: Hm. I wonder how we can automatically solve this. The user shouldn't be bothered about the cache-directory, as they are normally not creating it at all.

bastiandoetsch: Why can't we use mkdirall in conjunction with a file lock (flock library)?

PeterSchafer: @bastiandoetsch we can use flock but we would need to place the lock file outside of the cache directory, which we are going to create. We can do this, just that we need to consider this in the docs about filesystem access etc.

PeterSchafer: And yes, I agree, the user shouldn't be bothered at all. Per default the CLI determines the cache directory based on OS defaults, these directories exists and we just place a subfolder inside of it, which we can create relatively safe with mkdir. so per default, the assumption is that this error would never happen.

It might happen, if a user specifies a custom path.

cliv2/internal/cliv2/cliv2.go
```diff
@@ -75,6 +76,10 @@ func (c *CLI) Init() (err error) {
 		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
 			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
 			if err != nil {
+				// If the path given contains sub directories, ask user to create them (as using MkdirAll will result in concurrency issues)
+				if strings.Contains(c.CacheDirectory, "/") {
```
PeterSchafer, line 80: BTW, not all OS use "/" as a separator, better use something like https://pkg.go.dev/path/filepath#SplitList

cliv2/internal/cliv2/cliv2.go
bastiandoetsch: Maybe a lockfile in the temp dir?

cliv2/internal/cliv2/cliv2.go
Avishagp: Are you sure that'd work? using the example here it looks like a string like `folder1/folder2/folder3` does not get separated. Am I doing something wrong?

cliv2/internal/cliv2/cliv2.go
PeterSchafer: Definitely Possible!

cliv2/internal/cliv2/cliv2.go
PeterSchafer: You are right, I miss understood the docs. This function splits for example what is in the PATH environment variable.

PeterSchafer: This looks more promising https://pkg.go.dev/path/filepath#ToSlash

Avishagp: I think the separators here are also env variables separators as it's not working for `\`

cliv2/internal/cliv2/cliv2.go
Avishagp: After a discussion we decided that this solution should be ok for now as it handles this case, and it's something next to no users will do. We can make this better in the future by changing the location of the lock file like you said @bastiandoetsch @PeterSchafer

cliv2/internal/cliv2/cliv2.go
```diff
@@ -75,7 +75,7 @@ func (c *CLI) Init() (err error) {
 		if _, err = os.Stat(c.CacheDirectory); os.IsNotExist(err) {
 			err = os.Mkdir(c.CacheDirectory, local_utils.CACHEDIR_PERMISSION)
 			if err != nil {
-				return err
+				return fmt.Errorf("Cache directory path is invalid:\n" + err.Error())
```
PeterSchafer, line 78: Please use fmt.Errorf() the following way, see [the documentation](https://pkg.go.dev/fmt@go1.19.5#Errorf) for the reasoning
`fmt.Errorf("Cache directory path is invalid: %w", err)`

PeterSchafer (approved):

PR #4374, opened 2023-01-26
fix: host URL for instances in different regions

PeterSchafer:

#### What does this PR do?
Previously we assumed that the root hostname shouldn't start with "app." but this is actually the consistent endpoint for  cross region usage.

### DIFF
--- a/src/lib/config/api-url.ts
+++ b/src/lib/config/api-url.ts
@@ -180,3 +180,15 @@ export function getHiddenApiUrl(restUrl: string): string {
 
   return parsedBaseUrl.toString();
 }
+
+export function getRootUrl(apiUrlString: string): string {
+  // based on https://docs.snyk.io/snyk-processes/data-residency-at-snyk#what-regions-are-available the pattern is as follows
+  // https://app.[region.]snyk.io
+  // given an api url that starts with api means, that we can replace "api" by "app".
+
+  const apiUrl = new URL(apiUrlString);
+  apiUrl.host = apiUrl.host.replace(/^api\./, '');
+
+  const rootUrl = apiUrl.protocol + '//' + apiUrl.host;
+  return rootUrl;
+}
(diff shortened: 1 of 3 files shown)


src/lib/config/api-url.ts
```diff
@@ -180,3 +180,15 @@ export function getHiddenApiUrl(restUrl: string): string {
...
   return parsedBaseUrl.toString();
 }
+
+export function getRootUrl(apiUrlString: string): string {
+  // based on https://docs.snyk.io/snyk-processes/data-residency-at-snyk#what-regions-are-available the pattern is as follows
+  // https://app.[region.]snyk.io
+  // given an api url that starts with api means, that we can replace "api" by "app".
+
+  const apiUrl = new URL(apiUrlString);
+  apiUrl.host = apiUrl.host.replace(/^api\./, 'app.');
```
bastiandoetsch, line 190: what about `https://snyk.io/api`? That should become https://app.xx.snyk.io/api, right?

bastiandoetsch: I guess, what I want to say is that you don't always have an `api` subdomain ;).

bastiandoetsch (approved):

src/lib/config/api-url.ts
PeterSchafer: Absolutely correct, we might not have "api" sub-domains which is fine, they remain unaltered. As far as I understood, `https://snyk.io/api` is the legacy, still supported url.

aron (approved):

src/lib/config/api-url.ts
PeterSchafer: After discussing with @bastiandoetsch offline, we decided to reduce the impact of the change a bit, by replacing "api." by an empty string only, not replacing by "app.". Which is basically what was there before just that we keep Urls that have "app." already unaltered.

bastiandoetsch (approved):

PR #4376, opened 2023-01-31
chore: add additional debug info

PeterSchafer:

#### What does this PR do?
Improves logging format and contents of the Extensible CLI.

Now printing a common debug header like this

> 2023/01/31 12:06:53.176236 main - Version:        1.1092.0
> 2023/01/31 12:06:53.176256 main - Platform:       darwin arm64
> 2023/01/31 12:06:53.176260 main - API:            https://app.snyk.io/api
> 2023/01/31 12:06:53.176264 main - Cache:          /Users/username/Library/Caches/snyk
> 2023/01/31 12:06:53.176266 main - Token-Hash:     8f6a2dc9a14238c5bc319c6919e7ea89
> 2023/01/31 12:06:53.176267 main - Organization:   aaaaaaa-bbbb-cccc-dddd-eeeeeeeeeee
> 2023/01/31 12:06:53.176269 main - Insecure HTTPS: false

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -1,11 +1,14 @@
 package main
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 
 	"github.com/snyk/cli-extension-sbom/pkg/sbom"
@@ -197,6 +200,32 @@ func displayError(err error) {
 	}
 }
 
+func writeLogHeader(config configuration.Configuration) {
+	tokenShaSum := []byte{}
+	if token := config.GetString(configuration.AUTHENTICATION_TOKEN); len(token) > 0 {
+		temp := sha256.Sum256([]byte(token))
+		tokenShaSum = temp[0:16] // using a partial shasum to avoid sharing a token when sharing debug logs
+	}
+
+	org := config.GetString(configuration.ORGANIZATION)
+	insecureHTTPS := "false"
+	if config.GetBool(configuration.INSECURE_HTTPS) {
+		insecureHTTPS = "true"
+	}
+
+	tablePrint := func(name string, value string) {
+		debugLogger.Printf("%-15s %s", name+":", value)
+	}
+
+	tablePrint("Version", cliv2.GetFullVersion())
+	tablePrint("Platform", runtime.GOOS+" "+runtime.GOARCH)
+	tablePrint("API", config.GetString(configuration.API_URL))
+	tablePrint("Cache", config.GetString(configuration.CACHE_PATH))
+	tablePrint("Token-Hash", hex.EncodeToString(tokenShaSum))
+	tablePrint("Organization", org)
+	tablePrint("Insecure HTTPS", insecureHTTPS)
+}
+
 func MainWithErrorCode() int {
 	var err error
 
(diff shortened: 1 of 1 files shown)


cliv2/cmd/cliv2/main.go
```diff
@@ -197,6 +200,32 @@ func displayError(err error) {
...
+
+	tablePrint := func(name string, value string) {
+		debugLogger.Printf("%-15s %s", name+":", value)
+	}
+
+	tablePrint("Version", cliv2.GetFullVersion())
+	tablePrint("Platform", runtime.GOOS+" "+runtime.GOARCH)
+	tablePrint("API", config.GetString(configuration.API_URL))
+	tablePrint("Cache", config.GetString(configuration.CACHE_PATH))
+	tablePrint("Token-Hash", hex.EncodeToString(tokenShaSum))
```
michelkaporin, line 224: What is the purpose of sharing token hash?

cliv2/cmd/cliv2/main.go
```diff
@@ -197,6 +200,32 @@ func displayError(err error) {
...
+		tokenShaSum = temp[0:16] // using a partial shasum to avoid sharing a token when sharing debug logs
+	}
+
+	org := config.GetString(configuration.ORGANIZATION)
+	insecureHTTPS := "false"
+	if config.GetBool(configuration.INSECURE_HTTPS) {
+		insecureHTTPS = "true"
+	}
+
+	tablePrint := func(name string, value string) {
```
michelkaporin, line 216: Does this belong to `main.go`, or rather to some logger abstraction we have?

cliv2/cmd/cliv2/main.go
```diff
@@ -197,6 +200,32 @@ func displayError(err error) {
 	}
 }
 
+func writeLogHeader(config configuration.Configuration) {
+	tokenShaSum := []byte{}
+	if token := config.GetString(configuration.AUTHENTICATION_TOKEN); len(token) > 0 {
+		temp := sha256.Sum256([]byte(token))
```
michelkaporin, line 206: If we really intend to share token hash, crc32 or md5 functions are faster than computing sha256.

bastiandoetsch: yes, but they are not secure hash functions. speed of sha256 is more than fast enough, and it's not compromised. MD5, CRC32 are not remotely secure and not guaranteed to be unidirectional anymore (or CRC has never been).

cliv2/cmd/cliv2/main.go
bastiandoetsch: Looks to me it could also be inlined below into the tablePrint command for `insecure`

cliv2/cmd/cliv2/main.go
PeterSchafer: The token hash helps us compare the used credentials between different runs. In general this PR tries to capture the configuration of the application and token is just one of them.

cliv2/cmd/cliv2/main.go
PeterSchafer: We can optimize and just do it when `--debug` is used

cliv2/cmd/cliv2/main.go
PeterSchafer: Maybe it needs to go somewhere else at some point, currently we don't have a logger abstraction. 
The helper function was just meant to keep the code clean and I wanted to keep the PR as KISS as possible. I can also move it out of writeLogHeader() if needed.

PeterSchafer:
@michelkaporin @bastiandoetsch do you think there is something for me to change here?

cliv2/cmd/cliv2/main.go
bastiandoetsch: Maybe a random salt should be used on top to prevent attacks via rainbow tables. But yeah, just guard it with `--debug` and add a random UUID as salt.

bastiandoetsch: As it's not meant to change - maybe the random salt per instantiation is not possible.

cliv2/cmd/cliv2/main.go
michelkaporin: Makes perfect sense.

PR #4377, opened 2023-01-31
chore: check if the platform is supported

PeterSchafer:

#### What does this PR do?
Introduces a check and warning if the platform the CLI runs on is actually not supported.

#### Screenshots
Example of the message. The shown platform is supported.
image

### DIFF
--- a/src/lib/common.ts
+++ b/src/lib/common.ts
@@ -1,3 +1,9 @@
+import * as os from 'os';
+import * as alerts from './alerts';
+import * as Sentry from '@sentry/node';
+import * as version from './version';
+import * as analytics from './analytics/index';
+
 export async function sleep(ms: number): Promise<void> {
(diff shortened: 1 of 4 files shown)


src/lib/common.ts
```diff
@@ -6,3 +9,36 @@ export const reTryMessage =
...
+    'win32 arm64',
+  ];
+
+  const currentPlatform = os.platform() + ' ' + os.arch();
+  if (!supportedPlatforms.includes(currentPlatform)) {
+    const platformWarning =
+      '------------------------------- Warning -------------------------------\n' +
+      ' The current platform (' +
+      currentPlatform +
+      ') is not supported by Snyk.\n' +
```
bastiandoetsch, line 32: You may want to consider if using docker is possible for using Snyk.

bastiandoetsch (approved):

PR #4383, opened 2023-02-03
feat: Release Extensible CLI via npm

PeterSchafer:

#### What does this PR do?
Publishing a thin TS Wrapper for the Golang based Extensible CLI to NPM, as a replacement for the previously published TS CLI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,13 +20,16 @@ help:
 	@echo 'Use `npm run` for CLIv1 scripts.'
 
 $(BINARY_RELEASES_FOLDER_TS_CLI):
-	@mkdir $(BINARY_RELEASES_FOLDER_TS_CLI)
+	@mkdir -p $(BINARY_RELEASES_FOLDER_TS_CLI)
 
 $(BINARY_RELEASES_FOLDER_TS_CLI)/version: | $(BINARY_RELEASES_FOLDER_TS_CLI)
 	./release-scripts/next-version.sh > $(BINARY_RELEASES_FOLDER_TS_CLI)/version
 
 ifneq ($(BINARY_OUTPUT_FOLDER), $(BINARY_RELEASES_FOLDER_TS_CLI))
-$(BINARY_OUTPUT_FOLDER)/version: $(BINARY_RELEASES_FOLDER_TS_CLI)/version
+$(BINARY_OUTPUT_FOLDER):
+	@mkdir -p $(BINARY_OUTPUT_FOLDER)
+
+$(BINARY_OUTPUT_FOLDER)/version: $(BINARY_OUTPUT_FOLDER) $(BINARY_RELEASES_FOLDER_TS_CLI)/version
 	@cp $(BINARY_RELEASES_FOLDER_TS_CLI)/version $(BINARY_OUTPUT_FOLDER)/version
 endif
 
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "snyk",
+  "name": "snyk-ts-cli",
```
PeterSchafer, line 2: The only thing I'm not completely sure is, whether we should really rename the TS CLI project. There is a sanity check showing that two projects have the same name. Not sure whether this might become an issue.

ts-binary-wrapper/package.json
```diff
@@ -14,7 +18,7 @@
     "node": ">=16"
```
michelkaporin, line 18: Did we intentionally bump the engine version suggestion?

ts-binary-wrapper/package.json
```diff
@@ -1,8 +1,12 @@
 {
-  "name": "ts-binary-wrapper",
-  "version": "1.0.0",
-  "description": "Wrapper for Snyk's Golang based Extensible CLI.",
+  "name": "snyk",
+  "version": "1.0.0-monorepo",
+  "description": "snyk library and cli utility",
   "main": "dist/index.js",
+  "files": [
```
michelkaporin, line 6: Some of the files are missing if we diff to top-level package.json. E.g. LICENSE isn't present. Can it lead to any problems?

ts-binary-wrapper/package.json
PeterSchafer: Good catch! No, it automatically happened when I initiated the project. I will set it back to what is being used currently.

ts-binary-wrapper/package.json
PeterSchafer: To be honest I don't know. I attempted to use only files that I can see being used, to not just copy all. For License it might be okay to add it but there is a explicit license entry in the package.log file. I can go through the files once again.

Makefile
```diff
@@ -26,7 +26,10 @@ $(BINARY_RELEASES_FOLDER_TS_CLI)/version: | $(BINARY_RELEASES_FOLDER_TS_CLI)
 	./release-scripts/next-version.sh > $(BINARY_RELEASES_FOLDER_TS_CLI)/version
 
 ifneq ($(BINARY_OUTPUT_FOLDER), $(BINARY_RELEASES_FOLDER_TS_CLI))
-$(BINARY_OUTPUT_FOLDER)/version: $(BINARY_RELEASES_FOLDER_TS_CLI)/version
+$(BINARY_OUTPUT_FOLDER):
+	@mkdir $(BINARY_OUTPUT_FOLDER)
```
bastiandoetsch, line 30: Not a blocker, but I'd always use `mkdir -p`

bastiandoetsch (approved):

michelkaporin (approved):

Makefile
PeterSchafer: Done!

ts-binary-wrapper/src/bootstrap.ts
```diff
@@ -12,6 +13,16 @@ if (process.argv.includes('exec')) {
     .downloadExecutable(downloadUrl, executable, filenameShasum)
     .then(process.exit)
     .catch((err) => {
+      const sentryError = new Error(
+        'Error downloading binary ' + downloadUrl + ' (' + err + ')',
+      );
+      Sentry.init({
```
michelkaporin, line 19: It'd be good to make DSN injectable during the build pipeline to prevent the misbehaviour. Although if it's not possible currently, it's not a big issue.

michelkaporin (approved):

ts-binary-wrapper/src/bootstrap.ts
```diff
@@ -1,18 +1,42 @@
...
+  Sentry.init({
+    dsn:
+      'https://3e845233db8c4f43b4c4b9245f1d7bd6@o30291.ingest.sentry.io/4504599528079360',
+    release: version,
+  });
 
-if (process.argv.includes('exec')) {
-  const filenameShasum = config.getShasumFile();
-  const downloadUrl = config.getDownloadLocation();
+  const sentryError = new Error('Error downloading binary (' + err + ')');
```
michelkaporin, line 13: We probably want to keep stack trace of the original error, thus I'd suggest reattaching it as:
`sentryError.stack = err.stack`.

https://stackoverflow.com/questions/42754270/re-throwing-exception-in-nodejs-and-not-losing-stack-trace

bastiandoetsch: Nice catch

ts-binary-wrapper/test/acceptance/basic.spec.ts
```diff
@@ -10,10 +10,10 @@ describe('Basic acceptance test', () => {
   it('Bootstrap binary & execute a command', async () => {
     const cliVersionForTesting = '1.1080.0';
     const envInfo = await prepareEnvironment(cliVersionForTesting);
-    const executable = bootstrap.executable.replace(
-      envInfo.inputfolder,
-      envInfo.outputfolder,
-    );
+    const config = common.getCurrentConfiguration();
+    const executable = config
+      .getLocalLocation()
```
bastiandoetsch, line 15: Formatting could be prettier ;)

ts-binary-wrapper/src/bootstrap.ts
bastiandoetsch: Don't think it's necessary, but yeah, one could do that.

bastiandoetsch (approved):
Looks good, added the stack reporting to Sentry and did a minor reformat in basic.spec.ts

PR #4386, opened 2023-02-03
feat: parallel test-dep-graph requests

dantarian:

#### What does this PR do?
Allows up to five requests to the `/test-dep-graph` endpoint to be run at a time, reducing `snyk test` wall-clock run times.

There should be no change to outputs or the running order of operations.

There's only one changed file - `src/lib/snyk-test/run-test.ts`.

On a sample repository with multiple Open Source projects within it (ideally ones for which the `/test-dep-graph` API call takes non-trivial amount of time to complete - nuget projects seem to be good for this), run `snyk test --all-projects` for this version of the CLI and compare with `snyk test --all-projects` for another version. The outputs should be identical, but this version should run significantly quicker - specifically in the phase where the spinner displays "Querying vulnerabilities database...".

We've had support requests (linked below) where customers are complaining about the amount of time taken to scan large monorepos using the `--all-projects` argument while using the Azure Pipelines Snyk task. While there is a separate focus on speeding up the underlying requests (which in the linked example are taking 30-90s each), this approach seemed to be a simple way to reduce the wall-clock duration of the overall task. I've picked 5 for the concurrency level arbitrarily - it can be adjusted as needed.

https://snyk.zendesk.com/agent/tickets/35462

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -71,6 +71,9 @@ import { hasUnknownVersions } from '../dep-graph';
 
 const debug = debugModule('snyk:run-test');
 
+// Controls the number of simultaneous test requests that can be in-flight.
+const MAX_CONCURRENCY = 5;
+
 function prepareResponseForParsing(
   payload: Payload,
   response: TestDependenciesResponse,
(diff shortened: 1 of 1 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -219,14 +222,49 @@ async function sendAndParseResults(
...
-    /** sendTestPayload() deletes the request.body from the payload once completed. */
-    const payloadCopy = Object.assign({}, payload);
-    const res = await sendTestPayload(payload);
+  };
+
+  const threads: Promise<void>[] = [];
+  for (let thread = 0; thread < MAX_CONCURRENCY; thread++) {
+    threads.push(execRequest());
+  }
+  await Promise.all(threads);
```
dotkas, line 264: I am not very familiar with this, but does this handle "bubbling" up errors to the user if some of the requests fail?

dantarian: Yes - in this case, the first error thrown will bubble back up to the user. Any requests that are still in flight will continue to be processed at the Snyk Platform end, but the responses to those will be ignored.

dantarian:
To address a couple of comments that have been raised here and on Slack:

- No additional tests because I haven't really changed how this works, just the order of operations inside a function boundary.
- I've manually implemented the concurrency rather than using p-map because (a) it means no additional dependencies, and (b) when I tried to add p-map as a dependency, a lot of things broke.
- I've chosen 5 as the number of concurrent requests in order to provide a reasonable amount of concurrency without spamming Snyk with requests (for example, if this gets run against a monorepo with a ton of sub-projects). The number is otherwise plucked from thin air.

src/lib/snyk-test/run-test.ts
gitphill: how about when `--json` is used? do we collect all that info and return it as we did when it was all synchronous?

dantarian: It should do - this is essentially just an order of operations change. Where before we were taking each request, executing the request then processing the response, then moving onto the next request, here we perform all of the requests and gather the responses, then process the responses one-by-one (in the same order as before). The actual data and how it is processed are unchanged.

gitphill (approved):

PR #4395, opened 2023-02-09
feat: Release of the Extensible CLI via NPM

PeterSchafer:

#### What does this PR do?
Release the Extensible CLI via NPM

### DIFF
--- a/test/smoke/spec/snyk_code_spec.sh
+++ b/test/smoke/spec/snyk_code_spec.sh
@@ -15,11 +15,6 @@ Describe "Snyk Code test command"
       The output should include "Static code analysis"
       The output should include "✗ [High] SQL Injection"
       The status should be failure
-      if should_have_deprecation_warnings; then
-        The stderr should not equal ""
-      else
-        The stderr should equal ""
-      fi
     End
   End
 
@@ -29,11 +24,6 @@ Describe "Snyk Code test command"
       The status should be failure # issues found
       The output should include '"$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json"'
       The output should include '"name": "SnykCode"'
-      if should_have_deprecation_warnings; then
-        The stderr should not equal ""
-      else
-        The stderr should equal ""
-      fi
       The result of function check_valid_json should be success
     End
   End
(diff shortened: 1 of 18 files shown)


ts-binary-wrapper/src/bootstrap.ts
```diff
@@ -1,18 +1,25 @@
...
-    .then(process.exit)
-    .catch((err) => {
-      console.error(err);
-      process.exit(1);
-    });
+    common
+      .downloadExecutable(downloadUrl, executable, filenameShasum)
+      .then((error) => {
+        if (error !== undefined) {
+          common.logError('Download Error', error).then(() => process.exit(1));
```
bastiandoetsch, line 16: `Download Error' could be a string constant

ts-binary-wrapper/src/index.ts
```diff
@@ -2,9 +2,39 @@
...
+    console.warn("Executable doesn't exist, trying to download.");
+    common
+      .downloadExecutable(
+        config.getDownloadLocation(),
+        executable,
+        config.getShasumFile(),
+      )
+      .then((error) => {
+        if (error !== undefined) {
+          common.logError('Runtime Error', error).then(() => process.exit(2));
```
bastiandoetsch, line 27: Same here, runtime error should be

- `Download Error (Runtime)`
- and a const string, as it's repeated 3 times

bastiandoetsch (approved):
LGTM, only minor clean-code stuff

test/smoke/spec/snyk_code_spec.sh
```diff
@@ -15,11 +15,6 @@ Describe "Snyk Code test command"
       The output should include "Static code analysis"
       The output should include "✗ [High] SQL Injection"
       The status should be failure
-      if should_have_deprecation_warnings; then
```
strassl-snyk, line 18: q: Why do these tests have to go? Just trying to understand the reasoning.

PeterSchafer: The Extensible CLI is a Golang based binary, this means the user node runtime can't and doesn't have to be checked anymore. The binary ships with its own defined runtime.

PeterSchafer: The TS Code we are going to deploy is relatively small and currently there seems to be no need to replicate Node deprecation logic there.

strassl-snyk (approved):

PR #4399, opened 2023-02-10
feat: unmanaged sbom support

danlucian:

#### What does this PR do?

Add support for the `snyk sbom --unmanaged` command.
The above command will enable customers to generate SBOMs for their unmanaged projects.

### DIFF
--- a/src/lib/ecosystems/test.ts
+++ b/src/lib/ecosystems/test.ts
@@ -11,6 +11,10 @@ import { assembleQueryString } from '../snyk-test/common';
 import { getAuthHeader } from '../api-token';
 import { resolveAndTestFacts } from './resolve-test-facts';
 import { isUnmanagedEcosystem } from './common';
+import { convertDepGraph, getUnmanagedDepGraph } from './unmanaged/utils';
+import { jsonStringifyLargeObject } from '../json';
+
+type ScanResultsByPath = { [dir: string]: ScanResult[] };
 
 export async function testEcosystem(
   ecosystem: Ecosystem,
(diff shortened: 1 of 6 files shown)


src/lib/ecosystems/test.ts
```diff
@@ -31,7 +35,7 @@ export async function testEcosystem(
       sarifRes,
     );
   }
-  const scanResultsByPath: { [dir: string]: ScanResult[] } = {};
+  const scanResultsByPath: ScanResultsByPath = {};
```
PeterSchafer, line 38: nitpick: Naming variables the same way as data structures can be confusing. Maybe just rename the variables here?

PeterSchafer (approved):
LGTM but please get somebody with the domain knowledge to take a look as well.

src/lib/ecosystems/test.ts
dekelund: I would argue for "result" as variable name, less stuttering. At least if it's the only "result" in this context. Otherwise just `scanResult`

src/lib/ecosystems/unmanaged/utils.ts
```diff
@@ -71,3 +74,27 @@ export function getSelf() {
...
+
+export async function getUnmanagedDepGraph(scans: {
+  [dir: string]: ScanResult[];
+}) {
+  const results: DepGraphDataOpenAPI[] = [];
+  const orgId = (await getSelf())?.data.attributes.default_org_context;
+
+  for (const [, scanResults] of Object.entries(scans)) {
+    for (const scanResult of scanResults) {
+      const id = await submitHashes(
```
dekelund, line 86: Maybe rename `id` to `taskId` or similar

dekelund: just to differentiate between orgId and taskId

dekelund (approved):

src/lib/ecosystems/test.ts
danlucian: Good catch! Renamed to `results`, thank you both! 🙏

src/lib/ecosystems/unmanaged/utils.ts
danlucian: Makes sense, updated! Thank you!

PR #4412, opened 2023-02-14
fix: ./iac-data/ location

YairZ101:

#### What does this PR do?
Following a [feedback](https://snyk.slack.com/archives/CEPPB1XNE/p1675794455260329) from @lirantal, we've decided to move the location of the `iac-data` folder from the cwd to the Snyk CLI cache directory.

In addition to that, this PR creates a sub-dir in `iac-data` with a random name to allow the IDE to run several scans concurrently without any problem. 

run IaC commands and check that they are still running and working as expected.

### DIFF
--- a/src/cli/commands/test/iac/local-execution/file-utils.ts
+++ b/src/cli/commands/test/iac/local-execution/file-utils.ts
@@ -23,7 +23,7 @@ export function createIacDir(): void {
   const iacPath: fs.PathLike = path.join(LOCAL_POLICY_ENGINE_DIR);
   try {
     if (!fs.existsSync(iacPath)) {
-      fs.mkdirSync(iacPath, '700');
+      fs.mkdirSync(iacPath, { recursive: true, mode: 0o700 });
     }
     fs.accessSync(iacPath, fs.constants.W_OK);
   } catch {
--- a/src/cli/commands/test/iac/local-execution/local-cache.ts
+++ b/src/cli/commands/test/iac/local-execution/local-cache.ts
@@ -9,10 +9,15 @@ import * as analytics from '../../../../../lib/analytics';
 import { getErrorStringCode } from './error-utils';
 import config from '../../../../../lib/config';
 import { streamRequest } from '../../../../../lib/request/request';
+import envPaths from 'env-paths';
 
 const debug = Debug('iac-local-cache');
 
-export const LOCAL_POLICY_ENGINE_DIR = '.iac-data';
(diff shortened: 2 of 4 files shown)


bastiandoetsch (dismissed):

YairZ101:
@ofekatr you have a little bit more context about how custom rules are using the `iac-data` folder(if they even use it, I'm not sure), do you know if this change will affect them in any way?

ofekatr:
>  @ofekatr you have a little bit more context about how custom rules are using the iac-data folder(if they even use it, I'm not sure), do you know if this change will affect them in any way?

@YairZ101 we currently use this directory when pulling custom rule bundles from the customer's registered OCI registry
https://github.com/snyk/cli/blob/8dff90b85ac9da992654f627111929bb84e675a2/src/cli/commands/test/iac/local-execution/rules/oci-pull.ts#L64-L67

YairZ101:
> > @ofekatr you have a little bit more context about how custom rules are using the iac-data folder(if they even use it, I'm not sure), do you know if this change will affect them in any way?
> 
> @YairZ101 we currently use this directory when pulling custom rule bundles from the customer's registered OCI registry
> 
> https://github.com/snyk/cli/blob/8dff90b85ac9da992654f627111929bb84e675a2/src/cli/commands/test/iac/local-execution/rules/oci-pull.ts#L64-L67

so does it mean that the custom rules are getting repulled for each scan? if this is the case then if I understand it correctly this change wouldn't affect them.

ofekatr:
> so does it mean that the custom rules are getting repulled for each scan? if this is the case then if I understand it correctly this change wouldn't affect them.

@YairZ101 This flow downloads the custom rules in every execution and ensures the `iac-data` directory exists before downloading. It assumes the IaC data directory path to be in the `LOCAL_POLICY_ENGINE_DIR` env variable. 
IIUC, this PR attempts to avoid the creation of this directory in the CWD. 
If that's the case, we may need to make additional changes here as well.

ofekatr:
I see now that the exported value for `LOCAL_POLICY_ENGINE_DIR` is updated in this PR and that we use the same value for custom rules. So this should work, and no additional changes are needed
https://github.com/snyk/cli/blob/5168df306f7f7709cde8208020e35ead2a0c5a3b/src/cli/commands/test/iac/local-execution/local-cache.ts#L20

ofekatr (dismissed):

ofekatr (approved):

PR #4422, opened 2023-02-17
feat: add report upload to code test [ZEN-494]

patricia-v:

#### What does this PR do?

- Adds the report upload/ignores feature to the CLI
- Adds tests for:
  - reportResults in Code Test results
  - ignored issues through `suppressions` in sarif

 - [CLI Uploads Feature Plan](https://www.notion.so/snyk/Snyk-Code-Publish-CLI-results-to-the-web-UI-7b48a23e44634c60a5063d2eb8b6c528)
 - [CLI Ignores Feature Plan](https://www.notion.so/snyk/Snyk-Code-CLI-Ignore-support-ae5eb72f6af940e087e801ade11154c9)

 - [ZEN-494](https://snyksec.atlassian.net/browse/ZEN-494)
 - Related [code-client PR](https://github.com/snyk/code-client/pull/170)

#### TODO ⚠️ 

- [x] Bump code-client version to latest once [code-client#170](https://github.com/snyk/code-client/pull/170) has been merged

[ZEN-494]: https://snyksec.atlassian.net/browse/ZEN-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/plugins/sast/analysis.ts
+++ b/src/lib/plugins/sast/analysis.ts
@@ -2,8 +2,6 @@ import {
   analyzeFolders,
   AnalysisSeverity,
   MAX_FILE_SIZE,
-  FileAnalysis,
-  AnalysisResultSarif,
 } from '@snyk/code-client';
 import { ReportingDescriptor, Result } from 'sarif';
 import { SEVERITY } from '../../snyk-test/legacy';
@@ -11,7 +9,12 @@ import { getAuthHeader } from '../../api-token';
 import config from '../../config';
 import { spinner } from '../../spinner';
 import { Options } from '../../types';
-import { SastSettings, Log, CodeTestResults } from './types';
+import {
+  SastSettings,
+  Log,
+  CodeTestResults,
+  CodeAnalysisResults,
+} from './types';
 import { analysisProgressUpdate } from './utils';
 import {
   FeatureNotSupportedBySnykCodeError,
@@ -47,7 +50,7 @@ export async function getCodeTestResults(
 
   return {
     reportResults: codeAnalysis.reportResults,
-    analysisResults: codeAnalysis.analysisResults as AnalysisResultSarif,
+    analysisResults: codeAnalysis.analysisResults,
   };
 }
 
@@ -56,7 +59,7 @@ async function getCodeAnalysis(
   options: Options,
   sastSettings: SastSettings,
   requestId: string,
-): Promise<FileAnalysis | null> {
+) {
   const isLocalCodeEngineEnabled = isLocalCodeEngine(sastSettings);
   if (isLocalCodeEngineEnabled) {
     validateLocalCodeEngineUrl(sastSettings.localCodeEngine.url);
@@ -132,7 +135,7 @@ async function getCodeAnalysis(
     );
   }
 
-  if (!result || result?.analysisResults.type !== 'sarif') {
+  if (!result || result.analysisResults.type !== 'sarif') {
     return null;
   }
 
@@ -147,7 +150,7 @@ async function getCodeAnalysis(
     );
   }
 
-  return result;
+  return result as CodeAnalysisResults;
 }
 
 function filterIgnoredIssues(codeAnalysis: Log): Log {
--- a/src/lib/plugins/sast/analysis.ts
+++ b/src/lib/plugins/sast/analysis.ts
@@ -2,6 +2,8 @@ import {
   analyzeFolders,
   AnalysisSeverity,
   MAX_FILE_SIZE,
+  FileAnalysis,
+  AnalysisResultSarif,
 } from '@snyk/code-client';
 import { ReportingDescriptor, Result } from 'sarif';
 import { SEVERITY } from '../../snyk-test/legacy';
@@ -9,7 +11,7 @@ import { getAuthHeader } from '../../api-token';
 import config from '../../config';
 import { spinner } from '../../spinner';
 import { Options } from '../../types';
-import { SastSettings, Log } from './types';
+import { SastSettings, Log, CodeTestResults } from './types';
 import { analysisProgressUpdate } from './utils';
 import {
   FeatureNotSupportedBySnykCodeError,
@@ -23,12 +25,12 @@ import { getCodeClientProxyUrl } from '../../code-config';
 
 const debug = debugLib('snyk-code');
 
-export async function getCodeAnalysisAndParseResults(
+export async function getCodeTestResults(
   root: string,
   options: Options,
   sastSettings: SastSettings,
   requestId: string,
-): Promise<Log | null> {
+): Promise<CodeTestResults | null> {
   await spinner.clearAll();
   analysisProgressUpdate();
   const codeAnalysis = await getCodeAnalysis(
@@ -38,7 +40,15 @@ export async function getCodeAnalysisAndParseResults(
     requestId,
   );
   spinner.clearAll();
-  return parseSecurityResults(codeAnalysis);
+
+  if (!codeAnalysis) {
+    return null;
+  }
+
+  return {
+    reportResults: codeAnalysis.reportResults,
+    analysisResults: codeAnalysis.analysisResults as AnalysisResultSarif,
+  };
 }
 
 async function getCodeAnalysis(
@@ -46,7 +56,7 @@ async function getCodeAnalysis(
   options: Options,
   sastSettings: SastSettings,
   requestId: string,
-): Promise<Log | null> {
+): Promise<FileAnalysis | null> {
   const isLocalCodeEngineEnabled = isLocalCodeEngine(sastSettings);
   if (isLocalCodeEngineEnabled) {
     validateLocalCodeEngineUrl(sastSettings.localCodeEngine.url);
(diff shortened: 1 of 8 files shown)


src/lib/plugins/sast/format/output-format.ts
```diff
@@ -106,7 +115,10 @@ function getIssues(
         const artifactLocationUri = location.artifactLocation.uri;
         const startLine = location.region.startLine;
         const text = res.message.text;
-        const title = ruleIdSeverityText;
+        let title = ruleIdSeverityText;
+        if (res.fingerprints?.['identity']) {
```
strassl-snyk, line 119: q: I assume this won't show up in regular scans as `identity` is not set?

src/lib/policy/find-and-load-policy.ts
```diff
@@ -11,7 +11,7 @@ const debug = debugModule('snyk');
 
 export async function findAndLoadPolicy(
   root: string,
-  scanType: SupportedPackageManagers | 'docker' | 'iac' | 'cpp',
+  scanType: SupportedPackageManagers | 'docker' | 'iac' | 'cpp' | 'code',
```
strassl-snyk, line 14: q: What does this change do?

src/lib/plugins/sast/analysis.ts
```diff
@@ -176,14 +197,23 @@ function getSecurityResultsOnly(
 ): Result[] {
   const securityResults = results.reduce((acc: Result[], result: Result) => {
     const securityRule = securityRules.find((sr) => sr === result?.ruleId);
-    if (securityRule) {
+    // TODO: move this out of here - it doesn't belong here and it's a hacky way to do it
+    // This was just the easiest place to put it for a POC
+    const ignored = (result.suppressions?.length ?? 0) > 0;
```
michelkaporin, line 202: Is this redundant with the `filterIgnoredSecurityRules` below now?

test/fixtures/sast/sample-analyze-folders-with-report-and-ignores-response.json
```diff
@@ -0,0 +1,1757 @@
+{
+  "connection": { 
+    "baseURL": "http://proxy.acme--snyk-url.io",
+    "sessionToken": "912D0461-5CCD-417B-8073-1305D1D896C2",
```
strassl-snyk, line 4: q: I assume that is sanitized data?

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -380,6 +386,55 @@ describe('Test snyk code', () => {
...
+        path: '',
+        code: true,
+      },
+      sastSettings,
+      'test-id',
+    );
+
+    const sarifWithoutIgnores =
+      resultWithoutIgnores?.analysisResults.sarif.runs[0].results;
+    if (!sarifWithoutIgnores) return;
```
strassl-snyk, line 409: enh: We should really, really be failing the test if this happens. Silently early-exiting tests is a terrible practice and a good way to get silently failing tests. See also below.

src/lib/plugins/sast/analysis.ts
```diff
@@ -38,15 +40,23 @@ export async function getCodeAnalysisAndParseResults(
...
   spinner.clearAll();
-  return parseSecurityResults(codeAnalysis);
+
+  if (!codeAnalysis) {
+    return null;
+  }
+
+  return {
+    reportResults: codeAnalysis.reportResults,
+    analysisResults: codeAnalysis.analysisResults as AnalysisResultSarif,
```
michelkaporin, line 50: Why casting when you can pass the right type from `getCodeAnalysis`?

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -723,6 +779,11 @@ describe('Test snyk code', () => {
         projectName: undefined,
       },
       languages: undefined,
+      reportOptions: {
```
strassl-snyk, line 782: req: This shouldn't be necessary as code-client should be backward compatible.

test/jest/unit/snyk-code/snyk-code-test.spec.ts
michelkaporin, line 409: Returning here will likely cause the test to succeed when mocked result is changed - is this intended?

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -380,6 +386,55 @@ describe('Test snyk code', () => {
...
+        path: '',
+        code: true,
+      },
+      sastSettings,
+      'test-id',
+    );
+
+    const sarifWithIgnores =
+      resultWithIgnores?.analysisResults.sarif.runs[0].results;
+    if (!sarifWithIgnores) return;
```
michelkaporin, line 427: Returning here will likely cause the test to succeed when mocked result is changed - is this intended?

strassl-snyk (changes requested):
q: Same as for the code-client PR - we have made triple-sure that this will not affect existing functionality?

src/lib/plugins/sast/format/output-format.ts
novalex: Correct.

src/lib/policy/find-and-load-policy.ts
patricia-v: Was not needed, removed

Avishagp (approved):

test/fixtures/sast/sample-analyze-folders-with-report-and-ignores-response.json
patricia-v: This file is exactly the same as `test/fixtures/sast/sample-analyze-folders-response.json` that we've had for quite a while, I just added the `suppressions` for the ignores. So I'm assuming yes

src/lib/plugins/sast/analysis.ts
patricia-v: I don't think there was a particular reason in this case, was more of a safeguard

@novalex do you remember if there was a reason for this at the time?

test/jest/unit/snyk-code/snyk-code-test.spec.ts
patricia-v: So you mean we should optionally send the `reportOptions` [in the analysis here](https://github.com/snyk/cli/pull/4422/files#diff-fc7edfb2a4c95c09bd721194b8c3e5f3775226c91c4ca0270e42d02046021140R103) and do something like the below?

```
...(options.report && {reportOptions: {
  enabled: options.report ?? false,
  projectName: options['project-name'],
  targetRef: options['target-reference'],
}}),
```

patricia-v: The changes I proposed above would be exactly that: only when `--report` is present, we send the `reportOptions` to `code-client` 😁

strassl-snyk: Yes, I was agreeing but trying to be explicit about it, so we catch any misunderstanding :D

src/lib/plugins/sast/analysis.ts
novalex: It's because in `FileAnalysis` the analysis results can be of type legacy or Sarif. For our use-cases we only use the Sarif type, but that's what the `code-client` lib returns. But fair point, it can be returned as the Sarif type from `getCodeAnalysis` and avoid casting that way.

src/lib/plugins/sast/analysis.ts
```diff
@@ -176,14 +197,23 @@ function getSecurityResultsOnly(
 ): Result[] {
   const securityResults = results.reduce((acc: Result[], result: Result) => {
     const securityRule = securityRules.find((sr) => sr === result?.ruleId);
-    if (securityRule) {
+    // TODO: move this out of here - it doesn't belong here and it's a hacky way to do it
```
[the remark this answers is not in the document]
strassl-snyk: We could apply the filtering further up in the stack outside `getSecurityResultsOnly` - I only put it there because it was the easiest place to put it, not the best.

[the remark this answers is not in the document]
patricia-v: Sounds promising, I'll give it a go 🤞

src/lib/plugins/sast/analysis.ts
```diff
@@ -110,10 +132,30 @@ async function getCodeAnalysis(
...
 
-  if (result?.analysisResults.type === 'sarif') {
-    return result.analysisResults.sarif;
+  if (!result || result?.analysisResults.type !== 'sarif') {
+    return null;
   }
-  return null;
+
+  // Filter ignored issues when using report
+  if (options.report) {
```
strassl-snyk, line 140: q: We originally had this after `praseSecurityRules`, right? Does the order matter?

strassl-snyk (approved):

src/lib/plugins/sast/analysis.ts
patricia-v: It shouldn't matter, but to be safe I switched them again

src/lib/plugins/sast/analysis.ts
novalex: Unfortunately TS doesn't seem to figure out the results should be of type Sarif after we do the `result?.analysisResults.type !== 'sarif'` check and early return, so an explicit typecast is still needed.

novalex (dismissed):

src/lib/plugins/sast/analysis.ts
michelkaporin: Have you tried using [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)?

michelkaporin (dismissed):

strassl-snyk (dismissed):

src/lib/plugins/sast/analysis.ts
novalex: Yes but they're not helping readability and since we know the type will be Sarif I opted to just cast from inside the `getCodeAnalysis` function and let TS infer the correct type.

novalex (approved):

michelkaporin (approved):

PR #4423, opened 2023-02-17
Feat/head 57 ext cli to npm 3 [HEAD-57]

j-luong:

#### What does this PR do?
Fixes various bugs from deploying cli to npm

### DIFF
--- a/ts-binary-wrapper/src/bootstrap.ts
+++ b/ts-binary-wrapper/src/bootstrap.ts
@@ -1,18 +1,27 @@
 import * as common from './common';
 import * as process from 'process';
 
-const config = common.getCurrentConfiguration();
-export const executable = config.getLocalLocation();
+const errorContextMessage = 'Download Error';
 
-if (process.argv.includes('exec')) {
-  const filenameShasum = config.getShasumFile();
-  const downloadUrl = config.getDownloadLocation();
+(async () => {
+  try {
+    const config = common.getCurrentConfiguration();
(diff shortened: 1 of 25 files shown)


ts-binary-wrapper/src/index.ts
```diff
@@ -2,9 +2,65 @@
...
+  }
+
+  const exitCode = common.runWrapper(executable, cliArguments);
+  return exitCode;
+}
+
+(async () => {
+  let fallbackToLegacyCLI = argv.includes('--legacy-cli');
+
+  if (fallbackToLegacyCLI == false) {
```
j-luong, line 26: todo: In JS, best practice is to use strict equality `===` for comparisons rather than standard `==` equality. 

`==` will try to perform type coercion so `false == 0` resolves `true` vs `false === 0` resolves `false`

ts-binary-wrapper/src/bootstrap.ts
```diff
@@ -1,18 +1,30 @@
...
+      );
+      if (downloadError !== undefined) {
+        throw downloadError;
+      }
+    }
+  } catch (err) {
+    await common.logError(errorContextMessage, err);
+
+    // exit with error code
+    process.exit(1);
```
PeterSchafer, line 28: chore: bootstrap is not supposed to exit with any other exitcode than 0, in order to be able to use the fallback mechanism. Otherwise the installation would already fail and npm would rollback the package.

PR #4424, opened 2023-02-17
feat: support extensions with nested subcommands [HEAD-86]

bastiandoetsch:

#### What does this PR do?
With this an extension can be registered in a subtree, e.g. `snyk iac capture`. 

`main_test.go`

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -4,7 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


cliv2/cmd/cliv2/main.go
```diff
@@ -126,26 +130,41 @@ func getGlobalFLags() *pflag.FlagSet {
...
+		parentCommand := rootCommand
+		var currentParts string
+		for level := 0; level < len(workflowParts); level++ {
+			subCmdName := workflowParts[level]
+			currentParts += " " + subCmdName
+			subCmd := commandMap[currentParts]
+			if subCmd == nil {
+				subCmd = &cobra.Command{
+					Use:    subCmdName,
+					Hidden: !workflowEntry.IsVisible(),
```
PeterSchafer, line 154: Suggestion: Let's set the default visibility to false and only derive the visibility based on the WorkflowEntry  to the leaf command, like we do with 'runCommand'

bastiandoetsch: Sure

PeterSchafer (approved):

PR #4429, opened 2023-02-21
fix: pass org to code-client

patricia-v:

#### What does this PR do?
Continuation for [code-client#169](https://github.com/snyk/code-client/pull/169):

> fix: fixed http library to properly set API header with organization name
>
>Currently, we don't set a header with organization name for getAnalysis call.
>These changes are all internal and don't affect public API

### DIFF
--- a/src/lib/plugins/sast/analysis.ts
+++ b/src/lib/plugins/sast/analysis.ts
@@ -70,6 +70,8 @@ async function getCodeAnalysis(
     ? sastSettings.localCodeEngine.url
     : getCodeClientProxyUrl();
 
+  const org = sastSettings.org;
+
   // TODO(james) This mirrors the implementation in request.ts and we need to use this for deeproxy calls
   // This ensures we support lowercase http(s)_proxy values as well
   // The weird IF around it ensures we don't create an envvar with
@@ -94,12 +96,14 @@ async function getCodeAnalysis(
   const severity = options.severityThreshold
     ? severityToAnalysisSeverity(options.severityThreshold)
     : AnalysisSeverity.info;
+
   const result = await analyzeFolders({
     connection: {
       baseURL,
       sessionToken,
       source,
       requestId,
+      org,
     },
     analysisOptions: { severity },
     fileOptions: { paths: [root] },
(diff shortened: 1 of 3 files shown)


src/lib/plugins/sast/analysis.ts
```diff
@@ -94,12 +94,16 @@ async function getCodeAnalysis(
   const severity = options.severityThreshold
     ? severityToAnalysisSeverity(options.severityThreshold)
     : AnalysisSeverity.info;
+
+  const org = options.org ?? sastSettings.org;
```
strassl-snyk, line 98: q: Can you explain when `options.org` is set vs when `sastSettings.org` is set?

strassl-snyk (approved):

src/lib/plugins/sast/analysis.ts
patricia-v: `options.org` is when you use `--org=...` along with `code test`

`sastSettings.org` is from where we get the sast setting from the registry API in [getSastSettingsForOrg](https://github.com/snyk/cli/blob/ab5d579216c279e5661e413f2ddc9d4731c168ca/src/lib/plugins/sast/checks.ts#L8)

Although now I've noticed that before the call to `getSastSettingsForOrg` we go through [getSastSettings](https://github.com/snyk/cli/blob/master/src/lib/plugins/sast/settings.ts#L13) which already sets the org from the `options.org` above, so only `sastSettings.org` would suffice here (I've tested this already and I'll amend it)

strassl-snyk: Sooo to make sure I understand our current flow correctly:

1. The user passes org name or org id to `--org`
2. The CLI calls registry to resolve the name/id to get the org name. This is stored in sastSettings
3. The CLI calls deeproxy with the org name from sastSettings
4. deeproxy calls registry with the org name and retrieves the org id
5. deeproxy passes the org id to upstream services.

Is that about right? 😅

strassl-snyk (approved):

Arvi3d (approved):

src/lib/plugins/sast/analysis.ts
patricia-v: 2 is done not only for checking the org, but also to get the code settings (like if code is enabled and LCE settings). If the org doesn't exist and we can't get the sastSettings, we get an error which is properly handled in the cli.

4 is also done with the registry verification call, so it's not like it's that call's only purpose 😅 

But overall yes, that's the gist of it 😁 the org just keeps hopping among our services 🐇

Avishagp (approved):

PR #4431, opened 2023-02-22
feat: add cloud custom rules to iac behind feature flag

jason-snyk:

#### What does this PR do?

This PR adds initial support for Snyk Cloud custom rules to `snyk iac test` behind a feature flag. I also needed to fix a bug that caused non-fatal parse errors to be treated as fatal, because the most recent policy-engine emits those non-fatal errors in more situations than it used to. After this PR, we'll just debug log those errors until we can design a better way to communicate them.

Please reach out to me and I can help you get setup to test.

`CLOUD-1176`

#### Screenshots

A custom rule in the results:
Screenshot 2023-02-22 at 8 31 16 AM

### DIFF
--- a/src/cli/commands/test/iac/v2/index.ts
+++ b/src/cli/commands/test/iac/v2/index.ts
@@ -55,6 +55,8 @@ async function prepareTestConfig(
   const cloudContext = getFlag(options, 'cloud-context');
   const snykCloudEnvironment = getFlag(options, 'snyk-cloud-environment');
   const insecure = options.insecure;
+  const customRules = options['custom-rules'];
+  const experimental = options.experimental;
 
   return {
     paths,
@@ -74,5 +76,7 @@ async function prepareTestConfig(
     snykCloudEnvironment,
     insecure,
     org,
+    customRules,
+    experimental,
   };
 }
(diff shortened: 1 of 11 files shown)


src/cli/commands/test/iac/v2/index.ts
```diff
@@ -54,6 +55,10 @@ async function prepareTestConfig(
   const varFile = options['var-file'];
   const cloudContext = getFlag(options, 'cloud-context');
   const snykCloudEnvironment = getFlag(options, 'snyk-cloud-environment');
+  const customRulesEnabled = await isFeatureFlagSupportedForOrg(
```
ofekatr, line 58: 1. The CLI is meant to serve solely as the UI for the IaC test, and should not be exposed to implementation details such as engine-related features gating. Could you move the move the feature flag check to [`snyk/snyk-iac-test`](https://github.com/snyk/snyk-iac-test) please?
2. Should we also wrap this feature with the `experimental` CLI flag? Doing so provides us with an additional layer of gating to ensure the feature is not activated for pre-feature-release versions when it's not fully complete and ready for use by customers.

PR #4433, opened 2023-02-23
chore: optimize pipeline to use build images and be more resilient and fast

bastiandoetsch:
#### What does this PR do?

This PR cleans up the CLI pipeline and improves its resiliency. 

- Dependencies are pre-installed into a docker build image that is used for most jobs to reduce runtime dependencies.
- The build image is built once a week with a github action and pushed to docker hub (currently, we're still waiting for creds to push to snyk/images)
- Removed jobs without function
- Merged some jobs into the build job at the very beginning
- Tests are run only against one Node version (16), as extensible CLI bundles the TypeScript part with Node 16

- Look at the CircleCI run of this PR to see how the pipeline looks now

Can't be manually tested

We had a lot of failures in the pipeline stemming from runtime installations, especially when sdkman introduced API rate limiting. Moving dependency installation into a docker image decouples runtime from environment setup for the build.


cliv2/scripts/sign_windows.sh
```diff
@@ -12,6 +12,16 @@ APP_PATH_UNSIGNED="$APP_PATH.unsigned"
 SIGNING_SECRETS=secrets.p12
 
 LOG_PREFIX="--- $(basename "$0"):"
+
+sudo apt-get update && sudo apt-get install -y cmake libssl-dev libcurl4-openssl-dev faketime
```
PeterSchafer, line 16: Suggestion: We should build the dependencies when building the docker images and not in the script that shall just sign.

.circleci/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM --platform=$BUILDPLATFORM golang:1.20
+
+# install "normal" stuff
+RUN apt-get update
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
```
PeterSchafer, line 5: Suggestion/Question: Can we derive this from .nvmrc instead of hard-coding it here. It'll make maintenance much easier if we transition towards a single source of truth.

.circleci/config.yml
```diff
@@ -31,91 +23,29 @@ parameters:
...
+  arm64:
     machine:
-      # https://circleci.com/developer/machine/image/ubuntu-2004
-      image: ubuntu-2004:2022.04.1
-  linux-arm64:
-    machine:
-      # https://circleci.com/docs/2.0/arm-resources/
-      image: ubuntu-2004:2022.04.1
-    resource_class: arm.medium
+      image: ubuntu-2004:current
```
PeterSchafer, line 39: Question: Why are we using an ubuntu machine image here? Don't we have an arm64 cli-build?

.circleci/config.yml
```diff
@@ -135,181 +65,93 @@ commands:
...
-            - ~/.sdkman/archives
-      - aws-cli/install:
-          version: << pipeline.parameters.aws_version >>
+            brew install go gradle python elixir composer gradle@6 maven sbt
+
   install_sdks_linux:
     steps:
       - run:
-          name: Installing Python
+          name: Installing sdks and tools
```
PeterSchafer, line 78: Question: In relation to a previous question, shouldn't this be part of the arm64 build image?

.circleci/config.yml
```diff
@@ -31,91 +23,29 @@ parameters:
...
-      - when:
-          condition: << parameters.npm_global_sudo >>
-          steps:
-            run:
-              name: Installing npm
-              command: sudo npm install -g npm@<< parameters.npm_version >>
-      - run:
-          name: Configuring npm
-          command: |
-            npm config set '//registry.npmjs.org/:_authToken' '${NPM_TOKEN}'
```
PeterSchafer, line 105: Issue: I think this needs to be there for the release steps.

.circleci/config.yml
bastiandoetsch: The images that can be used with a machine runner are predefined - so we can't use our image (we could only use docker in docker).

.circleci/Dockerfile
bastiandoetsch: If that's guaranteed to be available, we could do that. But why is .nvmrc the truth?

.circleci/config.yml
bastiandoetsch: not possible on a machine runner (you can only run a docker image from the machine, doing all env and fs mappings manually)

.circleci/Dockerfile
PeterSchafer: I'm happy with any single source of truth. .nvmrc is just a quick shot proposal.

dotkas (commented):
Could you please fill out the PR description with any relevant context that might help anyone who wants to review this?

asaf92 (approved):

PR #4449, opened 2023-03-03
feat: add context flag to Snyk App create command

jgresty:

#### What does this PR do?

Add a new flag `--context` to the Snyk Apps create command.

Run `apps create`, both with and without `--interactive` is supported.

The context flag is just passed to the create app endpoint, at this time we accept 'tenant' (the old option) and 'user'. This is to support Snyk Apps that act on behalf of users rather than utilising a bot user.

### DIFF
--- a/src/lib/apps/create-app/index.ts
+++ b/src/lib/apps/create-app/index.ts
@@ -1,4 +1,5 @@
 import {
+  AppContext,
   AppsErrorMessages,
   createAppPrompts,
   ICreateAppRequest,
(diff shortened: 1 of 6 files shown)


src/lib/apps/create-app/index.ts
```diff
@@ -45,6 +52,7 @@ export function createAppDataScriptable(
         .replace(/\s+/g, '')
         .split(','),
       snykAppScopes: options.scopes.replace(/\s+/g, '').split(','),
+      context: options.context,
```
michelkaporin, line 55: context can be returned as `undefined` at this point - is this a problem?

jgresty: Nope, it is optional in the api

michelkaporin (approved):
LGTM ✅

gablaxian (approved):

PR #4476, opened 2023-03-20
fix: checksum comparison failed [HEAD-202]

j-luong:

#### What does this PR do?
This PR should resolve the issue of failed shasum checks by
- Committing to using the hash duplex stream to produce the computed hash

The [Hash class](https://nodejs.org/docs/latest-v12.x/api/crypto.html#crypto_class_hash) used to compute the shasum can be implemented in two ways:

- As a stream that is both readable and writable, where data is written to produce a computed hash digest on the readable side, or
- Using the `hash.update()` and `hash.digest()` methods to produce the computed hash.

We currently mix both implementation methods by using a piped stream with `hash.digest()`, leading to shasum check failures in certain node versions


bastiandoetsch (approved):

ts-binary-wrapper/src/common.ts
```diff
@@ -245,7 +245,7 @@ export function downloadExecutable(
     const options = new URL(downloadUrl);
     const temp = path.join(__dirname, Date.now().toString());
     const fileStream = fs.createWriteStream(temp);
-    const shasum = createHash('sha256');
+    const shasum = createHash('sha256').setEncoding('hex');
```
michelkaporin, line 248: That's a default afair, so shouldn't need to be specified.

PR #4479, opened 2023-03-21
chore: ensure adding authentication headers to the correct requests

PeterSchafer:

#### What does this PR do?
Legacy CLI and Extensible CLI read configuration from the same configuration sources but may process and use these values differently. For example the canonical API URL that is derived from the user input might look differently.
This PR updates the application framework to ensure that authentication header are used correctly.


j-luong (approved):
👍

PR #4496, opened 2023-03-29
fix: Help behaviour for Extensions

PeterSchafer:

#### What does this PR do?
Fix logic to ensure that in case of usage errors for commands that are implemented as Extension, the help is properly shown.

image

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -36,6 +36,14 @@ var debugLogger = log.New(os.Stderr, "", 0)
 
 const unknownCommandMessage string = "unknown command"
 
+type HandleError int
+
+const (
+	handleErrorFallbackToLegacyCLI HandleError = iota
+	handleErrorShowHelp            HandleError = iota
+	handleErrorUnhandled           HandleError = iota
+)
+
 func getDebugLogger(config configuration.Configuration) *log.Logger {
 	debug := config.GetBool(configuration.DEBUG)
 	if !debug {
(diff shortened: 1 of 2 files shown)


cliv2/cmd/cliv2/main.go
```diff
@@ -116,13 +124,15 @@ func sendAnalytics(analytics analytics.Analytics, debugLogger *log.Logger) {
 	}
 }
 
-func help(cmd *cobra.Command, args []string) {
+func help(cmd *cobra.Command, args []string) error {
 	helpProvided = true
-	defaultCmd(cmd, args) // TODO handle error
+	args = append(os.Args[1:], "--help")
+	return defaultCmd(cmd, args) // TODO handle error
```
michelkaporin, line 130: Is TODO irrelevant now?

cliv2/cmd/cliv2/main.go
```diff
@@ -116,13 +124,15 @@ func sendAnalytics(analytics analytics.Analytics, debugLogger *log.Logger) {
...
 }
 
 func defaultCmd(cmd *cobra.Command, args []string) error {
 	config.Set(configuration.WORKFLOW_USE_STDIO, true)
+	config.Set(configuration.RAW_CMD_ARGS, args)
```
michelkaporin, line 135: What does this do?

cliv2/cmd/cliv2/main.go
```diff
@@ -219,12 +238,14 @@ func doFallback(err error) (fallback bool) {
 		commandError := strings.Contains(errString, unknownCommandMessage)
 
 		// filter for known cobra errors, since cobra errors shall trigger a fallback, but not others.
-		if commandError || flagError {
-			fallback = true
+		if commandError {
+			resultError = handleErrorFallbackToLegacyCLI
+		} else if flagError {
+			resultError = handleErrorShowHelp
```
michelkaporin, line 244: It'd be great to comment why flag error is always a case to show help. 

Something along the lines: "Cobra has it's own help mechanism, however since we have help documentation only in legacy cli, we should fallback to calling it."

cliv2/cmd/cliv2/main.go
PeterSchafer: it add the current raw args to the configuration to be consumed by the legacy CLI.

michelkaporin: Makes sense, thanks! Wonder how it work without it before? 👀

PeterSchafer: RAW_CMD_ARGS is filled by default with os.Args[1:] but here we might want to add additional args, like "--help".

michelkaporin (approved):
🔥

PR #4497, opened 2023-03-29
feat: add `IAC_RULES_CLIENT_URL` to override the URL of the rules client [CCSC-1536]

ipapast:
This is a "hidden" (aka for development) env var to override the url of the rulesClient that is being used later in snyk-iac-test, to create a client for opa-rules-distribution. With this flag, we will be able to test bundles in development, while going through the opa-rules-distribution automatic download process.

To use locally:
``` 
SNYK_IAC_RULES_CLIENT_URL=https://something.com snyk-dev iac test test/fixtures/iac/terraform/sg_open_ssh.tf
```

To use with this branch: https://github.com/snyk/snyk-iac-test/pull/187: build the executable and run:

```
 SNYK_IAC_RULES_CLIENT_URL=https://something.com SNYK_IAC_POLICY_ENGINE_PATH=~/snyk-repos/snyk-iac-test/snyk-iac-test  snyk-dev iac test test/fixtures/iac/terraform/sg_open_ssh.tf

```

### Screenshots
Override with the bucket we already have:
image

Override with a non existing URL:
image

No flag:
image

### DIFF
--- a/src/lib/iac/test/v2/scan/index.ts
+++ b/src/lib/iac/test/v2/scan/index.ts
@@ -22,6 +22,7 @@ export async function scan(
   options: TestConfig,
   policyEnginePath: string,
   rulesBundlePath: string,
+  rulesClientURL: string,
 ): Promise<TestOutput> {
   const {
     tempOutput: outputPath,
@@ -33,8 +34,9 @@ export async function scan(
       options,
       policyEnginePath,
       rulesBundlePath,
-      outputPath,
+      rulesClientURL,
       tempPolicyPath,
+      outputPath,
     );
   } finally {
     await deleteTemporaryFiles(tempDirPath);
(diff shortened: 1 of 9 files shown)


src/lib/iac/test/v2/scan/index.ts
```diff
@@ -165,6 +175,16 @@ function processFlags(
...
 
+  if (options.userRulesClientURL) {
+    flags.push('-rulesClientURL', rulesClientURL);
+  }
+
+  // if (options.userRulesBundlePath && options.userRulesClientURL) {
+  //   debug(
+  //     'SNYK_IAC_RULES_CLIENT_URL can not be used together with IAC_BUNDLE_PATH',
+  //   );
+  // }
```
ipapast, line 186: oops leftover. I think I will just throw an error here as they can't logically be used together

francescomari: Why can't they be used together? I would say that `userRulesBundlePath` should override the behaviour configured by `userRulesClientURL` if both are specified.

ipapast: I think there's no logic priority for this case. It doesn't make sense that someone wants to try both a bundle automatically downloaded but also one that has locally.

francescomari (approved):

Avishagp (approved):

PR #4498, opened 2023-03-29
feat: add debug for snyk request id header

DoronElir:

#### What does this PR do?
Add debug log for `snyk-request-id` on monitor request

This is needed for debugging some support tickets where the `snyk-request-id` is needed

### DIFF
--- a/src/lib/monitor/index.ts
+++ b/src/lib/monitor/index.ts
@@ -268,7 +268,10 @@ async function monitorDepTree(
     if (!userMessage && res.statusCode === 504) {
       throw new ConnectionTimeoutError();
     } else {
-      debug("the header snyk-request-id is" + res.headers["snyk-request-id"] ?? 'undefined');
+      debug(
+        'the header snyk-request-id is' + res.headers['snyk-request-id'] ??
+          'undefined',
+      );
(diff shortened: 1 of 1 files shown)


src/lib/monitor/index.ts
```diff
@@ -268,6 +268,10 @@ async function monitorDepTree(
     if (!userMessage && res.statusCode === 504) {
       throw new ConnectionTimeoutError();
     } else {
+      debug(
+        'the header snyk-request-id is' + res.headers['snyk-request-id'] ??
+          'undefined',
+      );
```
gitphill, line 274: is it not also interesting for connection timeout error (above)?

gitphill: i think `monitorDepGraph` is really where you want this though

DoronElir: Absolutely, thanks for the comment 🤟

gitphill: or within `makeRequest` itself?

DoronElir: not agreeing there, we are not the only one who is using `makeRequest ` and maybe it should be there, but I think it makes this change much bigger

PR #4505, opened 2023-03-31
feat: add retry to sendTestPayload

xzhou-snyk:
Currently for monorepos, the `snyk test` command succeeds only when all projects in the monorepo succeeds in the `test-dep-graph` request. If one project fails, the whole `snyk test` command will fail.

This PR adds retry logic to the `test-dep-graph` request to increase the chance for `snyk test` success.


test/acceptance/fake-server.ts
```diff
@@ -79,6 +84,10 @@ export const fakeServer = (basePath: string, snykToken: string): FakeServer => {
     nextStatusCode = code;
   };
 
+  const setStatusCode = (code: number) => {
```
PeterSchafer, line 87: Question/Suggestion: Currently there is no test that actually tests retry attempts that eventually succeed. How about providing a possibility to express a sequence of response/status codes rather than having a single constant value. Based on this, we could/should add an additional test that can return 500 initially and later 200 ...

src/lib/snyk-test/run-test.ts
```diff
@@ -234,13 +237,35 @@ async function sendAndParseResults(
...
   };
   const requests: (() => Promise<TestResponse>)[] = [];
-  for (const payload of payloads) {
+  for (const originalPayload of payloads) {
     const request = async (): Promise<TestResponse> => {
-      /** sendTestPayload() deletes the request.body from the payload once completed. */
-      const originalPayload = Object.assign({}, payload);
-      const response = await sendTestPayload(payload);
-      return { payload, originalPayload, response };
+      let remainingAttempts = 3;
```
PeterSchafer, line 242: nitpick: maybe move this constant to a more prominent and central location.

src/lib/snyk-test/run-test.ts
```diff
@@ -234,13 +237,35 @@ async function sendAndParseResults(
...
+          const response = await sendTestPayload(payload);
+          return { payload, originalPayload, response };
+        } catch (err) {
+          error = err;
+          if (
+            err instanceof InternalServerError ||
+            err instanceof BadGatewayError ||
+            err instanceof ServiceUnavailableError
+          ) {
+            await sleep(500);
```
PeterSchafer, line 259: nitpick: maybe move this constant to a more prominent and central location.

src/lib/snyk-test/run-test.ts
danlucian: Hey, @PeterSchafer! Definitely! Moved them to the `common.ts` file, hope that's a better place for these invariants.

src/lib/snyk-test/run-test.ts
danlucian: Hey, @PeterSchafer! Definitely! Moved them to the `common.ts` file, hope that's a better place for these invariants.

test/acceptance/fake-server.ts
danlucian: Agreed, we will add the option to configure an array of status code on the fake server and add a test to check if the retry mechanism works as expected.

PeterSchafer: Looks great! I'm just wondering, do we really need both setStatusCode() and setStatusCodes()?

danlucian: I think yes, there are some tests that require a total failure and it would be convinient to just use`setStatusCode(500)` rather than `setStatusCodes([500, 500, 500])` or `setStatusCodes(Array(RETRY_ATTEMPTS).fill(500))`, what do you say?

PeterSchafer (approved):

PeterSchafer (changes requested):
Please take a look at the test it fails in CI.

PeterSchafer (approved):

danlucian (approved):

PR #4509, opened 2023-04-02
fix: custom rules json output missing `resolve` field

YairZ101:

#### What does this PR do?
Users and other Snyk tools(`snyk-to-html` for example) expect to have a `resolve` field with remediation advice for each IaC issue.

Currently, IaC issues that got generated by using custom rules will not have this field, this PR fix this issue.

https://snyk.slack.com/archives/C034QFM0DH8/p1680119186872949

#### Screenshots
Before
image

After
image

### DIFF
--- a/src/cli/commands/test/iac/local-execution/file-scanner.ts
+++ b/src/cli/commands/test/iac/local-execution/file-scanner.ts
@@ -30,6 +30,12 @@ export async function scanFiles(
       const { validatedResult, invalidIssues } = validateResultFromCustomRules(
         result,
       );
+      validatedResult.violatedPolicies.forEach((policy) => {
+        // custom rules will have a remediation field that is a string, so we need to map it to the resolve field.
+        if (typeof policy.remediation === 'string') {
+          policy.resolve = policy.remediation;
+        }
+      });
       scannedFiles.push(validatedResult);
       failedScans = [...failedScans, ...invalidIssues];
     } else {
--- a/src/cli/commands/test/iac/local-execution/types.ts
+++ b/src/cli/commands/test/iac/local-execution/types.ts
@@ -169,6 +169,7 @@ export interface PolicyMetadata {
   resolve: string;
   references: string[];
   // Included only in new policies
+  // Only custom rules will have a string remediation field
   remediation?: Partial<
     Record<'terraform' | 'cloudformation' | 'arm' | 'kubernetes', string>
   >;


src/lib/snyk-test/iac-test-result.ts
```diff
@@ -110,6 +110,10 @@ export interface MappedAnnotatedIacIssue
 export function mapIacIssue(
   iacIssue: AnnotatedIacIssue,
 ): MappedAnnotatedIacIssue {
+  // custom rules will have a remediation field that is a string, so we need to map it to the resolve field.
+  if (typeof iacIssue.remediation === 'string') {
```
ofekatr, line 114: Could you also update the type for this property to reflect that it could potentially be a string? 
https://github.com/snyk/cli/blob/45eefaa7e3adc8aa60d449080955471fe437e27e/src/lib/snyk-test/iac-test-result.ts#L23-L25

YairZ101: I can definitely make this change but I want to make sure that this type is not getting used in our v2 test flow functions.

ofekatr: It should be achievable with some type unions in the shared output flows. 
Let me know if you'd like to pair and look at it together. 
Worth mentioning that the type already seems to have properties that are not relevant for Integrated IaC, e.g., `resolve`

YairZ101: > Worth mentioning that the type already seems to have properties that are not relevant for Integrated IaC, e.g., resolve

yea you are correct but `remediation` is relevant so I don't want to change its type to an incorrect one if it is being used in the v2 functions.

YairZ101: BTW, off-topic to this PR but if we are not populating the `resolve` field in the integrated CLI json output then we need to find a solution for `snyk-to-html` which expects this field.
Most probably should be done as part of the transition project.

ofekatr: Would you mind writing about it in [the feature's channel](https://snyk.slack.com/archives/C04STDUN7JR) so we can fork the conversation about it to there?

YairZ101: Back to the original topic of this comment :)
I've checked if we are using this type in our v2 functions and it looks like we do.
[`formatSnykIacTestTestData`](https://github.com/snyk/cli/blob/master/src/lib/formatters/iac-output/text/formatters.ts#L89) which is being used in the [`buildTextOutput`](https://github.com/snyk/cli/blob/master/src/lib/iac/test/v2/output.ts#L126) function is returning the type [`IacTestData`](https://github.com/snyk/cli/blob/master/src/lib/formatters/iac-output/text/types.ts#L7) that is using the [`Issue`](https://github.com/snyk/cli/blob/master/src/lib/formatters/iac-output/text/types.ts#L37) type where you can find the `remediation` field.

Do you have any suggestions on how to update the type without adding to the v2 functions false information?
We can potentially duplicate this type and create one which is specific for the v2 flow but I'm not sure we want to do this as part of this fix PR.

YairZ101: Double-checked the v2 json output and it does include the `resolve` field.
image

I think this conversation we have just emphasizes how we need a json output contract.

ofekatr: Have you considered handling issues generated by custom rules in the issue mapping function? 
https://github.com/snyk/cli/blob/45eefaa7e3adc8aa60d449080955471fe437e27e/src/lib/formatters/iac-output/text/formatters.ts#L89-L132
Whenever `remediation` is a string value, you can replace it with a map that maps all Current IaC file types to the string value. 

The benefits would be: 
- This function is only called for Current IaC
- The initial input it uses is the `PolicyMetadata` type, which should only be used by Current IaC
- We already have different mappers for different experiences here, so we're not splitting it any more here.
- We would have a consistent shared flow in the output generation step for both experiences.

ofekatr: > I think this conversation we have just emphasizes how we need a json output contract.

We do have contract tests for Integrated IaC. The test for the JSON output can be found here:
https://github.com/snyk/cli/blob/08c30f1000099ed2be3532f03238e7e332f7924f/test/jest/unit/iac/process-results/fixtures/integrated-json-output.json

YairZ101: > We do have contract tests for Integrated IaC. 

I totally forgot about that :)

> Have you considered handling issues generated by custom rules in the issue mapping function?

No, but that's a great suggestion, I'll look into it!

YairZ101: so the issue with this approach is the fact that we don't call this function if the user provided the `json` flag.

QQ, what would it take to fix the source of the issue which is `snyk-iac-rule`? by fix I mean to change the name of the field from `remediation` to `resolve`, you think it's a change we can do in less than an hour?
(asking you because you have the context and background around this area)

ofekatr: Making changes in `snyk-iac-rules` would be the same as breaking an API contract, as it means all existing bundles stop working. 
It would require all of the customers with existing bundles to make generate their bundles again. 
Fixing the CLI on the other hand would only require them to update their CLI version to have the bug fix. 
We also have public documentation describing the string `remediation` property
https://docs.snyk.io/scan-cloud-deployment/snyk-infrastructure-as-code/custom-rules/getting-started-with-the-sdk/writing-a-rule

ofekatr: We could potentially map `remediation` to `resolve` for custom rule bundles in https://github.com/snyk/cli/blob/2942500ea215722a9e0255f87c0d84c8e3b2e848/src/cli/commands/test/iac/local-execution/file-scanner.ts

YairZ101: > Making changes in snyk-iac-rules would be the same as breaking an API contract, as it means all existing bundles stop working.

why is that? it is just a change to the metadata, new bundles with generate a metadata with `resolve` and old ones with `remediation`, how will break something?

YairZ101: @ofekatr and I sat down and we ended up moving the logic to the [`scanFiles`](https://github.com/snyk/cli/blob/fix/custom-rules-json-output/src/cli/commands/test/iac/local-execution/file-scanner.ts#L34-L38) function as he suggested in the comment above.

ofekatr (approved):
Great work! Thanks for pushing through the adjustments! 🚀

PR #4566, opened 2023-04-23
fix: align container sarif output

minsiyang:

#### What does this PR do?

Align Container sarif output with Open Source sarif output

* enable `licenses` entitlement

* Create a Dockerfile with following content
```bash
FROM eclipse-temurin:19-jre
USER 1000
```

* build the image
```bash
docker build -t local:test .
```

* Test the image locally
```bash
node dist/cli/index.js container test --file=Dockerfile --sarif local:test   
```

https://snyk.zendesk.com/agent/tickets/48269
https://snyk.zendesk.com/agent/tickets/46334

### DIFF
--- a/src/lib/formatters/sarif-output.ts
+++ b/src/lib/formatters/sarif-output.ts
@@ -1,12 +1,15 @@
 import * as sarif from 'sarif';
+import * as upperFirst from 'lodash.upperfirst';
 import { TestResult } from '../snyk-test/legacy';
 import { SEVERITY } from '../snyk-test/legacy';
-const upperFirst = require('lodash.upperfirst');
+import { getResults } from './get-sarif-result';
 
 export function createSarifOutputForContainers(
   testResults: TestResult[],
 ): sarif.Log {
   const sarifRes: sarif.Log = {
+    $schema:
+      'https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json',
     version: '2.1.0',
     runs: [],
   };
@@ -48,7 +51,7 @@ export function getTool(testResult): sarif.Tool {
         return;
       }
       const level = getIssueLevel(vuln.severity);
-      const cve = vuln['identifiers']['CVE'][0];
+      const cve = vuln.identifiers?.CVE?.join();
       pushedIds[vuln.id] = true;
       return {
         id: vuln.id,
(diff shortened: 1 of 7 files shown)


ChristinaDara (approved):

src/lib/formatters/sarif-output.ts
```diff
@@ -1,12 +1,16 @@
 import * as sarif from 'sarif';
 import { TestResult } from '../snyk-test/legacy';
 import { SEVERITY } from '../snyk-test/legacy';
-const upperFirst = require('lodash.upperfirst');
+import { getResults } from './open-source-sarif-output';
```
michelkaporin, line 4: Would it make sense to make `open-source-sarif-output' formatter more general-purpose in that case, instead of importing open source function here? 

I question myself on the difference between the two formatters.

cc @PeterSchafer @Avishagp

src/lib/formatters/sarif-output.ts
```diff
@@ -1,12 +1,16 @@
...
+// eslint-disable-next-line no-restricted-imports
+import { upperFirst } from 'lodash';
```
Jdunsby, line 6: should use `lodash.upperfirst` here. I see the CLI has `lodash` as a dev dependency, but otherwise we're using the per method packages.

src/lib/formatters/sarif-output.ts
```diff
@@ -70,39 +74,15 @@ export function getTool(testResult): sarif.Tool {
           level: level,
         },
         properties: {
-          tags: ['security', ...vuln.identifiers.CWE],
+          tags: [
+            'security',
+            ...(vuln.identifiers?.CWE || []),
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            testResult.packageManager!,
```
michelkaporin, line 80: Is there no case when packageManager isn't available in the tool output? Why asserting as non-null?

orsagie (approved):

src/lib/formatters/sarif-output.ts
minsiyang: It is more of a `copy and paste` job. I tried it locally by building Dockerfile with `scratch` image
```bash
npm init
npm install fuzzball@1.4.0
```
Dockerfile
```bash
FROM scratch

COPY package.json .
COPY package-lock.json .
```
removing the non-null check or not makes no difference. I rather keep it the same as open source for now 🙏🏻

src/lib/formatters/sarif-output.ts
minsiyang: Hey @Jdunsby, I tried that locally, sadly at runtime, it is not happy with `import { upperFirst } from 'lodash.upperfirst';`. The local CLI return to me 
```bash
lodash_upperfirst_1.upperFirst is not a function
```

src/lib/formatters/sarif-output.ts
michelkaporin: Where do you copy and paste it from?

minsiyang: https://github.com/snyk/cli/blob/master/src/lib/formatters/open-source-sarif-output.ts#L80-L84

src/lib/formatters/sarif-output.ts
minsiyang: Right you are 🥳 Thank you!!

src/lib/formatters/sarif-output.ts
michelkaporin: Thanks! Not an expert in Container/Open Source SARIF output, just wanted to flag this 👍

michelkaporin (approved):
LGTM, agree to the points in the offline discussion but nothing blocking from my side.

PR #4569, opened 2023-04-25
feat: include suppressions in sarif output

amirsibat:

#### What does this PR do?

Including suppressions (ignored rules) in SARIF outputs (`--sarif` and `--sarif-file-output`).

`snyk code test --report --project-name=<PROJECT_NAME> --sarif`

[ZEN-577](https://snyksec.atlassian.net/browse/ZEN-577)

#### Additional questions

[ZEN-577]: https://snyksec.atlassian.net/browse/ZEN-577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/plugins/sast/format/output-format.ts
+++ b/src/lib/plugins/sast/format/output-format.ts
@@ -10,16 +10,27 @@ import { CodeTestResults } from '../types';
 
 const debug = Debug('code-output');
 
-export function getCodeDisplayedOutput(
-  testResults: CodeTestResults,
-  meta: string,
-  prefix: string,
-): string {
+function filterIgnoredIssues(analysisResults: Sarif.Result[]): Sarif.Result[] {
+  return analysisResults.filter(
+    (rule) => (rule.suppressions?.length ?? 0) === 0,
+  );
+}
+
+export function getCodeDisplayedOutput(args: {
+  testResults: CodeTestResults;
+  meta: string;
+  prefix: string;
+  shouldFilterIgnored: boolean;
+}): string {
   let issues: { [index: string]: string[] } = {};
 
-  const sarif = testResults.analysisResults.sarif;
+  const sarif = args.testResults.analysisResults.sarif;
   if (sarif.runs[0].results) {
-    const results: Sarif.Result[] = sarif.runs[0].results;
+    // Filter ignored issues (suppressions) from the sarif to display in the cli output
+    // The sarif will remain unchanged and contain the suppressions
+    const results: Sarif.Result[] = args.shouldFilterIgnored
+      ? filterIgnoredIssues(sarif.runs[0].results)
+      : sarif.runs[0].results;
 
     const rulesMap: {
       [ruleId: string]: Sarif.ReportingDescriptor;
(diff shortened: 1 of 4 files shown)


src/lib/plugins/sast/format/output-format.ts
```diff
@@ -10,16 +10,25 @@ import { CodeTestResults } from '../types';
...
+  return analysisResults.filter(
+    (rule) => (rule.suppressions?.length ?? 0) === 0,
+  );
+}
+
 export function getCodeDisplayedOutput(
   testResults: CodeTestResults,
   meta: string,
   prefix: string,
+  shouldFilterIgnored = false,
```
strassl-snyk, line 23: enh: I find multiple unnamed function arguments (particularly strings and booleans) to be quite hard to read, as the call site quickly starts to look like `doTheThing("foo", "bar", false, true, false)`. Creating an argument object could be helpful here.

src/lib/plugins/sast/format/output-format.ts
```diff
@@ -10,16 +10,25 @@ import { CodeTestResults } from '../types';
...
 ): string {
   let issues: { [index: string]: string[] } = {};
 
   const sarif = testResults.analysisResults.sarif;
   if (sarif.runs[0].results) {
-    const results: Sarif.Result[] = sarif.runs[0].results;
+    const results: Sarif.Result[] = shouldFilterIgnored
```
strassl-snyk, line 29: enh: A comment explaining that we only want to filter displayed results (and why) may be useful.

strassl-snyk (commented):
In general LGTM, a few questions/suggestions.

We should make sure [Gareth's question](https://snyk.slack.com/archives/C0370TRPF7F/p1682422640052119) is addressed before releasing this.

src/lib/plugins/sast/index.ts
```diff
@@ -44,7 +45,9 @@ export const codePlugin: EcosystemPlugin = {
         throw new NoSupportedSastFiles();
       }
 
-      const sarifTypedResult = testResults?.analysisResults?.sarif;
+      const sarifTypedResult = options['report']
+        ? cloneDeep(testResults?.analysisResults?.sarif)
```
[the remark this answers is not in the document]
patricia-v: The `cloneDeep` is done so it doesn't have the same reference for `testResults` as `readableResult` [in line 60](https://github.com/snyk/cli/blob/feat/output-sarif-suppressions/src/lib/plugins/sast/index.ts#L60), where the output filtering is done, i.e. so the filtering doesn't affect `sarifTypedResults` as well.

Since we don't need to clone the sarif without the `report` option and we wanted to make sure that the behaviour of the output would not change other than for the `report`, we decided to add a check. AFAIK the clone would not affect other flows though.

src/lib/plugins/sast/format/output-format.ts
patricia-v: Do you mean here we could just send the `options` object as an argument and check for `options['report']`, for example?

strassl-snyk: I was suggesting something like:

```ts
export function getCodeDisplayedOutput(options: {
  testResults: CodeTestResults,
  meta: string,
  prefix: string,
  shouldFilterIgnored: boolean,
})
```

You could then call the function like this:
```ts
getCodeDisplayedOutput({
        testResults,
        meta,
        prefix,
        shouldFilterIgnored: options['report'],
})
```

Which I find quite a bit easier to understand

src/lib/plugins/sast/index.ts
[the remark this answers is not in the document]
strassl-snyk: So is there any reason why we couldn't just always clone? This would make the behavior consistent and make it easier to understand. Also, explaining why it is necessary would make a great comment :P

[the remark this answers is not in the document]
patricia-v: I don't think there's any real reason why we couldn't always clone, other than trying to keep the previous behaviour consistent 😄 I'll also add a comment

src/lib/plugins/sast/format/output-format.ts
patricia-v: Sure, that's a simple enough change 👍

strassl-snyk (approved):

PR #4597, opened 2023-05-11
chore: disable standalone iac for fedramp

andreeaneata:

#### What does this PR do?

Enforces the usage of integrated IaC (disables standalone IaC) for FedRAMP.

-

1. Set the endpoint like: `snyk-dev config set endpoint=http://api.snykgov.io/v1`
2. `snyk-dev iac test (-- report)` and see integrated IaC is used

In the future we will transition all our customers to the integrated IaC and decommission the standalone one. 
So no real disablement here, just enforcing the new flow on which we focus to make compliant.

[CFG-2291](https://snyksec.atlassian.net/browse/CFG-2291)

#### Screenshots
-

#### Additional questions
-

[CFG-2291]: https://snyksec.atlassian.net/browse/CFG-2291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/feature-flags/index.ts
+++ b/src/lib/feature-flags/index.ts
@@ -28,6 +28,13 @@ export async function hasFeatureFlag(
   featureFlag: string,
   options: Options,
 ): Promise<boolean | undefined> {
+  if (
+    featureFlag === 'iacIntegratedExperience' &&
+    config.API.includes('snykgov')
+  ) {
+    return true;
+  }
+
   const { code, error, ok } = await isFeatureFlagSupportedForOrg(
     featureFlag,
     options.org,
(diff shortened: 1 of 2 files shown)


src/lib/disabled-features.ts
```diff
@@ -0,0 +1,6 @@
+import config from './config';
+
+// disable standalone IaC for FedRAMP
+export const isStandaloneIacDisabled = config.API.includes('snykgov')
```
PeterSchafer, line 4: Suggestion: since this is just another type of feature flag, how about moving this into `hasFeatureFlag()`? 

Rational: As a developer, I'm only interested in finding out if a feature is enabled or not, not knowing the implementation behind it. Moving this logic into the existing hasFeatureFlag() method will enable to treat all checks the same. Just run this check before/instead querying the API endpoint.

PeterSchafer:
Also please provide a test for the functionality :)

PeterSchafer (approved):
LGTM!

PR #4598, opened 2023-05-11
fix: remove feature flag used to toggle old and new unmanaged service

dekelund:

#### What does this PR do?

Removes check of unused feature flag to decide which endpoint to use for tests of unmanaged source code.

- Removes support for USE_UNMANAGED_DEPS environment variable
- Removes support for snykNewUnmanagedTest feature flag

0. Use `snyk test --unmanaged` with an installation that has not been authenticated
1. Verify that the user receives a human readable reply, informing the user it's not authenticated
2. Run `snyk auth` to authenticate
3. Test the build with an unmanaged project, with following command `snyk test --unmanaged`
4. Verify the output correspond to the expected output

### DIFF
--- a/src/lib/ecosystems/resolve-test-facts.ts
+++ b/src/lib/ecosystems/resolve-test-facts.ts
@@ -1,3 +1,4 @@
+import { AuthFailedError } from '../errors';
 import { Options, PolicyOptions } from '../types';
 import { spinner } from '../../lib/spinner';
 import {
@@ -23,7 +24,6 @@ import { extractAndApplyPluginAnalytics } from './plugin-analytics';
 import { findAndLoadPolicy } from '../policy';
 import { filterIgnoredIssues } from './policy';
 import { IssueDataUnmanaged, Issue } from '../snyk-test/legacy';
-import { hasFeatureFlag } from '../feature-flags';
 import {
   convertDepGraph,
   convertMapCasing,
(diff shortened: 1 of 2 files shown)


src/lib/request/promise.ts
```diff
@@ -42,8 +42,11 @@ export async function makeRequestRest<T>(payload: any): Promise<T> {
           body: JSON.parse(body as any),
         });
       } else if (res.statusCode >= 401) {
+        const { error } = res.body;
+
         return reject({
           code: res.statusCode,
+          error: error,
```
aron, line 49: I'm not sure how this rejected object is used across the codebase. If it's output directly could this change leak internal error messages externally? @snyk/hammerhead will hopefully be able to provide some guidance.

Otherwise we could just pass a new error to the AuthFailedError constructor.

aron (approved):
Looks good, glad to see the improvement in error messaging for un-authenticated requests too 👍

src/lib/request/promise.ts
Avishagp: Agree with @aron, this change could be risky and I would refrain from making it.

Avishagp (changes requested):
Please change `src/lib/request/promise.ts`

src/lib/request/promise.ts
dekelund: It seems like it's expected from the  rest of the code base, if I don't do it it will end up with another error later. I fully understand that it might be dangerous, due to the language (in combination with the design).

I will look into it tomorrow and see what I can do to eliminate the risk.

src/lib/ecosystems/resolve-test-facts.ts
```diff
@@ -40,16 +40,17 @@ export async function resolveAndTestFacts(
...
+    return await resolveAndTestFactsUnmanagedDeps(scans, options);
+  } catch (error) {
+    const unauthorized = error.code === 401 || error.code === 403;
 
-  const featureFlagEnabled = await hasFeatureFlag(
-    'snykNewUnmanagedTest',
-    options,
-  );
+    if (unauthorized) {
+      throw AuthFailedError(error.error, error.code);
```
aron, line 49: Based on the feedback earlier, I'd suggest just passing in a new error here to satisfy the wrapper. There's not much additional information we'd need for a 401/403 error that we don't already have from the request.

```suggestion
      throw AuthFailedError(new Error("Unauthorized request to unmanaged service"), error.code);
```

dekelund: AuthFailedError requires a string, so I will use a similar approach without the `NewError` call.

Avishagp (approved):

PR #4603, opened 2023-05-18
fix: handle suppressed issues (no fail when only ignored issues are returned)

novalex:

#### What does this PR do?

Fix handling of Code ignored issues (suppressions).

When all found issues are suppressed, the exit code should be 0 (success).

Look at first commit for the relevant business logic changes.

The second (chore) commit is just an optimisation to run stringification only once, but does not affect functionality.

The third (refactor) commit is adding types and moving the ignored issue filtering function, it also does not affect functionality.

1. Run a test with `--report` and `--project-name` flags.  
  `snyk code test --report --project-name="test-project"`
  For any issues found, the CLI should exit with code 1.

2. Go to the project and ignore all found issues. Run the CLI command again.
  This time, as all found issues are ignored, the CLI should exit with code 0.

- [ZEN-626](https://snyksec.atlassian.net/browse/ZEN-626)

#### Screenshots

Before:
Screenshot 2023-05-18 at 13 17 44

After:
Screenshot 2023-05-18 at 13 17 57

#### Additional questions

[ZEN-626]: https://snyksec.atlassian.net/browse/ZEN-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/plugins/sast/format/output-format.ts
+++ b/src/lib/plugins/sast/format/output-format.ts
@@ -10,7 +10,9 @@ import { CodeTestResults } from '../types';
 
 const debug = Debug('code-output');
 
-function filterIgnoredIssues(analysisResults: Sarif.Result[]): Sarif.Result[] {
+export function filterIgnoredIssues(
+  analysisResults: Sarif.Result[],
+): Sarif.Result[] {
   return analysisResults.filter(
     (rule) => (rule.suppressions?.length ?? 0) === 0,
   );
--- a/src/lib/plugins/sast/format/output-format.ts
+++ b/src/lib/plugins/sast/format/output-format.ts
@@ -7,16 +7,9 @@ import { colorTextBySeverity, SEVERITY } from '../../../snyk-test/common';
 import { rightPadWithSpaces } from '../../../right-pad';
 import { Options } from '../../../types';
 import { CodeTestResults } from '../types';
+import { filterIgnoredIssues } from '../utils';
 
 const debug = Debug('code-output');
-
-export function filterIgnoredIssues(
-  analysisResults: Sarif.Result[],
-): Sarif.Result[] {
-  return analysisResults.filter(
-    (rule) => (rule.suppressions?.length ?? 0) === 0,
-  );
-}
 
 export function getCodeDisplayedOutput(args: {
   testResults: CodeTestResults;
--- a/src/lib/plugins/sast/index.ts
+++ b/src/lib/plugins/sast/index.ts
@@ -78,16 +78,26 @@ export const codePlugin: EcosystemPlugin = {
           delete message.markdown;
         });
       }
+
+      const shouldStringifyResult =
+        options['sarif-file-output'] ||
+        options.sarif ||
+        options['json-file-output'] ||
+        options.json;
+      const stringifiedResult = shouldStringifyResult
+        ? jsonStringifyLargeObject(sarifTypedResult)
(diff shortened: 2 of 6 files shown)


src/lib/plugins/sast/index.ts
```diff
@@ -49,8 +50,16 @@ export const codePlugin: EcosystemPlugin = {
       // cloneDeep is used so the sarif is not changed when using the testResults getting the displayed output
       const sarifTypedResult = cloneDeep(testResults?.analysisResults?.sarif);
```
vasser, line 51: nit. This is called typed, but actually has type of `any`.
Wanted to check the shape of sarif results, because constructions like `sarifTypedResult.runs?.[0].results` look a bit too complex for a comprehension 😬

src/lib/plugins/sast/index.ts
```diff
@@ -49,8 +50,16 @@ export const codePlugin: EcosystemPlugin = {
...
 
-      const numOfIssues = sarifTypedResult.runs?.[0].results?.length || 0;
+      // Report flow includes ignored issues (suppressions) in the results.
+      const hasIgnoredIssues = options['report'] ?? false;
+
+      // If suppressions are included in results filter them out to get real issue count
+      const foundIssues = hasIgnoredIssues
+        ? filterIgnoredIssues(sarifTypedResult.runs?.[0].results)
```
vasser, line 58: And we always have 1 run?

vasser (approved):
LGTM 🚀 
Left some non-blocking questions

src/lib/plugins/sast/index.ts
novalex: Good catch, will add the proper type 👌 
As for the shape, it's the basic Sarif format you can see [here](https://github.com/microsoft/sarif-tutorials/blob/main/docs/2-Basics.md#results). Not the most readable for sure 😅

src/lib/plugins/sast/index.ts
novalex: Yep, assuming the analysis was successful, and since we only have 1 tool producing SAST results, the Sarif we return will always have 1 run.

src/lib/plugins/sast/index.ts
novalex: Changed, types added and a new variable for the inner run results to hopefully improve readability.

src/lib/plugins/sast/format/output-format.ts
```diff
@@ -7,15 +7,10 @@ import { colorTextBySeverity, SEVERITY } from '../../../snyk-test/common';
 import { rightPadWithSpaces } from '../../../right-pad';
 import { Options } from '../../../types';
 import { CodeTestResults } from '../types';
+import { filterIgnoredIssues } from '../index';
```
vasser, line 10: Oh, this will unfortunately create a circular dependency index <-> format/output-format

novalex: Good catch, will change 🙏 

This shouldn't cause issues by itself, Node does handle circular imports by providing partial exports and it still works fine for now, but it also makes shooting ourselves in the foot easier (like if `filterIgnoredIssues` would at some point also depend on something from `output-format`).

Wonder why there's no lint rules for this 🤔

PR #4607, opened 2023-05-18
[SUP-1093] `p-map` some network requests

dotkas:
#### What does this PR do?

A while back [we introduced](https://github.com/snyk/cli/pull/4386) some pseudo-parallelization (to the extend Node allows) of promises when sending requests to our backend. The solution sped up things a bit, but the Promise-chunk would wait until the slowest promise had finished before starting a new batch.

`pMap` accounts for this problem. The performance increase will not be that significant for smaller chunks of requests, but for bigger ones the lack of holes in the chain of unbroken ought to speed things up a bit. (Testing the performance gain on 72 projects reduced the time with around 5 minutes - so not extreme, but better than nothing I guess)

This change (like the one before) comes on the back of another [customer support ticket](https://snyk.zendesk.com/agent/tickets/49249), where scanning a large project takes over an hour. 

The bottlenecks are numerous, but one of the biggest, as far as I have been able to identify, reveals itself around the requests to vulndb.


src/lib/plugins/nodejs-plugin/yarn-workspaces-parser.ts
```diff
@@ -93,7 +93,7 @@ export async function processYarnWorkspaces(
         ? pathUtil.dirname(yarnWorkspacesFilesMap[packageJsonFileName].root)
         : pathUtil.dirname(packageJsonFileName);
       const rootYarnLockfileName = pathUtil.join(rootDir, 'yarn.lock');
-      const yarnLock = await getFileContents(root, rootYarnLockfileName);
+      const yarnLock = getFileContents(root, rootYarnLockfileName);
```
dotkas, line 96: Just random, it ain't async.

admons (approved):

danlucian (approved):

bastiandoetsch (commented):
Could you please add an acceptance test for this or point out, where the logic of request parallelization/chaining is (successfully) tested?

dotkas:
> Could you please add an acceptance test for this or point out, where the logic of request parallelization/chaining is (successfully) tested?

Thanks for reviewing!

Virtually all tests in https://github.com/snyk/cli/tree/e164c1dfcdd3c3889d1abcc009253da60b7a2215/test/jest/acceptance/snyk-test cover this, when they get to sending a dependency graph to vulndb.

bastiandoetsch (approved):

PR #4609, opened 2023-05-22
feat: [SUP-1093] improve cli yarn performance for yarn workspaces

JamesPatrickGill:

#### What does this PR do?

We have recently introduced a fix in our yarn parsing logic of Node package files that now creates a depGraph directly, instead of the legacy depTree logic, which required doing a lot of conversions back and forward.

Now we're introducing the use of this parser in the CLI for generating yarn-workspace results, which increases the performance of scanning yarn workspaces tremendously.

## Comparison

Scanning a yarn workspace with 72 projects in it using `snyk test --yarn-workspaces --detection-depth=3`

Previous: 32 minutes
This fix: 43 seconds

An increase of 45x.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,7 @@
         "snyk-gradle-plugin": "3.26.4",
         "snyk-module": "3.1.0",
         "snyk-mvn-plugin": "2.32.3",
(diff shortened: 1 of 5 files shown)


dotkas (approved):

bastiandoetsch (approved):
LGTM

PR #4615, opened 2023-05-23
feat: support --all-projects for sbom command

xzhou-snyk:
This PR uses the latest version of [snyk/cli-extension-sbom](https://github.com/snyk/cli-extension-sbom) which contains support for the `--all-projects` flag for SBOM generation.


mcombuechen (approved):

bastiandoetsch:
Would you please add an acceptance test that tests SBOM with `--all-projects`?

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      const { code, stdout, stderr } = await runSnykCLI(
+        `sbom --org aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee --format cyclonedx1.4+json --debug --all-projects`,
+        {
+          cwd: project.path(),
+          env,
+        },
+      );
+
+      expect(code).toEqual(0);
+      expect(stderr).toMatch('Generating documents for 7 depgraph(s)');
```
PeterSchafer, line 52: Question: I know similar patterns can be found all over the tests, but relying on debug output is always a but tricky. Do you think it is possible to match for content of stdout, which is actually a part of the application interface.

mcombuechen: content of `stdout` is the response of the mocked server endpoint; so at the most it would ensure that we talk to the server, but not that we are sending a certain payload. we could introduce some logic into the endpoint mock that would yield a different response based on what payload is being received. e.g.:

```ts
app.post('/orgs/:orgId/sbom', (req, res) => {
  if (res.body === '...some payload shape...') {
    res.status(200).json({ some_response_shape: true });
  } else {
    res.status(200).json({ some_other_response_shape: true });
  }
})
```

do you think this would be a better approach?

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+import { isCLIV2 } from '../../util/isCLIV2';
+
+jest.setTimeout(1000 * 60);
+
+describe('snyk sbom --all-projects (mocked server only)', () => {
+  let server;
+  let env: Record<string, string>;
+
+  beforeAll((done) => {
+    const port = process.env.PORT || process.env.SNYK_PORT || '12345';
```
bastiandoetsch, line 13: Hm. Should this maybe be randomized, as the server could be waiting for the port to be free? Other tests are using exactly the same port - maybe a convenience function that tries to get free ports and is used by all tests (not in scope here, but in general) would be good. WDYT?

mcombuechen: @bastiandoetsch sounds good, to be honest we copied the `port` part (and other bits) over from other tests. What do you think about adding [`portfinder`](https://www.npmjs.com/package/portfinder) as a dev dependency for this?

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  afterAll((done) => {
+    server.close(() => {
+      done();
+    });
+  });
+
+  test('`sbom mono-repo-project` generates an SBOM for multiple projects', async () => {
+    const project = await createProjectFromWorkspace('mono-repo-project');
```
bastiandoetsch, line 40: maybe something even more test-specific?

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (isCLIV2()) {
```
bastiandoetsch, line 42: What is the expected behaviour, if it's not CLIV2? Should the test fail?

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
bastiandoetsch: I would not specify payload, as that's whitebox testing this functionality. So the fakeserver should have a complete mock response for the sbom of the fixture. Then the sbom extension json output and its functionality could be asserted. Does that make sense?

PeterSchafer: Since acceptance tests are blackbox tests, it would be good to test end to end. When the project is the input to the blackbox the sbom is the output. currently the resulting sbom is empty, which could be true but could also be a bug in the system, somewhere.

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
mcombuechen: we took inspiration from this test: https://github.com/snyk/cli/blob/95449b22e20ddb3a9c3bc54c26ab0b757c3267fd/test/jest/acceptance/extra-certs.spec.ts#L84-L109
if it's not CLIV2, this test should not assert anything, since we are testing a CLIV2 extension. I'm not sure this makes sense, admittedly we are not very experienced with how the acceptance tests work and how all the different test environments are set up. If you could advise and help us with what we should best be doing here, it would be much appreciated!

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
mcombuechen: trying this out here: https://github.com/snyk/cli/pull/4615/commits/ccf6b9d65deaa940128f293355aec77acddf0320

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
PeterSchafer: The test shouldn't fail. It could be skipped or just as it is now. Currently these tests still need to for both legacy CLI and Extensible CLI.

bastiandoetsch: Dunno either, just looked weird ;)

test/acceptance/fake-server.ts
```diff
@@ -515,9 +515,23 @@ export const fakeServer = (basePath: string, snykToken: string): FakeServer => {
     res.status(200).send({});
   });
 
-  app.post(basePath.replace('v1', 'hidden') + '/orgs/:org/sbom', (req, res) => {
-    res.status(200).send({});
-  });
+  app.post(
```
bastiandoetsch, line 518: Looks good to me, as this is what the real API is supposed to do.

test/acceptance/fake-server.ts
```diff
@@ -515,9 +515,23 @@ export const fakeServer = (basePath: string, snykToken: string): FakeServer => {
...
+    basePath.replace('v1', 'hidden') + '/orgs/:org/sbom',
+    express.json(),
+    (req, res) => {
+      let bom: Record<string, unknown> = { bomFormat: 'CycloneDX' };
+
+      if (Array.isArray(req.body.depGraphs) && req.body.subject) {
+        // Return a fixture of an all-projects SBOM.
+        bom = {
```
bastiandoetsch, line 526: Is the returned bom specific to `--all-projects`? 

I'd honestly return a bom with some dep data in it for different depgraphs. Else the mock return is not comparable to a real response.

mcombuechen: this one is specific enough to distinguish it from a bom generated without `--all-projects`. only `--all-projects` will create a "subject", and have an array of depgraphs instead of just a single depgraph. this should be safe enough for now. i'd like to deliver more acceptance tests for `snyk sbom` in the future, to cover more invocations. until then i think this can suffice.

test/jest/acceptance/snyk-sbom/all-projects.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        {
+          cwd: project.path(),
+          env,
+        },
+      );
+
+      expect(code).toEqual(0);
+      expect(stdout).toEqual(
+        expect.stringContaining(
+          '{"bomFormat":"CycloneDX","metadata":{"component":{"name":"mono-repo-project"}}}',
```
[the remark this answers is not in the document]
mcombuechen: I get what you mean; at this point I’m a bit hesitant to bake too much logic into the fake endpoint. The fact that it spits out a response specific to `--all-projects` (the `metadata.component.name` being the current working directory) is already something that only `--all-projects` would give you. As mentioned in a previous comment above, let's use this as a basis for more acceptance tests coming up in future housekeeping PRs.

bastiandoetsch (approved):

bastiandoetsch (approved):

PR #4616, opened 2023-05-23
fix: support —DISABLE_ANALYTICS

PeterSchafer:

#### What does this PR do?
* Add support for —DISABLE_ANALYTICS
* Fix a test


test/jest/acceptance/analytics.spec.ts
```diff
@@ -518,7 +519,9 @@ describe('analytics module', () => {
       env: envWithDisabledAnalytics,
     });
 
-    const { code } = await runSnykCLI(`version`);
+    const { code } = await runSnykCLI(`version`, {
+      env: env,
```
bastiandoetsch, line 523: Why does it need env on this call?

bastiandoetsch (approved):

test/jest/acceptance/analytics.spec.ts
PeterSchafer: because it sets the api endpoint to be the local server, without it, requests go to snyk.io and therefore the test is not doing anything useful.

PR #4632, opened 2023-05-30
test: add acceptance test for `snyk sbom`

mcombuechen:
#### What does this PR do?

This extends the acceptance test suite with a test for `snyk sbom`. This is to increase test covrage and build further on the recently introduced acceptance test for `snyk sbom --all-projects`.


bastiandoetsch (approved):
LOVE IT!

mcombuechen:
@bastiandoetsch @PeterSchafer while creating these integration tests, we realized that some test runners were not set up with all tools required to run `snyk test --all-projects` on some monorepos. This caused inconsistent SBOM results, since some dependencies weren't detected (by default `snyk test` silently skips over problematic manifest files). So we turned this PR into a `fix`, the SBOM Extension and DepGraphWorkflow have been extended to accept and forward the `fail-fast` option; which will fail SBOM generation if there are any issues during the SCA phase.

PeterSchafer (approved):

PR #4633, opened 2023-05-31
fix: use prettier version from package for user docs formatting

chdorner-snyk:
use prettier version from package for user docs formatting because the GH sync docs action ends up using the latest prettier version, not the one pinned in `package.json`, which ends up not formatting changes that then our CircleCI linting finds.

### DIFF
--- a/.github/workflows/sync-cli-help-to-user-docs.yml
+++ b/.github/workflows/sync-cli-help-to-user-docs.yml
@@ -25,7 +25,8 @@ jobs:
           if [[ $(git -C ./cli status --porcelain) ]]; then
             echo "Documentation changes detected"
             cd ./cli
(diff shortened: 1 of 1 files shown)


PeterSchafer (approved):

.github/workflows/sync-cli-help-to-user-docs.yml
```diff
@@ -25,7 +25,8 @@ jobs:
           if [[ $(git -C ./cli status --porcelain) ]]; then
             echo "Documentation changes detected"
             cd ./cli
-            npx prettier --write ./help/cli-commands
+            npm i
```
PeterSchafer, line 28: nitpick: maybe `npm ci` is better

PR #4646, opened 2023-06-07
chore: implement --print-graph for container

tommyknows:
This commit adds support for the `--print-graph` argument that will print the dependency graph after a container scan. We need this feature in order to build the `snyk container sbom` command later on.

The flag works the exact same way as the normal `snyk test --print-graph` does.

### DIFF
--- a/src/lib/snyk-test/common.ts
+++ b/src/lib/snyk-test/common.ts
@@ -1,5 +1,7 @@
 import config from '../config';
 import { color } from '../theme';
+import { DepGraphData } from '@snyk/dep-graph';
+import { jsonStringifyLargeObject } from '../json';
 
 export function assembleQueryString(options) {
   const org = options.org || config.org || null;
@@ -68,3 +70,16 @@ export type FailOn = 'all' | 'upgradable' | 'patchable';
 
 export const RETRY_ATTEMPTS = 3;
 export const RETRY_DELAY = 500;
+
+// depGraphData formats the given depGrahData with the targetName as expected by
+// the `depgraph` CLI workflow.
+export function depGraphToOutputString(
+  dg: DepGraphData,
+  targetName: string,
+): string {
+  return `DepGraph data:
+${jsonStringifyLargeObject(dg)}
+DepGraph target:
+${targetName}
+DepGraph end`;
+}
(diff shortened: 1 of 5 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -747,7 +747,7 @@ async function assembleLocalPayloads(
 
       // print dep graph if `--print-graph` is set
       if (options['print-graph'] && !options['print-deps']) {
-        await spinner.clear<void>(spinnerLbl)();
+        spinner.clear<void>(spinnerLbl)();
```
mcombuechen, line 750: is it safe to remove the `await` here?

shlomiSnyk: I would assume @tommyknows probably validate this but i'll have a double look.

tommyknows: Uhm yeah this has been a long time ago. I'm actually not sure, if I removed it, it's probably been because the language server told me it's not an awaitable function...but better check again :)

shlomiSnyk: I had validated it.. while the `addSpinner` object is an async function, the `clear` function is not.
Hence the await is redundant in the above line.

attaching function signatures here:
```
async function addSpinner(label: string): Promise<void> {...};
addSpinner.clear = <T>(label): ((valueToPassThrough: T) => T) => {...};
```

src/lib/snyk-test/common.ts
```diff
@@ -68,3 +70,13 @@ export type FailOn = 'all' | 'upgradable' | 'patchable';
 
 export const RETRY_ATTEMPTS = 3;
 export const RETRY_DELAY = 500;
+
+// depGraphData formats the given depGrahData with the targetName as expected by
+// the `depgraph` CLI workflow.
+export function depGraphData(dg: DepGraphData, targetName: string): string {
```
Jdunsby, line 76: nit:
Function name could be more descriptive of the process/output, 
e.g. depGraphToOutputString

shlomiSnyk: Thanks @Jdunsby 🙏 
Besides that is there anything else needed to approve and merge it?

PR #4649, opened 2023-06-08
feat: container sarif flag support without file

rmutuleanu:
#### What does this PR do?

Adds support for `--sarif` and `--sarif-output-file` flags for container scans without the `--file` flag. When the `--file` flag is not specified, the `locations` property is not included in the result.

Issue: LUM-257


test/tap/cli-test/cli-test.docker.spec.ts
```diff
@@ -745,9 +745,7 @@ export const DockerTests: AcceptanceTests = {
       }
     },
 
-    '`test --docker --file=Dockerfile --sarif `': (params, utils) => async (
-      t,
-    ) => {
+    '`container test --sarif `': (params, utils) => async (t) => {
```
tommyknows, line 748: I'm a bit confused, I thought it wasn't possible to generate sarif-documents without a dockerfile, but this test doesn't seem to have specified one so far? 🤔

tommyknows (approved):
Only a small question, not directly related to your changes.

Nice work!

test/tap/cli-test/cli-test.docker.spec.ts
rmutuleanu: You are correct, it wasn't possible. Now, it should be possible. More details here: https://snyksec.atlassian.net/browse/LUM-257

Avishagp (approved):

ChristinaDara (approved):

gitphill (approved):

src/lib/formatters/get-sarif-result.ts
```diff
@@ -18,7 +18,7 @@ export function getResults(testResult): sarif.Result[] {
         {
           physicalLocation: {
             artifactLocation: {
-              uri: testResult.displayTargetFile,
+              uri: testResult.displayTargetFile || testResult.path,
```
tommyknows, line 21: I don't know about this, isn't it a bit weird to have `Dockerfile` there if you specified one, and the container name otherwise? 🤔 

Interestingly, in the test fixture you added (`sarif-with-file-container-result.json`) I can't even find the image name!
Shouldn't we add the image name as the artifactLocation by default?

rmutuleanu: My knowledge of sarif is limited, but my understanding is that in the results location, we should have the location where the result occurred in the file.

Before this task, this was always possible because specifying the --sarif flag required the --file flag to be present.

In "sarif-with-file-container-result.json," you should see the target file and not the image name since that fixture is used when a Dockerfile is provided.

The image name (actually the test results path) is used when the file is not provided, which should be possible now after this task. Basically we are falling back to the path if no file is provided.
Why did we need to do this?
We've first tried to remove the locations completely where there was no file provided, but even though that seemed to be a valid sarif output, GitHub Code Scanning was not happy with that and was throwing an error. So we had to fall back on something there.

We've also checked what `trivy` is doing with the locations in the sarif output when no file is specified, and they do something similar.

Avishagp (approved):

PR #4670, opened 2023-06-19
fix: container sarif replace colon within location uri

rmutuleanu:
When using sarif as output and the Dockerfile is not specified, the `path` is used as the sarif results location uri (added with PR: https://github.com/snyk/cli/pull/4649). 

But the `path` can contain colon characters (e.g. between repo and tag) and GitHub Code Scanning seems to return an error in that case. 

#### What does this PR do?

Replaces the colon characters from the sarif results location uri with underscore characters.

Issue: LUM-257

### DIFF
--- a/src/lib/formatters/get-sarif-result.ts
+++ b/src/lib/formatters/get-sarif-result.ts
@@ -18,7 +18,10 @@ export function getResults(testResult): sarif.Result[] {
         {
           physicalLocation: {
             artifactLocation: {
-              uri: testResult.displayTargetFile || testResult.path,
+              uri: getArtifactLocationUri(
+                testResult.displayTargetFile,
+                testResult.path,
+              ),
             },
             region: {
               startLine: vuln.lineNumber || 1,
@@ -41,4 +44,16 @@ export function getLevel(vuln: AnnotatedIssue) {
     default:
       return 'note';
   }
+}
+
+function getArtifactLocationUri(targetFile: string, path: string): string {
+  if (targetFile) {
+    return targetFile;
+  }
+
+  // For container tests there might be cases when the target file (i.e. Dockerfile passed with the --file flag) is not
+  // present. In this case we use the test result path which contains the image reference (e.g. alpine:3.18.0).
+  // Also, Github Code Scanning returns an error when the artifact location uri from the uploaded sarif file contains
+  // a colon (e.g. alpine:3.18.0 is not valid, but alpine_3.18.0 is valid), so we are replacing colon characters.
+  return path.replace(/:/g, '_');
 }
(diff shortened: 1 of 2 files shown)


test/jest/unit/lib/formatters/get-sarif-result.spec.ts
```diff
@@ -5,7 +5,7 @@ describe('Retrieving sarif result', () => {
   it('should use the test results path as the location uri when target file is not present', () => {
     let result = getResults(
       getTestResult({
-        path: 'alpine:3.18.0',
+        path: 'alpine',
```
tommyknows, line 8: I'm a bit confused, isn't this the input to the fix, and shouldn't that include the tag? doesn't it *need* to include the tag to get the below results? 🤔

src/lib/formatters/get-sarif-result.ts
```diff
@@ -18,7 +18,9 @@ export function getResults(testResult): sarif.Result[] {
         {
           physicalLocation: {
             artifactLocation: {
-              uri: testResult.displayTargetFile || testResult.path,
+              uri:
+                testResult.displayTargetFile ||
+                getArtifactLocationUriFromPath(testResult.path),
```
tommyknows, line 23: This `path` variable is weirdly named, I would think that an image name would rather be in `displayTargetFile` than in `path`....can we add a comment to explain that? 🤔

test/jest/unit/lib/formatters/get-sarif-result.spec.ts
rmutuleanu: Yes, you are right, I've added 3 runs inside that test with different inputs each. I will split the test into 3 individual tests.

src/lib/formatters/get-sarif-result.ts
rmutuleanu: The `displayTargetFile` is what we are passing to the `--file` flag (i.e. Dockerfile in our case)

The test result `path` seem to be the image reference used and with an optional project name appended in some cases (e.g. `alpine`, `alpine:3.18.0`, `alpine/kubernetes-monitor`, last one has the project name appended)

rmutuleanu: I can add a comment to explain this

tommyknows: Yeah I think that would be nice 😅

src/lib/formatters/get-sarif-result.ts
```diff
@@ -42,3 +42,15 @@ export function getLevel(vuln: AnnotatedIssue) {
       return 'note';
   }
 }
+
+function getArtifactLocationUri(targetFile: string, path: string): string {
```
tommyknows, line 46: Nice solution, thanks!

tommyknows (approved):
Very nice! :)

Avishagp (approved):

dotkas (approved):

PR #4726, opened 2023-06-28
chore: bump nodejs parser 1.52.1

JamesPatrickGill:
This bumps nodejs parser 1.52.1

The changes in this lib update allow more accurate dependency resolution of transitive dependencies in npm lockfile v2 and v3 projects that use bundled dependencies. 

There is an [existing acceptance test](https://github.com/snyk/cli/blob/master/test/tap/cli-test/cli-test.npm.spec.ts#L36-L46) that should cover any breaking behavior and we have added a new one that takes a test fixture from the library to add another layer of testing on this specific feature.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,7 @@
         "snyk-gradle-plugin": "3.26.4",
         "snyk-module": "3.1.0",
         "snyk-mvn-plugin": "2.33.0",
-        "snyk-nodejs-lockfile-parser": "1.51.1",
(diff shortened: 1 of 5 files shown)


dotkas (approved):

darscan (dismissed):
This is not a `chore`. `chore`s do not trigger releases and carry no functional changes.. updating a dep is a functional change (or you wouldn't be doing it)

PeterSchafer (approved):

PR #4727, opened 2023-06-28
fix: build image

bastiandoetsch:


PeterSchafer (commented):
Suggestion: Should we change the base image to ubuntu instead of golang since go is installed externally, which can cause confusion and uncertainty about the go version used.

bastiandoetsch:
> Suggestion: Should we change the base image to ubuntu instead of golang since go is installed externally, which can cause confusion and uncertainty about the go version used.

I would rather use the installed version, instead of installing it at runtime - that's the whole point of a build image, to deliver the tools, not having them installed from externally :).

PeterSchafer (approved):

PeterSchafer (approved):

PR #4731, opened 2023-06-29
fix: updates snyk-policy to lastest version

jcsackett:
#### What does this PR do?

This updates snyk-policy to the latest version, to remove an error
with ignores failing because of ID casing.

Ignores are not meant to be case sensitive, but the IDs were being
interpreted as though they were, leading to ignores that didn't work.

This has already been updated in app.snyk.io, but is still an issue
in the CLI.

https://snyksec.atlassian.net/browse/NARW-2259

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,7 +76,7 @@
         "snyk-nodejs-lockfile-parser": "1.52.1",
         "snyk-nuget-plugin": "1.24.1",
         "snyk-php-plugin": "1.9.2",
-        "snyk-policy": "^1.25.0",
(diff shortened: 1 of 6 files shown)


PeterSchafer:
Could you please add an acceptance test for the CLI to ensure that we can detect a regression in the future.

jcsackett:
>Could you please add an acceptance test for the CLI to ensure that we can detect a regression in the future.

For the case sensitivity issue? That's tested in snyk-policy; adding tests here seems redundant, and I would need a fair bit of help adding it and am running out of time in my support rota to get this sorted.

PR #4742, opened 2023-07-13
fix: restore ability for snyk fix to make network requests [FIX-138]

aron:
Currently the `snyk fix` command fails to run due to the subshell running `poetry` and `pipenv` being unable to make network requests to pypy. This is due to the proxy configuration used by the Go wrapper. We've now updated the `@snyk/child-process` library in snyk/python-fix to reset these environment variables before invoking the subshell. This fixes the issue.

The change here is to bump the sub dependencies in the `snyk-fix` package to use the latest @snyk/child-process library (see: https://github.com/snyk/python-fix/pull/68).

This was tested on the above pipenv project and with the changes we can now successfully run `snyk fix` with no errors.

NOTE: We also update the CODEOWNERS to move ownership of snyk-fix from Tech Services over to Team Fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,7 @@ commands:
           name: Installing sdks and tools
           command: |
             brew install go gradle python elixir composer gradle@6 maven sbt
+            python3 -m pip install pipenv --user
 
   install_sdks_linux:
     steps:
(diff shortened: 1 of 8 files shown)


lili2311 (approved):
🎉

PeterSchafer (approved):

PR #4748, opened 2023-07-19
fix: bump snyk-docker-plugin

tommyknows:
Contains an important fix for Debian source package lookups.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "semver": "^6.0.0",
     "snyk-config": "4.0.0",
     "snyk-cpp-plugin": "2.22.2",
-    "snyk-docker-plugin": "6.3.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -115,7 +115,7 @@
     "semver": "^6.0.0",
     "snyk-config": "4.0.0",
     "snyk-cpp-plugin": "2.22.2",
-    "snyk-docker-plugin": "6.3.4",
+    "snyk-docker-plugin": "^6.5.1",
```
gitphill, line 118: we usually pin these to a specific version to avoid unintentional releases

gitphill (approved):

package.json
tommyknows: Good catch, done 👍

j-luong (approved):

PR #4763, opened 2023-07-30
fix: make IaC issue resource file optional

ofekatr:

#### What does this PR do?

- Defines the file path for resources associated with IaC+ test issues as optional.

- Test an IaC file that generates issues without an associated file origin, e.g., an issue for a non-existing resource that needs to be added.
- Ensure the test output for the CLI test is generated as expected.

- [A support thread](https://snyk.slack.com/archives/C034QFM0DH8/p1690280367869079) revealed that we currently incorrectly assume that all the resources associated with a generated IaC+ issue have a file origin.
- Looking at the [CLI code](https://github.com/snyk/cli/blob/bc7ecc2a5552e92c12b514c3ef37341ac0106eb4/src/lib/iac/test/v2/scan/results.ts#L117-L118), we currently assume the file path is required for resources.
The resource file paths for issues are retrieved from `snyk-iac-test`, after being propagated from policy-engine [here](https://github.com/snyk/snyk-iac-test/blob/064ddf70f2c243df9c71b897a0bdeb9b53c32d69/pkg/results/results.go#L163-L169).

- [Zendesk ticket 55895](https://snyk.zendesk.com/agent/tickets/55895)

#### Additional information

- [Slack thread from #ask-cloud](https://snyk.slack.com/archives/C04CQ0D2VFH/p1690722526092069)
- [Slack thread from #team-cloud-engines](https://snyk.slack.com/archives/C034QFM0DH8/p1690280367869079)

### DIFF
--- a/src/lib/formatters/iac-output/text/issues-list/index.ts
+++ b/src/lib/formatters/iac-output/text/issues-list/index.ts
@@ -41,8 +41,8 @@ export function getIacDisplayedIssues(
       const issuesOutput = severityResults
         .sort(
           (severityResult1, severityResult2) =>
-            severityResult1.targetFile.localeCompare(
-              severityResult2.targetFile,
+            severityResult1.targetFile?.localeCompare(
+              severityResult2.targetFile as string,
             ) ||
             severityResult1.issue.id.localeCompare(severityResult2.issue.id),
         )
--- a/src/lib/formatters/iac-output/text/types.ts
+++ b/src/lib/formatters/iac-output/text/types.ts
@@ -16,8 +16,8 @@ export type FormattedOutputResultsBySeverity = {
 
 export type FormattedOutputResult = {
   issue: Issue;
-  targetFile: string;
   projectType: IacProjectType | State.InputTypeEnum;
+  targetFile?: string;
 };
 
 export interface IacTestCounts {
--- a/src/lib/iac/test/v2/analytics/iac-type.ts
+++ b/src/lib/iac/test/v2/analytics/iac-type.ts
@@ -75,7 +75,9 @@ function getFilesCountByPackageManager(
         acc[packageManager] = new Set();
       }
 
-      acc[packageManager].add(resource.file);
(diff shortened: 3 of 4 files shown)


ofekatr:
Pending the confirmation from Cloud Engines that this assumption is incorrect ([thread](https://snyk.slack.com/archives/C034QFM0DH8/p1690280367869079))

YairZ101:
@ofekatr can you share a TF file that I can use to test this change?

ofekatr:
@YairZ101 According to Jasper ([thread](https://snyk.slack.com/archives/C04CQ0D2VFH/p1690799274774749?thread_ts=1690722526.092069&cid=C04CQ0D2VFH)), any TF plan scan should reproduce the case of not having a file path for the resource associated with the issue

YairZ101:
> any TF plan scan should reproduce the case of not having a file path for the resource associated with the issue

I don't think it is for any tf-plan file, I just tested it out and the scan works for me without any problem
image

ofekatr:
Communication on this thread is going slow, and I can't find a specific good example on my own. However, we were able to confirm with Jasper that there are cases where the path is not included. 
He also added that the input object from `policy-engine` should always include a file path, and that we should consider fetching it from there later on. 

With that said, seeing as this change unblocks the customer issue and only makes the property treated as optional, I think we should still merge it until we get the example

ofekatr:
@YairZ101 Communication on this thread is going slow, and I can't find a specific good example on my own. However, we were able to confirm with Jasper that there are cases where the path is not included. 
He also added that the input object from `policy-engine` should always include a file path, and that we should consider fetching it from there later on. 

With that said, seeing as this change unblocks the customer issue and only makes the property treated as optional, I think we should still merge it until we get the example

YairZ101 (approved):
Approving to unblock a customer.
We should have a better understand of which situation can lead to an issue without a file path and how to reproduce this situation in case we will need it in the future.

PR #4769, opened 2023-08-02
chore: add snyk check to e2e tests

bastiandoetsch:


.circleci/config.yml
```diff
@@ -788,35 +828,46 @@ jobs:
...
-      - failed-release-notification
+                name: Checking Snyk CLI
+                shell: bash.exe
+                command: |
+                  PIP_BREAK_SYSTEM_PACKAGES=1 pip install --user --upgrade requests
+                  python scripts/install-snyk.py $(cat binary-releases/version)
+      - when:
+          condition:
+            matches:
+              pattern: '^[docker|macos|alpine|linux].*'
```
PeterSchafer, line 863: Question: should [] be ()?

.circleci/config.yml
```diff
@@ -788,35 +828,46 @@ jobs:
...
+            - run: apt update && apt install -y sudo curl python3 python3-pip python3-requests
+
       - when:
           condition:
             matches:
-              pattern: '^[docker|win|macos].*'
+              pattern: '^win.*'
               value: << parameters.executor >>
           steps:
+            - install-deps-windows-full
```
PeterSchafer, line 853: Question: this takes quite some time, do we really this? If it is just to install python, python should be on the circleci machine images.

scripts/install-snyk.py
```diff
@@ -0,0 +1,118 @@
...
+            os.system(f"{executable} -v")
+
+        else:
+            os.remove(downloaded_file_path)
+            print("SHA256 checksum verification failed. Downloaded file deleted.")
+    else:
+        print(f"Failed to download Snyk CLI v{version}")
+
+
+def get_filename(arch_type, os_type):
```
PeterSchafer, line 84: Suggestion: A future improvement would be to use the file content in `binary-deployments.json`

PeterSchafer: to not hard code file name everywhere

PeterSchafer (approved):
LGTM! Please remove the temporary comments

scripts/install-snyk.py
PeterSchafer: to not hardcode the filenames everywhere.

PR #4770, opened 2023-08-02
feat: build windows and linux binaries with fips enabled Microsoft Go Fork

asaf92:

#### What does this PR do?
* This PR introduces a new way to download and install the golang version used for building the CLI. 
* This enables to use different golang versions besides the google version, we are looking towards using the MS Go Fork in the future.
* It also includes some pipeline refactorings to unify the build job.
* In addition it supports detecting the GOEXPERIMENT values introduced in the [MS Go Fork](https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips#usage-build) to enable FIPS validated crypto.

The different build steps should include log messages like this
```
-- Pre-paring to build
--  Using Go:        ***/go/bin/go
--  Building Binary: ***/snyk/cliv2/_bin/snyk-linux
--  GOEXPERIMENT:    
```
The _download go binary_ step should include
`--base_url=https://go.dev/dl/`

Currently this doesn't include a behavioural or functional change.

HEAD-392, HEAD-395

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,28 @@
 version: '2.1'
 
+parameters:
+  go_download_base_url:
+    type: string
+    default: 'https://storage.googleapis.com/golang/'
+  fips_go_download_base_url:
+    type: string
+    default: 'https://aka.ms/golang/release/latest/'
+  go_version:
+    type: string
+    # https://go.dev/doc/devel/release
+    default: '1.20.6'
+  aws_version:
+    type: string
+    # https://github.com/aws/aws-cli/blob/v2/CHANGELOG.rst
+    default: '2.4.12'
+  gh_version:
+    type: string
+    # https://github.com/cli/cli/releases
+    default: '2.7.0'
+
 orbs:
   prodsec: snyk/prodsec-orb@1.0
   snyk: snyk/snyk@1.7.0
-  go: circleci/go@1.7.1
   win: circleci/windows@5.0
   aws-cli: circleci/aws-cli@2.0.3
   gh: circleci/github-cli@2.1.0
@@ -22,11 +41,6 @@ executors:
       - image: bastiandoetsch209/cli-build-arm64:20230806-080507
     working_directory: /mnt/ramdisk/snyk
     resource_class: arm.large
-  linux-amd64:
-    machine:
-      image: ubuntu-2204:2023.02.1
-    working_directory: /mnt/ramdisk/snyk
-    resource_class: large
   linux-ubuntu-mantic-amd64:
     docker:
       - image: ubuntu:mantic
@@ -77,8 +91,63 @@ executors:
       # https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions
       xcode: '14.3.1'
     resource_class: macos.m1.large.gen1
+  win-amd64:
+    machine:
+      image: windows-server-2022-gui:2023.07.1
+    resource_class: windows.large
+    shell: powershell
 
 commands:
+  install-go:
+    parameters:
+      go_os:
+        type: string
+      go_target_os:
+        type: string
+      go_arch:
+        type: string
+      base_url:
+        type: string
+      extraction_path:
+        type: string
+      cache_key_file:
+        type: string
+        default: go_cache_key.txt
+    steps:
+      - run:
+          name: Create Cache Key
+          command: |
+            echo << parameters.extraction_path >>-<< parameters.base_url >>-<< parameters.go_target_os >>-<< parameters.go_arch >>-<< pipeline.parameters.go_version >> > << parameters.cache_key_file >>
+            cat << parameters.cache_key_file >>
+      - restore_cache:
+          name: Restoring go binary cache
+          keys:
+            - go-binary-cache-{{ checksum "<< parameters.cache_key_file >>" }}
+      - run:
+          name: Download go binary
+          command: python ./scripts/download_go.py << pipeline.parameters.go_version >> --go_os=<< parameters.go_os >> --go_arch=<< parameters.go_arch >> --base_url=<< parameters.base_url >> --extraction_path=<< parameters.extraction_path >>
+      - save_cache:
+          name: Caching go binary
+          key: go-binary-cache-{{ checksum "<< parameters.cache_key_file >>" }}
+          paths:
+            - << parameters.extraction_path >>/go
+      - unless:
+          condition:
+            equal: ['windows', << parameters.go_os >>]
+          steps:
+            - run:
+                name: Add go binary to PATH
+                command: echo "export PATH=$(realpath << parameters.extraction_path >>/go/bin):\$PATH" >> "$BASH_ENV"
+      - when:
+          condition:
+            equal: ['windows', << parameters.go_os >>]
+          steps:
+            - run:
+                name: Add go binary to PATH
+                command: |
+                  New-Item -Path $profile -ItemType File -Force
+                  '$Env:Path = "<< parameters.extraction_path >>\go\bin;" + $Env:Path' >> $profile
+
   install-deps-windows-full:
     steps:
       - restore_cache:
@@ -158,20 +227,6 @@ commands:
       - attach_workspace:
           at: .
 
-parameters:
-  go_version:
-    type: string
-    # https://go.dev/doc/devel/release
-    default: '1.20.6'
-  aws_version:
-    type: string
-    # https://github.com/aws/aws-cli/blob/v2/CHANGELOG.rst
-    default: '2.4.12'
-  gh_version:
-    type: string
-    # https://github.com/cli/cli/releases
-    default: '2.7.0'
-
 ####################################################################################################
 # WORKFLOWS
 ####################################################################################################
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,16 @@ commands:
           key: chocolatey-cache-v2-{{ arch }}-{{ checksum ".circleci/chocolatey.config" }}
           paths:
             - ~\AppData\Local\Temp\chocolatey
+      - run:
+          name: Download and Install Microsoft Go
+          shell: powershell.exe
+          command: |
+            $url = "https://aka.ms/golang/release/latest/go1.20.6-1.src.tar.gz"
```
PeterSchafer, line 105: Pleas use the pipeline parameter go_version instead of a hard coded version.

Avishagp: Not sure what you mean as it seems like I cannot change the pipeline parameter's value in a job?

PeterSchafer: $url = "https://aka.ms/golang/release/latest/go<< pipeline.parameters.go_version >>.src.tar.gz"

PeterSchafer (approved):

j-luong (approved):

.circleci/config.yml
```diff
@@ -159,6 +208,9 @@ commands:
           at: .
 
 parameters:
+  go_base_url:
+    type: string
```
j-luong, line 212: nitpick: can we `enum` this?
```suggestion
    type: enum
    enum: ["https://go.dev/dl/", "https://aka.ms/golang/release/latest/"]
```

PeterSchafer: We already have this in the python script, I don't feel comfortable to have it in both places. where do you think it makes most sense?

.circleci/config.yml
```diff
@@ -563,60 +622,46 @@ jobs:
...
+        type: string
       go_arch:
         type: string
       c_compiler:
         type: string
         default: ''
       executor:
         type: string
-      artifact:
+      go_base_url:
```
j-luong, line 638: actually, having the enum here might be more secure

j-luong, line 638: actually, having the enum here might be more secure

.circleci/config.yml
bastiandoetsch: Should we use `latest` or pin it? There's a new 1.21 release for microsoft go which seems to be "safer" to use. https://github.com/microsoft/go/blob/microsoft/release-branch.go1.21/docs/go1.21.md

.circleci/config.yml
```diff
@@ -563,60 +622,46 @@ jobs:
...
     steps:
       - prepare-workspace
-      - go/install:
-          version: << pipeline.parameters.go_version >>
+      - install-deps-<< parameters.install_deps_extension >>
+      - install-go:
+          go_os: << parameters.go_os >>
+          go_target_os: << parameters.go_target_os >>
+          go_arch: << parameters.go_arch >>
+          base_url: << parameters.go_base_url >>
```
bastiandoetsch, line 655: this should be named differently - it's not obvious what it is. Suggestion: `go_download_base_url`

Makefile
```diff
@@ -13,10 +13,10 @@ BINARY_RELEASES_FOLDER_TS_CLI = binary-releases
...
 SHASUM_CMD = shasum
 GOHOSTOS = $(shell go env GOHOSTOS)
-PYTHON = python
+export PYTHON = python
 
 PYTHON_VERSION = $(shell python3 --version)
-ifdef (PYTHON_VERSION)
+ifneq (, $(PYTHON_VERSION))
 	PYTHON = python3
 endif
```
bastiandoetsch, line 21: a bit more concise:
```
export PYTHON=$(which python 2>/dev/null || which python3)

File: cliv2/Makefile [hunk 6]
```diff
@@ -99,6 +101,12 @@ ifeq ($(GOHOSTOS), windows)
 	SHASUM_CMD = $(SPECIAL_SHELL) $(WORKING_DIR)/scripts/shasum.ps1
 	SIGN_SCRIPT = $(SPECIAL_SHELL) $(WORKING_DIR)/scripts/sign_$(GOHOSTOS).ps1
 	ISSIGNED_SCRIPT = $(SPECIAL_SHELL) $(WORKING_DIR)/scripts/issigned_$(GOHOSTOS).ps1
+	GOEXPERIMENT_DEFAULT_CRYPTO = cngcrypto
+endif
+
+GO_SYSTEMCRYPTO_SUPPORTED = $(shell GOEXPERIMENT=$(GOEXPERIMENT_DEFAULT_CRYPTO) go version 2> /dev/null)
+ifneq (,$(GO_SYSTEMCRYPTO_SUPPORTED))
+	GOEXPERIMENT = $(GOEXPERIMENT_DEFAULT_CRYPTO)
```
bastiandoetsch, line 109: This could be simplified, too (see above)

cliv2/Makefile
```diff
@@ -159,13 +167,20 @@ _validate_sha_v1: $(V1_DIRECTORY)/$(V1_EXECUTABLE_NAME).$(HASH_STRING)
 .PHONY: dependencies
 dependencies: $(V1_DIRECTORY)/$(V1_EXECUTABLE_NAME) $(V1_DIRECTORY)/$(V1_EXECUTABLE_NAME).$(HASH_STRING) _validate_sha_v1
 
+.PHONY: summary
+summary:
+	@echo "$(LOG_PREFIX) Pre-paring to build"
+	@echo "$(LOG_PREFIX)  Using Go:        $(shell which go)"
+	@echo "$(LOG_PREFIX)  Building Binary: $(BUILD_DIR)/$(V2_EXECUTABLE_NAME)"
+	@echo "$(LOG_PREFIX)  GOEXPERIMENT:    $(GOEXPERIMENT)"
+
```
bastiandoetsch, line 176: Nice!

scripts/download_go.py
```diff
@@ -0,0 +1,101 @@
...
+
+# download go binary
+def download_go_binary(filepath, filename, base_url):
+    url = base_url + filename
+    tmp_file = os.path.join(filepath, filename)
+
+    print("Downloading " + url + " to " + tmp_file)
+    
+    try:
+        urllib.request.urlretrieve(url, tmp_file)
```
bastiandoetsch, line 31: Consider using gzip encoding: https://gist.github.com/Manouchehri/0ce55d239fb07c41c92f

scripts/download_go.py
```diff
@@ -0,0 +1,101 @@
...
+        choices=["linux", "windows", "darwin"],
+        required=True)
+    parser.add_argument("--go_arch",
+        help="Architecture to download for (e.g., amd64, arm64)",
+        choices=["amd64", "arm64", "armv6l"],
+        required=True)
+    parser.add_argument("--base_url",
+        help="Base URL to download from (e.g., https://go.dev/dl/)",
+        choices=["https://go.dev/dl/", "https://aka.ms/golang/release/latest/"],
+        default="https://go.dev/dl/")
```
bastiandoetsch, line 76: Interesting question: should we default to MS or to Golang? I'd probably choose MS as a default.

scripts/download_go.py
```diff
@@ -0,0 +1,101 @@
...
+        default=os.getcwd())
+
+    return  parser.parse_args()
+
+if __name__ == "__main__":
+    args = init_argparse()
+
+    # check if go is cached, before trying to download and extract it
+    file_dir = os.path.join(os.path.abspath(args.extraction_path), "go")
+    if  os.path.exists(file_dir):
```
bastiandoetsch, line 88: This should most likely check the shasum?

scripts/download_go.py
```diff
@@ -0,0 +1,101 @@
...
+# download go binary
+def download_go_binary(filepath, filename, base_url):
+    url = base_url + filename
+    tmp_file = os.path.join(filepath, filename)
+
+    print("Downloading " + url + " to " + tmp_file)
+    
+    try:
+        urllib.request.urlretrieve(url, tmp_file)
+        print("Download complete")
```
bastiandoetsch, line 32: Shasum checking would be awesome!

bastiandoetsch: Also, as we're talking about FIPS functionality, do the go providers provide GPG keys to check?

bastiandoetsch (approved):

.circleci/config.yml
PeterSchafer: There is a pipeline parameter that is being used to specify the go version, which we should continue. The 'latest' in the URL might be incorrect, at least it is not intended to use latest. Let me check.

PeterSchafer: Okay, 'latest' is actually part of the base url, there are the following download links
* https://aka.ms/golang/release/latest/go1.20.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.19.linux-amd64.tar.gz
* https://aka.ms/golang/release/latest/go1.21.linux-amd64.tar.gz

PeterSchafer: latest probably relates to the latest version of the fork. It might make sense to pin this.

scripts/download_go.py
PeterSchafer: Since MS doesn't provide binaries for OSes like darwin, the default would be best set to google. So we could explicitly set when we use something else. But I guess there is a case to do it the other way around. Maybe we postpone this for now?!

Makefile
PeterSchafer: the usage of which is a bit problematic when aliases are involved, so I'd rather stick to the less concise version to be honest.

cliv2/Makefile
PeterSchafer: see above :)

scripts/download_go.py
j-luong: added shasum checking of the downloaded `.tar.gz` / `.zip`.

scripts/download_go.py
bastiandoetsch: Lets do gpg in a follow-up PR/Story, using aharc-coded public keys in config.yaml and validate on each platform using the given key.

bastiandoetsch (approved):

PR #4788, opened 2023-08-10
chore: logging SCLE version in -d mode [NEBULA-1329]

metju90:
Prints SCLE version when scanning with -d mode.

image

Prints SCLE errors when there are issues and run with -d mode.

image


src/lib/plugins/sast/analysis.ts
```diff
@@ -294,3 +299,42 @@ function validateLocalCodeEngineUrl(localCodeEngineUrl: string): void {
...
+
+  // `makeRequest` function converts `http` calls to `https`. In some cases, SCLE might be running on http.
+  // This problem is fixed by setting `options.rejectUnauthorized = true`.
+  // Setting `global.ignoreUnknownCA` to true adds rejectUnauthorized=true as an option in `makeRequest`.
+  global.ignoreUnknownCA = true;
+
+  await makeRequest({
+    url: `${scleBaseUrl}/status`,
+    method: 'get'
+  }, callback)
```
nick-watson, line 335: Q: do we actually need to have a callback here?

src/lib/plugins/sast/analysis.ts
```diff
@@ -294,3 +299,42 @@ function validateLocalCodeEngineUrl(localCodeEngineUrl: string): void {
...
+      debug(chalk.yellow(`API response: statusCode:${statusCode}, body: ${body}`))
+    }
+  }
+
+  const ignoreUnknownCAoriginalValue = global.ignoreUnknownCA;
+
+  // `makeRequest` function converts `http` calls to `https`. In some cases, SCLE might be running on http.
+  // This problem is fixed by setting `options.rejectUnauthorized = true`.
+  // Setting `global.ignoreUnknownCA` to true adds rejectUnauthorized=true as an option in `makeRequest`.
+  global.ignoreUnknownCA = true;
```
nick-watson, line 330: Q: how is this handled when calling the analysis/bundle endpoints?

nick-watson (commented):
@metju90 

1. Should we only run the request if -d option is passed?
2. With more SCLE related code do you think we should refactor to a separate module?

src/lib/plugins/sast/analysis.ts
metju90: Removed

metju90:
> @metju90
> 
> 1. Should we only run the request if -d option is passed?
> 2. As we are adding more SCLE related code do you think we should refactor to a separate module?

@nick-watson 

1 - Good point. But then, if SCLE is unhealthy, we want to output this information even if it's run without `-d`. Right? What if we run a health check on each scan and only output SCLE version if:
A) -d is provided
B) health check fails, regardless if i's in debug mode.

2 - Absolutely. Done.

src/lib/plugins/sast/analysis.ts
metju90: CLI calls analysis/bundle via deeproxy on HTTPS protocol. It uses `code-client` library to do that, so it doesn't have to deal with `makeRequest` custom logic.

@nick-watson

nick-watson:
> > @metju90
> > 
> > 1. Should we only run the request if -d option is passed?
> > 2. As we are adding more SCLE related code do you think we should refactor to a separate module?
> 
> @nick-watson
> 
> 1 - Good point. But then, if SCLE is unhealthy, we want to output this information even if it's run without `-d`. Right? What if we run a health check on each scan and only output SCLE version if: A) -d is provided B) health check fails, regardless if i's in debug mode.
> 
> 2 - Absolutely. Done.

@metju90 

1. IMO we are adding this for debugging purposes right now and any output should be with a debug log (no point in making the request if we don't show anything?). It can also be that the SCLE deployment can have 1 unhealthy pod but still be able to provide analysis so in this sense the developer doesn't care about SCLE health and warning them for an intermittent pod issue could be a red herring/confusing. They only care when errors happen which should then be debugged using the -d flag  and the SCLE version and health log can assist with the debugging.

metju90:
> @metju90
> 
> 1. IMO we are adding this for debugging purposes right now and any output should be with a debug log (no point in making the request if we don't show anything?). It can also be that the SCLE deployment can have 1 unhealthy pod but still be able to provide analysis so in this sense the developer doesn't care about SCLE health and warning them for an intermittent pod issue could be a red herring/confusing. They only care when errors happen which should then be debugged using the -d flag  and the SCLE version and health log can assist with the debugging.

Fair enough. Only logging when in debug mode now ✌️

src/lib/plugins/sast/localCodeEngine.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+export async function logLocalCodeEngineVersion(lceUrl = ""): Promise<void> {
+    const scleBaseUrl = lceUrl.replace('/api', '')
+
+    const ignoreUnknownCAoriginalValue = global.ignoreUnknownCA;
+    // `makeRequest` function converts `http` calls to `https`. In some cases, SCLE might be running on http.
+    // This problem is fixed by setting `options.rejectUnauthorized = true`.
+    // Setting `global.ignoreUnknownCA` to true adds rejectUnauthorized=true as an option in `makeRequest`.
```
novalex, line 32: > In some cases, SCLE might be running on http.

What about the cases where it's actually running on HTTPS and we inadvertently bypass CA checks when we shouldn't?

> Setting `global.ignoreUnknownCA` to true adds rejectUnauthorized=true as an option in `makeRequest`.

That may be the case now, but what if the behaviour for this global flag changes in the future and no longer does what it should or does something else unexpected?

src/lib/plugins/sast/analysis.ts
```diff
@@ -64,6 +66,9 @@ export async function getCodeTestResults(
   const isLocalCodeEngineEnabled = isLocalCodeEngine(sastSettings);
   if (isLocalCodeEngineEnabled) {
     validateLocalCodeEngineUrl(sastSettings.localCodeEngine.url);
+    if (process.env.DEBUG) {
```
nick-watson, line 69: req: should we not use `options.debug`?

metju90: Should we?  process.env.DEBUG is the simplest soultion I found. I also found :

1.  args.options.debug [here](https://github.com/snyk/cli/blob/b9658942496c2572db8aa916abb92c759a0d3cc0/src/cli/main.ts#L123)  - which would need the `global args` to get the `debug` value.

2. debugEnabled() [here](https://github.com/snyk/cli/blob/cde8ca6a33c9b0b7faaae867321a451cfcceceff/ts-binary-wrapper/src/common.ts#L155). This requires the CLI args to work, which works similar to the above.

`process.env.DEBUG`  is set [here](https://github.com/snyk/cli/blob/b22ea69c414b6eff4eb389dc49182205a741b4fe/src/cli/args.ts#L108). Looks good enough to me.

Which solution do you think it's best?

nick-watson: Are the options not passed down into the function [here](https://github.com/snyk/cli/pull/4788/files) then?

metju90: Yep. Thanks for the reference.

src/lib/plugins/sast/localCodeEngine.ts
metju90: > What about the cases where it's actually running on HTTPS

Good Q. I wrapped this logic in a non-https condition.

> what if the behaviour for this global flag changes in the future and no longer

I would expect whoever plans to change it to reach us before implementing a breaking change.

shirlupo:
@metju90 I see many people are already looking at this, tell me if you still need my review here

src/lib/plugins/sast/localCodeEngine.ts
```diff
@@ -0,0 +1,60 @@
...
+        global.ignoreUnknownCA = ignoreUnknownCAoriginalValue;
+    }
+
+    if (body?.ok && body?.version) {
+        debug(chalk.green(`Snyk Code Local Engine version: ${body.version}`))
+        return;
+    }
+
+    if (body?.ok === false) {
+        console.log(chalk.red(`Snyk Code Local Engine health check failed. statusCode:${statusCode}, version: ${body?.version}`))
```
nick-watson, line 55: 1. Q: any reason for `console.log` here vs `debug` above?
2. Sug: do we need to handle the scenario where `makeRequest` throws an error?

metju90: 1. Used `console.log` when I wanted to print without -d. Switched back to `debug()` now.
2. This was being handled elsewhere, now I added a try-catch within the SCLE file.

src/lib/plugins/sast/analysis.ts
```diff
@@ -61,15 +63,17 @@ export async function getCodeTestResults(
...
+    if (options.debug) {
+      await logLocalCodeEngineVersion(baseURL)
+    }
   }
 
-  const baseURL = isLocalCodeEngineEnabled
-    ? sastSettings.localCodeEngine.url
-    : getCodeClientProxyUrl();
-
   // TODO(james) This mirrors the implementation in request.ts and we need to use this for deeproxy calls
```
metju90, line 77: Are we still expecting James to do this change? cc @novalex  😬

metju90: @novalex since the last time you review this file, I changed the definition of `baseURL`. You might want to have a look at it.

test/jest/unit/snyk-code/snyk-code-test.spec.ts
```diff
@@ -929,6 +934,33 @@ describe('Test snyk code', () => {
...
+        allowCloudUpload: true,
+        enabled: true,
+      },
+    };
+    await getCodeTestResults(
+      '.',
+      {
+        path: '',
+        code: true,
+      },
```
nick-watson, line 952: Q: does this test past without the debug option being set?

metju90: @nick-watson nope, that wasn't working without the debug option. I fixed it.

nick-watson: @metju90 👍 do we also need a test cases for when it's not passed as well as if an error is thrown?

metju90: I think it's a bit of an overkill to test every aspect of a logging feature?

nick-watson: I think we should at least have a test checking/documenting that any errors thrown don't cause the analysis to fail early

metju90: Done

src/lib/plugins/sast/localCodeEngine.ts
```diff
@@ -0,0 +1,65 @@
...
+    try {
+        const { res: { body, statusCode } } = await makeRequest({
+            url: `${lceBaseUrl}/status`,
+            method: 'get'
+        })
+
+        if (isNonSecureHttp) {
+            // Resetting `global.ignoreUnknownCA` to whatever value it had before I changed it above.
+            global.ignoreUnknownCA = ignoreUnknownCAoriginalValue;
+        }
```
novalex, line 48: `req`: if the `makeRequest` call throws this will never be reset.

Should be moved into a "finally" block instead.

src/lib/plugins/sast/localCodeEngine.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        if (body?.ok === false) {
+            debug(chalk.red(`Snyk Code Local Engine health is not ok. statusCode:${statusCode}, version: ${body?.version}`))
+            debug(chalk.red(`Message: ${JSON.stringify(body?.message)}`))
+            return
+        }
+
+        debug(chalk.red(`Snyk Code Local Engine health check failed. statusCode:${statusCode}, version: ${JSON.stringify(body)}`))
+    } catch (err) {
+        debug("Snyk Code Local Engine health check failed.", err)
```
novalex, line 63: `q`: any chance of logging potentially sensitive data/something we shouldn't here?

src/lib/plugins/sast/localCodeEngine.ts
metju90: Yep, good catch!

src/lib/plugins/sast/localCodeEngine.ts
```diff
@@ -0,0 +1,65 @@
...
+        throw new MissingConfigurationError(
+            'Snyk Code Local Engine. Refer to our docs on https://docs.snyk.io/products/snyk-code/deployment-options/snyk-code-local-engine/cli-and-ide to learn more',
+        );
+    }
+}
+
+export async function logLocalCodeEngineVersion(lceUrl = ""): Promise<void> {
+    const lceBaseUrl = lceUrl.replace('/api', '')
+    const isNonSecureHttp = lceBaseUrl.match(/^http:/);
+    let ignoreUnknownCAoriginalValue;
```
nick-watson, line 29: nit: `ignoreUnknownCAOriginalValue` ?

nick-watson (approved):

novalex (approved):

Avishagp (approved):

PR #4792, opened 2023-08-11
fix: refactor to make filterIgnoredIssues match vuln ids case insensitively

pmlyng:
#### What does this PR do?

Ignored issues are not filtered out case insensitively. 
This adds a test and updates the logic to match the vulnID (named as issue?.issueId or as the key of the issuesData object), with the policy file's ignore object's keys in a case-insensitive manner.

~~This is a draft PR~~

Originally we thought this could be resolved by bumping the version of the policy lib (https://github.com/snyk/cli/pull/4731), however through acceptance testing, we found that the logic here means filtering is still case-sensitive so that wouldn't completely solve the problem.

~~This is still a draft PR, but when ready, we can cherry-pick the changes from this PR into #4731.~~

cherry-picked the changes from #4731 into this one 

Support ticket here: https://snyksec.atlassian.net/browse/NARW-2259

### DIFF
--- a/src/lib/ecosystems/policy.ts
+++ b/src/lib/ecosystems/policy.ts
@@ -42,14 +42,32 @@ export function filterIgnoredIssues(
     return [issues, issuesData];
   }
 
-  const filteredIssuesData: IssuesData = { ...issuesData };
+  const upperCaseKeys = (obj) => {
+    return Object.keys(obj).reduce((acc, k) => {
+      acc[k.toUpperCase()] = obj[k];
+      return acc;
+    }, {});
+  };
+
+  const policyIgnoreKeyUpperCased = upperCaseKeys({ ...policy.ignore });
+
+  const issuesDataCopy = { ...issuesData };
+  const returnUnfilteredIssuesData =
+    !issues?.length || !Object.keys(policy.ignore)?.length;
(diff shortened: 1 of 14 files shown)


darscan:
👋 This change introduces a _lot_ more processing, and tight nested loops into a synchronous operation. Are you sure there isn't a neater and cheaper implementation?

src/lib/ecosystems/policy.ts
```diff
@@ -42,23 +42,48 @@ export function filterIgnoredIssues(
...
 
-  const filteredIssuesData: IssuesData = { ...issuesData };
+  const upperCaseKeys = (obj) => {
+    return Object.keys(obj).reduce((acc, k) => {
+      acc[k.toUpperCase()] = obj[k];
+      return acc;
+    }, {});
+  };
+
+  const policyIgnoreKeyUpperCased = upperCaseKeys({ ...policy.ignore });
```
gclapperton, line 52: Should we normalize the keys in the policy lib?

YairZ101 (approved):
Approving for cloud-dev-ex

jcsackett (approved):

pmlyng:
Closing for now as summarised beautifully by @jcsackett  here https://snyk.slack.com/archives/CL6DF1CDT/p1694098802648089

PR #4813, opened 2023-08-25
test: unexpected-error test expects non-colored output

candrews:

#### What does this PR do?

The assertions in unexpected-error.spec.ts expect the output being tested to not contain ANSI color codes. To ensure that is always the case, set `FORCE_COLOR="0"` so node doesn't output colors.

Without this change, this test will pass or fail depending upon the terminal being used to run the tests. With this change, the test runs consistently.

The diff for the PR is small so I'd start with the diff.

Run the test with `TERM` unset, `TERM=dumb`, `TERM=xterm`, `TERM=xterm-256color`, and any other desired values. Note that with this change, the `unexpected-error` tests consistently pass.

Relevant node documentation: https://nodejs.org/dist/latest-v18.x/docs/api/cli.html#force_color1-2-3

No, but I can make one if that would be helpful - just point me to the instructions to do so :)

#### Screenshots

n/a

### DIFF
--- a/test/jest/unit/lib/unexpected-error.spec.ts
+++ b/test/jest/unit/lib/unexpected-error.spec.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import { runCommand } from '../../util/runCommand';
+import { RunCommandOptions, runCommand } from '../../util/runCommand';
 import { getFixturePath } from '../../util/getFixturePath';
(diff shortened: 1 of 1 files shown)


thisislawatts:
Closed via: https://github.com/snyk/cli/pull/5275

PR #4815, opened 2023-08-30
feat: trigger GHA workflow via CircleCI [HEAD-350]

cat2608:
#### What does this PR do?

This PR adds a CircleCI job to trigger a GitHub Actions workflow in the `snyk/snyk-images` repository using `repository_dispatch` [event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows). This will build and push Snyk Docker images after Upload NPM job is completed.

This new job is triggered when Upload NPM job is completed. 

`cURL` the same command written in `build-and-publish-snyk-images`

HEAD-350

#### Screenshots

|CICD to GHA|
|-|
|cicd-to-gha|

#### Additional questions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1042,20 +1042,6 @@ jobs:
     steps:
       - prepare-workspace
       - run:
-          name: Dispatch build-and-publish workflow at snyk-images
-          command: |
-            RESPONSE=$(curl -L \
-              -X POST \
-              -H "Accept: application/vnd.github+json" \
-              -H "Authorization: Bearer $HAMMERHEAD_GITHUB_PAT" \
-              -H "X-GitHub-Api-Version: 2022-11-28" \
-              https://api.github.com/repos/snyk/snyk-images/dispatches \
-              -d '{"event_type":"build_and_push_images"}' \
-              -w "%{http_code}" \
-              -o /dev/null)
-            if [ "$RESPONSE" -eq 204 ]; then
-              echo "Successfully triggered the workflow."
-            else
-              echo "Failed to trigger the workflow. HTTP response code: $RESPONSE."
-              exit 1
-            fi
+          name: Trigger build-and-publish workflow at snyk-images
+          command: ./release-scripts/upload-artifacts.sh
+      - failed-release-notification
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1043,5 +1043,5 @@ jobs:
       - prepare-workspace
       - run:
           name: Trigger build-and-publish workflow at snyk-images
-          command: ./release-scripts/upload-artifacts.sh
+          command: ./release-scripts/upload-artifacts.sh trigger-snyk-images
       - failed-release-notification
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -652,8 +652,8 @@ workflows:
               only:
                 - master
 
-      - build-and-publish-snyk-images:
-          name: build and publish snyk images
+      - trigger-building-snyk-images:
+          name: Trigger building snyk-images
           context: team-hammerhead-common-deploy-tokens
           requires:
             - upload npm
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -652,6 +652,13 @@ workflows:
               only:
                 - master
 
+      - build-and-publish-snyk-images:
+          requires:
+            - prepare-build
+          filters:
+            branches:
+              ignore: master
+
```
cat2608, line 661: Testing. To be replaced by `only: master`

.circleci/config.yml
```diff
@@ -652,6 +652,13 @@ workflows:
               only:
                 - master
 
+      - build-and-publish-snyk-images:
+          requires:
```
bastiandoetsch, line 656: Looks like your not injecting the context here that contains the token. Try adding `context: team-hammerhead-common-deploy-tokens`

cat2608: Ouch 🤦‍♀️ 
Thank you!!!! 🚀

.circleci/config.yml
```diff
@@ -1026,3 +1036,26 @@ jobs:
...
       - failed-release-notification
+
+  build-and-publish-snyk-images:
+    executor: docker-amd64
+    steps:
+      - prepare-workspace
+      - run:
+          name: Dispatch build-and-publish workflow at snyk-images
+          command: |
+            RESPONSE=$(curl -L \
```
PeterSchafer, line 1047: Suggestion: As you can see, we have been moving logic out of the circleci config into scripts and made available via make commands. This enables to better run these steps somewhere else for example locally. How about making this code part of `./release-scripts/upload-artifacts.sh` and add a new target eg. `snyk-images`.

.circleci/config.yml
```diff
@@ -652,6 +652,16 @@ workflows:
               only:
                 - master
 
+      - build-and-publish-snyk-images:
+          name: build and publish snyk images
```
PeterSchafer, line 656: nitpick: maybe just call it `Trigger building snyk-images` since this job doesn't actually build. This running successfully doesn't guarantee that snyk-images successfully build.

.circleci/config.yml
cat2608: That makes sense! 🙌
I'll change that :)
Thank you!

.circleci/config.yml
```diff
@@ -1026,3 +1036,12 @@ jobs:
...
           command: ./release-scripts/upload-artifacts.sh npm
       - failed-release-notification
+
+  trigger-building-snyk-images:
+    executor: docker-amd64
+    steps:
+      - prepare-workspace
+      - run:
+          name: Trigger build-and-publish workflow at snyk-images
+          command: ./release-scripts/upload-artifacts.sh
```
PeterSchafer, line 1046: Issue: I think you are missing the parameter `trigger-snyk-images`

cat2608: Oh! That's true! Thank you!

.circleci/config.yml
cat2608: Done 🎉

.circleci/config.yml
cat2608: Done 🎉

release-scripts/upload-artifacts.sh
```diff
@@ -56,6 +56,12 @@ show_help() {
   echo ""
   echo "  This will perform a dry run of uploading artifacts to GitHub, npm, and S3 for version v1.0.0"
   echo ""
+  echo -e "\033[1;33mTrigger Build and Publish Snyk Images:\033[0m"  # Set color to yellow
+  echo ""
+  echo "  upload-artifacts.sh trigger-snyk-images"
+  echo ""
+  echo "  This will trigger the build-and-publish workflow in the snyk-images repository."
+  echo ""
```
cat2608, line 64: This looks like the following:

show_help

PeterSchafer (approved):

PR #4837, opened 2023-09-05
fix: fix project name override with sub projects

xzhou-snyk:

#### What does this PR do?

When we scan a mono repo where in the root dir there's no project config but only sub dirs containing sub projects, e.g. like following
```
my-mono-repo
- project-a
  - settings.gradle
  - build.gradle 
- project-b
  - settings.gradle
  - build.gradle
```

We noticed that when doing `snyk monitor` instead of creating two projects, it creates one `my-mono-repo` project and the vulnerabilities found in the sub projects are overriding, and the `Explore this snapshot at` message printed out locally points to exactly the same remote URL.

We found that this issue is caused by that the `project-name` is missing as part of the meta data.

This PR supplies that missing `project-name`. The fixing has been tested locally and works well.

#### Screenshots

Before: 
Screenshot 2023-09-05 at 17 29 28

After:
Screenshot 2023-09-05 at 17 29 34

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -280,7 +280,11 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
           const res: MonitorResult = await promiseOrCleanup(
             snykMonitor(
               path,
-              generateMonitorMeta(options, extractedPackageManager),
+              generateMonitorMeta(
+                options,
+                extractedPackageManager,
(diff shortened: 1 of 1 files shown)


Jdunsby (approved):

Greg-Smulko:
Hey, is it possible that this change resulted in my list of projects looking like this:

image

as opposed to this:

image

?

PR #4844, opened 2023-09-08
fix: don't use `latest` in download, try to get actual version

bastiandoetsch:

### DIFF
--- a/scripts/install-snyk.py
+++ b/scripts/install-snyk.py
@@ -124,6 +124,23 @@ def verify_checksum(file_path, expected_checksum):
     return sha256.hexdigest() == expected_checksum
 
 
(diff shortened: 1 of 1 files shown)


scripts/install-snyk.py
```diff
@@ -134,10 +152,13 @@ def verify_checksum(file_path, expected_checksum):
 
     for retry in range(1, args.retry + 1):
         print("Trying to download: #" + str(retry) + " of #" + str(args.retry))
+
+        get_latest_version(args.version)
```
PeterSchafer, line 156: Question: Don't we want to use the return value here?

PeterSchafer (approved):

PR #4847, opened 2023-09-08
fix: update strip-ansi, update ansi-regex version

Avishagp:
Update strip-ansi version to update vulnerable ansi-regex version.


j-luong:
resolved in #4855

PR #4848, opened 2023-09-08
fix: update wrap-ansi to update ansi-regex

Avishagp:
Update wrap-ansi version to update vulnerable transitive dependency ansi-regex version.


j-luong:
resolved in #4855

PR #4852, opened 2023-09-11
Feat: output additional SARIF data

schottsfired:

#### What does this PR do?
Adds the following data to container and open source SARIF output:
* Number of artifacts scanned
* CVSS score
* Fix information
* Package info in pURL format

Please take a look at the thread referenced below for additional info.

* Build the CLI
* Run a command similar to `./binary-releases/snyk-macos test --file=/Users/schottsfired/Desktop/vulnerable/pom.xml --sarif-file-output=/Users/schottsfired/Desktop/test-output.sarif`

See thread [here](https://snyk.slack.com/archives/C1NREQSG0/p1694113046308119) for additional information on why these fields were added.

#### Additional questions
* I noticed that the `vuln` object includes a `cvssScore` in the SARIF output, but I couldn't find a type/declaration for `cvssScore`, so I casted `vuln` to a custom `CvssScore` type to extract the `cvssScore` value from the `vuln` object. I'm more of a Java developer than TS, so please forgive my ignorance and let me know if there's a better way.


bastiandoetsch (commented):
Hey David, looks good so far - would you please add / extend acceptance tests for these additional information?

schottsfired:
Acceptance test added!

bastiandoetsch (approved):

src/lib/formatters/open-source-sarif-output.ts
```diff
@@ -28,6 +28,9 @@ export function createSarifOutputForOpenSource(
       tool: {
         driver: {
           name: 'Snyk Open Source',
+          properties: {
+            artifactsScanned: testResult.dependencyCount
```
RotemS, line 32: Is this specific to your use case? 
Was wondering if we should add this to the container flow as well but didn't see it in the SARIF spec.

schottsfired: Hi @RotemS, great catch. You're right, `artifactsScanned` doesn't appear to be in the SARIF spec. So it's specific to my use case - populating the "Libraries scanned" integer in a security reporting UI.

schottsfired: How should we handle this @RotemS ? If we have a requirement to conform 100% to the SARIF spec, then I suppose this field should be removed. If not, then I suppose it can stay. Please LMK either way.

pecodez (approved):

src/lib/formatters/open-source-sarif-output.ts
pecodez: I don't think this is a blocker, we can re-visit if need be.

danielroymoore (approved):

schottsfired:
@bastiandoetsch @pecodez @danielroymoore apologies, I just received a document that may warrant tiny adjustments to this PR. Please hold on merging this until I've had a chance to review and post back.

schottsfired:
@bastiandoetsch @pecodez @danielroymoore thanks for your patience, we're all set now.

In my last commit ([fedaa58](https://github.com/snyk/cli/pull/4852/commits/fedaa58c0d8d8cbffdb9a4c2fdd6a7ce0be96554)) I moved the CVSS score to a different field, as per the updated docs/requirements I received last week. Updated the unit & acceptance tests as well.

schottsfired:
Hi @bastiandoetsch @pecodez @danielroymoore, all set to merge this?

src/lib/formatters/open-source-sarif-output.ts
```diff
@@ -84,12 +87,17 @@ ${vuln.description}`.replace(/##\s/g, '# '),
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             testResult.packageManager!,
           ],
+          cvssv3_baseScore: (vuln as CvssScore).cvssScore,
```
yurinka1, line 90: Suggestion: avoid extracting this field inline.

Alternative: extract the value of `cvssScore` outside without casting it to `CvssScore` ( is it really of this type??? )

yurinka1 (approved):

src/lib/formatters/open-source-sarif-output.ts
schottsfired: Actually this is better 😄 https://github.com/snyk/cli/pull/4852/commits/1692077736e5f3a6f219d9eb88b05f522c0d2ba5

schottsfired: I double checked with the team who owns the reporting UI and their expectation is to receive `null` if the score is unavailable: https://github.com/snyk/cli/pull/4852/commits/11f5898862fd40d71e5e8024096c44279e50b20f.

schottsfired:
✔️  fyi I went ahead and mirrored the changes to container SARIF output in addition to open source.

(fixed a few tests, too)

src/lib/formatters/get-sarif-result.ts
```diff
@@ -20,15 +20,33 @@ export function getResults(testResult): sarif.Result[] {
             artifactLocation: {
               uri: getArtifactLocationUri(
                 testResult.displayTargetFile,
-                testResult.path,
+                (testResult as any).path,
```
bastiandoetsch, line 23: suggestion: don't use `as any` if possible. Why do you need it here?

src/lib/formatters/get-sarif-result.ts
```diff
@@ -20,15 +20,33 @@ export function getResults(testResult): sarif.Result[] {
...
             },
           },
+          logicalLocations: [
+            {
+              fullyQualifiedName: `${vuln.packageName}@${vuln.version}`,
+            }
+          ]
         },
       ],
+      fixes: vuln.upgradePath.length >= 2
```
bastiandoetsch, line 37: suggestion: not a javascript expert, but maybe only call `.length` after checking if it actually is not `undefined`?

bastiandoetsch: suggestion: to get the same advice as in VSCode & Eclipse, you could even extend it more - see: https://github.com/snyk/snyk-ls/blob/a52f9d047ea29210636d8bb01ef9ea9d1233f4b2/infrastructure/oss/issue_html.go#L146

src/lib/formatters/get-sarif-result.ts
```diff
@@ -20,15 +20,33 @@ export function getResults(testResult): sarif.Result[] {
...
+            {
+              description: {
+                text: `Upgrade to ${vuln.upgradePath[1]}`,
+              },
+              artifactChanges: [],
+            },
+          ]
+        : undefined,
+      properties: {
+        cvssv3_baseScore: vuln.cvssScore,
```
bastiandoetsch, line 48: suggestion: add cwe, cve, too?

test/jest/acceptance/snyk-test/output-formats/sarif.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    const { code, stdout, stderr } = await runSnykCLI('test --sarif', {
+      cwd: project.path(),
+      env,
+    });
+    expect(code).toEqual(1);
+
+    expect(stdout).toContain('"artifactsScanned": 1');
+    expect(stdout).toContain('"cvssv3_baseScore": 5.3');
+    expect(stdout).toContain('"fullyQualifiedName": "lodash@4.17.15"');
+    expect(stdout).toContain('Upgrade to lodash@4.17.17');
```
bastiandoetsch, line 55: suggestion: please add expectation regarding the previous sarif output, e.g. ensuring that existing output does not change, and only additional information is made available. We wouldn't want to have an unvoluntary output (=API) change. Alternatively, if we already have an Acceptance Test for that, just point it out to me :).

bastiandoetsch:
Hm. I don't know why the outstanding workflows are not running after I approved them. Maybe a github/circleci thing, but maybe we need to open a new PR for this - maybe even one of the maintainers need to do this :(.

bastiandoetsch:
Ok, let's do it differently. CircleCI is special about this. So, I'll pull your branch in from your repo and push it into this repo and create a PR - I'll get back with more information on this later.

bastiandoetsch:
Pulled it in, now it's https://github.com/snyk/cli/pull/4966.

src/lib/formatters/get-sarif-result.ts
schottsfired: Thanks @bastiandoetsch , nice catch. When `as any` is removed, we get a compilation error ("path does not exist on type `testResult`"). I had tried adding `path` to the `TestResult` interface but broke a bunch of tests, so I worked around it with `as any` - and felt guilty doing that 😬 . Upon revisiting this, I found a better way to add `path` to the `TestResult` type, and will push that out shortly.

src/lib/formatters/get-sarif-result.ts
schottsfired: > not a javascript expert, but maybe only call .length after checking if it actually is not undefined?

Same here. I noticed that `upgradePath` is not optional in the `interface` and assumed it would always be defined, but after doing more research it's certainly better to do a null/undefined check first. Will add that.

src/lib/formatters/get-sarif-result.ts
schottsfired: Sorry, can you be more specific? The integration I'm working on only requires `cvssv3_baseScore`.

schottsfired:
My reflog is a mess 😅 but I managed to squash all my commits. @bastiandoetsch can you please update https://github.com/snyk/cli/pull/4966? All the warnings should go away 🙏

andreeaneata (approved):

PR #4853, opened 2023-09-12
feat: determine, inject and publish the LS protocol version [HEAD-692]

bastiandoetsch:

#### What does this PR do?
This change causes each deployment to also publish the LsProtocolVersion to static.snyk.io. Also, it injects it into the built CLI based on the language server version compiled into it.

https://static.snyk.io/fips/cli/v1.1217.0-dev.61ef58845c766b628151ea0e13504ef3dd03eafe/ls-protocol-version-10

Then config.yml, upload-to-s3.sh, both Makefiles

Run the pipeline, look at create-artifacts and upload jobs


.circleci/config.yml
```diff
@@ -181,6 +181,8 @@ commands:
           key: chocolatey-cache-v3-{{ arch }}-{{ checksum ".circleci/chocolatey.config" }}
           paths:
             - ~\AppData\Local\Temp\chocolatey
+      - install-deps-python:
+          executor: win-something
```
PeterSchafer, line 185: Question: win-something?

PeterSchafer: Oh got it, it is just to trigger another case. This is a bit confusing. Can we please at least have a comment what we do here?

PeterSchafer: Actually it might be better to rename the parameter from executor to something like OS, since executor is an existing construct that has a special meaning.

release-scripts/write-ls-protocol-version.py
```diff
@@ -0,0 +1,53 @@
...
+        return f"Failed to retrieve commit information: {response.status_code}"
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 4:
+        print("Usage: python script.py <commit_hash> <version> <output-directory>")
+        sys.exit(1)
+
+    commit_hash = sys.argv[1]
+    ls_protocol_version = get_goreleaser_yaml(commit_hash)
```
PeterSchafer, line 49: Suggestion: there should be an actual error handling. Currently everything returned is a string and the script succeeds if an error occurred or not.

release-scripts/write-ls-protocol-version.py
```diff
@@ -0,0 +1,53 @@
...
+
+            # Parse the YAML content
+            yaml_data = yaml.safe_load(yaml_content)
+
+            # Extract the value of LS_PROTOCOL_VERSION under env
+            env_variables = yaml_data.get('env', [])
+            for env_var in env_variables:
+                if isinstance(env_var, str) and env_var.startswith("LS_PROTOCOL_VERSION="):
+                    prot_version = env_var.split('=')[1]
+                    return prot_version
```
PeterSchafer, line 34: Suggestion: how about returning an integer, this would add input validation to the data.

.circleci/config.yml
bastiandoetsch: yeah, already thought about it and then forgot :)

release-scripts/write-ls-protocol-version.py
bastiandoetsch: Sure, why not :)

.circleci/config.yml
```diff
@@ -921,6 +966,11 @@ jobs:
       - run:
           name: Validating artifacts
           command: ./release-scripts/validate-checksums.sh
+      - run:
```
PeterSchafer, line 969: Question: the metadata files are already generated during the build, why do we need to regenerate them here?

cliv2/Makefile
```diff
@@ -208,12 +208,20 @@ $(V1_WORKING_DIR)/$(V1_BINARY_FOLDER)/$(V1_EXECUTABLE_NAME):
...
+
 .PHONY: build-ts-cli
 build-ts-cli: $(V1_WORKING_DIR)/$(V1_BINARY_FOLDER)/$(V1_EXECUTABLE_NAME)
 	$(eval CLI_V1_VERSION_TAG := $(shell cat $(V1_WORKING_DIR)/$(V1_BINARY_FOLDER)/version))
 	$(eval CLI_V1_LOCATION := $(V1_WORKING_DIR)/$(V1_BINARY_FOLDER)/)
-	
+
+.PHONY: generate-ls-protocol-metadata
+generate-ls-protocol-metadata:
+	$(eval CLI_V1_VERSION_TAG := $(shell cat $(DESTDIR)$(bindir)/version))
```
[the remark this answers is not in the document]
bastiandoetsch: Yes, because BUILD_DIR and INSTALL is not always available (e.g. when running create deployment artifacts)

PeterSchafer (approved):

PR #4861, opened 2023-09-16
chore: Make license script used in build more portable

garethr:
The current script `cliv2/ scripts/prepare_licenses.sh` assumes it can run python directly on the host.  If someone is using python actively for development that's likely disabled. It's good practice to disable pip from installing in the global namespace using one of:

```
pip config set global.require-virtualenv True
PIP_REQUIRE_VIRTUALENV=1
```

That means when running a build it will fail like so:

```
-- Preparing 3rd Party Licenses
make[1]: *** [/Users/garethr/Documents/cli/cliv2/_cache/prepare-3rd-party-licenses] Error 3
make: *** [build] Error 
```

This PR changes the script to use `virtualenv` if it's installed. It will create a venv if needed, or reuse an existing one. If virtualenv is not installed it will fall back on the current behaviour.

#### What does this PR do?

Makes the build process more portable, by using virtualenv when present to avoid clashes with the system python configuration.

By running the script directly, followed by running a full build.

* Run a build in a previously working environment, this shouldn't change anything there
* Run the script directly, again, this should work as previously
* Install virtualenv on your path, and then run the script directly. Note the use and creation of a virtualenv
* Run a full clean build in the presense of the virtualenv

### DIFF
--- a/cliv2/scripts/prepare_licenses.sh
+++ b/cliv2/scripts/prepare_licenses.sh
@@ -1,11 +1,30 @@
 #!/usr/bin/env bash
 set -euo pipefail
+
+VENV_DIR="./venv"
+
 BASEDIR=$(dirname "$0")
-PYTHON_VERSION=""
 
-if python3 -c 'print("python3")' > /dev/null 2>&1; then
-    PYTHON_VERSION="3"
+# check to see if virtualenv is installed
+command -v virtualenv >/dev/null 2>&1
+VIRTUALENV_INSTALLED=$?
+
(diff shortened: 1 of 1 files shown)


PeterSchafer:
Hey @garethr, thanks a lot for the contribution. I double checked and it is actually not required anymore to install from within the script. Since it is always better to remove code, I removed the install step in another PR and would propose to close this PR. 
I hope this makes sense.

PeterSchafer:
Closing this PR, since we removed the installation completely in #4863 so we don't require a virtual environment anymore.

PR #4882, opened 2023-09-28
chore: prepare support for Node 18 [HEAD-497]

cat2608:
#### What does this PR do?

This is one more step towards adding support for Node 18. In this PR we only updated config for smoke tests and added OPENSSL to the forbidden list

HEAD-497


Makefile
```diff
@@ -114,27 +114,27 @@ $(BINARY_RELEASES_FOLDER_TS_CLI)/snyk-protect.tgz: prepack | $(BINARY_RELEASES_F
 	mv $(shell npm pack --workspace '@snyk/protect') $(BINARY_RELEASES_FOLDER_TS_CLI)/snyk-protect.tgz
 
 $(BINARY_RELEASES_FOLDER_TS_CLI)/snyk-alpine: prepack | $(BINARY_RELEASES_FOLDER_TS_CLI)
-	$(PKG) -t node16-alpine-x64 -o $(BINARY_RELEASES_FOLDER_TS_CLI)/snyk-alpine
+	$(PKG) -t node18-alpine-x64 -o $(BINARY_RELEASES_FOLDER_TS_CLI)/snyk-alpine
```
PeterSchafer, line 117: Suggestion: How about using the `.mvnrc` file to determine the major version, instead of having it hardcoded here. Take a look at https://github.com/snyk/cli/blob/master/scripts/create-build-image.sh#L9C3-L9C32 for example

cat2608: Updated, thanks for the suggestion 🙏

PeterSchafer (approved):

PR #4887, opened 2023-10-02
feat: make organization ID generally available [HEAD-855]

PeterSchafer:

#### What does this PR do?
In the Legacy CLI, introduce an additional config value that will always hold the orgId provided by the Extensible CLI.

* cliv2/pkg/basic_workflows/legacycli.go
* src/lib/config/index.ts


cliv2/internal/proxy/proxy.go
```diff
@@ -49,12 +50,15 @@ const (
...
 )
 
-func NewWrapperProxy(insecureSkipVerify bool, cacheDirectory string, cliVersion string, debugLogger *log.Logger) (*WrapperProxy, error) {
+func NewWrapperProxy(config configuration.Configuration, cliVersion string, debugLogger *log.Logger) (*WrapperProxy, error) {
 	var p WrapperProxy
 	p.DebugLogger = debugLogger
 	p.cliVersion = cliVersion
 	p.addHeaderFunc = func(request *http.Request) error { return nil }
 
+	cacheDirectory := config.GetString(configuration.CACHE_PATH)
```
bastiandoetsch, line 59: praise: nice to include this here

cliv2/internal/cliv2/cliv2.go
```diff
@@ -47,7 +49,10 @@ const (
 	V2_ABOUT   Handler = iota
 )
 
-func NewCLIv2(cacheDirectory string, debugLogger *log.Logger) (*CLI, error) {
+func NewCLIv2(config configuration.Configuration, debugLogger *log.Logger) (*CLI, error) {
+
+	cacheDirectory := config.GetString(configuration.CACHE_PATH)
```
bastiandoetsch, line 54: question: should we ensure that the directory is writeable?

bastiandoetsch (approved):

cliv2/internal/cliv2/cliv2.go
PeterSchafer: Good idea, I think this should be done more centrally. In general the default cache path is checked for being writeable.

PR #4892, opened 2023-10-04
feat: support -- arguments for sbom and other extensions [HEAD-861]

PeterSchafer:

#### What does this PR do?
* For CLI Extensions, enable accessing arguments passed behind double dash (--)
* The legacycli Extension is automatically forwarding these arguments to the legacy CLI.

* cmd/cliv2/main.go
* pkg/basic_workflows/legacycli.go

Both invocations should give an error that they can't go list "arbitrary content"
`snyk sbom cliv2 --format cyclonedx1.4+json -- arbitrary content`
`snyk test cliv2 -- arbitrary content`

* Arguments behind -- are considered positional arguments even if they start with a `-` 
* They should not interfere with the positional argument used by the CLI to define a folder/package to scan

HEAD-861


cliv2/cmd/cliv2/main.go
```diff
@@ -141,23 +142,44 @@ func getFullCommandString(cmd *cobra.Command) string {
 	return name
 }
 
+func updateConfigFromParameter(config configuration.Configuration, args []string, rawArgs []string) {
+	// extract everything behind --
+	doubleDashArgs := []string{}
+	doubleDashPosition := -1
```
bastiandoetsch, line 148: suggestion (non-blocking): this doesn't need to be a number, could just be a boolean, which would make the logic I think easier to understand.

cliv2/cmd/cliv2/main.go
```diff
@@ -141,23 +142,44 @@ func getFullCommandString(cmd *cobra.Command) string {
...
+
+	// only consider the first positional argument as input directory if it is not behind a double dash.
+	if len(args) > 0 && !utils.Contains(doubleDashArgs, args[0]) {
+		config.Set(configuration.INPUT_DIRECTORY, args[0])
+	}
+}
+
 // main workflow
 func runCommand(cmd *cobra.Command, args []string) error {
+	return runMainWorkflow(globalConfiguration, cmd, args, os.Args)
```
bastiandoetsch, line 166: nitpick: I would add a command why we pass the os.Args array on top of the args. Either in the func declaration or this call.

cliv2/cmd/cliv2/main.go
```diff
@@ -197,15 +219,15 @@ func defaultCmd(args []string) error {
 	// prepare the invocation of the legacy CLI by
 	// * enabling stdio
 	// * by specifying the raw cmd args for it
-	config.Set(configuration.WORKFLOW_USE_STDIO, true)
-	config.Set(configuration.RAW_CMD_ARGS, args)
+	globalConfiguration.Set(configuration.WORKFLOW_USE_STDIO, true)
```
bastiandoetsch, line 222: praise: I like this rename!

cliv2/cmd/cliv2/main.go
```diff
@@ -197,15 +219,15 @@ func defaultCmd(args []string) error {
...
+	globalConfiguration.Set(configuration.RAW_CMD_ARGS, args)
 	_, err := engine.Invoke(basic_workflows.WORKFLOWID_LEGACY_CLI)
 	return err
 }
 
 func getGlobalFLags() *pflag.FlagSet {
-	globalConfiguration := workflow.GetGlobalConfiguration()
-	globalFLags := workflow.FlagsetFromConfigurationOptions(globalConfiguration)
+	globalConfigurationOptions := workflow.GetGlobalConfiguration()
```
bastiandoetsch, line 229: suggestion (non-blocking): let's rename `GetGlobalConfiguration` to `GetGlobalConfigurationOptions` in a follow-up PR.

cliv2/cmd/cliv2/main.go
```diff
@@ -485,21 +507,21 @@ func MainWithErrorCode() int {
...
+		writeLogHeader(globalConfiguration, networkAccess)
 	}
 
 	// init Analytics
 	cliAnalytics := engine.GetAnalytics()
 	cliAnalytics.SetVersion(cliv2.GetFullVersion())
 	cliAnalytics.SetCmdArguments(os.Args[1:])
 	cliAnalytics.SetOperatingSystem(internalOS)
-	if config.GetBool(configuration.ANALYTICS_DISABLED) == false {
+	if globalConfiguration.GetBool(configuration.ANALYTICS_DISABLED) == false {
```
bastiandoetsch, line 524: question (non-blocking): don't we need a fedramp check here?

PeterSchafer: The fedramp check is in the [gaf](https://github.com/snyk/go-application-framework/blob/main/pkg/analytics/analytics.go#L313)

bastiandoetsch (approved):
Works. 🎉

PR #4899, opened 2023-10-10
chore: remove smoke-tests.yml from github actions

bastiandoetsch:
#### What does this PR do?

This PR removes the periodic smoke tests. The tests are already covered in our pipeline regression tests


thisislawatts:
Closing for the time being. There is an ongoing discussion around the interaction between the smoke tests and acceptance tests which needs to be resolved before we make a decision on this.

PR #4865, opened 2023-09-19
fix: fix wrong project name change

xzhou-snyk:
Due to a previous [fixing for gradle](https://github.com/snyk/cli/pull/4837/files), we accidentally also changed the name of many existing CLI projects.

That affects ecosystems such as dotnet, cocoapads, golang projects not using go module, swift and composer.

This PR reverts the name change for those projects while still preserving the expected behavior for gradle.

PR #4906, opened 2023-10-13
fix: add condition to preserve project-name with gradle monitor

danielroymoore:
Following on from https://github.com/snyk/cli/pull/4865 this change tightens the method to explicitly check if allProjects is set and acts accordingly, either by making use of project name or falling back to the projectName flag.

### DIFF
--- a/test/jest/unit/cli-monitor-utils.spec.ts
+++ b/test/jest/unit/cli-monitor-utils.spec.ts
@@ -155,6 +155,24 @@ describe('cli-monitor-utils test', () => {
     expect(res).toEqual('meta-gradle-project');
   });
 
+  it('getProjectName returns project name when project-name is provided via option', () => {
+    const scannedProject: ScannedProject = {
+      depGraph: {} as any,
+      meta: { gradleProjectName: 'my-gradle-project' },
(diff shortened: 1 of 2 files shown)


DOlufemi (approved):

darscan (dismissed):
Please rebase instead of merging mainline back into your feature branch (which creates a messy history).

Also, please add or amend some tests. The previous "fix" added some tests, so just modify or add to those

test/jest/unit/cli-monitor-utils.spec.ts
```diff
@@ -155,6 +155,24 @@ describe('cli-monitor-utils test', () => {
     expect(res).toEqual('meta-gradle-project');
   });
 
+  it('getProjectName returns project name when project-name is provided via option', () => {
+    const scannedProject: ScannedProject = {
+      depGraph: {} as any,
+      meta: {},
```
darscan, line 161: You need this to actually test your change:
```suggestion
      meta: {
        gradleProjectName: 'my-gradle-project',
      },
```
Without `meta.gradleProjectName` you would have fallen through to returning `meta['project-name']`. i.e. you haven't run this test without the change and seen it fail.

PR #4914, opened 2023-10-19
fix: [OSM-347] Showing `displayTargetFile` before `targetFile` if defined

dotkas:
We are working on displaying multiple scan results for .NET projects, if the project targets multiple `TargetFramework`s. 

However, there isn't currently a good way to display it in the CLI's response. The user will just see multiple scan results looking identical, see attached screenshot:

Screenshot 2023-10-18 at 18 32 34

If we allow to use the `displayTargetFile` first (which also seems more like what variable name suggests it should?) I can send a overwritten `targetFile` from the plugin that will become the `displayTargetFile` while still retaining the correct `targetFile` in the final result object, as such:

Screenshot 2023-10-19 at 15 12 07

### DIFF
--- a/src/lib/formatters/format-test-meta.ts
+++ b/src/lib/formatters/format-test-meta.ts
@@ -10,7 +10,7 @@ export function formatTestMeta(
 ): string {
   const padToLength = 19; // chars to align
   const packageManager = res.packageManager || options.packageManager;
-  const targetFile = res.targetFile || res.displayTargetFile || options.file;
+  const targetFile = res.displayTargetFile || res.targetFile || options.file;
   const openSource = res.isPrivate ? 'no' : 'yes';
   const meta = res.org
     ? [chalk.bold(rightPadWithSpaces('Organization: ', padToLength)) + res.org]


dotkas:
> You've modified files in src/ directory, but haven't updated anything in test folder. Is there something that could be tested?

Yes ... yes .. will ask for comments first.

bastiandoetsch:
From the console output, I think it improves the UX. Regarding JSON output, how would it look like? IDEs assume a real file name in the JSON output, so navigation and in-editor highlighting can happen. This would most likely mean, that the displayTargetFile in the JSON output should not change.

dotkas:
Looking at the JSON output calling `--json` it seems `targetFile` is still retained in the original value, i.e.:

```json
    "packageManager": "nuget",
    "ignoreSettings": {
      "adminOnly": false,
      "reasonRequired": false,
      "disregardFilesystemIgnores": false
    },
    "summary": "No known vulnerabilities",
    "filesystemPolicy": false,
    "uniqueCount": 0,
    "targetFile": "obj/project.assets.json", <---
    "projectName": "runtime-deps-differ",
    "displayTargetFile": "obj/project.assets.json [net6.0]",
    "hasUnknownVersions": false,
    "path": "/Users/kasparmoss/git/playground/reproduce-user-errors/runtime-deps-differ"
  },
```

PR #4915, opened 2023-10-20
fix: use default HTTP headers consistently [HEAD-936]

PeterSchafer:

#### What does this PR do?
* update GAF to consistently use default HTTP headers
* make log level configurable to optionally increase information

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -77,12 +77,29 @@ const (
 	handleErrorUnhandled           HandleError = iota
 )
 
-func getDebugLogger(config configuration.Configuration) *zerolog.Logger {
+func getDebugLevel(config configuration.Configuration) zerolog.Level {
+	loglevel := zerolog.DebugLevel
+	if loglevelString := config.GetString("snyk_loglevel"); loglevelString != "" {
+		var err error
+		loglevel, err = zerolog.ParseLevel(loglevelString)
+		if err == nil {
+			debugLogger.Log().Msgf("Setting log level to %s", loglevelString)
+		} else {
+			debugLogger.Log().Msgf("%v", err)
+			loglevel = zerolog.DebugLevel
+		}
+	}
+	return loglevel
+}
+
+func initDebugLogger(config configuration.Configuration) *zerolog.Logger {
 	debug := config.GetBool(configuration.DEBUG)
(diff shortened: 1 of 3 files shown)


cliv2/cmd/cliv2/main.go
```diff
@@ -78,11 +78,18 @@ const (
...
+	} else {
+		if loglevelString := config.GetString("snyk_loglevel"); loglevelString != "" {
+			loglevel, _ = zerolog.ParseLevel(loglevelString)
+			debugLogger.Log().Msgf("Setting log level to %s", loglevelString)
+		}
 	}
 
+	debugLogger = debugLogger.Level(loglevel)
 	return &debugLogger
 }
```
cat2608, line 94: Do you think we could do an early return when `!debug` so we reduce one level of complexity? Something like:

```go
	debug := config.GetBool(configuration.DEBUG)

	if !debug {
		debugLogger = debugLogger.Output(io.Discard)
		return &debugLogger
	}
```

`nitpick`: do you think we could rename `getDebugLogger` to `getDebugLoggerLevel`?

`nitpick`: what id we don't ignore the `err` from `ParseLevel` and we `debugLogger.Log().Err(err).Msg("Failed to parse log level, using default debug level")`?

cat2608 (approved):

cliv2/cmd/cliv2/main.go
```diff
@@ -77,13 +77,31 @@ const (
...
 		debugLogger = debugLogger.Output(io.Discard)
+		return &debugLogger
+	} else {
+		loglevel := getDebugLevel(config)
+		debugLogger = debugLogger.Level(loglevel)
+		return &debugLogger
 	}
-
-	return &debugLogger
 }
```
cat2608, line 105: Great extraction :)

`nitpick`: What about:

```go
func initDebugLogger(config configuration.Configuration) *zerolog.Logger {
	debug := config.GetBool(configuration.DEBUG)
	if !debug {
		debugLogger = debugLogger.Output(io.Discard)
	} else {
		loglevel := getDebugLevel(config)
		debugLogger = debugLogger.Level(loglevel)
	}
	return &debugLogger
}
```

With your new solution we can have a single return 🙌

PeterSchafer: done!

PR #4916, opened 2023-10-24
fix: add `targetRuntime` meta information to monitor

dotkas:
The entire foundation for utilizing the `targetRuntime` exists in the Snyk backend, but is not utilized by `snyk monitor`.

Adding the `targetRuntime` will decorate the `snyk monitor` results with the Target Runtime information, used in multiple language ecosystems.

Here an example from .NET, before:
Screenshot 2023-10-24 at 21 58 53

After:
Screenshot 2023-10-24 at 21 58 56

This will also pave the way for support of `snyk monitor` of a single project with multiple target runtimes attached, without overwriting the same project multiple times.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,7 +74,7 @@
         "snyk-module": "3.1.0",
         "snyk-mvn-plugin": "3.1.0",
         "snyk-nodejs-lockfile-parser": "1.52.1",
-        "snyk-nuget-plugin": "1.39.1",
+        "snyk-nuget-plugin": "2.0.0",
         "snyk-php-plugin": "1.9.2",
         "snyk-policy": "^1.25.0",
         "snyk-python-plugin": "^2.0.1",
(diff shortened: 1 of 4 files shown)


alina-ilie (approved):

Avishagp (approved):

darscan (dismissed):
Blocking. We need to verify that this will not break existing pipelines (i.e. that tests map to the same projects that they used to, and that this won't create a flood of new projects in the system)

dotkas:
@darscan I already did that.

darscan:
So, you've manually tested every combination of cli-plugin and input argument, for every ecosystem against existing projects in the system?

dotkas:
@darscan [I checked which CLI plugin](https://github.com/search?q=org:snyk+targetRuntime+language:TypeScript&type=code&l=TypeScript) would populate `meta.targetRuntime`, could not find anything other than `snyk-nuget-plugin` that is used by the CLI currently. What have I missed?

dotkas:
@darscan it only applies to runs done with the beta-flag `--dotnet-runtime-resolution`, meaning existing .NET projects are not affected.

darscan:
Thanks, that clears things up a lot 👍

dotkas:
Disabling auto-merge and double-checking everything. @darscan made me paranoid.

darscan:
> it only applies to runs done with the beta-flag --dotnet-runtime-resolution, meaning existing .NET projects are not affected.

But what happens when that feature goes GA?

dotkas:
Continuing the discussion outside this PR.

PR #4934, opened 2023-11-16
fix: allow http requests towards code local engine [NEBULA-1541]

soniqua:
### What does this PR do?

- Allows requests made towards the Snyk Code Local Engine `/status` endpoint to respect the protocol of the Local Engine URL set on the org/group
- Fixes conditions where a Snyk Code Local Engine installation responds on both `http` and `https` and we want to query `http` only
- Allows certificate trust to remain intact

### DIFF
--- a/src/lib/plugins/sast/localCodeEngine.ts
+++ b/src/lib/plugins/sast/localCodeEngine.ts
@@ -1,11 +1,8 @@
 import * as debugLib from 'debug';
 import chalk from 'chalk';
-
 import { makeRequest } from '../../request';
-import { Global } from '../../../cli/args';
 import { SastSettings } from './types';
 
-declare const global: Global;
 const debug = debugLib('snyk-code');
 
 export function isLocalCodeEngine(sastSettings: SastSettings): boolean {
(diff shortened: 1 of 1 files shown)


shirlupo:
from the code seems like we already calling this function with baseURL
https://github.com/snyk/cli/blob/b2682f4d8b226f07f1badb831481926f762b57db/src/lib/plugins/sast/analysis.ts#L72
my assumption is we do not, 
can you maybe change this as well?

I am a bit confused what is happening here if no broker in the picture? are we just failing this as `Snyk Code Local Engine health is not ok` ?

shirlupo (approved):
Approved with some sug and a question

neonnoon:
I am not sure I understand what you mean with this @shirlupo:
> from the code seems like we already calling this function with baseURL
> my assumption is we do not,

Do you mean, why do we set a default parameter (`=''`)?
That is a fair question as this will clearly not work. So can `localEngineUrl` be undefined? In that case we'd have to properly handle it instead of defaulting to empty string.

> I am a bit confused what is happening here if no broker in the picture? are we just failing this as `Snyk Code Local Engine health is not ok` ?

There is no broker in the picture either way. This is CLI connecting directly to local engine.
Where would you expect broker (server or client?) to be involved?

neonnoon (commented):
I am not sure I understand what you mean with this @shirlupo:
> from the code seems like we already calling this function with baseURL
> my assumption is we do not,

Do you mean, why do we set a default parameter (`=''`)?
That is a fair question as this will clearly not work. So can `localEngineUrl` be undefined? In that case we'd have to properly handle it instead of defaulting to empty string.

> I am a bit confused what is happening here if no broker in the picture? are we just failing this as `Snyk Code Local Engine health is not ok` ?

There is no broker in the picture either way. This is CLI connecting directly to local engine.
Where would you expect broker (server or client?) to be involved?

shirlupo:
> I am not sure I understand what you mean with this @shirlupo:
> 
> > from the code seems like we already calling this function with baseURL
> > my assumption is we do not,
> 
> Do you mean, why do we set a default parameter (`=''`)? That is a fair question as this will clearly not work. So can `localEngineUrl` be undefined? In that case we'd have to properly handle it instead of defaulting to empty string.
> 


I meant the naming of the parameter 'baseUrl' 
https://github.com/snyk/cli/blob/b2682f4d8b226f07f1badb831481926f762b57db/src/lib/plugins/sast/analysis.ts#L72

from Matt and Matthew's code it seems like they parse the baseUrl so calling the parameter here baseURL is odd to me

_________

> > I am a bit confused what is happening here if no broker in the picture? are we just failing this as `Snyk Code Local Engine health is not ok` ?
> 
> There is no broker in the picture either way. This is CLI connecting directly to local engine. Where would you expect broker (server or client?) to be involved?


There is a call here to get the version by calling the health check no?

PR #4986, opened 2024-01-05
chore(ci): upload legacy tap test results to circle

jbowes:
This will allow for (probably) better parallelism, seeing what tests failed, and tracking flakey tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -832,9 +832,13 @@ jobs:
           command: node ./bin/snyk config set "api=${TEST_SNYK_TOKEN}" # many tests require the token to be in the config
       - run:
           name: Running Tap tests
-          command:
-            npx tap -j 1 -C --timeout=300 --node-arg=-r --node-arg=ts-node/register --allow-incomplete-coverage \
+          no_output_timeout: '30m' # the whole set of tests regularly takes 25+ minutes.
+          command: >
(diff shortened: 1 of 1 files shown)


jbowes:
After running this a bunch, here's the good and bad:

# Good
- IMO the `tap` output we have is unreadable for the number of tests we have. it's very hard to figure out what test actually failed in CI. storing the results in circle's test viewer makes it easier.
- tracking flakey tests and seeing execution times is very nice

# Bad
- Every test run prints an error, because the stock `tap` xunit reporter includes elements that circle doesn't like.
- zero output is sent to stdout/stderr, meaning circle has no idea about progress, and we have to bump up its wait time.

# Ugly
- The test splitting that was configured already for tap tests wasn't doing anything, and *can't* do anything with `tap`'s reporter, as filenames are no associated with the test output -- you don't get time-based splitting without result uploads that include timing *and* the test file name.

---

Overall I still think this is worthwhile, even with the display issues (ignored error, and no output), and it not being useful for parallelism. Though we could certainly find/write a better `tap` reporter...

thisislawatts (approved):

thisislawatts:
Looks good to me @jbowes, chatting with @PeterSchafer we are happy to take it for a spin and iterate.
Minor edit, adjusted to chore commit.

PR #4987, opened 2024-01-05
chore: don't checkout code if we have a prepared workspace

jbowes:
The workspace already has the code checked out, no need to do it, then
overwrite it with the workspace data.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -308,7 +308,6 @@ commands:
         type: string
         default: linux-base-edge-amd64
     steps:
-      - checkout
       - attach_workspace:
           at: .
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -758,12 +757,7 @@ jobs:
       - persist_to_workspace:
           root: .
           paths:
-            - binary-releases/*
```
PeterSchafer, line 761: Context: We tried to optimize the runtime by being very selective when persisting the workspace. We observed performance issues on Windows runners when restoring the workspace. This is why it was more efficient to checkout the code instead of persisting it in the workspace.

jbowes: time's a wash anyways 🤷 

any idea what caused the issues for windows?

jbowes:
gonna delete this, it saves a few seconds on restore, but adds a few to build, so it's pretty even.

.circleci/config.yml
PeterSchafer: > any idea what caused the issues for windows?
Unfortunately not! A wild guess would be, based on general observations on the windows runners, that they have a reduced performance when it comes to disk IO. Restoring the workspace is basically just an untar operation. Build and test duration are also not comparable to other runners.
But this is just guessing.

PR #4988, opened 2024-01-05
chore: upgrade conventional-changelog-cli to latest

jbowes:
This gets us away from a *very* old version of uuid in the dep tree, that
leaves a warning when installing. Less warnings are good.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -103,7 +103,7 @@
         "@typescript-eslint/parser": "^4.30.0",
         "ajv": "^6.12.6",
         "body-parser": "^1.19.0",
-        "conventional-changelog-cli": "^2.2.2",
+        "conventional-changelog-cli": "^4.1.0",
         "copy-webpack-plugin": "^9.0.1",
         "cross-spawn": "^6.0.5",
         "danger": "10.9.0",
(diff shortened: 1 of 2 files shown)


thisislawatts:
Initial investigation shows:
* We have deliberately pinned this to the current version: https://github.com/snyk/cli/pull/2693
* Used via `npx` in Makefile https://github.com/search?q=repo%3Asnyk%2Fcli%20conventional-changelog-cli&type=code
* Two sets of breaking changes which need to be verified https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/CHANGELOG.md

PeterSchafer:
I just looked into the points above.

> We have deliberately pinned this to the current version: https://github.com/snyk/cli/pull/2693

I think we are okay to upgrade, I don't see a value being stuck on the old version. Other than at some point Node version support might force us to upgrade.

> Used via npx in Makefile 

This works fine, the version specified in the package file is used by npx.

> Two sets of breaking changes which need to be verified 

The breaking changes seem to be limited to APIs and Runtime which are fine for us.

PeterSchafer (approved):

PR #4995, opened 2024-01-08
chore: tap knows about typescript now

jbowes:
... so we can remove the explicit import hook for it from test running


thisislawatts (approved):

thisislawatts:
Rebased to pull in updates and skip failing test cases.

dotkas (approved):

PR #5007, opened 2024-01-17
fix: Fix handling of large json data when writing to file via --json-file-output [CLI-46]

j-luong:
## Pull Request Submission

The pull request must:

- **Reviewer Documentation**
    - [x] follow [CONTRIBUTING](https://github.com/snyk/cli/blob/master/CONTRIBUTING.md) rules
    - [x] contain a link to the automatic tests that cover the updated functionality.
    - [x] contain testing instructions in case that the reviewer wants to manual verify as well, to add to the manual testing done by the author.
- **Testing**
    - [x] Changes, removals and additions to functionality must be covered by acceptance / integration tests or smoke tests - either already existing ones, or new ones, created by the author of the PR.

## Pull Request Review

## Pull Request Approval

by our content writers.

## What does this PR do?
This PR fixes an error occurring when using `--json-file-output` when writing large JSON objects into the specified file. The error surfaced as a exception with a stack trace similar to the one below

```
RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at Object.jsonStringifyLargeObject (/snapshot/project/dist/cli/webpack:/snyk/src/lib/json.ts:17:16)
    at Object.extractDataToSendFromResults (/snapshot/project/dist/cli/webpack:/snyk/src/lib/formatters/test/format-test-results.ts:105:27)
```

The root cause is a size limitation of Nodejs [V8 string size limit](https://github.com/nodejs/node/blob/fbf1fb3bed7015ed54c4019982d307f6cd302c4a/deps/v8/include/v8-primitive.h#L124-L127) when converting objects to strings using `JSON.stringify` in [json.ts](https://github.com/snyk/cli/blob/master/src/lib/json.ts).

The solution introduced in this PR uses a streaming mechanism to work around this limitation.

In order to minimise risk, the new streaming implementation is used only when:
- the originial `JSON.stringify` fails in [json.ts](https://github.com/snyk/cli/blob/master/src/lib/json.ts).
- only for `--json-file-output`

The existing behaviour when `JSON.stringify` fails is to throw a `RangeError` error, and occurs when using the `--json` and `--json-file-output` flags. 

This PR should also provide a better error message, instead of throwing a `RangeError`, when `JSON.stringify` fails with the `--json` flag.

For the above reasons, the **risk is assumed to be LOW**

Looking at the acceptance test [test --json-file-ouput handles responses larger than 512Mb string size limit in v8](https://github.com/snyk/cli/blob/622389afb1ad8608736aae849e47e3544501c227/test/jest/acceptance/cli-json-file-output.spec.ts#L114) to get an idea of what causes the issue.

It can be tested via building a local version of the CLI binary:
- checkout branch
- `make clean build`
- run local binary `./binary-releases/<local-binary> container test --json-file-output=./output.json <image:tag>`

It can also be quickly tested by running the legacy CLI as the changes are there:
- checkout branch
- from project root run: `npx ts-node ./src/cli/index.ts container test --json-file-output=./output.json <image:tag>`

Manually testing the fix will be difficult without manipulating the JSON object in code, as it depends on the image having an extremely high number of issues. The acceptance test `test --json-file-ouput handles responses larger than 512Mb string size limit in v8` shows how to do this if desired.


src/lib/json.ts
```diff
@@ -1,20 +1,25 @@
...
   let res = '';
   try {
-    // first try pretty-print
     res = JSON.stringify(obj, null, 2);
     return res;
   } catch (err) {
-    // if that doesn't work, try non-pretty print
-    debug('JSON.stringify failed - trying again without pretty print', err);
-    res = JSON.stringify(obj);
+    if (options?.json) {
```
PeterSchafer, line 16: Issue: From my understanding I think we need the non-pretty-print part of the original solution as well and only if this fails as well, we should fallback on the new solution. Removing the original code will cause issues for use cases of --json, which worked due to the non-pretty-print part.

PeterSchafer: The associated unit test needs to be fixed and an additional one needs to be created.

test/jest/unit/json.spec.ts
```diff
@@ -24,7 +24,7 @@ describe('jsonStringifyLargeObject', () => {
       });
 
     const s = jsonStringifyLargeObject(largeObject);
-    expect(jsonStringifyMock).toHaveBeenCalledTimes(2);
-    expect(s).toEqual(`{"name":"Brian","isGoodBoy":true,"type":"big"}`);
+    expect(jsonStringifyMock).toHaveBeenCalledTimes(1);
```
PeterSchafer, line 27: As mentioned before, this test should remain unchanged and a new one should be added.

test/jest/unit/json-file-output.spec.ts
```diff
@@ -48,3 +49,58 @@ describe('writeContentsToFileSwallowingErrors', () => {
...
+ * Format bytes as human-readable text.
+ *
+ * @param bytes Number of bytes.
+ * @param si True to use metric (SI) units, aka powers of 1000. False to use
+ *           binary (IEC), aka powers of 1024.
+ * @param dp Number of decimal places to display.
+ *
+ * @return Formatted string.
+ */
+function humanFileSize(bytes, si = false, dp = 1) {
```
PeterSchafer, line 84: Question: This function is now twice in the code, could we please move this to a helper file and avoid the duplication?

test/jest/acceptance/cli-json-file-output.spec.ts
```diff
@@ -109,4 +110,74 @@ describe('test --json-file-output', () => {
...
+ * Format bytes as human-readable text.
+ *
+ * @param bytes Number of bytes.
+ * @param si True to use metric (SI) units, aka powers of 1000. False to use
+ *           binary (IEC), aka powers of 1024.
+ * @param dp Number of decimal places to display.
+ *
+ * @return Formatted string.
+ */
+function humanFileSize(bytes, si = false, dp = 1) {
```
PeterSchafer, line 161: Same as the other comment, please move this in a helper file.

src/lib/json.ts
```diff
@@ -1,20 +1,25 @@
...
+
 const debug = require('debug')('snyk-json');
 
 /**
  * Attempt to json-stringify an object which is potentially very large and might exceed the string limit.
  * If it does exceed the string limit, try again without pretty-print to hopefully come out below the string limit.
  * @param obj the object from which you want to get a JSON string
  */
-export function jsonStringifyLargeObject(obj: any): string {
+export function jsonStringifyLargeObject(obj: any, options?: Options): string {
```
PeterSchafer, line 10: Question: Could we avoid passing in the options and move the check outside, where it is being invoked? For example by checking if the returned value is an empty string. To me it just feels like that this function shouldn't be responsible for showing a warning message related to a cmd flag.

src/lib/formatters/test/format-test-results.ts
```diff
@@ -102,7 +102,7 @@ export function extractDataToSendFromResults(
 
   let stringifiedJsonData = '';
   if (options.json || options['json-file-output']) {
-    stringifiedJsonData = jsonStringifyLargeObject(jsonData);
+    stringifiedJsonData = jsonStringifyLargeObject(jsonData, options);
```
PeterSchafer, line 105: Related to my other comment, it feels more clear to not pass the options and move the logic around the warning up here.

src/cli/commands/test/index.ts
```diff
@@ -214,6 +214,7 @@ export default async function test(
         stringifiedData,
         stringifiedJsonData,
         stringifiedSarifData,
+        stringifiedJsonData.length === 0 ? dataToSend : null,
```
PeterSchafer, line 217: Suggestion: since this logic appears here multiple times, why don't we move it inside of createJsonTestCommandResult(), or even simpler, can't we just always assign `dataToSend`?

test/jest/acceptance/cli-json-file-output.spec.ts
```diff
@@ -109,4 +110,74 @@ describe('test --json-file-output', () => {
...
+      {
+        cwd: project.path(),
+        env,
+      },
+    );
+
+    console.debug({ stdout, stderr });
+
+    const outputPath = await project.path(outputFilename);
+    const fileSize = fs.statSync(outputPath).size;
```
PeterSchafer, line 140: Question: maybe should we validate that the file content is a valid json?

src/cli/main.ts
```diff
@@ -44,6 +47,7 @@ import { SarifFileOutputEmptyError } from '../lib/errors/empty-sarif-output-erro
 import { InvalidDetectionDepthValue } from '../lib/errors/invalid-detection-depth-value';
 import { obfuscateArgs } from '../lib/utils';
 import { EXIT_CODES } from './exit-codes';
+const isEmpty = require('lodash/isEmpty');
```
thisislawatts, line 50: **nitpick (non-blocking):** Can we align on using imports rather than requires? Marking this as non-blocking as I ideally this is something picked up by tooling. Likely a bunch of other places across the code base this needs to be resolved.

src/cli/commands/test/index.ts
j-luong: We can pass `dataToSend` as we already check the payload in `main.ts > saveJsonResultsToFile`, will update and test to confirm.

src/lib/json.ts
j-luong: updated

test/jest/acceptance/cli-json-file-output.spec.ts
j-luong: This can be done at the unit test level in `json-file-output.spec.ts`

src/cli/main.ts
j-luong: This is actually a[ linting rule](https://github.com/snyk/cli/blob/279677acf1b7f596e1916d68afeb9c8cc4803f6d/.eslintrc.json#L35) specifically for lodash. Not sure why it's there, but didn't want to change it in this PR.

src/cli/commands/test/index.ts
j-luong: refactored to set `stringifiedJsonData.length === 0 ? dataToSend : null` as a const. The way the CLI handles finding issues vs finding no issues means passing `dataToSend` is not as simple to implement

src/cli/main.ts
```diff
@@ -157,8 +162,18 @@ async function handleError(args, error) {
...
 
-  await saveResultsToFile(args.options, 'json', error.jsonStringifiedResults);
-  await saveResultsToFile(args.options, 'sarif', error.sarifStringifiedResults);
+  if (error.jsonPayload) {
+    // send raw jsonPayload instead of stringified payload
+    await saveResultsToFile(args.options, 'json', '', error.jsonPayload);
+  } else {
+    // fallback to original behaviour
+    await saveResultsToFile(args.options, 'json', error.jsonStringifiedResults);
+    await saveResultsToFile(
```
PeterSchafer, line 171: Question: Don't we need to move this out of the else case?

j-luong: `await saveResultsToFile(args.options, 'json', error.jsonStringifiedResults);` maintains the original behaviour for cases when `--json || --json-file-output` works using `JSON.stringify`. 

When/if we decide to stream to writeable destinations instead of using `JSON.stringify` all the branching logic checks like we have here can go away.

PeterSchafer (approved):

PR #5010, opened 2024-01-18
fix: SARIF: artifactChanges property cannot be empty

schottsfired:
We received a report that `artifactChanges: []` causes an error in GitHub:
> Unable to upload "xyz.sarif" as it is not valid SARIF:
> - instance.runs[2].results[0].fixes[0].artifactChanges does not meet minimum length of 1

cf. https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791134

This fix populates `artifactChanges` so it's not empty.


cmars (approved):
LGTM!

bastiandoetsch (approved):

PR #5022, opened 2024-01-26
chore: all tests should run against MT-EU

Avishagp:
Add debug flag to make sure they do


PeterSchafer:
> * Do we need to run both `whoami` and `woof` to verify the API endpoint?
Yes and no
* `whoami` is used to test the network stack (eg. for FIPS)
* `woof` is used to execute the legacy CLI and ensure it can be properly launched.

The tests are simple smoke-test like tests, that are being run in plenty of environments to catch issues that come from the environment. We could of course run another command, but both of these are simple and are low maintenance.
I hope this makes sense :)

cmars:
Makes a lot of sense @PeterSchafer , thanks for the context!

PR #5031, opened 2024-02-02
chore: add npm dev script

cmars:
To aid in debugging and development of the Typescript parts of the CLI, this adds an npm script entrypoint to the CLI.

This doesn't run the complete CLI but may be useful for isolating and debugging Typescript CLI issues.

This is not a user-facing change.

Usage:

`npm run dev test` to run the equivalent of `snyk test` from Typescript source.

`npm run dev -- -d` to pass the debug `-d` flag to the Typescript main entrypoint (and not `npm run`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ clean-prepack: clean-package-files
 	@rm -f prepack
 
 .PHONY: clean-ts
-clean-ts: 
+clean-ts:
 	@npm run clean
 	@rm -f -r $(BINARY_RELEASES_FOLDER_TS_CLI)
 
@@ -183,7 +183,7 @@ $(BINARY_WRAPPER_DIR)/src/generated/sha256sums.txt:
 build-binary-wrapper: pre-build-binary-wrapper $(BINARY_WRAPPER_DIR)/src/generated/version $(BINARY_WRAPPER_DIR)/src/generated/sha256sums.txt
 	@echo "-- Building Typescript Binary Wrapper ($(BINARY_WRAPPER_DIR)/dist/)"
 	@cd $(BINARY_WRAPPER_DIR); npm run build
-	
+
 .PHONY: clean-binary-wrapper
 clean-binary-wrapper:
 	@rm -f $(BINARY_WRAPPER_DIR)/config.default.json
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,7 +43,8 @@
     "test:unit": "jest --runInBand --testPathPattern '/test(/jest)?/unit/'",
     "test:acceptance": "jest --runInBand --testPathPattern '/test(/jest)?/acceptance/'",
     "test:tap": "tap -Rspec --timeout=300 test/tap/*.test.* ",
-    "test:smoke": "./scripts/run-smoke-tests-locally.sh"
+    "test:smoke": "./scripts/run-smoke-tests-locally.sh",
+    "dev": "ts-node ./src/cli/index.ts"
```
PeterSchafer, line 47: Suggestion: In order to facilitate the usage of the Makefile, how about adding this to the Makefile? For example like this:
```
make run // runs the Extensible CLI
make run-ts // runs only the TS/Legacy CLI
```

PeterSchafer: The final goal could be to not be required to use `npm` calls anymore.

cmars: Added. Both targets check for an ARGS variable, which is used to specify the CLI arguments through `make`.

cmars: My preference would be to keep the `npm` script, but also wrap it with make. The `npm` script is easier to use with a debug launch profile in VSCode, and it inherits the right $PATH.

Makefile
```diff
@@ -273,3 +273,17 @@ format:
...
 	@echo "-- Generating protocol metadata"
 	@pushd $(EXTENSIBLE_CLI_DIR) && $(MAKE) generate-ls-protocol-metadata bindir=$(WORKING_DIR)/$(BINARY_OUTPUT_FOLDER) && popd
+
+.PHONY: run
+run: build require-args
+	$(wildcard $(WORKING_DIR)/$(BINARY_OUTPUT_FOLDER)/snyk-*) $(ARGS)
+
+.PHONY: run-ts
+run-ts: require-args
+	npm run dev "$(ARGS)"
```
PeterSchafer, line 283: Question: Would it make sense to add `npm run build`?

package.json
```diff
@@ -172,7 +173,7 @@
     "tap": "^18.5.0",
     "ts-jest": "^27.0.4",
     "ts-loader": "^9.0.2",
-    "ts-node": "^8.0.0",
+    "ts-node": "^8.10.2",
```
PeterSchafer, line 176: Question: was this required?

Makefile
cmars: It shouldn't be necessary, as `ts-node` runs the CLI directly from source.

`npm install` would be worth running though, to ensure there's a `node_modules`.

package.json
cmars: I'll see if I can revert this...

PeterSchafer (approved):

PR #5050, opened 2024-02-09
fix: upgrade LS

acke:
## Pull Request Submission

We have a new release of Language Server here:
https://github.com/snyk/snyk-ls/releases/tag/v20240209.115701

## What does this PR do?

IDE-24 Disable Tracking via Segment and Amplitude for NON-MT-US customers


thisislawatts:
**issue(blocking):** @acke would you be able to adjust the commit message so that we communicate the customer value? We use the commit messages to generate our release notes, see here: https://github.com/snyk/cli/releases

Suggestion:
_fix: Disable tracking via Segment and Amplitude for customers hosted outside US_

PeterSchafer (approved):

PR #5066, opened 2024-02-22
re org

bikochan:
- **feat: update team and context for ci and metadata**
- **feat: re-org labels and security gates**


j-luong (changes requested):
issue: 
- could we re-add the PR template and fill it out?
- could we update the PR title from `re org` to something more descriptive? e.g. `chore: update ignores and security scans workflow for CICD`. These are used in the release notes so it's useful to have something informative here
- looks like the build is failing on lint issues, can we run `npm run format` to fix this?

thisislawatts:
@j-luong we already have an ignore for SNYK-JS-SEMVER-3247795. Is there a problem with this not being picked up correctly? 

```
  SNYK-JS-SEMVER-3247795:
    - '*':
        reason: No newer versions to upgrade to
        expires: 2024-09-13T00:00:00.000Z
        created: 2023-09-13T14:33:35.717Z
```
https://github.com/snyk/cli/blob/main/ts-binary-wrapper/.snyk

PeterSchafer:
> @j-luong we already have an ignore for SNYK-JS-SEMVER-3247795. Is there a problem with this not being picked up correctly?
> 
> ```
>   SNYK-JS-SEMVER-3247795:
>     - '*':
>         reason: No newer versions to upgrade to
>         expires: 2024-09-13T00:00:00.000Z
>         created: 2023-09-13T14:33:35.717Z
> ```
> 
> https://github.com/snyk/cli/blob/main/ts-binary-wrapper/.snyk

The integration of the prodsec orb specifies `--policy-path=.snyk` which overrides the automatic detection of the .snyk files in subfolders.  This is why the ignore needs to be additionally done in the root .snyk file now. I will push up a fix now and remove the additional ignore.

@bikochan any thoughts?

PeterSchafer:
In addition, I think this PR should remove/replace the pre-existing snyk scans.

PeterSchafer:
@j-luong @bikochan  I adapted this PR now, to what I think it should be. please review the changes and lift any change request if okay.

.snyk
```diff
@@ -17,6 +17,26 @@ ignore:
         reason: --about lists all dependency licenses which is a requirement of MPL-2.0
         expires: 2122-12-14T16:35:38.252Z
         created: 2022-11-14T16:35:38.260Z
+  SNYK-JS-AXIOS-6032459:
```
j-luong, line 20: question: is it better to try and fix these vulns than to create a temp ignore for the PR? IIRC, these vulns will block other PRs once this workflow is merged in.

Also, is the `expires: 2024-03-05T00:00:00.000Z` correct? It's in the past.

PeterSchafer: I removed them, they are not high so they don't appear.

.circleci/config.yml
```diff
@@ -906,18 +906,14 @@ jobs:
           name: Linting project
           command: |
             npm run lint
-            pushd cliv2
+            pushd cliv2
```
j-luong, line 909: nitpick: I think you added a whitespace here

PeterSchafer: `npm run format` did

.circleci/config.yml
```diff
@@ -906,18 +906,14 @@ jobs:
           name: Linting project
```
j-luong, line 906: question: is there any reason we can't/shouldn't move the code-analysis (or at least the prodsec checks part of it) to the start of the CICD job, in parallel with the `secrets-scan`. This was we can fail fast.

thisislawatts: This sounds like a good follow up PR, I think there are loads of interesting changes we could make to the CI/CD pipeline to speed up the feedback loop.

thisislawatts (approved):

PR #5082, opened 2024-02-29
fix: enables multi-platform support for OCI images

pecodez:
## What does this PR do?

Enables multi-platform support for OCI images.

Using Docker engine 25.x test an OCI image with a non-standard platform...

`snyk container test busybox:latest --platform=linux/arm64`

Docker engine 25.x now creates an archive with a compliant OCI layout when running `docker save`. We previously only supported `linux/amd64` for OCI layouts.

https://github.com/snyk/snyk-docker-plugin/pull/571

https://snyksec.atlassian.net/browse/SUP-2577
https://snyksec.atlassian.net/browse/SUP-2579
https://snyksec.atlassian.net/browse/SUP-2594
https://snyksec.atlassian.net/browse/SUP-2595
https://snyksec.atlassian.net/browse/SUP-2630

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,7 +68,7 @@
         "semver": "^6.0.0",
         "snyk-config": "^5.0.0",
         "snyk-cpp-plugin": "2.24.0",
-        "snyk-docker-plugin": "^6.10.2",
+        "snyk-docker-plugin": "6.10.3",
         "snyk-go-plugin": "1.23.0",
         "snyk-gradle-plugin": "4.1.0",
         "snyk-module": "3.1.0",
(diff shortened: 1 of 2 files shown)


thisislawatts:
We use the commit messages to generate our release notes. Could you update the message to briefly explain the changes you made in a way that's easy to understand for users who might not be familiar with the code itself?

For example, instead of `fix: bumps snyk-docker-plugin to v6.10.3` version you could say `fix: adds support for multi-platform for OCI images`. I don't have full context here but does this change represent a bug fix/regression or should it instead use `feat:` because we are introducing new behaviour? 

This will help us create even better release notes for everyone!

cmars (approved):
LGTM with @thisislawatts comment addressed.

PR #5093, opened 2024-03-06
fix: Fix handling of large json data when writing to file via --json [CLI-73]

j-luong:
## Pull Request Submission

The pull request must:

- **Reviewer Documentation**
    - [x] follow [CONTRIBUTING](https://github.com/snyk/cli/blob/master/CONTRIBUTING.md) rules
    - [x] contain a link to the automatic tests that cover the updated functionality.
    - [x] contain testing instructions in case that the reviewer wants to manual verify as well, to add to the manual testing done by the author.
- **Testing**
    - [x] Changes, removals and additions to functionality must be covered by acceptance / integration tests or smoke tests - either already existing ones, or new ones, created by the author of the PR.

## Pull Request Review

## Pull Request Approval

by our content writers.

## What does this PR do?
This PR fixes an error occurring when using `--json` when writing large JSON objects to stdout. The error surfaced as a exception with a stack trace similar to the one below

```
RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at Object.jsonStringifyLargeObject (/snapshot/project/dist/cli/webpack:/snyk/src/lib/json.ts:17:16)
    at Object.extractDataToSendFromResults (/snapshot/project/dist/cli/webpack:/snyk/src/lib/formatters/test/format-test-results.ts:105:27)
```

The root cause is a size limitation of Nodejs [V8 string size limit](https://github.com/nodejs/node/blob/fbf1fb3bed7015ed54c4019982d307f6cd302c4a/deps/v8/include/v8-primitive.h#L124-L127) when converting objects to strings using `JSON.stringify` in [json.ts](https://github.com/snyk/cli/blob/master/src/lib/json.ts).

The solution introduced in this PR uses a streaming mechanism to work around this limitation.

In order to minimise risk, the new streaming implementation was successfully implemented for `--json-file-output` first #5007. Additionally, the streaming implementation is only used when
- the original `JSON.stringify` fails
- when `test` finds issues, if there are no issues the streaming implementation is not used as we know the size of the JSON output depends on finding vulnerabilities

The existing behaviour when `JSON.stringify` fails is to throw a `RangeError` error, and occurs when using the `--json` and `--json-file-output` flags. 

For the above reasons, the **risk is assumed to be LOW**

Looking at the acceptance test [test --json handles responses larger than 512Mb string size limit in v8](https://github.com/snyk/cli/blob/a55056cca13d06e49d2b06daae954ecdeb3ed078/test/jest/acceptance/cli-json-output.spec.ts#L80) to get an idea of what causes the issue.

It can be tested via building an edited local version of the CLI binary:
- checkout branch
- modify the code to ensure we always set the `jsonPayload` [here](https://github.com/snyk/cli/blob/a760a00eebacd5ef96f740d0cc0472e63e9e3fbf/src/cli/commands/test/index.ts#L256-L259) so that we trigger the `jsonStreamify` option [here](https://github.com/snyk/cli/blob/e2d1ed5ff48eef7fe45841fce0259244e0afb3b6/src/cli/main.ts#L148) to stream JSON rather than stringify it
    - [optional] add some `debug` logs to validate that we're using the `jsonStreamify` option
- `make clean build`
- run local binary `./binary-releases/<local-binary> container test <image:tag> --json -d`

It can also be quickly tested by running the legacy CLI as the changes are there:
- checkout branch
- from project root run: `npx ts-node ./src/cli/index.ts container test <image:tag> --json -d`

Manually testing the fix will be difficult without manipulating the JSON object in code, as it depends on the image having an extremely high number of issues. The acceptance test `test --json handles responses larger than 512Mb string size limit in v8` shows how to do this if desired.


src/lib/json.ts
```diff
@@ -11,14 +11,7 @@ export function jsonStringifyLargeObject(obj: any): string {
...
-    try {
-      // if that doesn't work, try non-pretty print
-      debug('JSON.stringify failed - trying again without pretty print', err);
-      res = JSON.stringify(obj);
-      return res;
-    } catch (error) {
-      debug('jsonStringifyLargeObject failed: ', error);
-      return res;
-    }
+    debug('jsonStringifyLargeObject failed: ', err);
```
PeterSchafer, line 14: Question: In order to reduce the impact of the complete change, I'm wondering, if we should remove the pretty print and make non-pretty print the default, as it'll safe characters from the start and make less cases actually trigger the new implementation.

src/lib/request/request.ts
```diff
@@ -139,15 +140,15 @@ export async function makeRequest(
   payload: Payload,
 ): Promise<{ res: needle.NeedleResponse; body: any }> {
   const { method, url, data, options } = setupRequest(payload);
-  debug(data);
+  // debug(data);
```
PeterSchafer, line 143: Question: shall we delete this?

test/jest/util/runCommand.ts
```diff
@@ -1,13 +1,16 @@
 import { SpawnOptionsWithoutStdio } from 'child_process';
 import { spawn } from 'cross-spawn';
+// const debug = require('debug')('acceptance-test:runCommand');
```
PeterSchafer, line 3: Question: shall we remove this?

test/jest/util/runCommand.ts
```diff
@@ -1,13 +1,16 @@
...
 type RunCommandResult = {
   code: number;
   stdout: string;
   stderr: string;
+  stdoutBuffer?: Buffer;
+  stderrBuffer?: Buffer;
 };
 
-type RunCommandOptions = SpawnOptionsWithoutStdio;
+type RunCommandOptions = SpawnOptionsWithoutStdio & { bufferOutput?: boolean };
```
PeterSchafer, line 13: Nitpick: a short comment what the new flag does would be helpful to future us.

src/lib/json.ts
j-luong: I've no strong opinion either way, but I would guess the end-user would prefer a nicely formatted JSON output, and indeed would expect it to be nicely formatted as it is the default behaviour.

If we're OK changing the default JSON formatting, then we can use unformatted JSON output as default instead of formatted.

PeterSchafer: Your answer makes me want to withdraw my initial question 😅 
I came from the place that json output most probably will be consumed by non humans, so why bother formatting. But maybe the assumption is not correct. So let's not change the default behaviour.

test/jest/acceptance/cli-json-output.spec.ts
```diff
@@ -76,4 +76,27 @@ describe('test --json', () => {
...
+      code,
+      stdoutBuffer,
+    } = await runSnykCLI(`test --json`, {
+      cwd: project.path(),
+      env,
+      bufferOutput: true,
+    });
+
+    expect(code).toEqual(1);
+    expect(stdoutBuffer).toBeDefined();
```
PeterSchafer, line 100: Suggestion: This test should ensure that the content is a json data structure, for example by using a library like [this](https://www.npmjs.com/package/jsonparse). It should also ensure that the length of stdoutBuffer is greater 0.

.circleci/config.yml
```diff
@@ -48,7 +49,7 @@ executors:
     docker:
       - image: bastiandoetsch209/cli-build:20240214-145818
     working_directory: /mnt/ramdisk/snyk
-    resource_class: large
+    resource_class: xlarge
```
thisislawatts, line 52: **Issue (blocking):** This will increase the resource size for **every** job that uses this executor, as we only need the additional resources to support the test cases lets override the resource_class only for the `acceptance-test` jobs that require it.

test/jest/acceptance/cli-json-output.spec.ts
j-luong: updated tests to better check this.

test/jest/util/runCommand.ts
```diff
@@ -5,9 +5,13 @@ type RunCommandResult = {
   code: number;
   stdout: string;
   stderr: string;
+  stdoutBuffer?: Buffer;
```
PeterSchafer, line 8: Suggestion: This seems unused, shall we remove it?!

src/lib/request/request.ts
```diff
@@ -44,9 +45,9 @@ function setupRequest(payload: Payload) {
 
   if (body && !noCompression) {
     debug('compressing request body');
-    const json = JSON.stringify(body);
+    const json = jsonStringifyLargeObject(body);
```
PeterSchafer, line 48: Suggestion: Let's revert some of the changes here, since it is better to fail than send an empty string.

src/lib/request/request.ts
```diff
@@ -44,9 +45,9 @@ function setupRequest(payload: Payload) {
...
     if (json.length < 1e4) {
-      debug(JSON.stringify(body, null, 2));
+      debug(jsonStringifyLargeObject(body));
```
PeterSchafer, line 50: Suggestion: let's re-use the `json` variable to save some memory.

thisislawatts (approved):

src/cli/commands/test/index.ts
```diff
@@ -252,6 +252,10 @@ export default async function test(
     err.json = stringifiedData;
     err.jsonStringifiedResults = stringifiedJsonData;
     err.sarifStringifiedResults = stringifiedSarifData;
+    // set jsonPayload if we failed to stringify it
+    if (stringifiedJsonData.length === 0) {
+      err.jsonPayload = dataToSend;
+    }
```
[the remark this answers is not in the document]
PeterSchafer: I removed the duplicated logic.
I realized that `err.jsonPayload` was already there, so I'm not changing this.

PeterSchafer (approved):

PR #5097, opened 2024-03-08
chore(ci): introduce parallel test runs for our acceptance tests

thisislawatts:
## Pull Request Submission

The pull request must:

- **Reviewer Documentation**
    - [X] follow [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) rules
    - [X] be accompanied by a detailed description of the changes
- **Testing**

## What does this PR do?

This PR implements CircleCI parallelism for all acceptance test suites, significantly reducing the duration for the _acceptance-tests windows amd64_ suite from 41 minutes (p95 last 90 days) to ~14 minutes. 

* Reviewers should start by examining the changes made to the CircleCI configuration files to implement parallelism for acceptance test suites.
* https://github.com/snyk/cli/pull/5097/commits/76fa2da483d054a5b511bdcfb790adcd51912da4 Remove hard coded port from tests which introduced issues when parallelising them. 
* Making use of recently upgraded jest dep. and [adopted `--shard` flag](https://jestjs.io/blog/2022/04/25/jest-28#sharding-of-test-run). 
* Note that parallelism has been introduced as a configurable parameter, this allows us to increase it further for windows whilst balancing against cost for the other platforms. 

Review the pipelines associated with this PR. 

Previously, the acceptance-tests windows amd64 suite had a duration of 41 minutes. By adopting CircleCI parallelism for all acceptance test suites, we've managed to reduce the duration to 14 minutes, significantly improving our testing efficiency.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -585,6 +585,7 @@ workflows:
           test_snyk_command: binary-releases\\snyk-win.exe
           install_deps_extension: windows-full
           dont_skip_tests: 0
+          shards: 4
           pre_test_cmds: Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1; RefreshEnv
 
       - sign:
@@ -1028,7 +1029,11 @@ jobs:
       pre_test_cmds:
         type: string
         default: 'echo Running tests'
+      shards:
+        type: integer
+        default: 3
     executor: << parameters.executor >>
+    parallelism: << parameters.shards >>
     steps:
       - prepare-workspace
       - install-deps-<< parameters.install_deps_extension >>
@@ -1039,7 +1044,7 @@ jobs:
           no_output_timeout: 30m
           command: |
             << parameters.pre_test_cmds >>
-            npm run test:acceptance -- --selectProjects coreCli
+            npm run test:acceptance -- --selectProjects coreCli --shard=$(expr $CIRCLE_NODE_INDEX + 1)/<< parameters.shards >>
           environment:
             TEST_SNYK_FIPS: << parameters.fips >>
             TEST_SNYK_COMMAND: << parameters.test_snyk_command >>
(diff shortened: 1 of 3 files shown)


scripts/capture-used-ports.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
PeterSchafer, line 1: Question: I appreciate the need of having a better solution for the ports used on tests. The question for me is if this needs to be combined with the parallelization of tests. As long as the tests in one shard run synchronously, the previous hard-coded port should be fine. The reason I ask, is that the solution implemented here seems quite complex and I wonder if we actually need this.

PeterSchafer: I can expand a bit. As far as I see, there is a timing aspect involved here. The port state at the point in time when the file is populated and the time when random values are being filtered based on the file. It is well possible that new ports get used in between these points in time, which is why I wonder if a random value with a retry wouldn't be a simpler and more robust solution.

cmars: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use is the issue here. It's unlikely in a container but stranger things can happen over time that might confound future us.

Binding to port 0 is probably the best way to get a unique port, but this would be hard to implement in the existing codebase. Here's another idea though:

Reserve the port range in advance with `sysctl`, https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html#ip-variables, see `ip_local_reserved_ports`. This will prevent the ports from being automatically assigned.

Then use the `--sysctl` argument to `docker run`.

This might require more changes to how the executor is set up in CircleCI.

cmars: The "try and increment" approach @PeterSchafer also SGTM. The potential payoff in doing this (either approach) I see, is that it could potentially save us CircleCI credits if we can push more of the concurrency into an executor, rather than only scaling by adding more executor instances.

PeterSchafer: Concurrence inside of an executor might have multiple issues to address besides port usage. There will occur issues in the usage of shared filesystem resources, like the config file. Some tests need to change the config file and others might start to fail randomly. 
Parallelizing via the Executors is a first good approach and their isolation helps with the other topics of shared resources.

PeterSchafer (approved):

PeterSchafer:
Very nice!

PR #5168, opened 2024-04-16
chore: Improvements for local build process

mgyorke:
## Pull Request Submission

The pull request must:

- **Reviewer Documentation**
    - [ ] follow [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) rules
    - [ ] be accompanied by a detailed description of the changes
- **Testing**

## Pull Request Review

## Pull Request Approval

by our content writers.

## What does this PR do?

This PR simplifies the local development setup by automating dependency installation for Python and adding specific targets for local build.

**Changes:**

- **Dependency Management:** Updated the install-dev-dependencies.sh script to create a .venv and install required Python dependencies at the project root.

- **Makefile Targets:** Introduced two new Makefile targets:
  `make install-deps`Automates dependency setup, wraps the installation script
  `make build-local` Triggers the local build process using the local .venv

- **Documentation:** Updated CONTRIBUTING.md with clear instructions for local development setup and usage of the new targets.

scripts/install-dev-dependencies.sh
MakeFile

`make install-deps` it shouldn't fail even if the dependencies are already installed.
Check if the .venv is created and if it contains the dependencies.
`make build-local` it should prepare the build as usual

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ Clone this repository with git.
 
 ```sh
 git clone git@github.com/snyk/cli.git
-cd snyk
+cd cli
 ```
 
 You will now be on our `main` branch. You should never commit to this branch, but you should keep it up-to-date to ensure you have the latest changes.
@@ -35,16 +35,25 @@ make clean
 
 ## Building
 
-To build the project, run the following command in the root of the repository.
+To build the project, run the following commands in the root of the repository.
+
+The _install-dev-dependencies.sh_ script created a venv with the necessary dependencies, we need to activate it:
+
+```sh
+source .venv/bin/activate
+```
+
+Now we trigger the build process:
 
 ```sh
 make build
 ```
 
-Run the build binary like this.
+Run the build binary like this, depending on your architecture:
 
 ```sh
 ./binary-releases/snyk-macos --version
+./binary-releases/snyk-macos-arm64 --version
 ```
 
 ## Running tests
@@ -114,6 +123,14 @@ You can run acceptance tests with:
 
 ```
 npm run test:acceptance -- --selectProjects coreCli
+```
+
+The output is saved as _junit.xml_ inside the root folder.
+
+Another example, if we want the output in console and to run a single test suite by name:
+
+```
+npm run test:acceptance -- --selectProjects coreCli --reporters=default -t 'woof'
 ```
 
 ### Smoke Tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ To install the required development dependencies in homebrew based environments,
 The only additional prerequisite is having [homebrew](https://brew.sh/) installed.
 
 ```sh
-./scripts/install-dev-dependencies.sh
+make install-deps
 ```
 
 ## Setting up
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -288,3 +288,16 @@ require-args:
...
+# targets responsible for local dev environment
+# install local dependencies
+install-deps:
+	@bash $(WORKING_DIR)/scripts/install-dev-dependencies.sh
+
+# trigger build using the local .venv for python
+build-local: pre-build
+	@( \
+    	source $(WORKING_DIR)/.venv/bin/activate; \
+    	cd $(EXTENSIBLE_CLI_DIR); $(MAKE) build-full install bindir=$(WORKING_DIR)/$(BINARY_OUTPUT_FOLDER) USE_LEGACY_EXECUTABLE_NAME=1; \
```
PeterSchafer, line 301: Question: Do we really need a differentiation between build and build-local? I'm worried that local and CI build get out of sync at some point. I would like to see everyone using the same command to build.

CONTRIBUTING.md
```diff
@@ -37,14 +37,17 @@ make clean
 
 To build the project, run the following command in the root of the repository.
 
+Now we trigger the build process:
```
PeterSchafer, line 40: Question: Isn't this what the previous line says? Do you really want to keep both? Maybe merge their content?

CONTRIBUTING.md
```diff
@@ -116,6 +119,14 @@ You can run acceptance tests with:
 npm run test:acceptance -- --selectProjects coreCli
 ```
 
+The output is saved as _junit.xml_ inside the root folder.
+
+Another example, if we want the output in console and to run a single test suite by name:
+
+```
+npm run test:acceptance -- --selectProjects coreCli --reporters=default -t 'woof'
```
PeterSchafer, line 127: Suggestion: Let's remove the `jest-junit` from the package.json and only add it when running in CI. That way, the developer doesn't have to specify the default reporter, which improves the local developer experience.

CONTRIBUTING.md
```diff
@@ -8,7 +8,7 @@ To install the required development dependencies in homebrew based environments,
 The only additional prerequisite is having [homebrew](https://brew.sh/) installed.
 
 ```sh
-./scripts/install-dev-dependencies.sh
+make install-deps
```
PeterSchafer, line 11: Question: can we assume `make` to be available? Probably since we don't install it anywhere, we might want to mention it like we do with brew above.

Makefile
PeterSchafer: Just to expand the thought, I wonder if we can make the usage of the virtual environment invisible to the `make` user.

mgyorke: We could, the initial assumption in the MakeFile is that Python dependencies are installed globally before we run Make. This is handled by CircleCI in the preparatory steps.

We could change this, to use venv in the pipelines and then the build targets are the same everywhere. I see this as the ideal approach but I hesitated to touch the pipelines without discussing possible implications first.

The other two solutions I see are more pragmatic, as the current build path remains the same: 
- dedicated build target for local build, proposed in this PR;
- do some checks in the existing build target so that it activates the venv if it runs locally or proceed as usual when running in container;

CONTRIBUTING.md
mgyorke: You're right, I'll clean this up once we decide on the approach.

CONTRIBUTING.md
mgyorke: I think so, AFAIK it's bunded with the Xcode Command line tools that are a prerequisite for homebrew.
We could check for the dependencies that we assume are already installed and inform the user. 

I see how this may be necessary if the developer has different toolchains for managing dependencies and they touch PATH.

PR #5272, opened 2024-05-29
feat: perf enhancement for snyk monitor

anthogez:
Rather than have post monitor working 1 by 1, we are trying to achieve an initial 
and minimal concurrency of 5 monitor requests at once

### DIFF
--- a/src/cli/commands/monitor/process-chunks.ts
+++ b/src/cli/commands/monitor/process-chunks.ts
@@ -1,0 +1,130 @@
+import * as Debug from 'debug';
+import * as pathUtil from 'path';
+import * as pMap from 'p-map';
+import * as analytics from '../../../lib/analytics';
+import { FailedToRunTestError } from '../../../lib/errors/failed-to-run-test-error';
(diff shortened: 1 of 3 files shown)


src/cli/commands/monitor/process-chunks.ts
```diff
@@ -0,0 +1,130 @@
...
+  inspectResult: MultiProjectResultCustom | pluginApi.MultiProjectResult,
+  multiProjectResult: MultiProjectResultCustom,
+  contributors: Contributor[],
+  options: MonitorOptions & Options,
+  targetFile?: string,
+): Promise<Array<GoodResult | BadResult>> {
+  const { scannedProjects } = multiProjectResult;
+
+  const results: Array<GoodResult | BadResult> = [];
+  const MAX_CONCURRENCY = 5;
```
PeterSchafer, line 42: Suggestion: Let's be a bit more pessimistic to not increase the load too much and set it to a lower value first. Let's start with 2.

thisislawatts:
🧹 Closing as stale

PR #5346, opened 2024-06-26
fix: align auth failure error messages for oauth

cmars:
In the Typescript CLI, when OAuth fails due to an invalid or expired token, throw the same "missing auth" error that one would get if not logged in (no auth token).

The resulting error message will prompt the user to re-authenticate.

This has to be mediated by the legacycli http proxy because the Typescript side of the CLI is not involved in the OAuth flow at all; it is handled transparently by the go-application-framework's networking middleware.

A special HTTP request and response header is used by the legacycli proxy to indicate whether that middleware auth has failed. If it has, the functions which perform HTTP client requests with needle will throw an error that indicates "unauthenticated", resulting in an experience consistent with the legacy token auth flow in the CLI.

Outstanding TODOs on this:

- [x] Improve the header names
- [x] Improve the middleware auth error checking -- a specific error match would be better here 
- [x] Review the approach with the team
- [x] Tests?

---

## Pull Request Submission

The pull request must:

- **Reviewer Documentation**
    - [ ] follow [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) rules
    - [ ] be accompanied by a detailed description of the changes
- **Testing**

## Pull Request Review

## Pull Request Approval

by our content writers.

## What does this PR do?

See top posting.

- Build this branch locally.
- Authenticate with OAuth.
- "Damage" the token as follows:
  - Set `expired` in the past
  - Flip some characters in the refresh token, making it invalid.
- Observe that with this binary, all test commands will fail with the same error message as legacy token auth. Other than Code and SBOM, they'll tell you to `snyk auth` (that's a pre-existing issue that will be fixed in a followup).

CLI-392

### DIFF
--- a/cliv2/internal/proxy/proxy.go
+++ b/cliv2/internal/proxy/proxy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"log"
 	"net"
@@ -18,6 +19,7 @@ import (
 	pkg_utils "github.com/snyk/go-application-framework/pkg/utils"
 
 	"github.com/snyk/go-application-framework/pkg/networking/certs"
+	"github.com/snyk/go-application-framework/pkg/networking/middleware"
 	"github.com/snyk/go-httpauth/pkg/httpauth"
 
 	"github.com/snyk/cli/cliv2/internal/constants"
@@ -143,9 +145,28 @@ func (p *WrapperProxy) ProxyInfo() *ProxyInfo {
 	}
 }
 
+// headerSnykAuthFailed is used to indicate there was a failure to establish
+// authorization in a legacycli proxied HTTP request and response.
+//
+// The request header is used to propagate this indication from
+// NetworkAccess.AddHeaders all the way through proxy middleware into the
+// response.
+//
+// The response header is then used by the Typescript CLI to surface an
+// appropriate authentication failure error back to the user.
+//
+// These layers of indirection are necessary because the Typescript CLI is not
+// involved in OAuth authentication at all, but needs to know that an auth
+// failure specifically occurred. HTTP status and error catalog codes aren't
+// adequate for this purpose because there are non-authentication reasons an API
+// request might 401 or 403, such as permissions or entitlements.
+const headerSnykAuthFailed = "snyk-auth-failed"
+
 func (p *WrapperProxy) replaceVersionHandler(r *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
-	err := p.addHeaderFunc(r)
-	if err != nil {
+	if err := p.addHeaderFunc(r); err != nil {
+		if errors.Is(err, middleware.ErrAuthenticationFailed) {
+			r.Header.Set(headerSnykAuthFailed, "true")
+		}
 		p.DebugLogger.Printf("Failed to add header: %s", err)
 	}
 
@@ -177,6 +198,12 @@ func (p *WrapperProxy) Start() error {
 	proxy.Logger = log.New(&pkg_utils.ToZeroLogDebug{Logger: p.DebugLogger}, "", 0)
 	proxy.OnRequest().DoFunc(p.replaceVersionHandler)
 	proxy.OnRequest().HandleConnect(p)
+	proxy.OnResponse().DoFunc(func(resp *http.Response, ctx *goproxy.ProxyCtx) *http.Response {
+		if authFailed := resp.Request.Header.Get(headerSnykAuthFailed); authFailed != "" {
+			resp.Header.Set(headerSnykAuthFailed, authFailed)
+		}
+		return resp
+	})
 	proxy.Verbose = true
 	proxyServer := &http.Server{
 		Handler: proxy,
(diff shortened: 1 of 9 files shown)


cliv2/internal/proxy/proxy.go
```diff
@@ -146,6 +146,7 @@ func (p *WrapperProxy) ProxyInfo() *ProxyInfo {
 func (p *WrapperProxy) replaceVersionHandler(r *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
 	err := p.addHeaderFunc(r)
 	if err != nil {
```
cmars, line 148: Once https://github.com/snyk/go-application-framework/pull/213 lands,

```suggestion
	if errors.Is(err, middleware.ErrAuthenticationFailed) {
```

would be more accurate.

cliv2/internal/proxy/proxy.go
```diff
@@ -143,9 +145,28 @@ func (p *WrapperProxy) ProxyInfo() *ProxyInfo {
...
+//
+// The response header is then used by the Typescript CLI to surface an
+// appropriate authentication failure error back to the user.
+//
+// These layers of indirection are necessary because the Typescript CLI is not
+// involved in OAuth authentication at all, but needs to know that an auth
+// failure specifically occurred. HTTP status and error catalog codes aren't
+// adequate for this purpose because there are non-authentication reasons an API
+// request might 401 or 403, such as permissions or entitlements.
+const headerSnykAuthFailed = "x-snyk-auth-failed"
```
j-luong, line 163: question: do we want to stick with the `x-` prefix to denote custom header fields? The convention was [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) AFAIK.

cliv2/internal/proxy/proxy.go
```diff
@@ -143,9 +145,28 @@ func (p *WrapperProxy) ProxyInfo() *ProxyInfo {
 	}
 }
 
+// headerSnykAuthFailed is used to indicate there was a failure to establish
```
j-luong, line 148: praise: appreciate the context given here

cliv2/internal/proxy/proxy.go
cmars: Old habits die hard. Good shout; per [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648) I think it's reasonable to drop the `x-` prefix here.

j-luong:
LGTM:
Screenshot 2024-07-09 at 17 34 05

j-luong (approved):

cliv2/internal/proxy/proxy.go
```diff
@@ -177,6 +198,12 @@ func (p *WrapperProxy) Start() error {
 	proxy.Logger = log.New(&pkg_utils.ToZeroLogDebug{Logger: p.DebugLogger}, "", 0)
 	proxy.OnRequest().DoFunc(p.replaceVersionHandler)
 	proxy.OnRequest().HandleConnect(p)
+	proxy.OnResponse().DoFunc(func(resp *http.Response, ctx *goproxy.ProxyCtx) *http.Response {
+		if authFailed := resp.Request.Header.Get(headerSnykAuthFailed); authFailed != "" {
+			resp.Header.Set(headerSnykAuthFailed, authFailed)
```
PeterSchafer, line 203: Question: Why can't we just print the message here? Instead of delegating this to the TS CLI. This way we could get to a consistent error written in the Extensible CLI.

PeterSchafer: So basically what I'm wondering is, if we can't handle auth errors transparently to the TS CLI, similar to how we handle auth in general.

PeterSchafer:
FYI there is a failing test, which doesn't expect the new http header.

cliv2/internal/proxy/proxy.go
cmars: I worry about inter-process complexities with such an approach, at least while we're still dispatching command execution and console interactions entirely to the TS CLI.

Decomposing console interactions to the Go CLI and treating the TS CLI like a library would mitigate this risk, but that'll take followups beyond the scope of this PR.

If we did error out here in the middleware, I think we'd have to terminate the node subprocess immediately to prevent its output from getting mixed up with that of the parent process. I worry about the assumptions we'd be making about the order of operations scheduled in node's event loop, as well as what's been buffered for output in that process, how it handles signals, and resource cleanup.

Things can get tricky with inter-process coordination. It might be fine in the typical case, but hard to debug and reproduce edge cases which could be timing or platform-dependent.

cmars:
> FYI there is a failing test, which doesn't expect the new http header.

Fixed

sandor-trombitas (approved):

cliv2/internal/proxy/proxy.go
PeterSchafer: I understand the points and I should probably take one step back from proposing a solution to proposing/aligning on the goal.

My goal is to achieve a consistent behaviour for authentication failures throughout the CLI, no matter if the domain logic is implemented in TS or Golang and in a not too distant future for LS as well.

Some non-functional requirements for a solution would be, Single source of Truth and KISS (Keep it simple stupid).

PR #5399, opened 2024-08-02
fix: print all dep-graphs of a container scan

mcombuechen:
## What does this PR do?

When printing the dep-graph results from Snyk container SCA, the code would escape early and omit all the dep-graphs that might have been found during remote analysis. This commit moves the dep-graph printing during container SCA to after the remote analysis call, this way including additional dep-graphs of applications within the container image.

### DIFF
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -306,7 +306,9 @@ async function sendAndParseResults(
       options,
     );
 
-    depGraphs.set(constructProjectName(scanResult), depGraph.toJSON());
+    if (depGraph) {
+      depGraphs.set(constructProjectName(scanResult), depGraph.toJSON());
+    }
 
     const ecosystem = getEcosystem(options);
     if (ecosystem && options['print-deps']) {
--- a/src/lib/snyk-test/run-test.ts
+++ b/src/lib/snyk-test/run-test.ts
@@ -35,8 +35,14 @@ import {
 } from '../errors';
 import * as snyk from '../';
 import { isCI } from '../is-ci';
-import * as common from './common';
-import { RETRY_ATTEMPTS, RETRY_DELAY } from './common';
+import {
+  assembleQueryString,
+  constructProjectName,
+  depGraphToOutputString,
+  shouldPrintDepGraphs,
+  RETRY_ATTEMPTS,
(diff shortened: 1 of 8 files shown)


src/lib/snyk-test/run-test.ts
```diff
@@ -328,6 +341,28 @@ async function sendAndParseResults(
...
+      if (!depGraph || !scanResult) {
+        continue;
+      }
+
+      console.log(
+        depGraphToOutputString(
+          depGraph.toJSON(),
+          constructProjectName(scanResult),
+        ),
+      );
```
cmars, line 360: Can we stream to stdout here, rather than using `jsonStringifyLargeObject` (which is basically a JSON.stringify wrapper).

`jsonStringifyLargeObject` can fail for objects of a sufficient size, and stringifying large objects roughly doubles their memory footprint.

For container, we've had a few situations where a large number of apps can OOM the CLI.

Suggest adding a stream version of `depGraphToOutputString` -- perhaps call it `depGraphToOutputStream`? -- which uses JsonStreamStringify (from the already imported module dependency `json-stream-stringify`).

src/lib/snyk-test/run-test.ts
```diff
@@ -768,9 +807,7 @@ async function assembleLocalPayloads(
           );
         }
 
-        console.log(
-          common.depGraphToOutputString(root.toJSON(), targetFile || ''),
-        );
+        console.log(depGraphToOutputString(root.toJSON(), targetFile || ''));
```
cmars, line 810: Same suggestion as above here

cmars (commented):
First pass looks good! See suggestion below to help this scale better for large objects. Heap space is unfortunately capped in the NodeJS process though, which can lead to incorrect outputs (or possibly process crashes due to OOMs). We have a couple open tickets where containers have tripped this limit.

src/lib/snyk-test/run-test.ts
mcombuechen: I had no idea! Sounds great, will do.

mcombuechen:
Too much happening, too much failing, `tap` is driving me up the wall. Will open smaller-scoped PRs instead.

mcombuechen:
@cmars will close this in favour of two other PRs.

PR #5432, opened 2024-08-20
fix: Fix multiple parallelization issues 

PeterSchafer:
## Pull Request Submission

The pull request must:

- **Reviewer Documentation**
    - [x] follow [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) rules
- **Testing**

## Pull Request Review

## Pull Request Approval

by our content writers.

## What does this PR do?
This PR addresses multiple issues related to parallel execution of the CLI as a standalone executable on Windows OS as well as through the language-server command on any OS. The following changes are combined since they are strongly coupled. 

* Replace singleton usage for the internal proxy by a singleton for the internal certificate authority, which showed to be the actual shared instance that required this.
* Replace implicit cleanup of temporary directories by an explicit removal, which can be triggered through the workflow engine.

Risk:
* For CLI standalone execution, the risk should be low as the main change here is less cleanup.
* For the language-server command, the risk is that in some edge cases, like crashes, temporary files are not cleaned up. Which is in a lot of cases acceptable.

### DIFF
--- a/cliv2/pkg/basic_workflows/globalresources.go
+++ b/cliv2/pkg/basic_workflows/globalresources.go
@@ -1,0 +1,87 @@
+package basic_workflows
+
+import (
+	"os"
+	"sync"
+
+	"github.com/rs/zerolog"
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/spf13/pflag"
+
+	"github.com/snyk/cli/cliv2/internal/cliv2"
+	"github.com/snyk/cli/cliv2/internal/proxy"
+	"github.com/snyk/cli/cliv2/internal/utils"
+)
+
+var caSingleton *proxy.CaData
+var caMutex sync.Mutex
+
+var WORKFLOWID_GLOBAL_CLEANUP workflow.Identifier = workflow.NewWorkflowIdentifier("internal.cleanup")
+
+func initCleanup(engine workflow.Engine) error {
+	entry, err := engine.Register(WORKFLOWID_GLOBAL_CLEANUP, workflow.ConfigurationOptionsFromFlagset(pflag.NewFlagSet("cleanup", pflag.ContinueOnError)), globalCleanupWorkflow)
+	if err != nil {
+		return err
+	}
+	entry.SetVisibility(false)
+
+	return nil
+}
+
+func globalCleanupWorkflow(
+	invocation workflow.InvocationContext,
+	_ []workflow.Data,
+) (output []workflow.Data, err error) {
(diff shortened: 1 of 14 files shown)


test/jest/acceptance/language-server-extension.spec.ts
cmars, line 1: I'm nervous about the changes to attempt to recreate the race condition here and below. Concerns:

* More tests means more chances for the pipeline to fail, v1 APIs are backed by registry. (False negatives happen)
* Longer pipeline run time.
* No guarantee that a race condition will be reproduced; it's entirely up to the scheduler whether it happens. (False positives are possible) (!)

Without a reliably repeatable failure in the event there actually is a race, what will happen in practice, is someone will re-run the test until it passes. Some of those "passes" might still mask a concurrency problem, making the test results inconclusive.

I would remove these acceptance tests and shore up the Go tests to the point where we can force multiple processes into the critical section and prove they do not race in the test.

cliv2/pkg/basic_workflows/globalresources.go
cmars, line 1: I'd add test coverage to these functions in isolation, also test for mutual exclusivity with goroutines.

cliv2/pkg/basic_workflows/legacycli.go
cmars, line 1: Ideally, I'd test this workflow e2e. And there are ways we could do that -- refactor such that the parts which execute Node are mockable, and that you can inject checkpoints that force a race (I liken this to playing "Maxwell's Demon" with goroutines.. stand up goroutines such that if GOMAXPROCS > 1 they're certain to race).

But there might be a much simpler way to prove that the CA provisioning / de-provisioning is mutually exclusive within the same process, and the same usage in the legacycli workflow is likely goroutine-safe. With that we might feel comfortable enough to land the solution as-is with a followup to make legacycli more testable (I have a sense that's going to be more involved).

Write a simple "test workflow" that doesn't do anything, but provisions / de-provisions the CA at the same integration points as `legacycli` does. Inject "checkpoints" just prior to the critical section (the call to `GetGlobalCertAuthority`) with `<-ch` channel receives to "park" execution there. Start N of them, wait until they're all "parked" (can use bounded channels for this), then turn 'em all loose by closing that `ch` synchronization channel. Let them all finish and assert they all got the expected CA, completed as expected and cleaned up.

This was the approach I took with the oauth concurrency -- which is even harder to test e2e since it's an interactive browser interaction.

cmars (commented):
I have concerns with the acceptance tests -- we won't get a clear enough signal out of them to test race conditions properly.

I'd prefer to improve the Go tests to the point that we're confident in the primitives we've built upon.

cmars (approved):

PR #5667, opened 2025-01-17
chore: speed up clean build process

thisislawatts:
- [X] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [X] Includes detailed description of changes

## What does this PR do?

Removing the node_modules siginificantly slows down the rebuild
cycle when iterating locally as it requires reinstalling all deps

Removing the node_modules should be an optional step of last resort during debugging rather than default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "packages/*"
   ],
   "scripts": {
-    "clean": "npx rimraf node_modules dist binary-releases test-results *.tgz tsconfig.tsbuildinfo .eslintcache pysrc packages/*/node_modules packages/*/dist packages/*/tsconfig.tsbuildinfo packages/*/*.tgz",
+    "clean": "npx rimraf dist binary-releases test-results *.tgz tsconfig.tsbuildinfo .eslintcache pysrc packages/*/node_modules packages/*/dist packages/*/tsconfig.tsbuildinfo packages/*/*.tgz",
     "format": "prettier --write '**/*.{js,ts,json,yaml,yml,md}' && npm run lint:js -- --fix",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "packages/*"
   ],
   "scripts": {
-    "clean": "npx rimraf node_modules dist binary-releases test-results *.tgz tsconfig.tsbuildinfo .eslintcache pysrc packages/*/node_modules packages/*/dist packages/*/tsconfig.tsbuildinfo packages/*/*.tgz",
+    "clean": "npx rimraf dist binary-releases test-results *.tgz tsconfig.tsbuildinfo .eslintcache pysrc packages/*/node_modules packages/*/dist packages/*/tsconfig.tsbuildinfo packages/*/*.tgz",
```
PeterSchafer, line 30: Question: Since clean for me tries to create the same or at least similar state as a clean checkout, I'm wondering if we really want to not remove the `node_modules` folder. What use case are we optimizing for?

PeterSchafer: I'm trying to refer [this framing](https://www.gnu.org/software/make/manual/html_node/Cleanup.html)

> how to delete all the object files and executables so that the directory is ‘clean’.

thisislawatts: Development in the typescript CLI and only rebuilding the Typescript binary.

thisislawatts: So perhaps we need an alternative command intended for fast TS dev cycles

PeterSchafer: I think this makes the most sense. Maybe `make rebuild`?

thisislawatts:
Will revisit at a later date.

PR #5709, opened 2025-02-04
fix: propagate errors when --json is used

CatalinSnyk:
- [ ] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [ ] Includes detailed description of changes


test/setup-jest.ts
```diff
@@ -1,7 +1,9 @@
+const stripAnsi = require('strip-ansi');
```
thisislawatts, line 1: **issue:** Use imports here rather than commonjs require format.

test/jest/acceptance/snyk-sbom/npm-options.spec.ts
```diff
@@ -68,7 +68,7 @@ describe('snyk sbom: npm options (mocked server only)', () => {
 
     expect(code).toEqual(2);
     expect(stdout).toContainText(
-      'An error occurred while running the underlying analysis needed to generate the SBOM.',
+      'Dependency snyk was not found in package-lock.json.',
     );
```
thisislawatts, line 72: **question:** Why have we made this change to the asserted behaviour?

CatalinSnyk: That's the underlying error message. coming the dep-graph workflow (which fails because of the strict out of sync flag). It validates the changes properly forwards the error from that workflow back to the user.

src/cli/ipc.ts
```diff
@@ -22,16 +23,25 @@ export async function sendError(err: Error): Promise<boolean> {
...
+    err = new CLI.GeneralCLIFailureError(detail, {
+      suppressJsonOutput: isJson,
+    });
+    // @ts-expect-error Overriding the HTTP status field.
     err.metadata.status = 0;
   }
 
   const data = (err as ProblemError)
-    .toJsonApi('error-catalog-ipc-instance?')
+    .toJsonApi('error-catalog-ipc-instance')
```
thisislawatts, line 44: **question:** Why remove the `?` character from this string?

CatalinSnyk: I added that in the previous PR when I was reading about the [instance](https://github.com/snyk/error-catalog/blob/815ea0ceed014ee35934dd0832d339dfe9f1b4bb/packages/error-catalog-nodejs/src/types.ts#L20) parameter (and I was confused about what it did), and I missed removing that 😅

cliv2/internal/cliv2/cliv2.go
```diff
@@ -485,6 +485,14 @@ func (c *CLI) getErrorFromFile(errFilePath string) (data error, err error) {
 		errs := make([]error, len(jsonErrors)+1)
 		for _, jerr := range jsonErrors {
 			jerr.Meta["orign"] = "Typescript-CLI"
+
+			// NOTE: Only set if the invocation uses the --json flag. There's no need for output suppression if the invocation does NOT use StdIO
+			// (meaning the TS CLI is handling the user output directly), so we can disable the suppression in order to surface this errors properly.
+			useStdIO := c.globalConfig.GetBool(configuration.WORKFLOW_USE_STDIO)
```
thisislawatts, line 491: **question:** The comment here mentions: Only set if the invocation uses the --json flag. But below it looks like we are checking, `configuration.WORKFLOW_USE_STDIO`, how does that relate to the --json flag?

CatalinSnyk: I was referring to the "suppressJsonOutput" field being set, will update this to be a bit more expressive about that.

cliv2/cmd/cliv2/main.go
```diff
@@ -539,6 +562,9 @@ func MainWithErrorCode() (int, []error) {
 	outputWorkflow, _ := globalEngine.GetWorkflow(localworkflows.WORKFLOWID_OUTPUT_WORKFLOW)
 	outputFlags := workflow.FlagsetFromConfigurationOptions(outputWorkflow.GetConfigurationOptions())
 	rootCommand.PersistentFlags().AddFlagSet(outputFlags)
+	// add output flags as persistent flags
+	_ = rootCommand.ParseFlags(os.Args)
```
thisislawatts, line 566: **Question:** What is the incorrect behaviour you were seeing which required this change? I am not sure I understand how it is related to the change you are looking to make here?

src/cli/ipc.ts
thisislawatts: Can we remove that as part of this change?

cliv2/cmd/cliv2/main.go
```diff
@@ -458,9 +459,14 @@ func displayError(err error, userInterface ui.UserInterface, config configuratio
 		}
 
 		if config.GetBool(output_workflow.OUTPUT_CONFIG_KEY_JSON) {
```
CatalinSnyk, line 461: This should be the only remaining changes to main.

cliv2/cmd/cliv2/main.go
PeterSchafer: This can probably be removed

CatalinSnyk: Note: this is still needed to properly populate the configuration, since we need it for the cliv2/internal/cliv2/cliv2.go: 
```go
useSTDIO := c.globalConfig.GetBool(configuration.WORKFLOW_USE_STDIO)
jsonEnabled := c.globalConfig.GetBool(output_workflow.OUTPUT_CONFIG_KEY_JSON)
hasBeenDisplayed := false
if useSTDIO && jsonEnabled {
    hasBeenDisplayed = true
}
```
If you remove it and test against `snyk-macos-arm64 test --print-graph --file=./nothere --json` or `snyk-macos-arm64 iac test ./nothere ./notthere --json`, but there would be json output coming from the TS CLI and human readable output from Golang.

src/cli/ipc.ts
```diff
@@ -43,3 +49,36 @@ export async function sendError(err: Error): Promise<boolean> {
...
+ * @param message The Error.message field.
+ * @returns The extracted message.
+ * @example
+ * Error { message: "{\"ok\":false,\"error\":\"This is the actual error message.\",\"path\":\"./\"}" }
+ */
+function extractMessageFromJson(message: string): string {
+  try {
+    let msgObject = JSON.parse(message);
+
+    /** IAC JSON mode can contain an array of error objects, we will use just the first one to send to our debug logs.
```
PeterSchafer, line 64: > we will use just the first one to send to our debug logs.

Question 1: Debug logs?
Question 2: Why not use all?

src/cli/main.ts
```diff
@@ -138,32 +138,34 @@ async function handleError(args, error) {
...
-    args.options.json || args.options.sarif || vulnsFound
+    vulnsFound || args.options.sarif
       ? false
-      : sendError(error);
-  if (!errorSent) {
-    if (args.options.debug && !args.options.json) {
+      : await sendError(error, args.options.json);
+
+  // JSON output flow
+  if (args.options.json) {
```
PeterSchafer, line 151: Question: was it on purpose to not have `!(error instanceof UnsupportedOptionCombinationError)` here?

cliv2/cmd/cliv2/errorhandling.go
```diff
@@ -32,3 +34,70 @@ func decorateError(err error) error {
...
+	if errors.As(err, &snykErr) {
+		message = snykErr.Detail
+	}
+
+	return message
+}
+
+// errorHasBeenShown return whether the error was already presented by the Typescript CLI or not.
+// This will iterate through the error chain to find if any of the errors in the chain has been shown.
+func errorHasBeenShown(err error) bool {
```
PeterSchafer, line 52: Suggestion: This should probably live in the cliv2 package, just considering the commend and the imports

cliv2/cmd/cliv2/main.go
thisislawatts: I removed these changes locally @CatalinSnyk, but have not yet been able recreate the behaviour you described. 
```
-       // add output flags as persistent flags
-       _ = rootCommand.ParseFlags(os.Args)
-       globalConfiguration.AddFlagSet(rootCommand.LocalFlags())
```

```
~/snyk/cli/binary-releases/snyk-macos iac test ./notthere --json
```
Output
```
{
  "ok": false,
  "code": 1010,
  "error": "Could not find any valid IaC files",
  "path": "./notthere"
}
```

src/cli/ipc.ts
CatalinSnyk: 1. Should be `send through IPC`
2. That need's a follow up to swap the ProblemError instance to a [CompositeError](https://github.com/snyk/error-catalog/blob/14f5b6b4cf514d7dd24877848bb0b0b5d70b0c5f/packages/error-catalog-nodejs/src/composite-error.ts#L4) for us to be able to write multiple errors to the IPC. Will address in the followup around sending the description field in the IPC.

test/setup-jest.ts
CatalinSnyk: This would require a version bump for the dependency, and it was used in a lot of places around the code and tests, some commands also had it scoped under a different version than the main package.json; so I think it should not be included in this PR.

src/cli/commands/test/iac/output.ts
```diff
@@ -183,6 +184,8 @@ export function buildOutput({
         ? new FormattedCustomError(
             errorResults[0].message,
             formatFailuresList(allTestFailures),
+            undefined,
+            new CLI.GeneralIACFailureError(formatFailuresList(allTestFailures)),
```
j-luong, line 188: suggestion: will need to remove formatting using `stripAnsi` here:
```suggestion
            new CLI.GeneralIACFailureError(stripAnsi(formatFailuresList(allTestFailures))),
```

PeterSchafer: I assume that FormattedCustomError is a customer error, so we might be able to do the assignment and stripping of formatting more centrally, similar to the CustomError.

cliv2/cmd/cliv2/main.go
PeterSchafer: We took a look together and understood why and when this is needed. The outcome is, that this is only needed when falling back to the legacy CLI, which is why we decided to move this logic only to the branch when the fallback actually happens rather then always.

PeterSchafer (approved):

PR #5748, opened 2025-02-21
feat: minimise how often countPathsToGraphRoot is called

JamesPatrickGill:

- [ ] Updates relevant GitBook documentation (PR link: \_\_\_)

## What does this PR do?

This PR makes two optimisations to the `run-test` flow. These are designed so as to be transparent to users except for increased performance for projects with dense dependency graphs - "dense" defined as graphs with many cross linking dependencies leading to very many paths to the root of the graph.

Our main culprit is `countPathsToRoot()` which goes through every pkg in a graph and accumulates a count of all paths to root from all nodes. This is then compared to our prune limits to see whether we should prune, the only other thing this result is used for is analytics.

**`--pruneRepeatedSubdependencies` short circuit:**

Previously if a user passed the `--pruneRepeatedSubdependencies` flag the prune functionality still counted paths to root and calculated whether we were dealing with a dense graph. This is redundant as no matter the response, if `--pruneRepeatedSubdependencies` is set, then we start the prune logic.
Now if we see this flag we just immediately attempt the prune.

**Early exit from, `countPathsToRoot()`:**

The `countPathsToRoot()` is incredibly expensive and scales badly. The way it was previously working, going through every pkg and counting _everything_ before then comparison, is overkill in the most expensive cases and can take _hours_. Instead, we now add to an accumulator every pkg processed and as soon as we breach the limit we exit as we know we are definitely going to prune. Additionally, the dep-graph library also provides the functionality to early exit within the context of a pkg. Applying both of these optimisations saves a lot of time in the worst cases.

There is only one function that has changed.

Compare against the current release version of `snyk` using a dense graph - a moderate speed up should be expected, the relative speed up shoul be inversely correlated with size and density of the graph.

An improvement to the handling of dense, highly connected dependency graph provides a moderate performance uplift for these projects.

### DIFF
--- a/src/lib/prune.ts
+++ b/src/lib/prune.ts
@@ -6,7 +6,6 @@ import config from './config';
 import { TooManyVulnPaths } from './errors';
 import * as analytics from '../lib/analytics';
 import { SupportedPackageManagers } from './package-managers';
-import { countPathsToGraphRoot } from './utils';
 
 const debug = _debug('snyk:prune');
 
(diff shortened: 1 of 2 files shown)


thisislawatts (approved):

src/lib/prune.ts
```diff
@@ -17,14 +16,11 @@ export async function pruneGraph(
...
   pruneIsRequired = false,
 ): Promise<DepGraph> {
-  const prePrunePathsCount = countPathsToGraphRoot(depGraph);
-  const isDenseGraph = prePrunePathsCount > config.PRUNE_DEPS_THRESHOLD;
+  const prune = shouldPrune(depGraph, pruneIsRequired);
 
   debug('rootPkg', depGraph.rootPkg);
-  debug('prePrunePathsCount: ' + prePrunePathsCount);
-  debug('isDenseGraph', isDenseGraph);
-  analytics.add('prePrunedPathsCount', prePrunePathsCount);
```
orsagie, line 26: The full prePrunePathsCount is now not going to be shown in analytics. Can we make sure we don't depend on this for anything other than debug purposes?

orsagie (approved):

src/lib/prune.ts
JamesPatrickGill: Yes these analytics lose meaning after these changes, I believe they _might_ be in dashboards but will check

PR #5793, opened 2025-03-24
fix: stop verifying token on ingore cmd

jlourenc:
- [] Highlights breaking API changes (if applicable)
- [] Links to automated tests covering new functionality
- [] Updates relevant GitBook documentation (PR link: ___)
- [] Includes product update to be announced in the next stable release notes

## What does this PR do?

Ignore command is the sole command (auth aside) that verifies against Snyk platform that a token is valid.

This is redundant as an invalid token would be identified as invalid on any requests performed against the platform with a 401 status code.

This change refactors the code to only perform local validation, verifying that the token is set appropriately either via configuration or environment variable.


dotkas (approved):

src/cli/commands/ignore.ts
```diff
@@ -14,16 +14,16 @@ import { MisconfiguredAuthInCI } from '../../lib/errors/misconfigured-auth-in-ci
...
 
-  return auth
-    .isAuthed()
-    .then((authed) => {
-      if (!authed) {
-        if (isCI()) {
-          throw MisconfiguredAuthInCI();
-        }
-      }
+  return new Promise(() => {
```
CatalinSnyk, line 17: suggestion: I think the `return` could be moved to the `authorization.actionAllowed('cliIgnore', options).then...` promise, so this `new Promise` won't be needed anymore.

src/cli/commands/ignore.ts
```diff
@@ -14,16 +14,16 @@ import { MisconfiguredAuthInCI } from '../../lib/errors/misconfigured-auth-in-ci
...
+    } catch (err) {
+      if (isCI()) {
+        throw MisconfiguredAuthInCI();
+      }
+      throw err;
+    }
+  })
     .then(() => {
       return authorization.actionAllowed('cliIgnore', options);
     })
```
CatalinSnyk, line 29: suggestion: this might need a catch in case the auth fails, the `actionAllowed` returns am object, that should return an auth error (AuthFailedError).

src/cli/commands/ignore.ts
jlourenc: Can you please clarify your proposal and propose some code change?

src/cli/commands/ignore.ts
jlourenc: That is true before this change as well, right? I believe this to be handled by the thing using the promise 🤔

src/cli/commands/ignore.ts
CatalinSnyk: I pushed a commit to take care of this.

src/cli/commands/ignore.ts
CatalinSnyk: Yes. I was just thinking that the API call to check the `cliIgnoreAuthorization` seem to not care about the status codes (neither does the verify function call), which seems like a bad idea.

I would say this is the last thing to be taken care of for this PR, just to make sure there's some better error messages for the users. Happy to pair on this if you want to, feel free to book a time in my calendar.

PR #5796, opened 2025-03-25
fix: adding a stale PR check script

dotkas:
## What does this PR do?

We should not have an open PR without any activity for a month. 

N/A.


PeterSchafer (approved):

PeterSchafer:
Praise: Nice Improvement

PR #5812, opened 2025-04-03
fix: Always set the API URL for legacy CLI invocations

PeterSchafer:

## What does this PR do?
The PR ensures to always specify the API URL when running the legacy Typescript CLI. Before the API URL was only set when the API was explicitly set but given the possibility to derive an URL from a OAuth claim, we need to always set the URL.

There was no closed box test for this, which is now there and one test had to be adapted since a side-effect of the change is, that an error message changed. Since the error message is only a side effect of one command and testing the edge case of an incorrectly specified `endpoint`, I removed these unrelated checks from the test.

The Risk of the change should be low, as it only ensures that both golang and typescript CLI actually use the same API URL, making the golang CLI as the single source of truth.

`cliv2/internal/cliv2/cliv2.go`

Run the test and observe that despite the fact that no API URL is configured the `SNYK_API` is set to `https://api.my.special.url.io `

```
npx jest test/jest/acceptance/api-auto-detect.spec.ts
```

### DIFF
--- a/test/jest/acceptance/api-auto-detect.spec.ts
+++ b/test/jest/acceptance/api-auto-detect.spec.ts
@@ -1,0 +1,34 @@
+import { runSnykCLI } from '../util/runSnykCLI';
+
+jest.setTimeout(1000 * 10);
+
+describe('Auto detect API URL', () => {
+  it('Based on OAuth token claim', async () => {
+    const expectedApiUrl = 'https://api.my.special.url.io';
+    const payload = {
+      sub: '1234567890',
+      name: 'John Doe',
+      iat: 1516239022,
+      aud: [expectedApiUrl],
+    };
+    const oauthToken = {
+      access_token: `a.${Buffer.from(JSON.stringify(payload))
+        .toString('base64')
+        .replace(/=+$/, '')}.b`,
+      token_type: 'Bearer',
+      refresh_token: 'configRefreshToken',
+      expiry: '3023-03-29T17:47:13.714448+02:00',
+    };
+
+    const envVars = {
+      ...process.env,
+      INTERNAL_OAUTH_TOKEN_STORAGE: JSON.stringify(oauthToken),
+    };
+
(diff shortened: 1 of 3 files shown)


test/jest/acceptance/api-auto-detect.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+describe('Auto detect API URL', () => {
+  it('Based on OAuth token claim', async () => {
+    const expectedApiUrl = 'https://api.my.special.url.io';
+    const oauthToken = {
+      access_token:
+        'a.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJhdWQiOlsiaHR0cHM6Ly9hcGkubXkuc3BlY2lhbC51cmwuaW8iXX0.b',
+      token_type: 'Bearer',
+      refresh_token: 'configRefreshToken',
+      expiry: '3023-03-29T17:47:13.714448+02:00',
+    };
```
dotkas, line 14: What do you think about something like this for test readability (didn't test it, just for suggestions)

```suggestion
    const payload = {
      sub: '1234567890',
      name: 'John Doe',
      iat: 1516239022,
      aud: [expectedApiUrl],
    };

    const oauthToken = {
      access_token:
        `a.${Buffer.from(
      JSON.stringify(readableTokenPayload)
    ).toString('base64')}.b`,
      token_type: 'Bearer',
      refresh_token: 'configRefreshToken',
      expiry: '3023-03-29T17:47:13.714448+02:00',
    };
```

dotkas (approved):

test/jest/acceptance/api-auto-detect.spec.ts
PeterSchafer: Awesome idea! Thank you, just needed some small changes to make it work 😄

PR #5869, opened 2025-05-01
fix: [OSM-2584] Bump PHP plugin version to fix project scanning

nico-incubiq:
- [x] Updates relevant GitBook documentation (PR link: ___)

## What does this PR do?
Bumps the version of the PHP Plugin used by the CLI, which fixes Open Source scanning of PHP projects.
Previously when  `composer`/ `composer.phar` is available on the machine, the command's result `stdout` was not handled properly and resulted in `[object Object]` being parsed as JSON.

Risk is **Low** as this only affects PHP, which has low traffic and it fixes a bug rather than changing any logic.

- Run the CLI on a directory containing PHP code, making sure that you have [composer](https://getcomposer.org/download/) either installed globally or the `composer.phar` file in the current working directory.
- This should output scan results as expected, while before this would have crashed with a JSON parsing error

Scanning open source dependencies of PHP projects when composer is installed on the user machine works again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,7 +77,7 @@
         "snyk-nodejs-lockfile-parser": "1.60.1",
         "snyk-nodejs-plugin": "1.4.4",
         "snyk-nuget-plugin": "2.7.15",
-        "snyk-php-plugin": "1.10.0",
+        "snyk-php-plugin": "1.12.1",
         "snyk-policy": "4.1.4",
         "snyk-python-plugin": "2.6.0",
         "snyk-resolve-deps": "4.8.0",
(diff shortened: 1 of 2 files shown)


PeterSchafer:
@nico-incubiq based on the description it appears to me that we don't have any relevant php related test here. Could you please take a look in `test/jest/acceptance/snyk-test` and see if we can add at least add a happy path test?

nico-incubiq:
@PeterSchafer interesting point; there is already an acceptance test for this, but it was exercising an overly permissive code path which was hiding the issue (interpreting the error as if we found a `[object` dependency with version `Object]`). I've removed this code branch from my PHP plugin bump, so now the existing test properly exercises everything there is to exercise, and a similar error would no longer be hidden.

PR #5922, opened 2025-05-26
Demonstration of stdio wrapping losing stderr

jbowes:
After building the binary, try these commands:

<binary> wraptest woof (works, stdout is captured and printed)
<binary> wraptest --usestdio woof (works, stdout is passed through)
<binary> wraptest --usestdio test (works, snyk test tests the current project, fails, and prints findings to stderr)
<binary> wraptest test (fails, snyk test tests the current project, fails, and stderr is lost)

### DIFF
--- a/cliv2/pkg/basic_workflows/wraptest.go
+++ b/cliv2/pkg/basic_workflows/wraptest.go
@@ -1,0 +1,70 @@
+package basic_workflows
+
+import (
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/spf13/pflag"
+)
+
+var WORKFLOWID_WRAPTEST workflow.Identifier = workflow.NewWorkflowIdentifier("wraptest")
+var DATATYPEID_WRAPTEST_STDOUT workflow.Identifier = workflow.NewTypeIdentifier(WORKFLOWID_WRAPTEST, "stdout")
+
+// initWraptest registers the wraptest workflow with the engine
+func initWraptest(engine workflow.Engine) error {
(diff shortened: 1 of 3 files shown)


cliv2/pkg/basic_workflows/wraptest.go
```diff
@@ -0,0 +1,70 @@
...
+	// Create a new configuration for the legacy CLI workflow
+	legacyConfig := config.Clone()
+	legacyConfig.Set(configuration.RAW_CMD_ARGS, args)
+
+	// Set the WORKFLOW_USE_STDIO flag based on the usestdio flag
+	legacyConfig.Set(configuration.WORKFLOW_USE_STDIO, useStdio)
+
+	// Execute the legacy CLI workflow
+	legacyData, legacyErr := invocation.GetEngine().InvokeWithConfig(WORKFLOWID_LEGACY_CLI, legacyConfig)
+	return legacyData, legacyErr
```
PeterSchafer, line 69: Issue: Forwarding the error from the legacy CLI will cause the CLI to exit early.

There is a bit of inconsistency here around error handling and the legacycli workflow. It returns the exit code error from the process as an error, which in case of 1 or 3 is not an actual error. 
If the error is returned unhandled here as well, the main workflow in the CLI ends and no output is written, in addition these errors are not printed which is why nothing is shown. 
Wrapping `legacyErr` with `fmt.Errorf("legacycli error: %w", legacyErr)` makes the error state visible, handling exit codes 1 and 3 would probably be the best way to deal with this currently. Temporarily swallowing `legacyErr` would also work, then the content of the stdout buffer and stderr buffer will be printed to the console.

jbowes: thank you!

PR #6042, opened 2025-07-15
chore: Update Release Candidate

j-luong:
Release Candidate

### DIFF
--- a/binary-releases/RELEASE_NOTES.md
+++ b/binary-releases/RELEASE_NOTES.md
@@ -22,8 +22,7 @@ The Snyk CLI is being deployed to different deployment channels, users can selec
 
 ### Bug Fixes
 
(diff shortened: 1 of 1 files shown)


binary-releases/RELEASE_NOTES.md
```diff
@@ -24,6 +24,7 @@ The Snyk CLI is being deployed to different deployment channels, users can selec
 
 * Bump `snyk-mvn-plugin` to require Node 20 ([c91cfbb](https://github.com/snyk/snyk/commit/c91cfbbb02a7319454a306f1889fd443288abf75))
```
CatalinSnyk: Small nitpick: we could probably add \*\*test**, \*\*general** and \*\*general** for these ones.

CatalinSnyk: Maybe even \*\*logging for for the last one since we use it bellow

PR #6168, opened 2025-09-15
feat: enable reachability checks in OS extension [OSF-91]

jacek-rzrz:
Bumping the version of the OS extension to bring in the check of reachability setting for tests with the `--reachability` param.

image

The latest version of the extension also brings in support for the `monitor` command. This PR relaxes the param validation for `monitor` in order to make routing via the extension work (same way as it had been done for `test`).


jacek-rzrz:
> @jacek-rzrz this change broke a lot of tests since it seems to change the behaviour, please take a look.

Thanks @PeterSchafer! I'm having a look :)

PR #6186, opened 2025-09-24
feat: [CN-249] support scanning system JARs

dan-arpino:
use the --include-system-jars option on the Snyk Docker Plugin to scan the usr/lib folder for JAR files

- [x] Commit messages

## What does this PR do?

Reads the `scanUsrLibJars` feature flag and correspondingly sets the `--include-system-jars` parameter when calling snyk-docker-plugin

Main change is in test (cli/commands/test/index.ts) and monitor (cli/commands/monitor/index.ts) command files.  The rest is updating testing.

A default `snyk container test/monitor` should return identical results (since the feature flag is off by default).  You can turn the feature flag on, then us `--print-deps` on a container image with JARs in the usr/lib (example openjdk:8-jdk-alpine), and ensure that they show up.  

No change to CLI users.  We will work with the clients who have requested the feature to get the feature flag & functionality turned on for them.

Low

This is because we are having issues reading JARs on certian base images.  Specifically Chainguard images ever since they switched to the usrmerge format back in May.

CN-352
CN-340
CN-337
CN-249

### DIFF
--- a/src/cli/commands/monitor/index.ts
+++ b/src/cli/commands/monitor/index.ts
@@ -28,7 +28,7 @@ import { apiOrOAuthTokenExists } from '../../../lib/api-token';
 import { maybePrintDepTree, maybePrintDepGraph } from '../../../lib/print-deps';
 import { monitor as snykMonitor } from '../../../lib/monitor';
 import { processJsonMonitorResponse } from './process-json-monitor';
-import snyk = require('../../../lib'); // TODO(kyegupov): fix import
+import * as snyk from '../../../lib';
 import { formatMonitorOutput } from '../../../lib/formatters';
 import { getDepsFromPlugin } from '../../../lib/plugins/get-deps-from-plugin';
 import { getExtraProjectCount } from '../../../lib/plugins/get-extra-project-count';
@@ -48,7 +48,17 @@ import { isMultiProjectScan } from '../../../lib/is-multi-project-scan';
 import { getEcosystem, monitorEcosystem } from '../../../lib/ecosystems';
 import { getFormattedMonitorOutput } from '../../../lib/ecosystems/monitor';
 import { processCommandArgs } from '../process-command-args';
-import { hasFeatureFlag } from '../../../lib/feature-flags';
+import {
+  hasFeatureFlag,
+  hasFeatureFlagOrDefault,
+} from '../../../lib/feature-flags';
+import {
+  SCAN_USR_LIB_JARS_FEATURE_FLAG,
+  CONTAINER_CLI_APP_VULNS_ENABLED_FEATURE_FLAG,
+  INCLUDE_SYSTEM_JARS_OPTION,
+  EXCLUDE_APP_VULNS_OPTION,
+  APP_VULNS_OPTION,
+} from '../constants';
 import {
   PNPM_FEATURE_FLAG,
   DOTNET_WITHOUT_PUBLISH_FEATURE_FLAG,
@@ -83,7 +93,7 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
   const results: Array<GoodResult | BadResult> = [];
 
   if (options.id) {
-    snyk.id = options.id;
+    (snyk as any).id = options.id;
   }
 
   if (options.allSubProjects && options['project-name']) {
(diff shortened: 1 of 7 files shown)


src/cli/commands/monitor/index.ts
```diff
@@ -28,7 +28,7 @@ import { apiOrOAuthTokenExists } from '../../../lib/api-token';
 import { maybePrintDepTree, maybePrintDepGraph } from '../../../lib/print-deps';
 import { monitor as snykMonitor } from '../../../lib/monitor';
 import { processJsonMonitorResponse } from './process-json-monitor';
-import snyk = require('../../../lib'); // TODO(kyegupov): fix import
+import * as snyk from '../../../lib';
```
ivan09069, line 31: Added logic to check the scanUsrLibJars feature flag. If enabled, the code sets the --include-system-jars option, allowing system JARs to be included in the scan.

dan-arpino:
> Warnings
> ⚠️	
> Since the CLI is unifying on a standard and improved tooling, we're starting to migrate old-style `import`s and `export`s to ES6 ones. A file you've modified is using either `module.exports` or `require()`. If you can, please update them to ES6 [import syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [export syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export). Files found:
> 
> * `src/cli/commands/test/index.ts`
> * `test/jest/unit/cli-commands/test-and-monitor.spec.ts`
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [2c76deb](https://github.com/snyk/cli/commit/2c76deb4270f5b7b0e1915cc95c6c99e2c3dd200)

Able to update a couple of these files, but these two remaining are a little more involved, affecting multiple files that was causing issues.  Leaving these two for now.

src/cli/commands/monitor/index.ts
```diff
@@ -126,6 +126,15 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
         console.log(theme.color.status.warn(appVulnsReleaseWarningMsg));
       }
     }
+
+    // Check scanUsrLibJars feature flag and add --include-system-jars parameter
+    const scanUsrLibJarsEnabled = await hasFeatureFlag(
```
PeterSchafer: Suggestion: hasFeatureFlag() might throw an exception, I would recommend to catch this here to avoid negative side effects by a feature flag change.

src/cli/commands/test/index.ts
```diff
@@ -127,6 +127,15 @@ export default async function test(
         console.log(theme.color.status.warn(appVulnsReleaseWarningMsg));
       }
     }
+
+    // Check scanUsrLibJars feature flag and add --include-system-jars parameter
+    const scanUsrLibJarsEnabled = await hasFeatureFlag(
```
PeterSchafer: Same as above.

src/cli/commands/monitor/index.ts
```diff
@@ -126,6 +126,15 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
...
+      'scanUsrLibJars',
```
PeterSchafer: Question: How do you feel about introducing some constants for the strings here and in all the other files?

src/cli/commands/monitor/index.ts
dan-arpino: good point.  I've added a new hasFeatureFlagOrDefault function to log the errors and return a default value.

src/cli/commands/monitor/index.ts
dan-arpino: Good Idea.  Let me do that and pull a couple of the other option strings in there.

src/cli/commands/monitor/index.ts
```diff
@@ -104,28 +114,39 @@ export default async function monitor(...args0: MethodArgs): Promise<any> {
...
+        options[EXCLUDE_APP_VULNS_OPTION] &&
         !options['json'] &&
         !options['sarif']
       ) {
         console.log(theme.color.status.warn(appVulnsReleaseWarningMsg));
       }
     }
+
+    // Check scanUsrLibJars feature flag and add --include-system-jars parameter
+    const scanUsrLibJarsEnabled = await hasFeatureFlagOrDefault(
```
PeterSchafer, line 142: Question: One additional question, from a logic perspective, should we skip the feature flag check if `--include-system-jars` is set?

PR #6697, opened 2026-04-01
chore: add lint pre-commit hook

pkey:

- [x] Commit messages

## What does this PR do?

Adds a `lint` pre-commit hook that runs `npm run lint` (eslint + prettier) locally before each commit. This matches the linting step in the [code-analysis](https://app.circleci.com/pipelines/github/snyk/cli?branch=main&filter=all) CI job.

Also updates `scripts/install-dev-dependencies.sh` to run `pre-commit install` so the hooks are automatically activated for new and existing developers.

### Why?

Linting issues that only surface in the `code-analysis` CI job increase the feedback loop — developers push, wait for CI, discover a formatting error, fix, push again. Running the same checks locally as a pre-commit hook catches these issues immediately, before the code even leaves the machine.

- `.pre-commit-config.yaml` — the new `lint` hook
- `scripts/install-dev-dependencies.sh` — added `pre-commit install`

1. Run `pre-commit install` (or `./scripts/install-dev-dependencies.sh`)
2. Introduce a formatting issue (e.g. bad indentation in a `.ts` file)
3. `git commit` — should fail with prettier error
4. Fix the formatting, commit again — should pass

N/A — internal developer tooling only.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,3 +5,23 @@ repos:
     rev: v8.17.0
     hooks:
       - id: gitleaks
+  - repo: local
+    hooks:
+      - id: lint-eslint
+        name: eslint
+        entry: npx eslint --quiet --cache
+        language: system
+        types_or: [javascript, ts]
+        pass_filenames: true
+      - id: lint-prettier
+        name: prettier
+        entry: npx prettier --check
+        language: system
+        types_or: [javascript, ts, json, yaml, markdown]
+        pass_filenames: true
+      - id: golangci-lint
+        name: golangci-lint
+        entry: bash -c 'cd cliv2 && CGO_ENABLED=1 make lint'
+        language: system
+        files: ^cliv2/
+        pass_filenames: false
--- a/scripts/install-dev-dependencies.sh
+++ b/scripts/install-dev-dependencies.sh
@@ -3,3 +3,5 @@ set -exuo pipefail
 
 # requires https://brew.sh/
 brew bundle --file=$(dirname "$0")/Brewfile
+
+pre-commit install


.pre-commit-config.yaml
```diff
@@ -5,3 +5,23 @@ repos:
     rev: v8.17.0
     hooks:
       - id: gitleaks
+  - repo: local
```
PeterSchafer, line 8: Suggestion: Nice idea! did you intentionally not use `npm run format`?

PeterSchafer: The only worry I have is performance/latency it adds.

pkey: @PeterSchafer yeah I originally had `npm run lint` which I believe covers formatting and is also what we run in the `code-analysis` job. The issue with that is that it will run the linter/formatter on all the files every time. 

This approach limits it only to what has changed before committing. Performance wise, we can't go any further I believe.

pkey: Locally on my 32GB M1 the whole `npm run lint` runs in 7s - so that's the absolute worst case scenario.

pkey: I guess the latency is not exactly linear (1 file still takes 2.5s) but still worth it IMO given that this can easily be 20+ minutes before you find out you have a linter error in the CI. 

image

j-luong: I think running the same commands as in the `code-analysis` step should suffice. i.e. `npm run lint` & `make lint`.

I agree that the ~15seconds it takes to run the pre-commit linting is better than waiting for the CI job to fail

PeterSchafer: okay let's run an experiment. when it gets too long, we can still change it!

pkey: @j-luong so would you prefer the full `npm run lint` command or this version? This version I believe covers everything though if you would change what `npm run lint` does it would fall through the cracks.

j-luong: > so would you prefer the full npm run lint command 
 
yes, running the same commands in the CICD step and pre-commit makes the most sense to me

pkey:
@PeterSchafer @j-luong hey folks, first of all, sorry for the delay. Our priorities shifted a bit and I won't be iterating on the CLI as much in the near future which means I won't be able to properly "monitor" this on a daily basis. 

Would you like me to merge and then do the observation yourself or you rather me close it and you add it as a task for the new joiners later?

pkey:
Closing so that this can be picked up by new-joiners.

PR #6728, opened 2026-04-15
fix(general): Improve error handling to surface the most relevant error

PeterSchafer:

## What does this PR do?

Implements a first iteration error prioritization to ensure the most relevant/severe error is displayed and used for exit code derivation when multiple errors occur during CLI execution.

**Changes:**
- Added [FindMostRelevantError()](cliv2/internal/errors/errors.go:100:0-125:1) function that finds the most relevant error from a list based on:
- Updated error processing to use [FindMostRelevantError()](cliv2/internal/errors/errors.go:100:0-125:1) for selecting which error to display and derive exit codes from
- Moved and extended maintenance error tests to a dedicated test file

1. Core logic for [FindMostRelevantError()](cliv2/internal/errors/errors.go:100:0-125:1) and priority comparison in the CLI errors package
2. Usage in the [processError()](cliv2/cmd/cliv2/main.go:719:0-746:1) function
3. Unit tests covering various scenarios

```
TEST_SNYK_COMMAND=binary-releases/snyk-macos-arm64 npx jest test/jest/acceptance/maintenance.spec.ts
```

### DIFF
--- a/cliv2/cmd/cliv2/main.go
+++ b/cliv2/cmd/cliv2/main.go
@@ -681,20 +681,23 @@ func MainWithErrorCode() int {
 		// ignore
 	}
 
-	if err != nil {
-		errorList, err = processError(err, errorList)
+	outputError := err
+	allErrors := errorList
 
-		for _, tempError := range errorList {
+	if err != nil {
+		allErrors, outputError = processError(err, errorList)
+
+		for _, tempError := range allErrors {
 			if tempError != nil {
 				cliAnalytics.AddError(tempError)
 			}
 		}
 	}
 
-	displayError(err, globalEngine.GetUserInterface(), globalConfiguration, ctx)
+	displayError(outputError, globalEngine.GetUserInterface(), globalConfiguration, ctx)
 
-	exitCode := cliv2.DeriveExitCode(err)
-	globalLogger.Printf("Deriving Exit Code %d (cause: %v)", exitCode, err)
+	exitCode := cliv2.DeriveExitCode(outputError)
+	globalLogger.Printf("Deriving Exit Code %d (cause: %v)", exitCode, outputError)
 
 	updateInstrumentationDataBeforeSending(cliAnalytics, startTime, ua, exitCode)
 
@@ -711,7 +714,7 @@ func MainWithErrorCode() int {
 	}
 
 	if debugEnabled {
-		writeLogFooter(exitCode, errorList, globalConfiguration, networkAccess)
+		writeLogFooter(exitCode, allErrors, globalConfiguration, networkAccess)
 	}
 
 	return exitCode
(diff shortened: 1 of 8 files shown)


cliv2/internal/errors/errors.go
```diff
@@ -9,3 +15,63 @@ type ErrorWithExitCode struct {
 func (e ErrorWithExitCode) Error() string {
 	return fmt.Sprintf("exit code: %d", e.ExitCode)
 }
+
+// FindMostRelevantError searches for the maintenance error (SNYK-0099) in a list of errors.
+// This ensures the maintenance error takes priority for exit code mapping when present.
+// Returns a pointer to the maintenance error if found, or nil otherwise.
+func FindMostRelevantError(errorList []error) error {
+	// 1. find known high priority errors
+	knownHighPriorityErrors := []string{snyk.NewMaintenanceWindowError("").ErrorCode}
```
danskmt: The function description currently says "FindMostRelevantError searches **for the maintenance error (SNYK-0099)** [...]", but it seems the purpose of the function was defined to be expanded at some point by having the variable **knownHighPriorityErrors** as a list.
Either make the function specific for the case of identifying SNYK-0099, or make it generic, possibly moving this definition of "high priority errors" to a constant outside of this function and by changing the function's description

cliv2/internal/errors/errors.go
```diff
@@ -9,3 +15,63 @@ type ErrorWithExitCode struct {
...
+			}
+		}
+	}
+
+	// fallback: create a single error from all errors
+	if len(errorList) == 1 {
+		return errorList[0]
+	} else if len(errorList) > 1 {
+		return errors.Join(errorList...)
+	}
```
danskmt, line 47: I believe we can not include such fallback for this function. The caller should decide whether to return the whole list or just the first error from the list, wdyt? (also thinking because of the function name and description)

cliv2/internal/errors/errors.go
```diff
@@ -9,3 +15,63 @@ type ErrorWithExitCode struct {
...
+			}
+		}
+		return nil
+	}
+
+	// Check if this error itself is the target snyk error
+	var snykErr snyk_errors.Error
+	if errors.As(err, &snykErr) && snykErr.ErrorCode == errorCode {
+		return &snykErr
+	}
```
danskmt: I believe this check should come before the previous one that involves recursive calls

cliv2/internal/errors/errors.go
PeterSchafer: > the caller should decide whether to return the whole list or just the first error from the list, wdyt? 
I'm not sure I understand, somewhere we need to make the decision. This is why it is here.

This logic was previously located in processError() to decide which is the error to display and derive the exit code from. IMHO I think it makes sense as I wanted to encapsulate the logic in one function, single source of truth. Maybe the naming is misleading but I really don't know what else to name the function.

How about calling this function `getFinalError()`?

cliv2/internal/errors/errors.go
PeterSchafer: Can you explain why?

PeterSchafer: Maybe we need to change the overall order here?!

danskmt: I think it makes sense to come first because this is the "first level", i.e. '_is this error the one we need to return?_' before going through unwrapping and recursive calls

danskmt: just makes it easier to read

cliv2/internal/errors/errors.go
danskmt: hmmm maybe makes sense to still be in **processError()**, as it "adds" extra functionalities to this function's purpose if leaving it here. You can see it can return a joined error instead of a single one as well... 

There is also the problem that this function at the moment is just looking for  **NewMaintenanceWindowError**, but accepts searching in a list of "priority errors". I'm wondering what will happen if someone ever adds a second one there, or multiple ones, they will need to keep in mind the ordering is the priority over the "priority errors"....

cliv2/cmd/cliv2/main.go
danskmt, line 746: we are mutating **err** at line 739, is that expected from what we're trying to achieve at line 735?

PeterSchafer: refactored and removed the mutation

cliv2/internal/errors/errors.go
PeterSchafer: Let's take a closer look together, I think the logic would be incorrect if we move this to the front.

cliv2/internal/errors/errors.go
PeterSchafer: Let's take a look toghether.

danskmt (approved):

danskmt (approved):

CatalinSnyk (approved):

PR #6747, opened 2026-04-24
feat: parallelize container test dependency requests

bgardiner:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages
  are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing
  _what_ was changed, not _how_.
- [x] Includes detailed description of changes
- [x] Contains risk assessment (Low | Medium | High)
- [ ] Highlights breaking API changes (if applicable)
- [x] Links to automated tests covering new functionality
- [x] Includes manual testing instructions (if necessary)
- [ ] Updates relevant GitBook documentation (PR link: ___)
- [x] Includes product update to be announced in the next stable release notes

## What does this PR do?

This PR updates request dispatch in the active `snyk container test` path (`src/lib/snyk-test/run-test.ts`) and keeps request execution parallelized via the existing pMap concurrency limit (`MAX_CONCURRENCY = 5`).

## Important routing correction

A previous iteration targeted `src/lib/ecosystems/test.ts`, but `snyk container test` currently routes through the legacy test flow (`snyk-test/run-test.ts`), not that ecosystems test function. This PR keeps changes in the correct path.

## Where should the reviewer start?

- `src/lib/snyk-test/run-test.ts`

## How should this be manually tested?

1. Run lint checks (matches CI code-analysis lint stage):
   - `npm run lint`
2. Run related monitor unit tests for regression confidence:
   - `npm run test:unit -- --runTestsByPath test/jest/unit/ecosystems-monitor-docker.spec.ts`
3. Optionally run a representative container test command against a multi-project image and compare duration:
   - `snyk container test <image>`

## What's the product update that needs to be communicated to CLI users?

Container test request orchestration changes are applied in the active legacy test path used by `snyk container test`.

## Risk assessment (Low | Medium | High)?

Low. The final change is narrowly scoped and maintains existing container request parallelism semantics.

## Any background context you want to provide?

The branch previously introduced a base-first ordering strategy for container payloads, but that was intentionally removed.

## What are the relevant tickets?

N/A

[...]

PR #6756, opened 2026-04-29
feat: introduce SNYK_REQUEST_CONCURRENCY for dependency request parallelism

bgardiner:
## What does this PR do?

Adds a tunable concurrency knob for in-flight dependency-test / dependency-monitor HTTP requests, configurable via the user-facing `SNYK_REQUEST_CONCURRENCY` env var (default **5**, clamped to `[1, 50]`).

The default is unchanged from the prior hard-coded `MAX_CONCURRENCY = 5` — this PR is purely a *configurability* change. Bumping the default can be revisited separately once we have telemetry on CPU and rate-limit impact (per the review thread).

The user-facing env var is owned by the Go CLI's GAF configuration, with the resolved value forwarded to the legacy CLI via the internal `SNYK_INTERNAL_REQUEST_CONCURRENCY` env var. This keeps Go as the single source of truth for application configuration (env vars, future config files / flags, etc.), per Peter's review feedback.

### Wiring

- **Go side**
  - `cliv2.ConfigKeyRequestConcurrency` (`internal_request_concurrency`) — new GAF config key.
  - `main.go` registers `snyk_request_concurrency` as an alternative key, so the env var feeds the config.
  - `fillEnvironmentFromConfig` writes the resolved value to `SNYK_INTERNAL_REQUEST_CONCURRENCY` for the legacy CLI process.
  - The internal var is added to the legacy-CLI env blacklist so a user can't bypass the Go config by setting it directly.
- **TS side**
  - `getRequestConcurrency()` in `src/lib/snyk-test/common.ts` reads `SNYK_INTERNAL_REQUEST_CONCURRENCY` (the internal contract from Go), defaults to 5, clamps to `[1, 50]`.
  - Replaces the hard-coded `MAX_CONCURRENCY = 5` constant at the existing `pMap` call site in `sendAndParseResults` (`src/lib/snyk-test/run-test.ts`).

A follow-up PR (#6757) adopts the same helper in `src/lib/ecosystems/monitor.ts:monitorDependencies`, which is currently fully sequential.

## Why?

`snyk container test` produces one `ScanResult` per directory of dependencies in the image (`lib/analyzer/applications/java.ts:groupJarFingerprintsByPath` in `snyk-docker-plugin`). For Java-heavy images this can be hundreds of ScanResults, each becoming a separate `POST /test-dependencies` request. With the prior hard-coded concurrency cap of 5, the request fan-out is the dominant wall-clock cost — and there was no escape hatch.

This PR adds the escape hatch without changing default behavior:
- Single-project tests and small `--all-projects` runs are unaffected.
- Container tests and large `--all-projects` runs default to the same throughput as today, but can opt into higher concurrency via the env var.

### Benchmark — `quay.io/wildfly/wildfly:34.0.1.Final-jdk21` (512 ScanResults)

`hyperfine --warmup 1 --runs 3` against a locally-built PR binary, varying only `SNYK_REQUEST_CONCURRENCY`. The `c=5` row is the default and reproduces the prior hard-coded behavior.

| Configuration | Mean wall-clock | vs default |
|---|---|---|
| default (`c=5`) | **68.04 s ± 1.72 s** | 1.00× |
| override (`c=10`) | **40.96 s ± 0.15 s** | **1.66× faster** (−40%) |
| override (`c=20`) | **25.13 s ± 0.48 s** | **2.71× faster** (−63%) |

Min/max within ±2% of mean across all configs — variance is tight. Standard deviation drops with higher concurrency (1.72s → 0.15s → 0.48s) because the wait-time component shrinks and the scan becomes more CPU-bound on the (much steadier) JAR-extraction work in `snyk-docker-plugin`.

(I re-validated end-to-end on the latest revision with 2 hyperfine runs per config; min(c=20)=53s < min(c=5)=64s confirms the env var continues to flow through Go → TS as expected.)

- `cliv2/internal/cliv2/cliv2.go` — new exported `ConfigKeyRequestConcurrency`; `fillEnvironmentFromConfig` forwarding; blacklist entry.
- `cliv2/internal/constants/constants.go` — `SNYK_INTERNAL_REQUEST_CONCURRENCY_ENV` constant.
- `cliv2/cmd/cliv2/main.go` — `AddAlternativeKeys` registration for the user-facing env var name.
- `src/lib/snyk-test/common.ts` — `getRequestConcurrency()` helper.
- `src/lib/snyk-test/run-test.ts` — single call-site swap.
- `cliv2/internal/cliv2/cliv2_test.go` — `Test_PrepareV1EnvironmentVariables_RequestConcurrency` (3 sub-cases: forwarded-when-set, omitted-when-unset, user-set-internal-var-stripped-and-reapplied).
- `test/jest/unit/lib/snyk-test/common.spec.ts` — 9 unit tests for the helper covering default, override, clamping, and invalid-input cases.

1. Run targeted unit tests:
   ```sh
   npx jest --selectProjects coreCli --testPathPattern 'test/jest/unit/lib/snyk-test/common.spec'
   cd cliv2 && go test ./internal/cliv2/ -run Test_PrepareV1EnvironmentVariables_RequestConcurrency
   ```
2. Optionally measure on a representative many-ScanResult image (Wildfly works well):
   ```sh
   hyperfine --warmup 1 --runs 3 \
     './binary-releases/snyk-... container test quay.io/wildfly/wildfly:34.0.1.Final-jdk21' \
     'SNYK_REQUEST_CONCURRENCY=10 ./binary-releases/snyk-... container test quay.io/wildfly/wildfly:34.0.1.Final-jdk21' \
     'SNYK_REQUEST_CONCURRENCY=20 ./binary-releases/snyk-... container test quay.io/wildfly/wildfly:34.0.1.Final-jdk21'
   ```
   (Use the Go-wrapped binary, not `node bin/snyk` directly — the Go wrapper is what reads `SNYK_REQUEST_CONCURRENCY` and forwards it as `SNYK_INTERNAL_REQUEST_CONCURRENCY`.)

## Risk assessment

**Low.** No default behavior change. The override is bounded to `[1, 50]`. The Go-side wiring is additive: the legacy CLI's env handling stays the same except for the new internal var (covered by the new Go test).

## Background

Supersedes #6747 (which targeted the wrong code path — `src/lib/ecosystems/test.ts:testDependencies` is unreachable from `snyk container test`; `getEcosystemForTest` returns null for docker, so container test goes through `runTest`/`pMap` instead).


bgardiner:
## Benchmark — `quay.io/wildfly/wildfly:34.0.1.Final-jdk21` (512 ScanResults)

`hyperfine --warmup 1 --runs 3` against the same locally-built PR A binary, varying only `SNYK_REQUEST_CONCURRENCY` to isolate the API concurrency change from any other build/runtime differences. The `c=5` row reproduces the prior hard-coded value, so it's an apples-to-apples baseline.

| Configuration | Mean wall-clock | vs baseline |
|---|---|---|
| baseline (`c=5`) | **68.04 s ± 1.72 s** | 1.00× |
| PR A default (`c=10`) | **40.96 s ± 0.15 s** | **1.66× faster** (−40%) |
| PR A override (`c=20`) | **25.13 s ± 0.48 s** | **2.71× faster** (−63%) |

Each config: 3 runs after 1 warmup; image pre-pulled. Min/max within ±2% of mean across all configs — variance is tight.

Standard deviation drops with higher concurrency (1.72s → 0.15s → 0.48s) because the wait-time component shrinks and the scan becomes more CPU-bound on the (much steadier) JAR-extraction work in `snyk-docker-plugin`.

bdemeo12 (approved):
Same questions as this PR: https://github.com/snyk/cli/pull/6757

But LGTM

jacek-rzrz (approved):

src/lib/snyk-test/common.ts
```diff
@@ -111,6 +111,28 @@ export type FailOn = 'all' | 'upgradable' | 'patchable';
...
+const MIN_REQUEST_CONCURRENCY = 1;
+const MAX_REQUEST_CONCURRENCY = 50;
+
+/**
+ * Returns the maximum number of in-flight Snyk dependency-test or
+ * dependency-monitor HTTP requests permitted at once. Override with the
+ * SNYK_REQUEST_CONCURRENCY environment variable; values are clamped to
+ * [MIN_REQUEST_CONCURRENCY, MAX_REQUEST_CONCURRENCY].
+ */
+export function getRequestConcurrency(): number {
```
PeterSchafer, line 126: Suggestion: There is already a similar configuration value in the [Golang CLI](https://github.com/snyk/go-application-framework/blob/main/pkg/app/app.go#L273). While this is currently not publicly documented, I think we should [inject it in the Legacy CLI](https://github.com/snyk/cli/blob/main/cliv2/internal/cliv2/cliv2.go#L389) and have the Golang configuration a single source of truth for this. We can also expose the config value but would probably want to frame it around `SNYK_MAX_CONCURRENCY` rather then focussed on requests.

src/lib/snyk-test/common.ts
```diff
@@ -111,6 +111,28 @@ export type FailOn = 'all' | 'upgradable' | 'patchable';
 export const RETRY_ATTEMPTS = 3;
 export const RETRY_DELAY = 500;
 
+const DEFAULT_REQUEST_CONCURRENCY = 10;
```
PeterSchafer: Thinking: Generally changing a default without visibility in the behaviour, we currently don't have any metrics on CPU consumption, makes me wonder if we should be more careful. The impact is quite huge as it affects all SCA and Container scans.

src/lib/snyk-test/common.ts
bgardiner: Thanks for the pointers. I didn't know about this config value. I agree having a single source of truth here makes sense. Before I plumb it through, I want to flag a semantic concern.

`MAX_THREADS` is documented as ["maximum number of threads that can be used by Extensions"](https://github.com/snyk/go-application-framework/blob/main/pkg/configuration/constants.go#L13) and defaults to `runtime.NumCPU()`. That makes sense for **CPU-bound parallelism**.

What this PR controls is **HTTP request concurrency** — how many in-flight `/test-dependencies` / `/monitor-dependencies` requests we allow at once. That's network I/O, not CPU-bound, so `runtime.NumCPU()` is a poor proxy and also somewhat unrelated to the number of threads being used.

Binding HTTP concurrency to `MAX_THREADS` would make the default machine-dependent. It would also tie an HTTP knob to a CPU knob, where in principle we might want to tune them independently (e.g., a small machine could still benefit from high HTTP concurrency without needing many CPU threads).

A few options I see:

1. **Keep as separate config** for now (`SNYK_MAX_CONCURRENCY`, Node-side env-var read). Acknowledges the semantic split; loses centralization.
2. **Bind to `MAX_THREADS` but override the default for the legacy CLI bridge** (e.g., `max(MAX_THREADS, 10)`). Preserves the source-of-truth idea for users who deliberately set it high, but also preserves the functional change in this PR. However, it's highjacking a single variable for two different uses.
3. **Create a new go-application-framework config key** (`MAX_HTTP_CONCURRENCY` or similar) that's separate from `MAX_THREADS`. Cleanest semantically; larger scope.

Am I understanding things correctly? What are you preferences from a CLI architecture perspective?

src/lib/snyk-test/common.ts
bgardiner: I don't think there'll be significant CPU impact. Node executes JS on a single thread, and the code path this PR touches spends nearly all its time waiting on HTTP responses, not in JS execution.

On the "affects all SCA and Container scans" concern, I agree. A few things that make me comfortable with the bump:
- headroom to throttling is seemingly large. I put more discussion [here](https://github.com/snyk/cli/pull/6757#discussion_r3182732814). The api-gateway rate limit on /v1/test* and /v1/monitor* is keyed per principal_id (per-token), default high bucket: 200 req/s burst, 2000/min, 60000/hour. Even at the override ceiling (50), a single CLI scan stays at ≤25% of the per-second cap.
- 5 is a conservatively low bound and 10 is still quite modest as a default.
- OS flows team reviewed and approved the PR 😄 

Other options:
- we could land a 5 to 8 bump first and revisit after we have telemetry?
- add a feature-flag-style override/escape hatch to bump it back to 5
- any other ideas?

What do you think?

PeterSchafer: Makes sense for a single CLI run. A lot of our user actually run the CLI highly concurrent and already experience rate limiting. In the combination in the best case, the increase of request parallelism and rate limiting might just equal out, in the worst case the rate limiting will eventually exceed the retry limit and fail the CLI.

src/lib/snyk-test/common.ts
PeterSchafer: I think semantically and technically you are right and point 3 would make sense. 

Besides this, there is a usability question, as a user specifying `MAX_HTTP_CONCURRENCY` I would expect that this value is actually the maximum. 
Currently given the codebase, I don't know how many request are actually happening in parallel and if we can guarantee this. A solution could be to limit it in the network stack but this could have some unforeseen impact as well.

Architecturally Application configuration via Environment Variables and configuration handling is done through the Golang CLI and its configuration package. This way, the decision on where values are coming from (Env var, config file, cmd arg ..) is left to the application. Making this available in the Typescript CLI via internal environment variables makes sense though.

PeterSchafer: Is there an option to separate concerns and decouple the configurability and the increase of the default value?

PeterSchafer (approved):

cliv2/internal/cliv2/cliv2.go
```diff
@@ -66,6 +66,13 @@ const (
 const (
 	configKeyErrFile         = "INTERNAL_ERR_FILE_PATH"
 	ERROR_HAS_BEEN_DISPLAYED = "hasBeenDisplayed"
+	// ConfigKeyRequestConcurrency is the configuration key holding the
```
CatalinSnyk, line 69: small nit: The comment is a bit verbose in my oppinion, not blocking

CatalinSnyk: nvm wrong PR 😄

PR #6748, opened 2026-04-24
feat: parallelize container monitor dependency requests

bgardiner:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages
  are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing
  _what_ was changed, not _how_.
- [x] Includes detailed description of changes
- [x] Contains risk assessment (Low | Medium | High)
- [ ] Highlights breaking API changes (if applicable)
- [x] Links to automated tests covering new functionality
- [x] Includes manual testing instructions (if necessary)
- [ ] Updates relevant GitBook documentation (PR link: ___)
- [x] Includes product update to be announced in the next stable release notes

## What does this PR do?

This PR improves performance for ecosystem/container monitoring by changing the `/monitor-dependencies` request flow from fully sequential to hybrid execution:

- For each path, it always executes the first `ScanResult` request first (base OS project ordering retained)
- It then executes the remaining `ScanResult` requests in parallel
- It preserves output ordering by collecting parallel responses and appending them in scan order
- It preserves existing error semantics:
  - 401 still throws `AuthFailedError`
  - other 4xx errors still throw `MonitorError`
  - non-4xx failures are collected as monitor errors and execution continues

## Where should the reviewer start?

- `src/lib/ecosystems/monitor.ts`
- `test/jest/unit/ecosystems-monitor-docker.spec.ts`

## How should this be manually tested?

1. Run monitor unit tests:
   - `npm run test:unit -- --runTestsByPath test/jest/unit/ecosystems-monitor-docker.spec.ts`
2. Optionally run a representative multi-project container monitor command and compare runtime:
   - `snyk container monitor <image>`

## What's the product update that needs to be communicated to CLI users?

Container monitor execution is now faster for images that produce multiple scan results. The CLI now keeps base OS processing first and parallelizes the remaining dependency API requests.

## Risk assessment (Low | Medium | High)?

Low. This is a scoped orchestration change with focused tests that verify ordering and existing error semantics.

## Any background context you want to provide?

Large images that produce many scan results can experience long monitor durations due to sequential API calls; this change addresses that bottleneck.

## What are the relevant tickets?

N/A

[...]

PR #6757, opened 2026-04-29
feat: parallelize container monitor dependency requests

bgardiner:
## What does this PR do?

Rewrites `monitorDependencies` in `src/lib/ecosystems/monitor.ts` to fan out per-`ScanResult` `PUT /monitor-dependencies` calls in parallel via `pMap`, bounded by the `getRequestConcurrency()` helper introduced in #6756 (default **5**, user-overridable via `SNYK_REQUEST_CONCURRENCY`).

Effective change for users: container monitor goes from sequential (1 in-flight request) to **5 in-flight by default**, with the same env-var override surface as `snyk container test`.

- Per-ScanResult work is extracted into a small `monitorOneScanResult` helper for testability.
- Outer loop over paths still runs sequentially (one path per `monitor` invocation is the common case anyway).
- Inner loop is now `pMap(scanResults, ..., { concurrency })`.

This PR is **based on `feat/snyk-request-concurrency` (#6756)** so its diff shows only the monitor-flow changes.

## Why?

`snyk container monitor` previously ran fully sequentially — one full RTT per ScanResult — because of a nested `for...of await` loop. For large images with many ScanResults (e.g. fat-JAR-heavy images that produce one ScanResult per directory of JARs in `snyk-docker-plugin`'s `groupJarFingerprintsByPath`), this is the dominant wall-clock cost.

### Benchmark — `quay.io/wildfly/wildfly:34.0.1.Final-jdk21` (512 ScanResults)

Sequential baseline measured against the released CLI (`snyk` v1.1303.2, prior code path). PR configurations measured against the locally-built PR binary, varying only `SNYK_REQUEST_CONCURRENCY`. Re-monitoring updates existing project snapshots rather than creating duplicates.

| Configuration | Wall-clock | vs sequential |
|---|---|---|
| sequential (`snyk` v1.1303.2, prior code) | **481 s (8m 1s)** | 1.00× |
| **default (`c=5`, this PR)** | **104 s** | **4.6× faster** |
| override (`c=10`) | 54 s | 8.9× faster |
| override (`c=20`) | 29 s | 16.6× faster |

Each PR run successfully updated all 512 snapshots. The sequential run produced 510 snapshots; possibly minor version drift between v1.1303.2 and the PR binary's bundled `snyk-docker-plugin`. No errors logged; doesn't affect the timing comparison.

The default-of-5 row is the new behavior users will see without any opt-in.

## Error semantics preserved

- **401** → still throws `AuthFailedError` (terminates the run).
- **Other 4xx** → still throws `MonitorError` (terminates the run via `pMap`'s default fail-fast on rejection).
- **5xx and other non-4xx** errors are accumulated per-ScanResult into the `errors` array, matching prior continue-on-error behavior.

## Result ordering preserved

`pMap` returns results in input order regardless of completion order, so the formatted output remains deterministic.

- `src/lib/ecosystems/monitor.ts` — `monitorDependencies` rewrite + new `monitorOneScanResult`.
- `test/jest/unit/ecosystems-monitor-docker.spec.ts` — 5 new tests in a `parallelization of monitor-dependencies requests` describe:
  - Concurrency cap honored at default 5.
  - `SNYK_INTERNAL_REQUEST_CONCURRENCY=3` override respected (the internal env var the TS code reads; the user-facing `SNYK_REQUEST_CONCURRENCY` is plumbed through Go in #6756).
  - Result ordering preserved when responses arrive out of order.
  - 4xx fail-fast (throws `MonitorError`).
  - 5xx accumulation (continues, returns errors).

1. Run unit tests:
   ```sh
   npx jest --selectProjects coreCli --testPathPattern 'test/jest/unit/ecosystems-monitor-docker'
   ```
2. Optional smoke test against a throwaway org with a multi-project image — use the Go-wrapped binary so `SNYK_REQUEST_CONCURRENCY` is plumbed through:
   ```sh
   ./binary-releases/snyk-... container monitor <image> --org=<throwaway-org>
   SNYK_REQUEST_CONCURRENCY=20 ./binary-releases/snyk-... container monitor <image> --org=<throwaway-org>
   ```

## Risk assessment

**Low.** The change is scoped to a single function. Error semantics are preserved (verified by tests). Concurrency is bounded and configurable. `pMap` is already used by `runTest` so it's a familiar dependency in this codebase.

## Background

**Depends on / based on #6756**, which:
- introduces the `getRequestConcurrency` helper and clamps,
- pipes the user-facing `SNYK_REQUEST_CONCURRENCY` env var through the Go-side GAF configuration so the legacy CLI receives the resolved value via the internal `SNYK_INTERNAL_REQUEST_CONCURRENCY` env var.

Supersedes #6748:
- That PR used unbounded `Promise.all`, which on a 500-ScanResult image would fire 500+ concurrent PUTs — a real risk of 429s, socket exhaustion, or thundering-herd backend load.
- That PR added a "base scan first, then `Promise.all` the rest" hybrid pattern; once results are kept in input order via `pMap`, the special case isn't load-bearing.

### DIFF
--- a/src/lib/ecosystems/monitor.ts
+++ b/src/lib/ecosystems/monitor.ts
@@ -1,8 +1,10 @@
 import { InspectResult } from '@snyk/cli-interface/legacy/plugin';
 import chalk from 'chalk';
+import * as pMap from 'p-map';
 import config from '../config';
 import { isCI } from '../is-ci';
 import { makeRequest } from '../request/promise';
+import { getRequestConcurrency } from '../snyk-test/common';
 import {
   Contributor,
   MonitorOptions,
@@ -148,52 +150,81 @@ async function monitorDependencies(
 ): Promise<[EcosystemMonitorResult[], EcosystemMonitorError[]]> {
   const results: EcosystemMonitorResult[] = [];
   const errors: EcosystemMonitorError[] = [];
+  const concurrency = getRequestConcurrency();
+
   for (const [path, scanResults] of Object.entries(scans)) {
     await spinner(`Monitoring dependencies in ${path}`);
-    for (const scanResult of scanResults) {
-      const monitorDependenciesRequest =
-        await generateMonitorDependenciesRequest(scanResult, options);
-
-      const configOrg = config.org ? decodeURIComponent(config.org) : undefined;
-
-      const payload = {
-        method: 'PUT',
-        url: `${config.API}/monitor-dependencies`,
-        json: true,
-        headers: {
-          'x-is-ci': isCI(),
-          authorization: getAuthHeader(),
-        },
-        body: monitorDependenciesRequest,
-        qs: {
-          org: options.org || configOrg,
-        },
-      };
-      try {
-        const response =
-          await makeRequest<MonitorDependenciesResponse>(payload);
-        results.push({
-          ...response,
-          path,
-          scanResult,
-        });
-      } catch (error) {
-        if (error.code === 401) {
-          throw AuthFailedError();
-        }
-        if (error.code >= 400 && error.code < 500) {
-          throw new MonitorError(error.code, error.message);
-        }
-        errors.push({
-          error: 'Could not monitor dependencies in ' + path,
-          path,
-          scanResult,
-        });
+    const perScanResults = await pMap(
+      scanResults,
+      (scanResult) => monitorOneScanResult(scanResult, options, path),
+      { concurrency },
+    );
+    for (const r of perScanResults) {
+      if (r.result) {
+        results.push(r.result);
+      }
+      if (r.error) {
+        errors.push(r.error);
       }
     }
     spinner.clearAll();
   }
   return [results, errors];
+}
+
+async function monitorOneScanResult(
+  scanResult: ScanResult,
+  options: Options & MonitorOptions,
+  path: string,
+): Promise<{
+  result?: EcosystemMonitorResult;
+  error?: EcosystemMonitorError;
+}> {
+  const monitorDependenciesRequest = await generateMonitorDependenciesRequest(
+    scanResult,
+    options,
+  );
+
+  const configOrg = config.org ? decodeURIComponent(config.org) : undefined;
+
+  const payload = {
+    method: 'PUT',
+    url: `${config.API}/monitor-dependencies`,
+    json: true,
+    headers: {
+      'x-is-ci': isCI(),
+      authorization: getAuthHeader(),
+    },
+    body: monitorDependenciesRequest,
+    qs: {
+      org: options.org || configOrg,
+    },
+  };
+
+  try {
+    const response = await makeRequest<MonitorDependenciesResponse>(payload);
+    return {
+      result: {
+        ...response,
+        path,
+        scanResult,
+      },
+    };
+  } catch (error) {
+    if (error.code === 401) {
+      throw AuthFailedError();
+    }
+    if (error.code >= 400 && error.code < 500) {
+      throw new MonitorError(error.code, error.message);
+    }
+    return {
+      error: {
+        error: 'Could not monitor dependencies in ' + path,
+        path,
+        scanResult,
+      },
+    };
+  }
 }
 
 export async function getFormattedMonitorOutput(
--- a/test/jest/unit/ecosystems-monitor-docker.spec.ts
+++ b/test/jest/unit/ecosystems-monitor-docker.spec.ts
@@ -285,4 +285,160 @@ describe('monitorEcosystem docker/container', () => {
     );
     expect(parsedOutput.projectName).not.toBe('my-custom-project-name');
   });
+
+  describe('parallelization of monitor-dependencies requests', () => {
+    const ORIGINAL_CONCURRENCY = process.env.SNYK_INTERNAL_REQUEST_CONCURRENCY;
+
+    afterEach(() => {
+      if (ORIGINAL_CONCURRENCY === undefined) {
+        delete process.env.SNYK_INTERNAL_REQUEST_CONCURRENCY;
+      } else {
+        process.env.SNYK_INTERNAL_REQUEST_CONCURRENCY = ORIGINAL_CONCURRENCY;
+      }
+    });
+
+    function makeMavenScanResult(targetFile: string): ScanResult {
+      const base = readJsonFixture(
+        'maven-project-0-dependencies-scan-result.json',
+      ) as ScanResult;
+      return {
+        ...base,
+        identity: { ...base.identity, targetFile },
+      };
+    }
+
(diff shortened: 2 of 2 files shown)


bgardiner:
## Benchmark — `quay.io/wildfly/wildfly:34.0.1.Final-jdk21` (512 ScanResults)

Sequential baseline measured against the released CLI in PATH (`snyk` v1.1303.2, prior code path). PR-B configurations measured against the locally-built PR binary, varying only `SNYK_REQUEST_CONCURRENCY`. Re-monitoring updates existing project snapshots rather than creating duplicates.

| Configuration | Wall-clock | vs sequential |
|---|---|---|
| sequential (`snyk` v1.1303.2, prior code) | **481 s (8m 1s)** | 1.00× |
| `SNYK_REQUEST_CONCURRENCY=5` (PR B) | **104 s** | **4.6× faster** |
| default `c=10` (PR B) | **54 s** | **8.9× faster** |
| `SNYK_REQUEST_CONCURRENCY=20` (PR B) | **29 s** | **16.6× faster** |

Each PR-B run successfully updated all 512 snapshots. The sequential run produced 510 snapshots; possibly minor version drift between v1.1303.2 and the PR binary's bundled `snyk-docker-plugin`. No errors logged; doesn't affect the timing comparison.

src/lib/snyk-test/common.ts
```diff
@@ -111,6 +111,28 @@ export type FailOn = 'all' | 'upgradable' | 'patchable';
 export const RETRY_ATTEMPTS = 3;
 export const RETRY_DELAY = 500;
 
+const DEFAULT_REQUEST_CONCURRENCY = 10;
+const MIN_REQUEST_CONCURRENCY = 1;
+const MAX_REQUEST_CONCURRENCY = 50;
+
+/**
+ * Returns the maximum number of in-flight Snyk dependency-test or
```
bdemeo12, line 119: Do you know if these APIs rate limit? Could we run into this if a customer (like Walmart) sets this to 50?

src/lib/snyk-test/run-test.ts
```diff
@@ -39,6 +39,7 @@ import { isCI } from '../is-ci';
 import {
   RETRY_ATTEMPTS,
   RETRY_DELAY,
+  getRequestConcurrency,
```
bdemeo12, line 42: Do we know why concurrency was not enabled for monitor-deps ?

bdemeo12 (commented):
LGTM! Just a few questions!

src/lib/snyk-test/common.ts
bgardiner: Good question. I am worried about that. I'm not sure how the rate limiting occurs (e.g., dest endpoint, source IP, etc.). I need to look into this. I'm hoping going from 5->10 as a default is relatively safe, but I don't think we can do more without some testing.

src/lib/snyk-test/run-test.ts
bgardiner: I assume just an oversight. I don't know

src/lib/snyk-test/common.ts
bgardiner: ## Rate-limit investigation summary

I dug into the api-gateway config (`github.com/snyk/api-gateway`) to confirm the bumped concurrency won't push CLI users over rate limits. Here's what I found.

### Rate-limit policy for our endpoints

`POST /v1/test-dependencies`, `POST /v1/test-dep-graph`, and `PUT /v1/monitor-dependencies` all match the default `/v1/test*` and `/v1/monitor*` prefix routes in [`helm/templates/route-tables/api-v1.yaml`](https://github.com/snyk/api-gateway/blob/main/helm/templates/route-tables/api-v1.yaml). None of them have a custom `envoyMetadata.snyk.ratelimiter.keys.bucket` override, so they fall into the default `high` bucket of the `api-v1` rate-limit config.

- **Keyed by `principal_id`** — per-authenticated-user (per-token), per [`docs/rate-limiting.md`](https://github.com/snyk/api-gateway/blob/main/docs/rate-limiting.md). Not per-IP, not per-cluster.
- **Default bucket: `high`**. Limits per token:

| Window     | Limit                                    |
| ---------- | ---------------------------------------- |
| Per second | **200 req/s**                            |
| Per minute | **2,000 req/min** (≈ 33/s sustained)     |
| Per hour   | **60,000 req/hour** (≈ 16.7/s sustained) |

All three apply simultaneously (burst, 1-min, 1-hour ceilings).

The only adjacent exception is `lowerDepGraphRateLimit.enabled` (`polaris-prod-pc-pc01-1` only), which drops `/v1/monitor/dep-graph` (with slash, **not** our endpoint) to `extra_low`. Doesn't affect us.

### Headroom analysis for this PR

The default change is 5 → 10 in-flight requests per CLI invocation. The proposed env override (`SNYK_REQUEST_CONCURRENCY`) caps at 50.

| Scenario                                    | Observed peak rate  | Limit           | Headroom |
| ------------------------------------------- | ------------------- | --------------- | -------- |
| Wildfly container test, c=10                | ~13 req/s sustained | 200 req/s burst | 15×      |
| Wildfly container test, c=20                | ~20 req/s sustained | 200 req/s burst | 10×      |
| Wildfly container test, c=50 (max override) | ~50 req/s burst     | 200 req/s burst | 4×       |

For a single invocation, no setting in the supported range gets close to the per-second ceiling.

### Other context from registry-side telemetry

For completeness, I checked Datadog for actual 429s on the registry pod:

| Endpoint                    | 30-day requests | 30-day 429s | 429 rate |
| --------------------------- | --------------- | ----------- | -------- |
| `POST /test-dep-graph`      | 9.0 M           | 24          | 0.00027% |
| `POST /test-dependencies`   | 1.65 M          | 22          | 0.0013%  |
| `PUT /monitor-dependencies` | 1.04 M          | 0           | 0%       |

Caveat: the Envoy gateway returns 429s to clients **before** the request reaches the registry pod, so registry-side 429 counts under-represent total rejections. At the cluster level, ~1% of total traffic to `static-default-registry-web-30100_api-gateway` (~807M requests over 30 days) is rejected — but that's all paths combined, and per-endpoint attribution isn't available at the metric layer.

### Verdict

**The bump from 5 → 10 (and the 50 cap on the env override) is well within the per-token rate limit policy for these endpoints.** Even at the highest supported override, a single CLI burst stays at ≤25% of the per-second ceiling and doesn't change per-minute/per-hour quota consumption.

### Recommended follow-up (optional, separate PR)

The current retry logic in `run-test.ts:265-296` retries on 500/502/503 but **not** 429. If a heavy user (CI fleet sharing a token) ever hits the per-hour bucket — which is plausible regardless of this PR — they get a hard failure today.

CatalinSnyk (approved):

PR #6758, opened 2026-04-30
chore: upgrade Node.js runtime from 20 to 24 LTS

danskmt:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing _what_ was changed, not _how_.
- [x] Includes detailed description of changes
- [x] Contains risk assessment (Low | Medium | High)
- [x] Highlights breaking API changes (if applicable)
- [ ] Links to automated tests covering new functionality
- [x] Includes manual testing instructions (if necessary)
- [ ] Updates relevant GitBook documentation (PR link: ___)
- [ ] Includes product update to be announced in the next stable release notes

## What does this PR do?

Upgrades the Snyk CLI to Node.js 24 LTS (v24.15.0). Node 20 reaches EOL on April 30, 2026.

### Changes

- `.nvmrc`: `20.11.1` → `24.15.0`
- `package.json` engines: `"^20"` → `"^24"`
- `package-lock.json`: regenerated with Node 24
- `scripts/windows/install-node.ps1`: version and SHA256 updated for `node-v24.15.0-x64.msi`
- `.circleci/config.yml`: Docker image tags updated to `20260430-081507` (images with Node 24)
- `.github/workflows/check-dependencies.yml`: `16.16.0` → `24.15.0`
- `.github/workflows/danger-zone.yml`: `18.19.1` → `24.15.0`
- `.github/workflows/snyk-protect-production-smoke-tests.yml`: added Node 22 and 24 to smoke test matrix

### Prerequisites

- New `cli-build-private` and `cli-build-private-arm64` Docker images must be built and pushed with tag `20260430-081507` before CI will pass.
- `snyk-mvn-plugin` and `snyk-nodejs-plugin` should be released with updated `engines.node: "^24"` to eliminate EBADENGINE warnings.

## Where should the reviewer start?

- `.nvmrc` and `package.json` — the core version change
- `scripts/windows/install-node.ps1` — Windows CI installer version + SHA
- `.circleci/config.yml` — Docker image tag update

## How should this be manually tested?

Branch name includes `e2e` to trigger the full CircleCI environment test matrix (Docker, Windows, macOS, Ubuntu, Alpine, FIPS, scratch container, static Linux).

## Risk assessment

[...]

PR #6764, opened 2026-05-04
chore: Windows CI Node install and snyk-env.sh BOM

danskmt:

## What does this PR do?

Hardens **Windows CI** for the Node toolchain used when building the CLI:

- **Node is installed only via nvm-windows** (`nvm install` / `nvm use`) using the version pinned in **`.nvmrc`**. There is **no MSI** path or SHA256 check. If **`nvm`** is not on `PATH`, the script **fails fast** with a clear error.
- **Diagnostic logging:** `nvm` binary path, **`NVM_HOME`**, **`NVM_SYMLINK`**, and **`[nvm-cache] HIT`** / **`[nvm-cache] MISS`** so logs show whether Node was downloaded this run or reused from the nvm store.
- **CircleCI cache:** Windows tools cache bumped to **`windows-tools-cache-v4-{{ arch }}-{{ checksum ".nvmrc" }}`** so caches invalidate when Node changes; **`save_cache`** includes **`<< windows_cache_dir >>`** and **`C:\ProgramData\nvm`** so the nvm install tree persists between jobs.

This addresses the class of failures where **`node.exe`** and **npm’s own files** came from **different** layouts (e.g. MSI vs NVM), by standardizing on **one** manager (nvm) on CI.

**Note:** `scripts/windows/install-node.ps1` in this PR **does not** append Node to `snyk-env.ps1` / `snyk-env.sh`; it only runs nvm and verifies **`C:\Program Files\nodejs\node.exe`**. Downstream steps rely on **`nvm use`** having switched the **junction** and on the image / default **`PATH`** including **`Program Files\nodejs`** as usual. If anything regresses, we can restore explicit PATH lines or document the assumption.

## Changes

- **`scripts/windows/install-node.ps1`** — nvm-only install from `.nvmrc`; verification; logging; **no MSI**.
- **`.circleci/config.yml`** — `install-deps-windows-native-build`: cache key **`v4`** + **`{{ checksum ".nvmrc" }}`**; **`save_cache.paths`** adds **`C:\ProgramData\nvm`**.

1. `scripts/windows/install-node.ps1`
2. `.circleci/config.yml` — `install-deps-windows-native-build` (restore/save cache)

1. Run a **Windows** job that uses **`install-deps-windows-native-build`** (e.g. full Windows build).
2. In **Install Node.js (native)**, confirm **`nvm-windows:`**, **`NVM_HOME:`**, and **`[nvm-cache]`** HIT/MISS.
3. Confirm **legacy TS / npm** steps still succeed (no **`Class extends value undefined`** from npm).
4. On first green run, confirm **`NVM_HOME`** in logs matches **`C:\ProgramData\nvm`**; if not, adjust the **`save_cache`** path in `.circleci/config.yml`.

None. CI/build plumbing only.

**Low** — CI scripts and config only; no change to shipped CLI behaviour.

CircleCI Windows images ship **nvm-windows**; a second **MSI** installer was redundant and risked inconsistent installs. **`{{ checksum ".nvmrc" }}`** ties the cache to the pinned Node version.

Related: [CLI-1469](https://snyksec.atlassian.net/browse/CLI-1469). Prior discussion: https://github.com/snyk/cli/pull/6758

[CLI-1469](https://snyksec.atlassian.net/browse/CLI-1469)

## Screenshots (if appropriate)

N/A

[CLI-1469]: https://snyksec.atlassian.net/browse/CLI-1469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -276,7 +276,7 @@ commands:
       - restore_cache:
           name: Restoring Windows tools cache
           keys:
-            - windows-tools-cache-v2-{{ arch }}
+            - windows-tools-cache-v4-{{ arch }}-{{ checksum ".nvmrc" }}
       - run:
           name: Install Node.js (native)
           shell: powershell
@@ -315,9 +315,10 @@ commands:
             .\scripts\windows\ensure-python-uv.ps1
       - save_cache:
           name: Saving Windows tools cache
-          key: windows-tools-cache-v2-{{ arch }}
+          key: windows-tools-cache-v4-{{ arch }}-{{ checksum ".nvmrc" }}
           paths:
             - << pipeline.parameters.windows_cache_dir >>
+            - C:\ProgramData\nvm
 
   install-deps-windows-native-full-signing:
     steps:
--- a/scripts/windows/install-node.ps1
+++ b/scripts/windows/install-node.ps1
@@ -13,10 +13,7 @@ if (Test-Path $envScript) {
 }
 
 try {
-  $expectedNodeVersion = '20.11.1'
-  $expectedSha256 = 'c54f5f7e2416e826fd84e878f28e3b53363ae9c3f60a140af4434b2453b5ae89'
-
-  # Resolve repo root from script location (script is in .circleci/windows)
+  # Resolve repo root from script location (script is in scripts/windows)
   $repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..\..')
   $nvmrcPath = Join-Path $repoRoot '.nvmrc'
 
@@ -24,42 +21,33 @@ try {
     throw ".nvmrc not found at $nvmrcPath"
   }
 
-  $nvmVersion = (Get-Content $nvmrcPath -Raw).Trim()
-  if ([string]::IsNullOrWhiteSpace($nvmVersion)) {
+  $nodeVersion = (Get-Content $nvmrcPath -Raw).Trim().TrimStart('v')
+  if ([string]::IsNullOrWhiteSpace($nodeVersion)) {
     throw ".nvmrc is empty at $nvmrcPath"
   }
 
-  if ($nvmVersion -ne $expectedNodeVersion) {
-    throw ".nvmrc version '$nvmVersion' does not match expected Node.js version '$expectedNodeVersion' used by Windows CI."
+  # nvm-windows is the only supported installer here. It manages
+  # C:\Program Files\nodejs as a junction; using a parallel MSI install on the
+  # same machine leaves stale npm internals.
+  $nvmExe = Get-Command nvm -ErrorAction SilentlyContinue
+  if (-not $nvmExe) {
+    throw "nvm-windows is required but was not found on PATH. Ensure the runner image provides nvm-windows."
   }
 
-  $nodeVersion = $expectedNodeVersion
+  Write-Host "nvm-windows: $($nvmExe.Source)"
+  Write-Host "NVM_HOME: $env:NVM_HOME"
+  Write-Host "NVM_SYMLINK: $env:NVM_SYMLINK"
 
-  $msiPath = Join-Path $cacheDir "node-v$nodeVersion-x64.msi"
-
-  if (-not (Test-Path $cacheDir)) {
-    New-Item -ItemType Directory -Path $cacheDir | Out-Null
+  $nvmList = & nvm list 2>&1 | Out-String
+  if ($nvmList -notmatch ('\b' + [regex]::Escape($nodeVersion) + '\b')) {
+    Write-Host "[nvm-cache] MISS: Node.js v$nodeVersion not present; installing via nvm..."
+    & nvm install $nodeVersion
+  } else {
+    Write-Host "[nvm-cache] HIT: Node.js v$nodeVersion already installed in nvm; skipping download."
   }
 
-  if (-not (Test-Path $msiPath)) {
-    Write-Host "Downloading Node.js v$nodeVersion (x64 MSI)..."
-    $url = "https://nodejs.org/dist/v$nodeVersion/node-v$nodeVersion-x64.msi"
-    curl.exe -L $url -o $msiPath
-  }
-
-  Write-Host 'Verifying Node.js installer checksum...'
-  $hash = Get-FileHash -Path $msiPath -Algorithm SHA256
-  if ($hash.Hash.ToLower() -ne $expectedSha256.ToLower()) {
-    throw "Checksum verification failed for $msiPath. Expected $expectedSha256 but got $($hash.Hash.ToLower())."
-  }
-
-  Write-Host "Installing Node.js v$nodeVersion..."
-  $msiArgs = "/i `"$msiPath`" /qn /norestart"
-  $process = Start-Process -FilePath msiexec.exe -ArgumentList $msiArgs -PassThru
-  $process.WaitForExit()
-  if ($process.ExitCode -ne 0) {
-    throw "Node.js MSI installer exited with code $($process.ExitCode)."
-  }
+  Write-Host "Activating Node.js v$nodeVersion via nvm..."
+  & nvm use $nodeVersion
 
   # Verify installation using the known default installation path
   $nodeExe = "C:\Program Files\nodejs\node.exe"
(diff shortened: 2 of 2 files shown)


danskmt:
> [**Integrity Verification Bypass**](https://github.com/snyk/cli/pull/6764/changes#diff-0c7f3b471c49319da48d1dcddcbf297dcf033654f431697c7a6801761a2779c3R35-R48) 🟠 [major]
> The new logic for `nvm-windows` detection bypasses the SHA256 checksum verification entirely. While the MSI fallback still checks the hash, the preferred `nvm install` path downloads and activates Node.js binaries without verifying their integrity. This is a regression compared to the previous version and other tool installation scripts in the same directory (e.g., `install-gradle.ps1`) that enforce checksums for all downloads.
> 
> ```powershell
> if ($nvmExe) {
>   Write-Host "nvm-windows detected; using nvm to install and activate Node.js v$nodeVersion..."
> 
>   $nvmList = & nvm list 2>&1 | Out-String
>   if ($nvmList -notmatch [regex]::Escape($nodeVersion)) {
>     Write-Host "Installing Node.js v$nodeVersion via nvm..."
>     & nvm install $nodeVersion
>   } else {
>     Write-Host "Node.js v$nodeVersion already installed in nvm."
>   }
> 
>   Write-Host "Activating Node.js v$nodeVersion via nvm..."
>   & nvm use $nodeVersion
> } else {
> ```
> 
> [**Maintenance Landmine**](https://github.com/snyk/cli/pull/6764/changes#diff-0c7f3b471c49319da48d1dcddcbf297dcf033654f431697c7a6801761a2779c3R16-R27) 🟠 [major]
> The hardcoded `$expectedSha256` for the MSI fallback is no longer guarded by an explicit version check. In the previous code, a mismatch between `.nvmrc` and `$expectedNodeVersion` would trigger a clear error. Now, updating `.nvmrc` without manually updating `$expectedSha256` will cause any CI runner using the MSI fallback to fail with an opaque checksum verification error during Node.js download.
> 
> ```powershell
> # MSI fallback only: SHA256 of node-v{version}-x64.msi for the version in .nvmrc (nodejs.org). Update when bumping Node in .nvmrc.
> $expectedSha256 = 'c54f5f7e2416e826fd84e878f28e3b53363ae9c3f60a140af4434b2453b5ae89'
> 
> # Resolve repo root from script location (script is in scripts/windows)
> $repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..\..')
> $nvmrcPath = Join-Path $repoRoot '.nvmrc'
> 
> if (-not (Test-Path $nvmrcPath)) {
>   throw ".nvmrc not found at $nvmrcPath"
> }
> 
> $nodeVersion = (Get-Content $nvmrcPath -Raw).Trim().TrimStart('v')
> ```
> 
> [**Loose Version Match**](https://github.com/snyk/cli/pull/6764/changes#diff-0c7f3b471c49319da48d1dcddcbf297dcf033654f431697c7a6801761a2779c3R39-R39) 🟡 [minor]
> The `nvm list` check uses a loose regex match against the version string. If `$nodeVersion` is '20.11.1' and `nvm list` contains '20.11.10', the `-notmatch` check will evaluate to false (a match is found), causing the script to skip installing the specific required version. It should use word boundaries or an exact match against the listed versions.

1. NVM manages its own downloads over HTTPS from nodejs.org; SHA256 verification is for the manual MSI fallback only.
2. That was the same behavior before, requiring to update both. Here we'll use .nvmrc as the source of node version.

scripts/windows/install-node.ps1
```diff
@@ -13,52 +13,66 @@ if (Test-Path $envScript) {
...
     throw ".nvmrc is empty at $nvmrcPath"
   }
 
-  if ($nvmVersion -ne $expectedNodeVersion) {
-    throw ".nvmrc version '$nvmVersion' does not match expected Node.js version '$expectedNodeVersion' used by Windows CI."
-  }
+  # Prefer nvm-windows if available — it manages C:\Program Files\nodejs as a junction.
+  # Using the MSI on an NVM-managed machine leaves stale npm internals from the old version.
+  $nvmExe = Get-Command nvm -ErrorAction SilentlyContinue
+  if ($nvmExe) {
```
PeterSchafer: Question: It seems that nvm is not using the cache path. We use the cache path to limit dependencies to 3rd party resources. Is there a reason why this is not cached?

scripts/windows/install-node.ps1
```diff
@@ -13,52 +13,66 @@ if (Test-Path $envScript) {
...
+      Write-Host "Installing Node.js v$nodeVersion via nvm..."
+      & nvm install $nodeVersion
+    } else {
+      Write-Host "Node.js v$nodeVersion already installed in nvm."
+    }
 
-  $nodeVersion = $expectedNodeVersion
+    Write-Host "Activating Node.js v$nodeVersion via nvm..."
+    & nvm use $nodeVersion
+  } else {
```
PeterSchafer, line 45: Question: Do we really need to maintain two different download solutions here? It seems redundant/

scripts/windows/install-node.ps1
```diff
@@ -13,52 +13,66 @@ if (Test-Path $envScript) {
...
+  # Prefer nvm-windows if available — it manages C:\Program Files\nodejs as a junction.
+  # Using the MSI on an NVM-managed machine leaves stale npm internals from the old version.
+  $nvmExe = Get-Command nvm -ErrorAction SilentlyContinue
+  if ($nvmExe) {
+    Write-Host "nvm-windows detected; using nvm to install and activate Node.js v$nodeVersion..."
+
+    $nvmList = & nvm list 2>&1 | Out-String
+    if ($nvmList -notmatch ('\b' + [regex]::Escape($nodeVersion) + '\b')) {
+      Write-Host "Installing Node.js v$nodeVersion via nvm..."
+      & nvm install $nodeVersion
```
PeterSchafer: Question: Does nvm introduce a new dependency that could introduce flakiness to the pipeline again?

PeterSchafer: But if we use caching, it is probably not a big issue.

scripts/windows/install-node.ps1
danskmt: Yeah, that's correct.

For the MSI, $cacheDir is used for loading/saving snyk-env.ps1 / snyk-env.sh and the downloaded .msi.

About NVM, it stores versions under its own layout (typically C:\ProgramData\nvm\...) and points C:\Program Files\nodejs at the active version. It does not put those bits under tools-cache. Would you like to have that in this PR?

scripts/windows/install-node.ps1
danskmt: It's like a "primary" vs "fallback". In this PR we are adding the "primary" getting from NVM, and as "fallback" the previous implementation with MSI. The conflict happens if the machine contains NVM.

scripts/windows/install-node.ps1
danskmt: Persisting C:\tools-cache helps env scripts + MSI reuse. It is currently not helping caching NVM's install.
About flakiness, I don't think so.

scripts/windows/install-node.ps1
PeterSchafer: I think we don't need the fallback, as we can define nvm as a dependency here. Otherwise the fallback will never be used and just bloats the code base.

scripts/windows/install-node.ps1
PeterSchafer: We need to use the cache, otherwise, we would make a step back and add another live dependency to our pipeline, that might block pipelines if not available.

scripts/windows/install-node.ps1
PeterSchafer: if you can use nvm with caching, let's use this as our one and only way to install nodejs here

.circleci/config.yml
```diff
@@ -261,7 +261,9 @@ commands:
                     New-Item -Path $bashEnvScript -ItemType File -Force | Out-Null
                   }
                   $bashPath = $goBin.Replace('\', '/').Replace('C:', '/c')
-                  'export PATH="' + $bashPath + ':$PATH"' | Out-File -FilePath $bashEnvScript -Append -Encoding UTF8
+                  # Write UTF-8 without BOM — a BOM makes Git Bash unable to parse "export", causing "export: command not found".
```
PeterSchafer: Question: did you consider fixing this issue when reading the file rather then every write?

scripts/windows/install-node.ps1
danskmt: yeah I agree with you. I thought about keeping the old behavior just to leave the original implementation here, even if as a fallback... But if we can ensure all images will contain NVM, then we can just go for it.

.circleci/config.yml
danskmt: I’d rather keep this as write-time: snyk-env.sh should be valid UTF-8 without BOM on disk, since we own all the producers. If we only remove the BOM when we read the file, we’re hiding bad files instead of fixing them, and we’d have to do that cleanup in every spot that runs source.

PeterSchafer (approved):

PR #6769, opened 2026-05-05
chore: upgrade Node.js runtime from 20 to 22 LTS

danskmt:
- [ ] Tests written and linted
- [ ] Documentation written / README.md updated [https://snyk.io/docs/snyk-for-node/]()
- [ ] Follows [CONTRIBUTING agreement](CONTRIBUTING.md)
- [ ] Commit history is tidy [https://git-scm.com/book/en/v2/Git-Branching-Rebasing]()
- [ ] Reviewed by Snyk team

Depends on https://github.com/snyk/cli/pull/6764

### What this does

Upgrades the **Snyk CLI** repository to **Node.js 22 LTS** ([CLI-1469](https://snyksec.atlassian.net/browse/CLI-1469)): `engines.node` → **`^22`**, `.nvmrc` and CI defaults aligned on the chosen **22.x** patch, dependency / lockfile updates as needed, and GitHub Actions + CircleCI config adjusted so Linux/Windows/macOS pipelines install and test against **22** (including native build / embedded CLI paths where applicable). Documentation is updated so supported Node versions match what we actually run in CI.

### Why Node 22 and not Node 24 (for now)

- **Still LTS, smaller jump:** **22** is a supported **LTS** line and is a deliberate step up from **20** without adopting the newest major immediately.
- **Avoid Node 24–specific test / stderr churn:** On **Node 24** we saw **deprecation warnings** on **stderr** (e.g. legacy `url.parse`, `child_process` + `shell` behaviour) from transitive dependencies. Many **Jest acceptance** tests assert **`stderr` is empty**, so those warnings fail CI unless we upgrade dependencies, change spawn behaviour, or filter stderr in tests. On **22**, we expect to **validate real behaviour** without that layer of suppression.
- **Stabilize CI first:** Land **22**, get green pipelines and product confidence, then plan **24** with dependency upgrades and any intentional code changes, rather than coupling the version bump to test-harness workarounds.

### Notes for the reviewer

- Confirm **CircleCI executor images**, **GitHub Actions `node-version`**, **`.nvmrc`**, and **`engines`** are all consistent with the same **22.x** line.
- Confirm **Windows** install path (`install-node.ps1` / NVM + MSI fallback) matches the pinned version and checksum where used.
- If we narrow or change **test matrices**, confirm that matches intended **support policy** for the CLI.

### More information

- [CLI-1469](https://snyksec.atlassian.net/browse/CLI-1469)

### Screenshots

N/A

[CLI-1469]: https://snyksec.atlassian.net/browse/CLI-1469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/check-dependencies.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v4
         with:
-          node-version: '16.16.0'
+          node-version: '22.22.2'
```
PeterSchafer: thinking out loud: this seems like we need a bit more single source of truth?!

PeterSchafer: Actually, I think this is solvable by using [node-version-file](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file) rather then `node-version`

.github/workflows/danger-zone.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v4
         with:
-          node-version: '18.19.1'
+          node-version: '22.22.2'
```
PeterSchafer: same as above

package.json
```diff
@@ -21,7 +21,7 @@
     "snyk": "bin/snyk"
   },
   "engines": {
-    "node": "^20"
+    "node": "^22"
```
PeterSchafer, line 24: Question: Is there a reason why we don't just add 22 in addition?

danskmt: @PeterSchafer we can. Since here is always just a single version (and not for example 18 & 20), I thought about keeping just 22. But I believe we can have something similar as we have [here](https://github.com/snyk/snyk-nodejs-plugin/blob/main/package.json), should we?
```json
"engines": {
    "node": "^20.18.1 || ^22.13.0 || >=24.0.0"
},
```

danskmt: Actually for 24 I would drop for the CLI for now for those reasons we discussed

PeterSchafer: let's do 20 and 22

danskmt: actually, we would need to ensure the CI is testing for both versions. Right now it considers only what we have in .nvmrc.
Shall we also invest time in supporting tests for multiple versions here?

PeterSchafer (approved):

PeterSchafer (approved):

package.json
danskmt: We decided to keep just 22

PR #6786, opened 2026-05-08
chore: update install-snyk script to allow os and arch overrides

j-luong:

- [ ] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [ ] Commit messages
- [ ] Includes detailed description of changes

## What does this PR do?
This PR updates the `install-snyk.py` script so that callers can provide an override the automatically detected OS and Architecture. 

This change is meant to support cross compilation or other scenarios where the host os and arch are not the target ones.

on a mac, run 
```python scripts/install-snyk.py 1.1304.2 --os linux --arch amr64```

None, this is an internal change

## Risk assessment - Low?
Low, as these a pure overrides for special cases

### DIFF
--- a/scripts/install-snyk.py
+++ b/scripts/install-snyk.py
@@ -42,12 +42,17 @@ def get_os_arch():
         return None, None
 
 
-def download_snyk_cli(download_version, base_url):
+def download_snyk_cli(download_version, base_url, os_override, arch_override):
     success = 0
     fail = 1
     abort = 2
 
     os_type, arch_type = get_os_arch()
+
+    if os_override:
+        os_type = os_override
+    if arch_override:
+        arch_type = arch_override
 
     if not os_type or not arch_type:
         return abort
@@ -108,7 +113,7 @@ def download_snyk_cli(download_version, base_url):
 
 
 # will try to download via the base_url 'retries' amount of times
-def download_with_retry(retries, base_url):
+def download_with_retry(retries, base_url, os_type, arch_type):
     for retry in range(1, retries + 1):
         print(
             "Trying to download version "
@@ -118,7 +123,7 @@ def download_with_retry(retries, base_url):
             + " of #"
             + str(retries)
         )
-        download_status = download_snyk_cli(args.version, base_url)
+        download_status = download_snyk_cli(args.version, base_url, os_type, arch_type)
 
         # download failed - retry
         if download_status == 1:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/install-snyk.py [hunk 1]
```diff
@@ -11,6 +11,16 @@
 secondary_url = "https://static.snyk.io"
 urls = [primary_url, secondary_url]
 
+VALID_FILENAMES = [
```
PeterSchafer: Suggestion: Let's try to incrementally get to a [single source of truth](https://github.com/snyk/cli/blob/main/binary-deployments.json) for binary filenames.

scripts/install-snyk.py
```diff
@@ -186,6 +191,14 @@ def verify_checksum(file_path, expected_checksum):
     parser.add_argument(
         "--base_url", help="Base URL to download from", default=primary_url
     )
+    parser.add_argument(
+        "--arch", help="Explicitly specify the architecture to download", default=None,
+        choices=["amd64", "arm64"]
+    )
+    parser.add_argument(
+        "--os", help="Explicitly specify the OS to download", default=None,
+        choices=["linux", "windows", "macos"]
```
PeterSchafer: nitpick: alpine

j-luong: alpine gets derived from the `linux` os:
```
def get_filename(arch_type, os_type):
    filename = ""
    output_filename = "snyk"
    suffix = ""

    if os_type == "linux" and arch_type == "arm64":
        filename = "snyk-linux-arm64"
        stat_result = os.path.exists("/lib/ld-musl-aarch64.so.1")
        if stat_result:
            filename = "snyk-alpine-arm64"
```

And is the reason for the `major`:
```
The get_filename function (not directly changed but invoked by updated logic) uses host-side filesystem checks (os.path.exists("/lib/ld-musl-aarch64.so.1")) to distinguish between Glibc and Alpine (musl) binaries. In a cross-installation scenario—such as the one described in the PR (downloading for arm64 on an amd64 host)—the check for the target architecture's musl library will always fail on the host. This forces the selection of the snyk-linux-arm64 (glibc) binary even if the intended target is an Alpine image, resulting in a broken installation for the target environment.
```

I left it as it's unchanged in this PR, but can update if it's a blocker.

PeterSchafer: but for an override, I would still want to be able to select it.

PeterSchafer (approved):

PR #6870, opened 2026-06-01
test(sbom-monitor): add Dragonfly user journey acceptance test

jacek-rzrz:
Adds a user journey acceptance test for the Dragonfly `sbom test` reporting flow, gated behind `INTERNAL_SNYK_CLI_ROLLOUT_DFLY_SBOM_MONITOR`. Like the existing `sbom test` reachability user journey test, it runs against a real (pre-prod) Snyk instance rather than the fake server.

### What it covers

Both flavours of the reporting flow, sharing one body via `describe.each` so the assertions are identical:

| Variant | Asset |
| --- | --- |
| `sbom test --report` | `cli-sbom-test-report-user-journey` |
| `sbom test --monitor` | `cli-sbom-test-monitor-user-journey` |

Each variant writes to its own asset so the two flows never share a target. Both assert that stdout contains `View your asset(s) at:` and that the exit code is 1 — the `snyk-goof` SBOM fixture carries known vulnerabilities. stderr is logged for diagnostics but not asserted on.

### How it runs

- Credentials come from `TEST_SNYK_API_DEV` / `TEST_SNYK_TOKEN_DEV`, supplied by the CircleCI contexts already attached to the `acceptance-tests` jobs. The suite skips itself when `TEST_SNYK_TOKEN_DEV` is absent, so local runs and forks stay green.
- It runs in the standard `acceptance-tests` jobs — no separate pre-prod job. An earlier pipeline confirmed the `--report` variant executing in all eight flavours (linux amd64 / arm64 / static arm64 / fips arm64, alpine amd64 / arm64, macOS arm64, windows amd64) at ~20s each; `--monitor` runs in the same jobs.
- Those jobs shard six ways, so the spec lands in a single container's log per flavour. To find it, grep for `[sbom-monitor-user-journey][--report]` or `[sbom-monitor-user-journey][--monitor]` — note that every shard prints a `No timing found for ...sbom-test-user-journey.spec.ts` line from the test splitter, which is not evidence of execution.

Also renames `all-projects.spec.ts` → `beta-all-projects.spec.ts` to distinguish the mocked-server beta tests from this user journey suite.

Relates to OSF-422.


danskmt:
Hi @jacek-rzrz is this PR still relevant?

test/jest/acceptance/snyk-sbom-test/sbom-test-user-journey.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+      const { code, stdout, stderr } = await runSnykCLI(command, {
+        env: dragonflyEnv,
+      });
+
+      console.log(`${logTag} exit code: ${code}`);
+      console.log(`${logTag} stdout:\n${stdout}`);
+      console.log(`${logTag} stderr:\n${stderr}`);
+
+      expect(stderr).toBe('');
```
PeterSchafer: Suggestion: this can be brittle as the CLI might print something to stderr. The general recommendation is to not test for things that are not part of the functionality you add.

jacek-rzrz: Thank you @PeterSchafer , dropped!

PeterSchafer (approved):

PR #6881, opened 2026-06-04
chore(npm): enforce minimum release age policy for packages

octavian-snyk:

- [x] Commit messages

Proposed order:
- `.npmrc`, `package.json`, `package-lock.json`
- `.circleci/Dockerfile` 
- `.circleci/config.yml`
- `scripts/windows/install-node.ps1`
- `test/jest/acceptance/snyk-test/reachability-user-journey.spec.ts`
- `.github/workflows/*`, `.nvmrc`
- `CONTRIBUTING.md`

None to CLI users.
CLI devs must upgrade to npm>=11.10

This PR enforces `min-release-age` locally. This allows us to bypass the restriction when needed, which is useful if a package has a CVE that we need to patch before the min-release-age cooldown is met.

If you agree, I believe implementing how the CI/CD pipeline should verify package age compliance (while still allowing a bypass for those critical CVE cases) would be best addressed in a separate, dedicated PR.

[CLI-1516](https://snyksec.atlassian.net/browse/CLI-1516)

[CLI-1516]: https://snyksec.atlassian.net/browse/CLI-1516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robertolopezlopez:
### `min-release-age` does not protect the primary CI install path

Most of our install paths use `npm ci` (`prepare-build`, GHA `check-dependencies`, `danger-zone`). Per npm's own guidance, `min-release-age` has **no effect on `npm ci`** because it installs the lockfile verbatim without re-resolution ([npm/cli#9281](https://github.com/npm/cli/issues/9281)).

Similarly, when installing from a committed lockfile, explicitly pinned versions are installed regardless of `min-release-age` ([npm/rfcs#856](https://github.com/npm/rfcs/issues/856)).

So the 4-day cooldown mainly applies when someone runs `npm install` / `npm update` and npm performs fresh resolution — e.g. the CONTRIBUTING flow of `npm ci` then `npm install <dep>`. Routine CI and most `npm ci` runs are not age-gated.

**Suggestion:** document this operational model explicitly (CONTRIBUTING or internal runbook): lockfile updates must be made with `min-release-age` active; `npm ci` is not a second line of defense. If the intent is CI-enforced cooldown, we may need an additional guard (e.g. validating the lockfile was generated under the policy, or another check in `check-dependencies`).

package.json
```diff
@@ -222,5 +223,10 @@
   },
   "jest-junit": {
     "reportTestSuiteErrors": "true"
+  },
+  "allowScripts": {
```
octavian-snyk: context: Future versions of npm will disable install scripts by default (a change currently signaled by warnings in npm v11). To prevent potential regressions, these scripts have been explicitly enabled here.

.nvmrc
```diff
@@ -1 +1 @@
-22.22.2
+22.22.3
```
octavian-snyk, line 1: context:
This node version errors on npm upgrades
nodejs/node#62425

package.json
```diff
@@ -21,7 +21,8 @@
     "snyk": "bin/snyk"
   },
   "engines": {
-    "node": "^22"
+    "node": ">=22",
```
octavian-snyk: context: [macOS](https://app.circleci.com/pipelines/github/snyk/cli/37645/workflows/326cf671-8177-4b74-b916-fb7856027876/jobs/589803) image already uses node24. This was silenced because engine-strict was false

.circleci/config.yml
```diff
@@ -922,7 +930,7 @@ workflows:
           test_snyk_command: ./binary-releases/snyk-alpine
           install_deps_extension: alpine-full
           dont_skip_tests: 0
-          pre_test_cmds: export BROWSER="wget -O -"
+          pre_test_cmds: export BROWSER="wget -O -" && export MAVEN_OPTS="--sun-misc-unsafe-memory-access=allow"
```
octavian-snyk: context Claude:
Bumping the alpine executor 3.20 → 3.23 also bumps the JDK that apk add maven pulls, from OpenJDK 21 → 25. On JDK 25, Maven's bundled Guice triggers sun.misc.Unsafe deprecation warnings to stderr. snyk-mvn-plugin builds its error from stderr || stdout, so that stderr noise masks Maven's real Non-parseable POM message (which is on stdout) — flipping the SBOM scanError from "error parsing the xml file" to the generic "cannot build maven dependency tree" and failing the allow-incomplete-sbom test.

The flag sets sun.misc.Unsafe memory access to allow, which silences the warning so stderr stays clean and the plugin reports the correct error.

.circleci/config.yml
```diff
@@ -47,11 +47,11 @@ orbs:
 executors:
   alpine:
     docker:
-      - image: alpine:3.20
+      - image: alpine:3.23
```
robertolopezlopez, line 50: is this change strictly necessary for this PR? the original task doesn't say anything of upgrading alpine version :-)

.circleci/config.yml
```diff
@@ -1713,7 +1721,7 @@ jobs:
           no_output_timeout: 30m
           command: |
             << parameters.pre_test_cmds >>
-            npm run test:acceptance -- --selectProjects coreCli --shard=<< parameters.shard_calc_cmd >>/<< parameters.shards >>
+            npm run test:acceptance "--" --selectProjects coreCli --shard=<< parameters.shard_calc_cmd >>/<< parameters.shards >>
```
robertolopezlopez, line 1724: please clean unnecessary changes from the diff :-)

.circleci/Dockerfile
```diff
@@ -70,6 +70,11 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     && apt-get auto-remove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 RUN node -v
 
+# Upgrade npm to match the version required in package.json "engines"
```
robertolopezlopez, line 73: wouldn't we have to upgrade also node in such case? (just wondering, I am not a JS person)

package.json
robertolopezlopez: Yeah but this gives too much room for future "unwanted" node upgrades. I would rather upgrade ^22->^24 so it stays pinned

.circleci/config.yml
octavian-snyk: It became necessary after implementing engine-strict=true. The current Alpine image (3.20) uses Node 20, it can't be upgraded further via apk, and it also reached EOL two months ago. Upgrading the Alpine version seemed like the most straightforward way to resolve it.

.circleci/config.yml
octavian-snyk: Just Windows weirdness :)
On Linux and macOS it doesn't do anything, but after the npm upgrade, Windows strictly requires those quotes "--" to parse the arguments correctly.

.circleci/Dockerfile
octavian-snyk: Not necessarily. npm can be upgraded independently of node.
We could also upgrade Node to v24. Starting with Node 24.14.1 it seems, it natively ships with npm >=11.10.

package.json
octavian-snyk: Sure, I believe we could do that in another Upgrade Node PR.

.circleci/config.yml
robertolopezlopez: 😄 almost as picky as me

.circleci/Dockerfile
robertolopezlopez: Yes, I think that may be a good idea

package.json
robertolopezlopez: Please create a new ticket for that - and mention also to update `.nvmrc` as well.

And also: please do it in this PR as well with the version which you will pin here.

robertolopezlopez (commented):
LGTM - but let's wait for other review

.circleci/config.yml
pieterocp: plus security focused companies shouldn't use EOL'ed things in their CI 😉

pieterocp (commented):
Approve, would only take 1 snyk dev to run npm install at the wrong time without this in place and boom, you might have some gnarly malware with the same permissions getting everything exfiltrated.

.github/workflows/check-dependencies.yml
```diff
@@ -4,18 +4,28 @@ on:
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check-dependencies:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
     steps:
-      - uses: actions/checkout@v2
```
pieterocp, line 11: **Suggestion**: Run zizmor over the top of all this projects GitHub actions + pin actions off the back of it + then get depdenabot/renovate/snyk to run updates to reduce the risk of upstream action hijacking.

.circleci/config.yml
j-luong: issue: this seems like an incompatibility of `snyk-mvn-plugin` with JDK 25. Adding `export MAVEN_OPTS="--sun-misc-unsafe-memory-access=allow"` seems to alter the behaviour in our environment that may not be configured in another users' env.

This could lead to different behaviours in envs where `export MAVEN_OPTS="--sun-misc-unsafe-memory-access=allow"` is not set, and there's no clear way to know that the difference is because of `export MAVEN_OPTS="--sun-misc-unsafe-memory-access=allow"`

test/jest/acceptance/snyk-test/reachability-user-journey.spec.ts
```diff
@@ -14,7 +14,10 @@ const TEMP_LOCAL_PATH = '/tmp/snyk-goof';
 beforeAll(() => {
   if (!existsSync(TEMP_LOCAL_PATH)) {
     execSync(
-      `git clone ${TEST_REPO_URL} ${TEMP_LOCAL_PATH} && cd ${TEMP_LOCAL_PATH} && npm install --ignore-scripts`,
+      // engine-strict=false: snyk-goof pins engines.node 6.14.1, and the CLI's
+      // .npmrc engine-strict leaks in via npm_config_* env from `npm run test:acceptance`,
+      // which would otherwise make this install fail (EBADENGINE) on modern Node.
+      `git clone ${TEST_REPO_URL} ${TEMP_LOCAL_PATH} && cd ${TEMP_LOCAL_PATH} && npm install --ignore-scripts --engine-strict=false`,
       {
```
j-luong, line 21: suggest: maybe check with prodsec, this looks like it could be trouble

.circleci/config.yml
j-luong: suggest: let's apply this change to the specific test by passing the `MAVEN_OPTS="--sun-misc-unsafe-memory-access=allow` into the env param when executing the CLI. This should scope the change to one place

package.json
j-luong: suggest: let's remove this for now and follow up with prodsec on the implications of adding this

j-luong (approved):
Pre-approved, pending changes we suggested

PR #6884, opened 2026-06-05
feat: add support for user-doc-based help routing and manifest generation

robertolopezlopez:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages
  are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing
  _what_ was changed, not _how_.
- [x] Includes detailed description of changes
- [ ] Contains risk assessment (Low | Medium | High)
- [ ] Highlights breaking API changes (if applicable)
- [x] Links to automated tests covering new functionality
- [x] Includes manual testing instructions (if necessary)
- [ ] Updates relevant GitBook documentation (PR link: ___)
- [ ] Includes product update to be announced in the next stable release notes

## What does this PR do?

Shows help generated by Cobra for CLI commands which don't have any GitBook help - existing GitBook help remains as source of truth.

### Changes

Help routing now follows this order:

1. GitBook user-doc exists -> existing markdown help (legacy CLI)
2. No user-doc -> Cobra usage + flags
3. Unknown command / root help -> generic README (unchanged)

How to find out: generic README contains How to get started. Cobra help contains Usage: and Flags:.

#### Before

```
snyk-macos-arm64 secrets test --help | head -20
CLI help
  Snyk CLI scans and monitors your projects for security vulnerabilities and license issues.

  For more information, visit the Snyk website https://snyk.io/

  For details, see the CLI documentation https://docs.snyk.io/snyk-cli

How to get started
  1. Authenticate by running snyk auth.
  2. Test your local project with snyk test.
  3. Get alerted for new vulnerabilities with snyk monitor.

Available commands
  To learn more about each Snyk CLI command, use the --help option, for example, snyk auth --help.

  Note: The help on the docs site is the same as the --help in the CLI.

  snyk auth
    Authenticate Snyk CLI with a Snyk account.
```

#### After

```
snyk-macos-arm64 secrets test --help           
Usage:
  snyk secrets test [flags]

Flags:
      --exclude string                        Ignores all issues originating from the specified file path.
  -h, --help                                  help for test
      --include-ignores                       Shows all discovered issues, including any that have been previously ignored.
      --json                                  Print results on the console as a JSON data structure.
      --json-file-output string               Save test output as a JSON data structure directly to the specified file, regardless of whether or not you use the --json option.
[...]

PR #6919, opened 2026-06-18
fix: replace manifest.txt help mechanism with embedded user docs directory

robertolopezlopez:
### **User description**

Follow-up to [CLI-1474](https://snyksec.atlassian.net/browse/CLI-1474) / PR #6884 ([CLI-1596](https://snyksec.atlassian.net/browse/CLI-1596)): replace the generated, committed `manifest.txt` with `go:embed` of `help/cli-commands`, removing the drift-prone manifest artifact and the `helpdocs-manifest` / `git add manifest.txt` workflow.

**Why this is necessary:** GitBook sync updates `help/cli-commands/` without triggering a build. A separate committed manifest could drift from the synced markdown and break help routing or CI until someone remembered to regenerate it.

**What changed technically:**
- `cliv2/internal/helpdocs` now uses `//go:embed cli-commands` and walks embedded `.md` basenames (same lookup logic as before), via a `CommandHelp` type instead of package-level state.
- `helpdocs-manifest` is removed; `_helpdocs-prepare` / `_helpdocs-clean` copy `help/cli-commands/*.md` into the embed tree **at build time only**, then remove copied `.md` files afterward. `make -C cliv2 test` and lint do not run the copy step.
- Committed `cli-commands/do-not-delete` keeps the embed directory present for tooling (`go list`, gopls) between builds; `cliv2/.gitignore` ignores transient copied markdown under `internal/helpdocs/cli-commands/`.
- Tests inject help lookup instead of relying on compile-time embed: `helpdocs` and `helprouting` use `CommandHelpFromRepo()` (reads `help/cli-commands/` from disk) or a small in-memory fixture when that directory is unavailable. No `TestMain` sync and no skip guards for bare `go test`.

Help routing behavior is unchanged: documented commands still get legacy GitBook markdown help; undocumented ones still get Cobra help. Production wiring uses `helpdocs.DefaultCommandHelp().HasUserDoc` from `main.go`.

- [ ] Highlights breaking API changes (if applicable) — N/A
- [ ] Updates relevant GitBook documentation (PR link: ___) — N/A (GitBook is source of truth)
- [ ] Includes product update to be announced in the next stable release notes — N/A (internal build/routing plumbing)

## What does this PR do?

- Deletes `cliv2/internal/helpdocs/manifest.txt` and manifest sync test.
- Embeds `help/cli-commands` markdown via `go:embed` instead of a basename list.
- Replaces `helpdocs-manifest` Makefile target with transient `_helpdocs-prepare` / `_helpdocs-clean` around **build only**.
- Introduces injectable `CommandHelp` and wires `HasUserDoc` through `helprouting.Router` and `main.go`.
- Updates `CONTRIBUTING.md` for the new workflow.

1. `cliv2/internal/helpdocs/command_help.go` + `embed.go` — lookup + compile-time embed (behavior should match pre-change)
2. `cliv2/internal/helpdocs/fixture.go` — disk/fixture fallback for tests
3. `cliv2/Makefile` — `_helpdocs-prepare` / `_helpdocs-clean` on build target only
4. `cliv2/internal/helprouting/helprouting.go` — injected `HasUserDoc` on `Router`

```sh
make -C cliv2 test
make build BUILD_MODE=public

./binary-releases/snyk-macos-arm64 help test        # documented → GitBook help
./binary-releases/snyk-macos-arm64 help agent-scan  # undocumented → Cobra help
./binary-releases/snyk-macos-arm64 secrets test --help | head -5    # Cobra Usage:
```

Verify post-build working tree: `cliv2/internal/helpdocs/cli-commands/` contains only `do-not-delete` (no copied `.md`).

None. User-facing help content and routing behavior are unchanged.

## Risk assessment

**Low** — same routing algorithm; only the embed data source, build sync, and test wiring changed.

`go:embed` cannot reference repo-root `help/cli-commands/` from `cliv2/internal/helpdocs/` (no `..` paths). The Makefile copies markdown into the package tree at build time; tests read the repo directory directly so they stay in sync with GitBook PRs without a prepare step.

- ~~CLI-1596~~
- [CLI-1596](https://snyksec.atlassian.net/browse/CLI-1616)

[CLI-1474]: https://snyksec.atlassian.net/browse/CLI-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robertolopezlopez:
> ## PR Reviewer Guide 🔍
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Regex mismatch risk**](https://github.com/snyk/cli/pull/6919/changes#diff-a0e0625f3d7e5700dbcacd19a793cdacdf8e785542f3d34cfffba37e9bfc0ec4R17-R17) 🟡 [minor]
> The `nonDocChars` regex `[^a-zA-Z0-9-]` strips all characters except alphanumeric and hyphens. If GitBook or the sync process produces filenames containing underscores or other symbols (which are common in some documentation systems), `HasUserDoc` will fail to find the file even if it exists in the embed. While the current tests use hyphens, this creates a tight coupling between the sync script's naming convention and the Go code that isn't explicitly enforced.
> 
> ```go
> var nonDocChars = regexp.MustCompile(`[^a-zA-Z0-9-]`)
> ```
> 
> [**Silent WalkDir failure**](https://github.com/snyk/cli/pull/6919/changes#diff-a0e0625f3d7e5700dbcacd19a793cdacdf8e785542f3d34cfffba37e9bfc0ec4R25-R31) 🟡 [minor]
> In `docFilesFromEmbed`, the error returned by the inner function of `fs.WalkDir` is returned to `WalkDir`, but the call to `WalkDir` itself on line 25 has its return value ignored (`_ = ...`). If the embedded filesystem is corrupted or inaccessible for an unexpected reason, the `docFiles` map will be silently incomplete, causing the CLI to fall back to the legacy help mechanism without any internal log or warning indicating a filesystem error.
> 
> ```go
> _ = fs.WalkDir(fsys, root, func(_ string, d fs.DirEntry, err error) error {
> 	if err != nil || d.IsDir() || !strings.HasSuffix(d.Name(), ".md") {
> 		return err
> 	}
> 	files[d.Name()] = struct{}{}
> 	return nil
> })
> ```
> 
> 📚 **Repository Context Analyzed**
> 
> **This review considered 14 relevant code sections from 11 files (average relevance: 0.80)**
> 
> 
> * [help/README.md](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/help/README.md/#L1-L19) (lines 1-19)
> * [cliv2/pkg/core/main.go](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/cliv2/pkg/core/main.go/#L209-L412) (2 segments, lines 209-412)
> * [cliv2/internal/embedded/cliv1/dummy_embedded_legacy_cli.go](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/cliv2/internal/embedded/cliv1/dummy_embedded_legacy_cli.go/#L1-L23) (lines 1-23)
> * [package.json](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/package.json/#L1-L229) (lines 1-229)
> * [scripts/upgrade-snyk-go-dependencies.go](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/scripts/upgrade-snyk-go-dependencies.go/#L1-L149) (lines 1-149)
> * [cliv2/scripts/prepare_licenses.go](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/cliv2/scripts/prepare_licenses.go/#L1-L189) (lines 1-189)
> * [package-lock.json](https://github.com/snyk/cli/blob/bd242f14f3ee7ec39bcb6498e6042ba5bcc5c7ab/package-lock.json/#L6-L266) (lines 6-266)
>   _... and 4 more file(s)_
> 
> 🤖 **Repository instructions applied** (from `AGENTS.md`)

> Regex mismatch risk

this is just mimicking the previous .ts behavior

> Silent WalkDir failure

this failure would just happen in build time, not runtime

cliv2/internal/helpdocs/cli-commands/do-not-delete
```diff
@@ -0,0 +1 @@
+Adding this file to the project will prevent IDE errors between builds
```
PeterSchafer, line 1: Question: does this directory need to go into gitignore

robertolopezlopez: mmmmh good one, adding - but you get the reasoning behind this file, right?

octavian-snyk: Could we perhaps ignore the directory contents entirely in case an interrupted `make build`/`test` leaves a dirty tree such that they cannot be committed by accident?

PeterSchafer: This is what I meant, gitignore the whole directory and just have this one file.

> but you get the reasoning behind this file, right?
I think this file is here so that a build would work even without using `make build` for it?!

cliv2/internal/helpdocs/helpdocs.go
```diff
@@ -1,71 +1,64 @@
...
 	"regexp"
 	"strings"
 )
 
-//go:embed manifest.txt
-var manifest string
+const cliCommandsDir = "cli-commands"
+
+//go:embed all:cli-commands
+var cliCommands embed.FS
```
octavian-snyk: Could we perhaps make the top level /help a go module itself, so we don't need this copy?

cliv2/Makefile
```diff
@@ -257,14 +267,18 @@ generate: $(WORKING_DIR)/internal/httpauth/generated/httpauth_generated_mock.go
...
+	@$(MAKE) _helpdocs-clean
 
 .PHONY: test
-test: helpdocs-manifest openboxtest
+test: openboxtest
 
 .PHONY: lint
 lint: $(TOOLS_BIN)/golangci-lint
-	golangci-lint run -v
+	@$(MAKE) _helpdocs-prepare
```
octavian-snyk: what's the purpose of _helpdocs-prepare in lint?

cliv2/internal/helpdocs/helpdocs_test.go
```diff
@@ -1,30 +1,91 @@
...
-		"container.md",
-		"container-test.md",
-		"iac-describe.md",
-		"redteam.md",
+const cliCommandsSourceDir = "../../../help/cli-commands"
+
+var testCLICommandDocFiles map[string]struct{}
+
+func TestMain(m *testing.M) {
+	teardown, err := setupCLICommandsForTest()
```
octavian-snyk: Could we test the logic against an in-memory `fstest.MapFS` to avoid the copy-delete process?

cliv2/internal/helpdocs/cli-commands/do-not-delete
robertolopezlopez: ok! for sure

cliv2/internal/helpdocs/helpdocs.go
```diff
@@ -1,71 +1,64 @@
...
+	"io/fs"
 	"regexp"
 	"strings"
 )
 
-//go:embed manifest.txt
-var manifest string
+const cliCommandsDir = "cli-commands"
+
+//go:embed all:cli-commands
```
PeterSchafer: Suggestion: `all` seems unnecessary and risky to include unwanted filesystem content

cliv2/Makefile
```diff
@@ -191,17 +191,27 @@ summary:
...
 	@set -e; \
-	md_count=$$(ls $(HELPDOCS_SOURCE) 2>/dev/null | wc -l | tr -d ' '); \
+	md_count=$$(ls $(HELPDOCS_SOURCE)/*.md 2>/dev/null | wc -l | tr -d ' '); \
 	if [ "$$md_count" -eq 0 ]; then \
-		echo "$(LOG_PREFIX) ERROR: no .md files found in help/cli-commands ($(HELPDOCS_SOURCE))"; \
+		echo "$(LOG_PREFIX) ERROR: no .md files found in $(HELPDOCS_SOURCE)"; \
 		exit 1; \
 	fi; \
-	ls $(HELPDOCS_SOURCE) | xargs -n1 basename | sort > $(HELPDOCS_MANIFEST)
+	rm -rf $(HELPDOCS_EMBED_DIR)/*.md; \
```
PeterSchafer, line 202: Suggestion: to be sure we start from an empty dir, let's just remove all file please.

cliv2/Makefile
robertolopezlopez: Leftover from previous approach - cleaning up

cliv2/internal/helpdocs/helpdocs_test.go
robertolopezlopez: I would prefer to mirror tHe `Makefile#_Helpdocs-prepare`. `fstest.MapFS` would only test against a made up FS tree, not the current state of `help/cli-commands/*.md`. I might add some extra test with `fstest.MapFS` - but probably this is redundant.

Please share your thoughts :-)

cliv2/internal/helpdocs/helpdocs.go
robertolopezlopez: Unfortunately not and this is why we bring up this `Makefile` wizardry. Root reason: it is off the `cliv2/go.mod` file tree. This is enforced by [the directive itself](https://pkg.go.dev/embed#hdr-Directives): `Patterns may not contain ‘.’ or ‘..’ or empty path elements`

robertolopezlopez: and we cannot move `help/` without extra trouble because there is a git hook which pushes directly to `help/cli-commands/`

cliv2/Makefile
```diff
@@ -191,17 +191,23 @@ summary:
...
-		echo "$(LOG_PREFIX) ERROR: no .md files found in help/cli-commands ($(HELPDOCS_SOURCE))"; \
+		echo "$(LOG_PREFIX) ERROR: no .md files found in $(HELPDOCS_SOURCE)"; \
 		exit 1; \
 	fi; \
-	ls $(HELPDOCS_SOURCE) | xargs -n1 basename | sort > $(HELPDOCS_MANIFEST)
+	rm -rf $(HELPDOCS_EMBED_DIR)/*; \
+	cp $(HELPDOCS_SOURCE)/*.md $(HELPDOCS_EMBED_DIR)/
 
-$(BUILD_DIR)/$(V2_EXECUTABLE_NAME): $(BUILD_DIR) $(SRCS) generate-ls-protocol-metadata $(HELPDOCS_MANIFEST)
+_helpdocs-clean:
```
octavian-snyk: Question: Now that we have `_helpdocs-prepare` and `_helpdocs-clean`. Could it be possible just to generate the `manifest.txt` with the filenames, embed that, and then remove it?

Will the contents of .md files end up being embedded twice in the binary as `//go:embed cli-commands` doesn't include strictly just the file names, and I believe they're also embedded by the legacy TS version?

robertolopezlopez: We are not generating the `manifest.txt` due to the user-docs hook issue; that's the whole purpose of this PR.

I do not follow your second comment :-)

octavian-snyk: On the manifest point, I think we're aligned that a committed manifest.txt is the problem: because of the weekly sync-cli-help-to-user-docs action, the manifest could drift.

What I was wondering is whether a transient manifest (generated in _helpdocs-prepare, embedded for compile, removed in _helpdocs-clean, never committed) would still have that problem. It feels like it would solve the same drift issue, just with a smaller Go-side embed than copying all the .md files. Does that match your thinking, or am I missing something about why even a build-time-only manifest is undesirable?

On the second part, Go's //go:embed cli-commands pulls in the full markdown files (with content), but HasUserDoc only uses their filenames for routing.

Mostly curious whether I'm reasoning about the transient-manifest alternative correctly, or if there's a constraint I'm not seeing. Happy to resolve the thread either way once I've understood your take. 🙂

robertolopezlopez: But what's the point of generating a `manifest.txt` in `_helpdocs-prepare`. I think that's adding extra complexity:

1. Iterate over all user-docs
2. Create `manifest.txt`
3. At the end, removing the manifest

Vs what we have here. `embed.FS` is basically an iterable construct which does not imply any complexity other than iterating it. 

I do not think it brings any advantage. If you could name any real, tangible advantage we may be in the same line ;-)

octavian-snyk: With this approach we're embedding the docs content twice in the shipped binary: Go embeds the full .md files but only uses filenames for routing, and the legacy TS binary already embeds the same content for rendering. With a transient manifest we'd only have the filenames on the Go side. The difference isn't big though 😅 .

octavian-snyk (approved):

cliv2/Makefile
robertolopezlopez: You may be right on one side: yes, double embed. The TS bundle will be there for long time I believe.

Your proposal may work and shrink the Go side; but reopens the initial approach in CLI-1596 and adds more problems to the PR, which is already large.

Planned scope: preparing+cleaning changes, embed, getting a green pipeline (heh), multiple new review threads.

Do you really think it is worth to start changing the whole PR? :)

octavian-snyk: Fair point. Thank you 😀

Approval already given. I believe you can merge this PR.

robertolopezlopez: thank youuuu

cliv2/Makefile
```diff
@@ -257,10 +265,12 @@ generate: $(WORKING_DIR)/internal/httpauth/generated/httpauth_generated_mock.go
 .PHONY: openboxtest
 openboxtest:
 	@echo "$(LOG_PREFIX) Running $@"
-	@$(GOCMD) test -cover ./...
+	@$(MAKE) _helpdocs-prepare
```
PeterSchafer: Question: can we test against synthetic data rather then relying on this pre-test step

robertolopezlopez: `_helpdocs-prepare` needs to be run before `go test`, so the test binary embeds the real `.md` (compile time vs run time). So we test with real world data :-)

Does this address your question, or what do you really mean with `synthetic data`?

cliv2/Makefile
PeterSchafer: just to add a closing thought here, we will be removing the TS implementation in the future, so the double embed is only temporary.

cliv2/Makefile
PeterSchafer: I think I meant what @octavian-snyk meant with [their comment](https://github.com/snyk/cli/pull/6919/changes#r3442826660).

cliv2/internal/helpdocs/helpdocs_test.go
```diff
@@ -1,30 +1,91 @@
...
-		"test.md",
-		"container.md",
-		"container-test.md",
-		"iac-describe.md",
-		"redteam.md",
+const cliCommandsSourceDir = "../../../help/cli-commands"
+
+var testCLICommandDocFiles map[string]struct{}
+
+func TestMain(m *testing.M) {
```
PeterSchafer: Question: do we really/still need this?

cliv2/internal/helpdocs/helpdocs_test.go
robertolopezlopez: I ended up applying this

cliv2/internal/helpdocs/embed.go
octavian-snyk, line 1: nitpick: For readability it might be better to integrate these small files (embed.go, and export_test.go) into  other relevant files; perhaps command_help.go, and fixture.go, respectively. 
 
Other than this, LGTM :)

cliv2/internal/helpdocs/command_help.go
```diff
@@ -0,0 +1,67 @@
...
+	files map[string]struct{}
+}
+
+// NewCommandHelp builds a lookup by walking root on fsys for *.md files.
+func NewCommandHelp(fsys fs.FS, root string) *CommandHelp {
+	return &CommandHelp{files: docFilesFromEmbed(fsys, root)}
+}
+
+// NewCommandHelpFromFiles builds a lookup from an existing filename set.
+func NewCommandHelpFromFiles(files map[string]struct{}) *CommandHelp {
```
PeterSchafer: Suggestion: Maybe use https://pkg.go.dev/golang.org/x/tools/godoc/vfs/mapfs and remove this New function.

cliv2/internal/helpdocs/command_help.go
```diff
@@ -0,0 +1,67 @@
...
+// HasUserDoc reports whether legacy user-doc help should be shown for command segments.
+// Empty segments -> true (top-level README via legacy help).
+// Non-empty segments -> true only if a matching .md exists (README excluded).
+func (h *CommandHelp) HasUserDoc(segments []string) bool {
+	return hasUserDoc(segments, h.files)
+}
+
+func docFilesFromEmbed(fsys fs.FS, root string) map[string]struct{} {
+	files := make(map[string]struct{})
+	_ = fs.WalkDir(fsys, root, func(_ string, d fs.DirEntry, err error) error {
```
PeterSchafer: Question: Why not surface the error?

cliv2/pkg/core/main.go
```diff
@@ -2,6 +2,7 @@ package core
 
 // !!! This import needs to be the first import, please do not change this !!!
```
PeterSchafer, line 3: Issue: please see the comment above!!! the fips import needs to be the first one

PeterSchafer:
@robertolopezlopez besides the remaining comments, I think this looks now very good! Thank you for the refactoring!

cliv2/internal/helpdocs/command_help.go
robertolopezlopez: ok!

cliv2/internal/helpdocs/command_help.go
robertolopezlopez: it is deprecated! from your own link :-)

cliv2/internal/helpdocs/embed.go
robertolopezlopez: ack :-) about embed.go: the Go convention is that `//go:embed` sits in a minimal file. I may merge it with `command_help.go` if you prefer but it is getting dangerously large IMHO and this is a sleepery road ;-)

robertolopezlopez: in snyk you can find similar examples https://github.com/snyk/policy-engine/blob/main/rego/embed.go

robertolopezlopez: about export_test.go: export_help.go is a kind of standard for test only package helpers. Example: os/export_test.go, fmt/export_test.go, etc.

octavian-snyk: TIL, thanks for the explanation!

I noticed CommandHelpFromRepo is already used directly in a couple of test files (helprouting_test.go:233 and externally at main_test.go:770), while CommandHelpForTest is only used internally at command_help_test.go:17. Do we need the extra re-export via CommandHelpForTest?

cliv2/internal/helpdocs/command_help.go
PeterSchafer: sorry, you are right. Please take a look at this https://pkg.go.dev/testing/fstest#MapFS

cliv2/internal/utils/interactive_test.go
```diff
@@ -11,8 +11,12 @@ import (
 func Test_isTerminal_pipeIsNotATerminal(t *testing.T) {
 	r, w, err := os.Pipe()
 	require.NoError(t, err)
-	defer r.Close()
-	defer w.Close()
+	defer func(r *os.File) {
+		_ = r.Close()
```
PeterSchafer: Suggestion: How about something like this `assert.NoError(t, r.Close())`?

PeterSchafer: same for all the other close calls.

cliv2/pkg/core/main.go
PeterSchafer: Thank you!

cliv2/internal/helpdocs/command_help.go
robertolopezlopez: yes, I was already using it `fstest.MapFS` @ `command_help_test.go:43` - thinking it over, I will move that to `fixture.go` - I like the change

cliv2/Makefile
```diff
@@ -191,17 +191,23 @@ summary:
 .PHONY: configure
 configure: _validate-build-mode summary $(CACHE_DIR) $(CACHE_DIR)/variables.mk $(V1_DIRECTORY)/$(V1_EMBEDDED_FILE_OUTPUT) dependencies $(CACHE_DIR)/prepare-3rd-party-licenses
 
-.PHONY: helpdocs-manifest
-helpdocs-manifest:
+.PHONY: _helpdocs-prepare _helpdocs-clean
```
robertolopezlopez: move the phony _helpdocs_clean to its own target

cliv2/Makefile
```diff
@@ -191,17 +191,23 @@ summary:
...
+	rm -rf $(HELPDOCS_EMBED_DIR)/*; \
+	cp $(HELPDOCS_SOURCE)/*.md $(HELPDOCS_EMBED_DIR)/
 
-$(BUILD_DIR)/$(V2_EXECUTABLE_NAME): $(BUILD_DIR) $(SRCS) generate-ls-protocol-metadata $(HELPDOCS_MANIFEST)
+_helpdocs-clean:
+	@git clean -fdX -- $(HELPDOCS_EMBED_DIR)
+	@git checkout -- $(HELPDOCS_EMBED_DIR)
+
+$(BUILD_DIR)/$(V2_EXECUTABLE_NAME): $(BUILD_DIR) $(SRCS) generate-ls-protocol-metadata
+	@$(MAKE) _helpdocs-prepare
```
robertolopezlopez: move to dependency

cliv2/Makefile
robertolopezlopez: remove

PeterSchafer (approved):
Pre-approve, expecting the discussed final changes to be done.

cliv2/Makefile
robertolopezlopez: not needed, just remove

PR #6948, opened 2026-06-25
feat: [CMPA-604] Adds sbom component metadata for maven behind an internal flag

calhar-snyk:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing _what_ was changed, not _how_.
- [x] Includes detailed description of changes
- [x] Contains risk assessment (Low | Medium | High)
- [x] Highlights breaking API changes (if applicable) — none
- [ ] Links to automated tests covering new functionality
- [x] Includes manual testing instructions (if necessary)
- [x] Updates relevant GitBook documentation — N/A, intentionally undocumented (internal, restricted testing)
- [x] Includes product update to be announced in the next stable release notes — N/A, no client-visible change

## What does this PR do?

Plumbs the internal `--include-component-metadata` flag through the legacy CLI as part of
CMPA-604 (SBOM component metadata).

This flag is **not a user-facing command option**. It is sourced from the feature-flag
gateway inside `cli-extension-sbom`, passed down through the `cli-extension-dep-graph`
workflow into the legacy CLI, which forwards it to the `mvn`/`npm` legacy plugins. When
set, the plugins attach hash and distribution-URL information as labels on the DepGraph,
which `cli-extension-sbom` uses to populate `components[].hashes[]` and `externalUrls`
(distribution) entries in CycloneDX SBOM output.

Changes here:
- Registers `include-component-metadata` as a recognised argument so it is forwarded to
  the plugins (`src/cli/args.ts`, `src/lib/types.ts`). It is deliberately undocumented.
- Bumps `snyk-mvn-plugin` `^4.7.0` → `4.8.0` — adds Maven artifact checksum collection
  and the `includeComponentMetadata` plugin option (snyk/snyk-mvn-plugin#220,
  renamed from `includeHashes` in snyk/snyk-mvn-plugin#222).
- Bumps the cliv2 extensions that drive this: `cli-extension-sbom`
  (FF-gated activation + SBOM enrichment, snyk/cli-extension-sbom#192) and
  `cli-extension-dep-graph/v2` `v2.6.0` → `v2.7.1` (flag pass-through,
  snyk/cli-extension-dep-graph#222), with the accompanying `go.mod`/`go.sum` tidy across
  both `cliv2/` and `cliv2-private/`.

[...]

PR #6950, opened 2026-06-25
feat: [CMPA-603] Add npm support for SBOM component metadata being FF

calhar-snyk:

- [ ] Updates relevant GitBook documentation (PR link: ___) — n/a, internal/undocumented flag
- [x] Includes product update to be announced in the next stable release notes — n/a, no user-facing change

## What does this PR do?

Extends `--include-component-metadata` support to **npm** projects, following on from the Maven support added in #6948.

When the flag is set, the npm plugin now forwards it to `snyk-nodejs-lockfile-parser`, which reads the install-time `integrity` and `resolved` fields already recorded in the lockfile and surfaces them as `hash:<algorithm>` and `distribution:url` labels on the dep-graph nodes. The SBOM extension then uses these labels to populate component hashes and external distribution URLs in the CycloneDX output.

Concretely:
- Forwards `includeComponentMetadata` through the legacy `npm-lock-parser` in both code paths — the v2/v3 dep-graph path (`parseNpmLockV2Project`) and the v1 depTree path (`buildDepTreeFromFiles`). For v1, the labels survive the downstream depTree→dep-graph conversion.
- Adds `includeComponentMetadata` to the plugin `Options` type.
- Bumps `snyk-nodejs-lockfile-parser` to `2.9.0`, the version that emits the component-metadata labels.

The flag remains internal/undocumented and is gated by the feature-flag gateway — the same surface as the Maven precursor. Unlike Maven, no artifact resolution step is needed: the metadata already lives in the lockfile, so nothing has to be installed/resolved first.

- `src/lib/plugins/nodejs-plugin/npm-lock-parser.ts` — the two-line forwarding change (v1 and v2/v3 paths).
- `src/lib/plugins/types.ts` — the new optional `Options` field.
- `test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts` — coverage across lockfile v1/v2/v3.

Note: npm intentionally stays on the in-tree legacy plugin here (the change just threads the flag through it).  

Against an npm project with a `package-lock.json` (v1, v2, or v3):
```
snyk test --print-graph --include-component-metadata --file=package-lock.json
```
Dep-graph nodes should carry `hash:<algorithm>` and `distribution:url` labels. Re-running without `--include-component-metadata` should produce no such labels.

Automated coverage: `test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts` exercises lockfile v1/v2/v3, each asserting the labels are present with the flag and absent without it.

The full `snyk sbom` flow can also be tested against the pre-prod API:
```
❯ ~/code/cli/binary-releases/snyk-macos-arm64 sbom --format=cyclonedx1.6+json | jq '.'
{
  "$schema": "http://cyclonedx.org/schema/bom-1.6.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.6",
  "serialNumber": "urn:uuid:acd67741-5c1d-4c3e-8591-713b95281254",
  "version": 1,
  "metadata": {
    "timestamp": "2026-06-25T19:28:08Z",
    "tools": {
      "components": [
        {
          "type": "application",
          "author": "Snyk",
          "name": "snyk-cli",
          "version": "1.1306.0-dev.59984f8c87451bff7664b9b1598c599fc51f881f"
        }
      ],
      "services": [
        {
          "provider": {
            "name": "Snyk"
          },
          "name": "SBOM Export API",
          "version": "v1.131.0"
        }
      ]
    },
    "component": {
      "bom-ref": "1-npm-component-metadata-v1@1.0.0",
      "type": "application",
      "name": "npm-component-metadata-v1",
      "version": "1.0.0",
      "purl": "pkg:npm/npm-component-metadata-v1@1.0.0"
    }
  },
  "components": [
    {
      "bom-ref": "2-lodash@4.17.15",
      "type": "library",
      "name": "lodash",
      "version": "4.17.15",
      "hashes": [
        {
          "alg": "SHA-512",
          "content": "f3139c447bc28e7a1c752e5ca705d05d5ce69a1e5ee7eb1a136406a1e4266ca9914ba277550a693ce22dd0c9e613ee31959a2e9b2d063c6d03d0c54841b340d4"
        }
      ],
      "licenses": [
        {
          "expression": "MIT"
        }
      ],
      "purl": "pkg:npm/lodash@4.17.15",
      "externalReferences": [
        {
          "url": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
          "type": "distribution"
        }
      ],
      "properties": [
        {
          "name": "snyk:npm:scope",
          "value": "prod"
        }
      ]
    }
  ],
  "dependencies": [
    {
      "ref": "1-npm-component-metadata-v1@1.0.0",
      "dependsOn": [
        "2-lodash@4.17.15"
      ]
    },
    {
      "ref": "2-lodash@4.17.15"
    }
  ]
}
```

None. The feature is inert by default and gated behind an internal feature flag; there is no user-facing change unless explicitly enabled.

**Low.** The behaviour is gated behind an internal feature flag from the feature-flag gateway and is inert by default — clients see no change unless it is explicitly enabled for them. The only always-on change is the `snyk-nodejs-lockfile-parser` 2.9.0 bump, which adds the label-emitting code path without altering default output. npm scans continue to run through the existing legacy plugin.

Precursor: #6948 (Maven component metadata + registering `include-component-metadata` as a recognised CLI argument forwarded to the plugins). This PR is the npm counterpart.

- [CMPA-603](https://snyksec.atlassian.net/browse/CMPA-603)

[CMPA-603]: https://snyksec.atlassian.net/browse/CMPA-603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/plugins/nodejs-plugin/npm-lock-parser.ts
```diff
@@ -89,6 +90,7 @@ export async function parse(
       strictOutOfSync,
       true,
       options.showNpmScope,
+      options.includeComponentMetadata || false,
```
danskmt: [Optional] instead of setting false as default in both of these **includeComponentMetadata**, I would suggest saving to a constant such as **defaultIncludeComponentMetadata**

src/lib/plugins/types.ts
```diff
@@ -22,6 +22,7 @@ export interface Options {
   scanAllUnmanaged?: boolean;
   showNpmScope?: boolean;
   allProjects?: boolean;
+  includeComponentMetadata?: boolean;
```
danskmt, line 25: can this hold values such as **null** and **undefined**? Just a sanity question

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    // Control: without the flag the same project must not produce the labels,
+    // proving they are driven by `--include-component-metadata`.
+    it('does not attach the labels without the flag', async () => {
+      const project = await createProjectFromFixture(fixture);
+
+      const { code, stdout } = await runSnykCLI(
+        'test --print-graph --file=package-lock.json',
+        { cwd: project.path() },
+      );
```
danskmt, line 74: since this test is in a loop, I'm wondering if we should add some timeout here or if default is enough

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      expect(code).toEqual(0);
+      expect(stdout).toContain('DepGraph data:');
+      expect(labelKeys(stdout, 'hash:')).toHaveLength(0);
+      expect(labelKeys(stdout, 'distribution:url')).toHaveLength(0);
```
danskmt: is there any chance that changes in the responses from **runSnykCLI** would cause these values to fail the test?
Maybe a better question is: are we testing what we want here?

danskmt: (maybe my comment is hard to understand the way I wrote - just wondering if this could be flaky)

src/lib/plugins/types.ts
calhar-snyk: Effectively yes, in the vast majority of cases it won't be getting set.

calhar-snyk: https://github.com/snyk/cli/pull/6948/changes#diff-eb8513f7e557af1081e96db6b6603ca292cc0c2bf383fb00c5f914431372cee3 is where we build the `Options` object that gets passed in to the plugin `inspect` method. So can be undefined

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
calhar-snyk: I don't think this should ever be flakey. The results here should not be dependent on external services as the underlying flows rely entirely on the `package-lock.json` contents

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
calhar-snyk: I think the default timeout _should_ be sufficient. The underlying behaviour is purely dependent on `package-lock.json` contents which are simple and small. This _should_ be fast.

CatalinSnyk (approved):

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      const project = await createProjectFromFixture(fixture);
+
+      const { code, stdout } = await runSnykCLI(
+        'test --print-graph --all-projects',
+        { cwd: project.path() },
+      );
+
+      expect(code).toEqual(0);
+      const graphs = parseDepGraphs(stdout);
+      expect(graphs.map((g) => g.target).sort()).toEqual([root, ...members]);
```
CatalinSnyk, line 146: issue: Seems like the windows path's didn't play nice for this assertion...
Image

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      const project = await createProjectFromFixture(fixture);
+
+      const { code, stdout } = await runSnykCLI(
+        'test --include-component-metadata --print-graph --all-projects',
+        { cwd: project.path() },
+      );
+
+      expect(code).toEqual(0);
+      const graphs = parseDepGraphs(stdout);
+      expect(graphs.map((g) => g.target).sort()).toEqual([root, ...members]);
```
CatalinSnyk, line 114: Ditto.

test/jest/acceptance/snyk-test/npm-include-component-metadata.spec.ts
PeterSchafer: This is actually an interesting problem as we should internally use `/`, so there might be some more central problem than just the assertion.

calhar-snyk: I believe this is pre-existing. From what I can see this seems to be pre-existing in `get-multi-plugin-result.ts`:
```
    const optionsClone = cloneDeep(options);
    optionsClone.file = pathLib.relative(root, targetFile);
    optionsClone.packageManager = detectPackageManagerFromFile(
      pathLib.basename(targetFile),
      featureFlags,
    );
    try {
      const inspectRes = await getSinglePluginResult(
        root,
        optionsClone,
        optionsClone.file,
        featureFlags,
      );
```
`pathLib.relative(root, targetFile)` internally uses `path.sep`. And that's what then gets passed in to the plugin and back out as the `targetFile`.

Similarly, in `npm-workspace-parser.ts`:
```
import * as pathUtil from 'path';
...

      const project: ScannedProjectCustom = {
        packageManager: 'npm',
        targetFile: pathUtil.relative(root, packageJson.fileName),
        depGraph: res as any,
        plugin: {
          runtime: process.version,
        },
      };
      result.scannedProjects.push(project);
```

calhar-snyk: My hunch here, would be that the `--print-graph` flag being used here to produce the raw json DepGraph output from the `snyk test` command is instead bypassing some path normalization that would normally be performed after plugin invocation.

PR #6967, opened 2026-07-06
fix: Exclude passthrough args after -- from scan input directories

danskmt:

## What does this PR do?

When a positional path is passed to a command together with `--` passthrough
args (e.g. `snyk test . -- -s /path/to/settings.xml`), the CLI was storing the
*entire* positional slice — including the tokens after `--` — into
`INPUT_DIRECTORY`. Those passthrough tokens then leak into the scan target list.

For `snyk test` with Risk Score enabled this caused a silent data loss: the
leaked `-s` token was misread as a package name by the downstream flow router,
which force-routed the scan to the legacy engine that never returns Risk Score.
The same leak also affects other slice consumers (SBOM test, monitor, secrets
test).

This fix trims the post-`--` passthrough args before setting `INPUT_DIRECTORY`,
so only real positional paths are treated as scan targets — matching the
existing behavior of `DetermineInputDirectory`.

- `cliv2/pkg/core/main.go` – `updateConfigFromParameter`: only positional args
  before `--` are stored as input directories.
- `cliv2/pkg/core/main_test.go` – `Test_runMainWorkflow_doubleDashArgsDoNotLeakIntoInputDirectory`
  covers `.`, relative, and absolute paths.

With an org that has Risk Score enabled
(`useExperimentalRiskScore` + `useExperimentalRiskScoreInCLI`):

1. `snyk test -d` (from a scannable project) → output contains `Risk Score:` lines.
2. `snyk test . -d -- -s <any-settings-file>` → output should now also contain
   `Risk Score:` lines (previously absent), and debug should show
   `Using legacy flow: false`.

Fixes a bug where passing an explicit path together with `--` passthrough
arguments could silently drop Risk Score data from `snyk test` results.

[CLI-1634]: https://snyksec.atlassian.net/browse/CLI-1634

### DIFF
--- a/cliv2/pkg/core/main.go
+++ b/cliv2/pkg/core/main.go
@@ -162,10 +162,18 @@ func updateConfigFromParameter(config configuration.Configuration, args []string
 	}
 	config.Set(configuration.UNKNOWN_ARGS, doubleDashArgs)
 
-	// only consider the first positional argument as input directory if it is not behind a double dash.
-	if len(args) > 0 && !utils.Contains(doubleDashArgs, args[0]) {
-		config.Set(configuration.INPUT_DIRECTORY, args)
-		config.Set(localworkflows.ConfigurationNewAuthenticationToken, args[0])
+	// Only positional args before "--" are input directories. Everything after "--" is
+	// passthrough (e.g. build-tool flags like `-s settings.xml`) and must not leak into
+	// INPUT_DIRECTORY, otherwise it gets misread as a path/package and force-routes to legacy.
+	// cobra appends the post-"--" tokens as the suffix of args, so trim them off.
+	positionalArgs := args
+	if len(doubleDashArgs) > 0 && len(args) >= len(doubleDashArgs) {
+		positionalArgs = args[:len(args)-len(doubleDashArgs)]
+	}
+
+	if len(positionalArgs) > 0 {
+		config.Set(configuration.INPUT_DIRECTORY, positionalArgs)
+		config.Set(localworkflows.ConfigurationNewAuthenticationToken, positionalArgs[0])
 	}
 }
 
--- a/cliv2/pkg/core/main_test.go
+++ b/cliv2/pkg/core/main_test.go
@@ -185,71 +185,104 @@ func Test_CreateCommandsForWorkflowWithSubcommands(t *testing.T) {
 	assert.True(t, cmd2.Hidden)
 }
 
-func Test_runMainWorkflow_unknownargs(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cliv2/pkg/core/main.go
```diff
@@ -162,10 +162,30 @@ func updateConfigFromParameter(config configuration.Configuration, args []string
 	}
 	config.Set(configuration.UNKNOWN_ARGS, doubleDashArgs)
 
-	// only consider the first positional argument as input directory if it is not behind a double dash.
-	if len(args) > 0 && !utils.Contains(doubleDashArgs, args[0]) {
-		config.Set(configuration.INPUT_DIRECTORY, args)
-		config.Set(localworkflows.ConfigurationNewAuthenticationToken, args[0])
```
danskmt, line 168: In the previous code, `len(args) > 0` is always true if `doubleDashArgs` is not empty. The condition evaluates to true if `doubleDashArgs` doesn't contain `args[0]`, attempting to check: "is the first argument a real positional path, or did it come from behind the double dash?" However, this implementation introduced two critical bugs:

- **Argument Leakage**: Since Cobra strips out the "--" string but appends everything after it directly into args, setting `INPUT_DIRECTORY` to the raw args slice accidentally leaks the underlying tool's passthrough flags (like "-s") into our directory configuration.

- **The Duplication Edge Case**: Using `utils.Contains` completely breaks if the same string is used on both sides of the double dash (for example: `snyk test some/path -- command some/path`). Because `some/path` exists in doubleDashArgs, the code falsely assumes the first `some/path` is a passthrough flag, skips the block, and leaves `INPUT_DIRECTORY` completely unset.

Switching to length-based slicing (`len(args) - len(doubleDashArgs)`) avoids inspecting string content entirely, cleanly solving both the leakage and duplication bugs.
Image

cliv2/pkg/core/main_test.go
```diff
@@ -185,71 +185,104 @@ func Test_CreateCommandsForWorkflowWithSubcommands(t *testing.T) {
...
 			}
 
-			positionalArgs := []string{expectedInputDir}
-			positionalArgs = append(positionalArgs, expectedUnknownArgs...)
-
-			rawArgs := []string{"app", "command", "--sad", expectedInputDir}
-			if len(expectedUnknownArgs) > 0 {
-				rawArgs = append(rawArgs, "--")
-				rawArgs = append(rawArgs, expectedUnknownArgs...)
+			if len(tc.wantUnknownArgs) == 0 {
```
PeterSchafer, line 281: nitpick: the Equal assert below should work in both cases

PeterSchafer (approved):

PR #6987, opened 2026-07-10
chore: Reduce Windows CI build time

danskmt:

## What does this PR do?

Reduces Windows CI build time. On main, the Windows build job installs the full toolchain (Node, Maven, Gradle, .NET, Python, Make) plus DigiCert signing tools, even though compiling only needs Node/Python/Make — and its Go module cache uses a Linux-only path that never hits on Windows, so modules re-download every run.
This PR:
- Adds a lean `windows-native-build-min` deps bundle (Node, Python, Make) with its own small cache for the build job. .NET/Maven/Gradle stay only on the acceptance-test job that needs them.
- Removes the unused DigiCert signing setup from the build job; signing is unchanged on `sign windows amd64`.
- Fixes the Go module/build cache so it works on Windows (pins GOMODCACHE/GOCACHE to the workspace instead of a Linux-only path).
- Builds the standard and FIPS binaries in one job, drops `clean-golang`, and adds a `verify-fips-artifacts` step that fails if the FIPS build did not take effect.

- `.circleci/config.yml` — `install-deps-windows-native-build-min`, `setup-go-build-cache`, the `build windows amd64` job, and the `build-artifact` Go cache
- `scripts/verify-fips-build.sh` and `Makefile` (`verify-fips-artifacts`) — the FIPS safety net
- `scripts/windows/*.ps1` — skip-if-already-installed logic

1. Push branch and run `test_and_release` in CircleCI.
2. Confirm `build windows amd64` restores the small build cache (not the ~3.6 GiB one) and does not install .NET/Maven/Gradle/DigiCert.
3. Confirm the `verify-fips-artifacts` step passes and logs distinct sha256 hashes for the normal vs FIPS binaries.
4. Confirm `acceptance-tests windows amd64` still passes (full tool bundle).
5. On a `main`/`release`/`*e2e*` branch, confirm `sign windows amd64` and the `e2e fips tests (win-server2022-amd64)` still pass.
6. Compare Windows build wall-clock/credits against the previous run.

None. CI-only change with no user-facing impact.

Low-Medium. CI-only, no user-facing changes. The main correctness concern — removing `clean-golang` while building normal + FIPS in one job — is guarded by `verify-fips-artifacts`, which fails the build if the FIPS binary is not distinct from the normal one. FIPS runtime behaviour is still covered by the existing `e2e fips tests` job.

The Windows build was dominated by cache/dependency overhead rather than compilation: a multi-GB tool cache (mostly .NET/Gradle unused by the build), duplicated DigiCert installs, and a Go module cache that never hit on Windows. These changes remove that overhead from the build path while keeping the tools available where they are actually used (tests/signing).

[CLI-1633](https://snyksec.atlassian.net/browse/CLI-1633)

## Screenshots (if appropriate)

N/A

[CLI-1633]: https://snyksec.atlassian.net/browse/CLI-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/verify-fips-build.sh
+++ b/scripts/verify-fips-build.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+# Verifies that the FIPS build actually produced a FIPS binary that is distinct
+# from the normal build.
+#
+# Why this exists: the normal and FIPS builds compile to the same intermediate
+# output path and are only distinguished by the crypto backend. If a regression
+# (e.g. removing `clean-golang` combined with a stale build output) caused the
+# FIPS build to be skipped, the FIPS binary would just be a copy of the normal
+# one. This script fails loudly in that case.
+#
+# Two independent gates, both must pass:
+#   1. The FIPS binary's embedded Go build info must show a FIPS crypto backend
+#      (GOEXPERIMENT=systemcrypto/cngcrypto/opensslcrypto/boringcrypto).
+#   2. The FIPS binary must be byte-different from the normal binary.
+#
+# The script never exits 0 unless it could actually perform these checks.
+#
+# Usage: verify-fips-build.sh <normal-dir> <fips-dir>
+set -euo pipefail
+
+normal_dir="${1:?normal binary directory required}"
+fips_dir="${2:?fips binary directory required}"
+
+# Best-effort: load the Windows env (PATH for go/sha tools) when running on the
+# Windows executor. The caller (e.g. the make-binary step) normally already
+# sources this and exports PATH, which this child process inherits; this is a
+# fallback for standalone runs. Override the location with SNYK_WIN_ENV_SCRIPT.
+win_env_script="${SNYK_WIN_ENV_SCRIPT:-/c/tools-cache/snyk-env.sh}"
+if [ -f "$win_env_script" ]; then
+  # shellcheck disable=SC1090
+  source "$win_env_script"
+fi
+
+# The verification depends on the Go toolchain to read embedded build info.
+if ! command -v go >/dev/null 2>&1; then
+  echo "[verify-fips] ERROR: 'go' not found on PATH; cannot inspect build info. Failing."
+  exit 1
+fi
+
+find_binary() {
+  local dir="$1"
+  local f
+  shopt -s nullglob
+  for f in "$dir"/snyk-*; do
+    case "$f" in
+      *.sha256 | *.sha256.asc | *.asc) continue ;;
+    esac
+    [ -f "$f" ] || continue
+    echo "$f"
+    return 0
+  done
+  return 1
+}
+
+hash_file() {
+  if command -v sha256sum >/dev/null 2>&1; then
+    sha256sum "$1" | awk '{print $1}'
+  elif command -v shasum >/dev/null 2>&1; then
+    shasum -a 256 "$1" | awk '{print $1}'
+  else
+    echo "NOHASH"
+  fi
+}
+
+log_build_settings() {
+  local label="$1"
+  local bin="$2"
+  echo "[verify-fips] Build settings ($label): $bin"
+  go version -m "$bin" 2>/dev/null | grep -E '^[[:space:]]*build[[:space:]]' || echo "  (no build info available)"
+}
+
+# True if the binary's Go build info records a FIPS crypto backend. Microsoft's
+# Go toolchain sets this via GOEXPERIMENT; the concrete value is systemcrypto,
+# or the platform backend it resolves to (cngcrypto on Windows, opensslcrypto on
+# Linux). boringcrypto is included for completeness.
+binary_has_fips_backend() {
+  go version -m "$1" 2>/dev/null \
+    | grep -E '^[[:space:]]*build[[:space:]]+GOEXPERIMENT=' \
+    | grep -qiE 'systemcrypto|cngcrypto|opensslcrypto|boringcrypto'
+}
+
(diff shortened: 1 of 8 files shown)


scripts/verify-fips-build.sh
```diff
@@ -0,0 +1,102 @@
...
+log_build_settings "fips" "$fips_bin"
+echo "------------------------------------------------------------------"
+
+# Informational: Microsoft's Go toolchain records GOEXPERIMENT=systemcrypto for
+# FIPS builds. Absence is not treated as fatal (toolchain behaviour may vary),
+# but it is surfaced clearly.
+if go version -m "$fips_bin" 2>/dev/null | grep -qi 'systemcrypto'; then
+  echo "[verify-fips] FIPS indicator: 'systemcrypto' FOUND in FIPS binary build info."
+else
+  echo "[verify-fips] WARNING: 'systemcrypto' NOT found in FIPS binary build info."
```
robertolopezlopez: are you sure that this should be surfaced just as a warning?

scripts/verify-fips-build.sh
```diff
@@ -0,0 +1,102 @@
...
+normal_hash="$(hash_file "$normal_bin")"
+fips_hash="$(hash_file "$fips_bin")"
+echo "[verify-fips] normal sha256: $normal_hash"
+echo "[verify-fips] fips   sha256: $fips_hash"
+
+if [ "$normal_hash" = "NOHASH" ]; then
+  echo "[verify-fips] WARNING: no sha256 tool available; skipping byte-difference gate."
+  echo "[verify-fips] Review the build settings above to confirm the FIPS backend."
+  exit 0
+fi
```
robertolopezlopez: same here: probably warning is not enough. `exit 0`???

scripts/verify-fips-build.sh
robertolopezlopez: as it is just a warning and not error/exit 1, I think this may lead to creating fips binaries without systemcrypto. If the validation script does not error out when validation does not pass: which is the point of having such?

this is just a fast idea, please think it over

scripts/verify-fips-build.sh
robertolopezlopez: same here: if we do not have sha256 and exit 0... can we assure that the binary is fips compliant?

scripts/verify-fips-build.sh
```diff
@@ -0,0 +1,102 @@
...
+set -euo pipefail
+
+normal_dir="${1:?normal binary directory required}"
+fips_dir="${2:?fips binary directory required}"
+
+# Load Windows env (PATH for go) if running on the Windows executor.
+if [ -f "/c/tools-cache/snyk-env.sh" ]; then
+  # shellcheck disable=SC1091
+  source "/c/tools-cache/snyk-env.sh"
+fi
```
robertolopezlopez: is it always the same path?

robertolopezlopez (approved):

scripts/windows/install-python3.ps1
```diff
@@ -14,27 +14,50 @@ if (Test-Path $envScript) {
...
+  $skipInstall = $false
+
+  foreach ($candidate in $pythonCandidates) {
+    if (-not (Test-Path $candidate)) {
+      continue
+    }
+    $reportedVersion = (& $candidate --version 2>&1).Trim()
+    # Require the exact version to be followed by whitespace or end-of-string so
+    # a different patch (e.g. 3.12.80) is not treated as 3.12.8.
+    if ($reportedVersion -match '3\.12\.8(\s|$)') {
```
robertolopezlopez, line 31: [as graphite pointed out](https://github.com/snyk/cli/pull/6987#issuecomment-4980161622), this matching may be ambiguous - or is it intentional?

danskmt: what part of the matching do you mean?

robertolopezlopez:
Should we worry about this warning? I would use the suggested `Get-Command dotnet` or `dotnet` (I guess this works in PowerShell as well)

> [**Redundant Installation Guard**](https://github.com/snyk/cli/pull/6987/changes#diff-beaa2249ad8bcd335f1c89e15f16f180e5c0afb7c73d6583b480d0ffe6d0ccc6R17-R20) 🟡 [minor]
> The logic to check for an existing .NET installation is coupled to a hardcoded path `C:\Program Files\dotnet\dotnet.exe`. If the SDK was previously installed to a different location (e.g., via a user-local installation or custom `DOTNET_ROOT`), the script will redundantly re-download and install the SDK even if `dotnet` is already on the PATH and satisfies the version requirement. Using `Get-Command dotnet` would be more robust.
> 
> ```powershell
> $dotnetExe = 'C:\Program Files\dotnet\dotnet.exe'
> $dotnetPath = $null
> 
> if (Test-Path $dotnetExe) {
> ```

danskmt:
> Should we worry about this warning? I would use the suggested `Get-Command dotnet` or `dotnet` (I guess this works in PowerShell as well)

In our flow we always install .NET to the default `C:\Program Files\dotnet` and cache that exact path, so the hardcoded check can't miss

robertolopezlopez:
> > Should we worry about this warning? I would use the suggested `Get-Command dotnet` or `dotnet` (I guess this works in PowerShell as well)
> 
> In our flow we always install .NET to the default `C:\Program Files\dotnet` and cache that exact path, so the hardcoded check can't miss

great!

PR #7019, opened 2026-07-21
fix: include distribution:url in mvn component metadata [CMPA-604]

calhar-snyk:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing _what_ was changed, not _how_.
- [x] Includes detailed description of changes
- [x] Contains risk assessment (Low | Medium | High)
- [ ] Highlights breaking API changes (if applicable)
- [x] Links to automated tests covering new functionality
- [ ] Includes manual testing instructions (if necessary)
- [ ] Updates relevant GitBook documentation (PR link: ___)
- [ ] Includes product update to be announced in the next stable release notes

## What does this PR do?

Bumps `snyk-mvn-plugin` to 4.9.1, pulling in [snyk/snyk-mvn-plugin#224](https://github.com/snyk/snyk-mvn-plugin/pull/224) and [snyk/snyk-mvn-plugin#228](https://github.com/snyk/snyk-mvn-plugin/pull/228). With this bump, dep-graph nodes for Maven projects carry a `distribution:url` label alongside the existing `hash:<algorithm>` labels when `--include-component-metadata` is set.

Extends `test/jest/acceptance/snyk-test/maven-include-component-metadata.spec.ts` to assert the new `distribution:url` label is present when the flag is set and absent otherwise, matching the existing hash-label coverage.

Review on the initial 4.9.0 bump asked for a regression test covering a repository configured with credentials embedded in its URL (e.g. `https://user:secret@example.com/maven2`). Chasing that down surfaced a real leak: `mvn dependency:list-repositories` prints such a URL verbatim, and nothing in `snyk-mvn-plugin` stripped it before building `distribution:url` — those credentials would have flowed straight into the label and downstream SBOM `externalReferences`. Fixed upstream in snyk-mvn-plugin#228 (strips embedded userinfo via the `URL` class, mirroring the same guard `nodejs-lockfile-parser` already applies to npm/yarn resolved URLs) and released as 4.9.1, which this PR now pins. The new regression test hand-constructs a local `.m2` entry for a repository with a credentialed URL and asserts the emitted label contains no userinfo.

## Where should the reviewer start?

[...]

Issue #44, opened 2016-09-26
Snyk test fails wtih error 500


timelf123:
- `node -v`: 6.6.0
- `npm -v`: 3.10.6
- `snyk -v`: 1.1931
- Command run: 
`snyk test`
### Actual behaviour
```
An unknown error occurred. Please include the trace below when reporting to Snyk:

Error: 500
    at node_modules\snyk\lib\test.js:104:23
    at node_modules\snyk\lib\request.js:64:7
    at lib$es6$promise$$internal$$tryCatch (node_modules\es6-promise\dist\es6-promise.js:362:16)
    at lib$es6$promise$$internal$$invokeCallback (node_modules\es6-promise\dist\es6-promise.js:374:17)
    at lib$es6$promise$$internal$$publish (node_modules\es6-promise\dist\es6-promise.js:345:11)
    at lib$es6$promise$asap$$flush (node_modules\es6-promise\dist\es6-promise.js:119:9)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
npm ERR! Test failed.  See above for more details.
```

### Steps to reproduce
Windows if it matters. This error was resolved by re-running snyk-wizard, then snyk test would run again

---

If applicable, please append the `--debug` flag on your command and include the output here **ensuring to remove any sensitive/personal details or tokens.

adrukh:
@timelf123 thanks for reporting this!

First of all, trying to understand if this was a persistent failure until you ran `snyk wizard`. Can you provide the name of the project as it appears in your `package.json` file so I can look through our logs? Also, a general timeframe of these errors will be helpful.

If this is was not persistent for the same project, it may relate to  some back-end instability we've had for 36 minutes between 11:38am and 12:14pm UTC today. Some clients may have received HTTP 5xx responses during this time. Again, the timeframe of the failure you observed will help me understand if this is related.

timelf123:
I experienced this issue ~10-20 minutes before I posted this issue. The name in the package is shared by other projects "mean"

timelf123:
Seems to be intermittent. I pushed to heroku to test on another env and the initial build failed on the `snyk protect` command with a 503, but I was able to run ` heroku run --app myApp  snyk protect --debug`

my id from the debug is: `c168efccdb0168f67e8cdcedc29b50066f4a3ab3`

adrukh:
[...]

PR #7048, opened 2026-07-28
test: cover --include-component-metadata for poetry projects

calhar-snyk:
## What does this PR do?

Pins the `--include-component-metadata` behaviour for **poetry** projects the way
it's already pinned for npm, maven, go and yarn, and bumps the plugin that emits it.

The CLI already forwards `--include-component-metadata` into plugin options through
its generic single-plugin path (`get-single-plugin-result.ts`), and poetry rides
that same path — so **no `src/` runtime change is needed**. What was missing was a
plugin release that threads the flag to the poetry parser, plus the test.

Three parts:

1. **Bump `snyk-python-plugin` `^3.2.1` → `^3.3.1`.** 3.3.0 first threaded
   `includeComponentMetadata` to `snyk-poetry-lockfile-parser`; **3.3.1** pulls in
   parser **1.10.1**, which tightens the `distribution:url` provenance label so a
   source URL's query string (not part of the artifact's identity) is no longer
   carried into it. Chain: snyk/snyk-poetry-lockfile-parser#44 →
   snyk/snyk-python-plugin#377.
2. **Add `poetry-include-component-metadata.spec.ts`.** With the flag the printed
   dep-graph carries `hash:<alg>` and PEP 503 `distribution:url` labels; without it
   neither appears. The interpreter is resolved `python → python3` the same way the
   sibling python specs do, so it runs on the Alpine acceptance executors that ship
   only `python3`.
3. **Add a provenance-hygiene regression test** (`poetry-creds-source` fixture),
   mirroring the maven and go coverage in #7019: a private index whose source URL
   embeds userinfo and a query string must yield a `distribution:url` reduced to its
   canonical identity — no userinfo, no query.

## Note on the lockfile / `min-release-age`

The lockfile is generated with `--min-release-age=0`. `3.3.1` / `1.10.1` were
published within this repo's 4-day freshness window, but they are **our own
releases, cut and verified in this same change set** (parser #44 → plugin #377) —
the freshness guard's provenance concern doesn't apply. The guard only gates fresh
*resolution*; `npm ci` installs `3.3.1` from the committed lockfile without
re-checking it (verified locally), so CI and end users are unaffected.

The spec file. Two poetry-specific differences from the npm sibling worth a look:

- Only one project is scanned (no npm-workspaces equivalent), so there's no
  `--all-projects` case.
- `distribution:url` for poetry is a PEP 503 **project-page** URL with a
  `#<filename>` fragment (e.g.
  `https://pypi.org/simple/jinja2/#Jinja2-2.11.2-py2.py3-none-any.whl`), not a
  direct download URL — `poetry.lock` never records one.

**Low.** Test and fixtures only; no CLI runtime code path changes. The single
production change is the `snyk-python-plugin` bump, which tightens the
`distribution:url` provenance label and remediates pre-existing transitive
advisories (`tmp`, `lodash`) rather than introducing any.

`snyk test --include-component-metadata` (and the SBOMs built from it) now emit
per-component `hash:` and `distribution:url` labels for **poetry** projects, as
they already do for npm/maven/go/yarn. Delivered by the `snyk-python-plugin` 3.3.1
bump in this PR; the test pins it.

> Commit type: typed `test:` (no CLI runtime code), but the bump does deliver the
> above capability. If we'd rather this surface in release notes, `fix:` or `feat:`
> is reasonable — release owner's call.

[CMPA-656](https://snyk.atlassian.net/browse/CMPA-656)

[CMPA-656]: https://snyksec.atlassian.net/browse/CMPA-656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/jest/acceptance/snyk-test/poetry-include-component-metadata.spec.ts
+++ b/test/jest/acceptance/snyk-test/poetry-include-component-metadata.spec.ts
@@ -1,0 +1,142 @@
+import { createProjectFromFixture } from '../../util/createProject';
+import { runSnykCLI } from '../../util/runSnykCLI';
+import { runCommand } from '../../util/runCommand';
+
+jest.setTimeout(1000 * 60);
+
+// `--include-component-metadata` makes the python plugin forward the flag to
+// snyk-poetry-lockfile-parser, which reads the per-package hashes already in
+// poetry.lock and surfaces them as `hash:<algorithm>` labels on the dep-graph
+// nodes. Like npm (and unlike maven) there is nothing to resolve first — the
+// hashes live in the lockfile — so this fixture needs no `poetry install`.
+//
+// Note the difference from npm: poetry.lock records no artifact *download* URL,
+// so `distribution:url` here is the PEP 503 project page for the package with a
+// `#<filename>` fragment naming the file whose hash is reported — provenance
+// rather than a fetch target. For PyPI-sourced deps (no `[package.source]`) that
+// root is pypi.org; private-index (`legacy`) deps use their recorded root.
+describe('`snyk test --include-component-metadata` (poetry)', () => {
+  interface PrintedGraph {
+    target: string;
+    graph: any;
+  }
+
+  const parseDepGraphs = (printGraphStdout: string): PrintedGraph[] =>
+    printGraphStdout
+      .split('DepGraph end')
+      .filter((block) => block.includes('DepGraph data:'))
+      .map((block) => ({
+        graph: JSON.parse(
+          block.split('DepGraph data:')[1].split('DepGraph target:')[0],
+        ),
+        target: block.split('DepGraph target:')[1].trim(),
+      }));
+
+  const labelKeys = (graph: any, prefix: string): string[] =>
+    graph.graph.nodes
+      .flatMap((node) => Object.keys(node.info?.labels ?? {}))
+      .filter((key) => key.startsWith(prefix));
+
+  const labelValues = (graph: any, key: string): string[] =>
+    graph.graph.nodes
+      .map((node) => node.info?.labels?.[key])
+      .filter((value): value is string => Boolean(value));
+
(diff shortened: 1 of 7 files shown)


danskmt (approved):

CatalinSnyk (approved):

test/jest/acceptance/snyk-test/poetry-include-component-metadata.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+jest.setTimeout(1000 * 60);
+
+// `--include-component-metadata` makes the python plugin forward the flag to
+// snyk-poetry-lockfile-parser, which reads the per-package hashes already in
+// poetry.lock and surfaces them as `hash:<algorithm>` labels on the dep-graph
+// nodes. Like npm (and unlike maven) there is nothing to resolve first — the
+// hashes live in the lockfile — so this fixture needs no `poetry install`.
+//
+// Note the difference from npm: poetry.lock records no artifact *download* URL,
```
CatalinSnyk, line 13: question: I'm guessing then the case for credentials being part of the `distribution:url` does not apply for this, right? Just cross-referencing with #7019 which I recalled added explicit tests for credentials being stripped as part of the acceptance tests.

CatalinSnyk: Also my local PR reviewer noted this down, I'm not as familiar with the poetry context, so I'll just share it here in case it's something indeed to be considered.
```
Query credentials can leak through distribution:url

Parser 1.10.0 preserves URL query parameters. Poetry stores source URLs unchanged, so tokens such as ?token=... or X-Amz-Signature can enter the dependency graph and be uploaded. Please strip or redact query parameters and add private-source coverage.
```

PR #7062, opened 2026-07-30
docs: document that unit tests require TEST_SNYK_TOKEN

rrama:

- [x] Commit messages are release-note ready
- [ ] Links to automated tests covering new functionality — n/a, docs only
- [ ] Updates relevant GitBook documentation — n/a, contributor docs
- [ ] Includes product update to be announced in the next stable release notes — n/a

## What does this PR do?

Documents that `npm run test:unit` requires `TEST_SNYK_TOKEN`.

Without it, four suites fail with `MissingApiTokenError`, because they drive command entry points that call `validateCredentials` before doing anything else:

- `test/jest/unit/lib/commands/fix/fix.spec.ts`
- `test/jest/unit/python/snyk-test-pyproject.spec.ts`
- `test/jest/unit/pnpm/snyk-test-pnpm-project.spec.ts`
- `test/jest/unit/snyk-code/snyk-code-test-report.spec.ts`

This wasn't written down anywhere. `CONTRIBUTING.md` mentions `TEST_SNYK_TOKEN` only under **acceptance** tests, and `AGENTS.md` showed a bare `npm run test:unit`.

The part worth documenting explicitly is that reaching for `SNYK_TOKEN` — the obvious variable, and the one the CLI itself uses — has no effect. `test/setup.js` deletes `SNYK_TOKEN` and `SNYK_API_KEY`, then writes `TEST_SNYK_TOKEN` into the CLI user config. That's sensible (tests run against a known configuration rather than whatever the developer is authenticated as), but from the outside it looks like the token is being ignored and the tests are simply broken.

This came out of an agent reporting these suites as failures during environment setup, having concluded from the error that `SNYK_TOKEN` was needed. It isn't, and setting it changes nothing.

No behaviour is changed here — this only documents what already happens. Whether these suites *should* need a token is a separate question: `CONTRIBUTING.md` states unit tests "should not test services outside the code itself", so arguably they belong at the acceptance layer. Happy to raise that separately if the team agrees.

`CONTRIBUTING.md` -> "Unit tests". The `AGENTS.md` change mirrors it.

```sh
npm run test:unit                            # 4 suites fail with MissingApiTokenError
TEST_SNYK_TOKEN=<token> npm run test:unit    # those 4 suites pass (22 tests)
SNYK_TOKEN=<token> npm run test:unit         # unchanged - still fails
```

None — contributor documentation only.

## Risk assessment: Low

Documentation only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -177,8 +177,8 @@ The CLI follows a layered testing pyramid. Each layer has a different goal, syst
 ## Running Tests
 
 ```sh
-# TypeScript unit tests
-npm run test:unit
+# TypeScript unit tests (some suites validate credentials, so a token is required)
+TEST_SNYK_TOKEN=<token> npm run test:unit
 
 # TypeScript acceptance/user journey tests (requires a built binary)
 TEST_SNYK_COMMAND=./binary-releases/snyk-macos-arm64 npm run test:acceptance
@@ -190,6 +190,8 @@ cd cliv2 && make test
 # A single TS test file
 npx jest --runInBand test/jest/unit/path/to/test.spec.ts
 ```
+
+`SNYK_TOKEN` is **not** an alternative to `TEST_SNYK_TOKEN` — `test/setup.js` removes `SNYK_TOKEN` (and `SNYK_API_KEY`) from the environment when either is set, and writes `TEST_SNYK_TOKEN` into the CLI user config so tests run against a known configuration.
 
 ## Running the CLI Locally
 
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -178,8 +178,8 @@ The CLI follows a layered testing pyramid. Each layer has a different goal, syst
 ## Running Tests
 
 ```sh
-# TypeScript unit tests
-npm run test:unit
+# TypeScript unit tests (some suites validate credentials, so a token is required)
+TEST_SNYK_TOKEN=<token> npm run test:unit
```
rrama, line 181: The docs do not instruct the agent that they need to build the CLI before they can run the tests. My agent didn't notice as it built the CLI first anyway, but would it be worth adding this requirement as well?

danskmt (approved):

PR #7064, opened 2026-07-31
fix: add user journey tests for FUA code path + bump code-client-go

florinMiro:
## Pull Request Submission Checklist

- [x] Follows [CONTRIBUTING](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) guidelines
- [x] Commit messages
  are [release-note ready](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md#writing-commit-messages), emphasizing
  _what_ was changed, not _how_.
- [ ] Includes detailed description of changes
- [ ] Contains risk assessment (Low | Medium | High)
- [ ] Highlights breaking API changes (if applicable)
- [ ] Links to automated tests covering new functionality
- [ ] Includes manual testing instructions (if necessary)
- [ ] Updates relevant GitBook documentation (PR link: ___)
- [ ] Includes product update to be announced in the next stable release notes

## What does this PR do?
* Add tests for Code CLI upload path
* Remove deprecated calls to `request`
* Bump code-client-go to pull in an observability fix https://github.com/snyk/code-client-go/pull/170

## Risk assessment (Low | Medium | High)?
Low

PR #7081, opened 2026-08-04
refactor: remove legacy Snyk Code client

fsargent:
## Summary
- Remove the unused TypeScript Snyk Code implementation and `@snyk/code-client` dependency.
- Keep Snyk Code on the native `code-client-go` workflow.
- Reject accidental Snyk Code use through legacy TypeScript entry points.
- Remove tests for the retired implementation and keep native workflow coverage.

This replaces the stale [PR #6944](https://github.com/snyk/cli/pull/6944). File Upload API support and user-journey coverage have since merged in [PR #7061](https://github.com/snyk/cli/pull/7061) and [PR #7064](https://github.com/snyk/cli/pull/7064).

## Test plan
- [x] `npm run lint`
- [x] `npx jest --runInBand test/jest/unit/ecosystems.spec.ts test/jest/unit/validate-fix-command-is-supported.spec.ts`
- [x] `npm run build:dev`
- [x] Snyk Code scan. No finding affects changed code.


src/cli/commands/test/index.ts
```diff
@@ -110,6 +111,12 @@ export default async function test(
     throw new MissingArgError();
   }
 
+  if (options.code) {
```
robertolopezlopez, line 114: if code-client-go resolves legacy implementation, the user will get an error. Are you sure that we want to throw an error here instead of fully disabling legacy fallback?

src/cli/commands/fix/validate-fix-command-is-supported.ts
```diff
@@ -4,6 +4,7 @@ import { getEcosystemForTest } from '../../../lib/ecosystems';
 
 import { isFeatureFlagSupportedForOrg } from '../../../lib/feature-flags';
 import { FeatureNotSupportedByEcosystemError } from '../../../lib/errors/not-supported-by-ecosystem';
+import { CustomError } from '../../../lib/errors/custom-error';
```
robertolopezlopez: not blocking: this is redundant with the import in line 9, merge

src/cli/commands/test/index.ts
robertolopezlopez: either disabling legacy fallback or making code-client-go select native entrypoint always

fsargent: `code-client-go` no longer selects between implementations. `codeWorkflowEntryPoint` always calls `EntryPointNative`. The CLI falls back only for unknown commands. `code test` is registered explicitly and accepts unknown flags. This guard protects direct TypeScript use and routing regressions.

src/cli/commands/fix/validate-fix-command-is-supported.ts
fsargent: Fixed in d1e322577. Both errors now use the existing barrel import.

PR #7102, opened 2026-08-09
chore: CLI-1758 trigger Endpoint Explorer refresh after releases

mattdolan-snyk:
## What changed

- trigger the Endpoint Explorer snapshot refresh after successful Preview and Stable S3 publication
- run each trigger as a terminal CircleCI job with no downstream release dependencies
- dispatch the existing `refresh-cli-data.yml` workflow with the published channel
- cap the dispatch request at five seconds and convert missing credentials, transport failures, and non-204 responses into warnings
- add release-script tests for accepted dispatches and each non-blocking failure path

## Why

The Endpoint Explorer should refresh when a public CLI channel changes, without waiting for the Explorer workflow or making CLI releases depend on Explorer availability.

The dispatch is best-effort. Once GitHub accepts the workflow request, the CLI pipeline does not follow or wait for the Explorer run. If the request cannot be made, the CLI release continues and the existing scheduled/manual Explorer refresh remains available as recovery.

## Validation

- `make format`
- `make lint`
- `go test ./... -skip TestBuildRecipeUsesGeneratedLSMetadataInLdflags` in `release-scripts`
- `bash -n release-scripts/trigger-endpoint-explorer-refresh.sh`
- YAML parse of `.circleci/config.yml`
- verified the live `snyk/endpoint-binary-explorer` workflow still accepts `workflow_dispatch` with `stable` and `preview` channel inputs

The unmodified `TestBuildRecipeUsesGeneratedLSMetadataInLdflags` currently fails because its temporary CLI fixture does not include the help Markdown now required by the build recipe; the remaining release-script suite passes.

## Deployment assumption

`HAMMERHEAD_GITHUB_PAT` must be able to dispatch Actions workflows in `snyk/endpoint-binary-explorer`. Missing or insufficient access is intentionally warning-only and does not fail the CLI release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1333,6 +1333,33 @@ workflows:
                 - main
                 - '/release.*/'
 
+      # Best-effort terminal jobs: no release job depends on these dispatches.
+      - trigger-endpoint-explorer-refresh:
+          name: Trigger Endpoint Explorer refresh (preview)
+          channel: preview
+          context:
+            - team-hammerhead-common-deploy-tokens
+          requires:
+            - upload preview
+          filters:
+            branches:
+              only:
+                - main
+                - '/release.*/'
+
+      - trigger-endpoint-explorer-refresh:
+          name: Trigger Endpoint Explorer refresh (stable)
+          channel: stable
+          context:
+            - team-hammerhead-common-deploy-tokens
+          requires:
+            - upload latest/stable
+          filters:
+            branches:
+              only:
+                - main
+                - '/release.*/'
+
       - release-github:
           name: upload github
           context:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1333,6 +1333,33 @@ workflows:
                 - main
                 - '/release.*/'
 
+      # Best-effort terminal jobs: no release job depends on these dispatches.
```
PeterSchafer, line 1336: Question: why didn't you just extend the [existing mechanism](https://github.com/snyk/cli/blob/main/release-scripts/upload-artifacts.sh#L290-L315) we use for other examples?

PeterSchafer (changes requested):

PR #7047, opened 2026-07-28
feat(container): surface provenance attestations behind feature flag [PRIM-100]

bdemeo12:
## What

Surfaces container image **provenance attestations** through the CLI, gated behind a feature flag.

- Bumps `snyk-docker-plugin` `^9.16.0` → `^9.18.0`, which adds provenance attestation extraction (emits the `provenanceMetadata` fact).
- Gates that fact behind the **`surfaceProvenanceAttestations`** feature flag in `filterDockerFacts`, mirroring the existing `allowNewContainerFacts` pattern. When the flag is off, `provenanceMetadata` is filtered out before facts are sent downstream; when on, it is forwarded to Registry for upsert.

The provenance flag is independent of `allowNewContainerFacts` so it can be rolled out separately.

## Changes

- `package.json` / `package-lock.json` — bump `snyk-docker-plugin` to `^9.18.0`.
- `src/cli/commands/constants.ts` — add `SURFACE_PROVENANCE_ATTESTATIONS_FEATURE_FLAG = 'surfaceProvenanceAttestations'`.
- `src/lib/ecosystems/common.ts` — in `filterDockerFacts`, filter `provenanceMetadata` unless the flag is enabled.

## Feature flag

`surfaceProvenanceAttestations` — defined in registry (snyk/registry#44819).

## Downstream

This is the CLI half of the end-to-end provenance work. Registry relays the fact and assets-api persists it.

## Notes
n/a

PR #7130, opened 2026-08-13
fix(cli): prevent handled errors from producing duplicate output [CLI-1765]

bdemeo12:

- [x] Commit messages
- [ ] Updates relevant GitBook documentation (PR link: ___) — n/a, no user-facing docs change

## What does this PR do?

CLI Ticket: https://snyksec.atlassian.net/browse/CLI-1765?actionerId=712020%3Aed5c88da-ffa1-47ff-9ea7-b4345eaadf9f&sourceType=assign

Fixes:

json output being unparseable because an extra error was printed after the result. This happens when the CLI receives an error (aside from exit status 1), it prints it — so once teardown joins the exit error together with the handled network errors, the result is no longer exit status 1 and gets printed after the real output.

Also pins `snyk-docker-plugin` to `9.20.0` so CI exercises the failure being fixed. (Original PR: https://github.com/snyk/cli/pull/7047 + discussion: https://snyksec.atlassian.net/servicedesk/customer/portal/64/CLIA-1576)

`cliv2/pkg/core/main.go` — `shouldSuppressDisplay` and the third return value from `processError`, then the guarded `displayError` call in `tearDown`.

Then `cliv2/pkg/core/main_test.go` for the two new cases in `Test_processError` and the new `Test_shouldSuppressDisplay`.

Against an image whose provenance fetch fails (SDP version 9.20.0 +):

```sh
snyk container test <image> --json > out.json
jq . out.json
```
The `container.spec.ts` acceptance job is the automated equivalent — it fails on 9.20.0 without this fix, which is why the pin is included here.

n/a

### DIFF
--- a/cliv2/pkg/core/main.go
+++ b/cliv2/pkg/core/main.go
@@ -772,7 +772,7 @@ func processError(err error, errorList []error) ([]error, error) {
 	return resultErrorList, mostRelevant
 }
 
-//  Reports whether err is the command's own exit-code result
+// Reports whether err is the command's own exit-code result
 func commandOwnsOutput(err error) bool {
 	if _, ok := err.(*exec.ExitError); ok {
 		return true
--- a/cliv2/pkg/core/main.go
+++ b/cliv2/pkg/core/main.go
@@ -456,11 +456,41 @@ func doctorTip(isCI bool) string {
 	return "Try snyk doctor: `snyk <command> -d 2>&1 | snyk doctor --stdin`"
 }
 
+// shouldSuppressDisplay reports whether err is worth printing.
+//
+// Joined errors match no direct type assertion, so they are unwrapped and checked
+// one at a time.
+func shouldSuppressDisplay(err error) bool {
+	if wrappedErr, ok := err.(interface{ Unwrap() []error }); ok {
+		unwrappedErrs := wrappedErr.Unwrap()
+		if len(unwrappedErrs) == 0 {
+			return false
+		}
+
+		for _, err := range unwrappedErrs {
+			var exitErr *exec.ExitError
+			if errors.As(err, &exitErr) && exitErr.ExitCode() < constants.SNYK_EXIT_CODE_ERROR {
+				return true
+			}
+		}
+
+		for _, err := range unwrappedErrs {
+			if !shouldSuppressDisplay(err) {
+				return false
+			}
+		}
+		return true
+	}
+
+	_, isExitError := err.(*exec.ExitError)
+	_, isErrorWithCode := err.(*cli_errors.ErrorWithExitCode)
+
+	return isExitError || isErrorWithCode || errorHasBeenShown(err)
+}
+
 func displayError(err error, userInterface ui.UserInterface, config configuration.Configuration, ctx context.Context, isCI bool) {
 	if err != nil {
-		_, isExitError := err.(*exec.ExitError)
-		_, isErrorWithCode := err.(*cli_errors.ErrorWithExitCode)
-		if isExitError || isErrorWithCode || errorHasBeenShown(err) {
+		if shouldSuppressDisplay(err) {
 			return
 		}
 
--- a/cliv2/pkg/core/main.go
+++ b/cliv2/pkg/core/main.go
@@ -456,18 +456,9 @@ func doctorTip(isCI bool) string {
 	return "Try snyk doctor: `snyk <command> -d 2>&1 | snyk doctor --stdin`"
 }
 
-// A successful command's exit-code error is not a reportable failure and does
-// not need to be printed.
+// Skip errors the command already reported
 func shouldSuppressDisplay(err error) bool {
-	if exitErr, ok := err.(*exec.ExitError); ok {
-		return exitErr.ExitCode() < constants.SNYK_EXIT_CODE_ERROR
-	}
-
-	if codeErr, ok := err.(*cli_errors.ErrorWithExitCode); ok {
-		return codeErr.ExitCode < constants.SNYK_EXIT_CODE_ERROR
-	}
-
-	return errorHasBeenShown(err)
+	return commandOwnsOutput(err) || errorHasBeenShown(err)
 }
 
 func displayError(err error, userInterface ui.UserInterface, config configuration.Configuration, ctx context.Context, isCI bool) {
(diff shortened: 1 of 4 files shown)


bdemeo12:
This PR originates from: https://github.com/snyk/cli/pull/7047

cliv2/pkg/core/main.go
```diff
@@ -455,11 +455,41 @@ func doctorTip(isCI bool) string {
...
+func shouldSuppressDisplay(err error) bool {
+	if wrappedErr, ok  := err.(interface{ Unwrap() []error }); ok {
+		unwrappedErrs := wrappedErr.Unwrap()
+		if len(unwrappedErrs) == 0 {
+			return false
+		}
+
+		for _, err := range unwrappedErrs {
+			var exitErr *exec.ExitError
+			if errors.As(err, &exitErr) && exitErr.ExitCode() < constants.SNYK_EXIT_CODE_ERROR {
```
bdemeo12: An exit code below SNYK_EXIT_CODE_ERROR means the command worked and anything alongside it is noise

cliv2/pkg/core/main.go
```diff
@@ -455,11 +455,41 @@ func doctorTip(isCI bool) string {
 	return "Try snyk doctor: `snyk <command> -d 2>&1 | snyk doctor --stdin`"
 }
 
+// shouldSuppressDisplay reports whether err is worth printing.
+//
+// Joined errors match no direct type assertion, so they are unwrapped and checked
+// one at a time.
+func shouldSuppressDisplay(err error) bool {
```
bdemeo12, line 460: All errors are combined here: https://github.com/snyk/cli/blob/cbc275e37a255b46f20556d163a16abb4ae558b9/cliv2/pkg/core/main.go#L771

shouldSuppressDisplay returns true when there's nothing to show: https://github.com/snyk/cli/blob/cbc275e37a255b46f20556d163a16abb4ae558b9/cliv2/pkg/core/main.go#L504

Previously displayError checked the error's type directly. it didnt account for a combined error, so it fell through to the printing code and emitted a second JSON object.: https://github.com/snyk/cli/blob/26f0543da50200ccf22d93adaf2a177dde0f22eb/cliv2/pkg/core/main.go#L462

Now this check suppresses the display err by returning early, before the printing

package.json
```diff
@@ -116,7 +116,7 @@
     "semver": "^6.0.0",
     "snyk-config": "^5.0.0",
     "snyk-cpp-plugin": "^2.24.3",
-    "snyk-docker-plugin": "9.19.0",
+    "snyk-docker-plugin": "9.20.0",
```
bdemeo12, line 119: bumped SDP to use as an acceptance test

bdemeo12: https://github.com/snyk/cli/pull/7047

cliv2/pkg/core/main.go
CatalinSnyk: I found a second reproduction after this PR was opened: running snyk code test can successfully return findings while an auxiliary organization-slug lookup returns 403. The scan returns ErrorWithExitCode{1}, which is then joined with the captured Forbidden error and printed after the results.

This implementation appears to fix the Container case (exec.ExitError{1}), but not this Code case because the early suppression only checks exec.ExitError. Could we determine suppression from the original command error before joining, covering both error types?

cliv2/pkg/core/main.go
CatalinSnyk: The Code reproduction also makes me think the original error is the better place to make this decision. Before processError, we know whether the command returned exec.ExitError or ErrorWithExitCode; after joining, we have to infer that recursively and risk suppressing unrelated sibling errors.

CatalinSnyk: I think we could calculate suppression from the original error in tearDown func, before `processError` joins anything?
```go
suppressDisplay := shouldSuppressDisplay(err)
if err != nil {
    allErrors, outputError = processError(err, errorList)
    // If processError selected a different error, such as maintenance or the
    // captured cause of TS_CLI_TERMINATED, that error still needs displaying.
    suppressDisplay = suppressDisplay && errors.Is(outputError, err)
}
if !suppressDisplay {
    displayError(outputError, ...)
}
```

`shouldSuppressDisplay` could then retain the simple direct checks for `exec.ExitError`, `ErrorWithExitCode`, and already-displayed errors. This covers both cases while avoiding recursive unwrapping.

cliv2/pkg/core/main_test.go
```diff
@@ -729,6 +707,42 @@ func Test_displayError(t *testing.T) {
 	})
 }
 
+func Test_shouldSuppressDisplay(t *testing.T) {
```
CatalinSnyk, line 710: suggestion: Could we also use the ticket mentioned tests here? They assert that both `exec.ExitError` and `ErrorWithExitCode` remain the returned outputError, while the auxiliary error remains in `allErrors`.  I think it better shows the separate concern for the user facing output and analytics error list.

cliv2/pkg/core/main.go
```diff
@@ -516,10 +524,14 @@ func tearDown(err error, errorList []error, startTime time.Time, ua networking.U
 
 	outputError := err
 	allErrors := errorList
+	suppressDisplay := false
 
 	if err != nil {
 		allErrors, outputError = processError(err, errorList)
 
+		// Determine suppression from the original exit code
```
CatalinSnyk: suggestion: I think we can keep this responsibility inside `processError` instead of returning a joined display error and having carrying a suppression state suppression?

`processError` already returns both `allErrors` and `outputError`. We can keep the joined errors in allErrors for the analytics payload, but return the original command error as `outputError` when the selected result still contains it. Something like this is what I'm thinking currently:

```go
mostRelevant := cli_errors.FindMostRelevantError(allErrors)
if commandOwnsOutput(originalError) &&
    errors.Is(mostRelevant, originalError) {
    outputError = originalError
} else {
    outputError = mostRelevant
}
```

`commandOwnsOutput` could match the two original command-result types: `exec.ExitError` and `ErrorWithExitCode`.

cliv2/pkg/core/main_test.go
bdemeo12: Yes! Added!

cliv2/pkg/core/main.go
bdemeo12: Yes! processError now owns this! Thanks for the help!

robertolopezlopez:
@bdemeo12 please have a look to the failing pipeline 👍

bdemeo12:
> @bdemeo12 please have a look to the failing pipeline 👍

I think this pipeline is flakey! I reran it and its succeeded!

robertolopezlopez (dismissed):

PeterSchafer:
Closing this PR as it was replaced by #7191

PR #7131, opened 2026-08-13
test: isolate CLI-1765 Go fix from the snyk-docker-plugin bump

bdemeo12:
Throwaway branch to isolate a variable — not for review.

Identical Go change to #7130, but `snyk-docker-plugin` stays at 9.16.0.

#7130 failed acceptance tests on all 8 platforms (auth errors printing nothing). That PR changes two things at once: the error-handling fix and the 9.16.0 → 9.20.0 bump. This separates them.

- Acceptance tests **pass** here → the 9.20.0 bump caused the failures, not the Go change
- Acceptance tests **fail** here → the Go change is at fault and needs rework

Will be closed once it has answered the question.


bdemeo12:
Answered its question: the failures came from the Go change, not the dependency bump. Superseded by the reworked fix in #7130.

PR #7132, opened 2026-08-14
chore: bump cli-extension-cos

robert-snyk:
Bumps `cli-extension-cos` to its current main: the target configuration file now
takes a `target.scope` block and rejects unknown keys, and grey-box repositories
can be selected by organization slug and SCM target name.

Tests: unchanged — `test/jest/acceptance/snyk-cos/cos.spec.ts` passes as is
(37/37) against the rebuilt binary.

Risk: low. Private build only; everything it touches sits behind
`snyk cos --experimental`, and no other command changes.

### DIFF
--- a/cliv2-private/go.mod
+++ b/cliv2-private/go.mod
@@ -5,7 +5,7 @@ go 1.26.6
 require (
 	github.com/snyk/ambient-canary v0.0.0-20260722064253-fba619a134a9
 	github.com/snyk/cli-extension-axi v0.0.0-20260812113949-210b479505a9
-	github.com/snyk/cli-extension-cos v0.0.0-20260804092425-a7e642ae7b3b
+	github.com/snyk/cli-extension-cos v0.0.0-20260813100139-4d5c43e0785f
 	github.com/snyk/cli/cliv2 v0.0.0
 	github.com/snyk/remy-cli-extension v1.36.1
 	github.com/snyk/rift-cli-extension v1.1.1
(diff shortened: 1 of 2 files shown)


cliv2-private/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.26.5
 require (
 	github.com/snyk/ambient-canary v0.0.0-20260722064253-fba619a134a9
 	github.com/snyk/cli-extension-axi v0.0.0-20260812113949-210b479505a9
-	github.com/snyk/cli-extension-cos v0.0.0-20260804092425-a7e642ae7b3b
+	github.com/snyk/cli-extension-cos v0.0.0-20260813100139-4d5c43e0785f
```
robertolopezlopez, line 8: While running a deep inspect into the changes for this bump, my agent found some potentially dangerous actions.

The old version was exposing just credential count. The new version:

- Copies keys from API response

    (https://github.com/snyk/cli-extension-cos/blob/4d5c43e0785f0c73c555a3aef58e1f12475901a7/internal/services/target/auth_models.go#L144-L146)

- Renders keys without sanitization

    (https://github.com/snyk/cli-extension-cos/blob/4d5c43e0785f0c73c555a3aef58e1f12475901a7/internal/commands/target/create.go#L171-L173)

- The only validation is comparing to empty key/value

    (https://github.com/snyk/cli-extension-cos/blob/4d5c43e0785f0c73c555a3aef58e1f12475901a7/internal/commands/config_auth.go#L82-L87)

Or maybe someone (me or my agent) is hallucinating here. Could you please shed some light? :-)

robertolopezlopez (changes requested):

cliv2-private/go.mod
robert-snyk: Not a hallucination, but a false alarm

`key` here is the name in a key/value pair (`username`, `X-Api-Token`), not a secret – the change is `credential_count: 2` → `credential_keys: [username, password]`, a count replaced by names.

Values exist only in the unexported types the API response decodes into. The `Target` that commands render carries a summary with no value field at all, `toSummary()` drops them at that boundary, and `target dump` exports the `target` block only. Nothing downstream can carry a secret.

robertolopezlopez: `the change is credential_count: 2 → credential_keys: [username, password]`

Just to be sure: this `password` is either sanitized or not the real password. Is this correct?

robert-snyk: Correct – it is not the real password, it is the field's name.

`credential_keys` carries the `key` of each credential pair, verbatim from the config file:

```yaml
- label: Standard user
  type: credentials_login
  credentials:
    - key: username
      value: testuser@acme.com
    - key: password
      value: "s3cr3t"
```

which renders as:

```
Authentication
  Standard user (credentials: username, password)
```

The values are dropped at `toSummary()`, so they are not in the struct the commands render – nothing is masked because nothing is there to mask.

robertolopezlopez (approved):

PR #7140, opened 2026-08-17
chore: [CLI-1758] trigger Endpoint Explorer refresh after releases

mattdolan-snyk:
## What changed

- trigger the Endpoint Explorer snapshot refresh after successful Preview and Stable S3 publication
- run each trigger as a terminal CircleCI job with no downstream release dependencies
- dispatch the existing `refresh-cli-data.yml` workflow with the published channel
- cap the dispatch request at five seconds and convert missing credentials, transport failures, and non-204 responses into warnings
- add release-script tests for accepted dispatches and each non-blocking failure path

## Why

The Endpoint Explorer should refresh when a public CLI channel changes, without waiting for the Explorer workflow or making CLI releases depend on Explorer availability.

The dispatch is best-effort. Once GitHub accepts the workflow request, the CLI pipeline does not follow or wait for the Explorer run. If the request cannot be made, the CLI release continues and the existing scheduled/manual Explorer refresh remains available as recovery.

## Validation

- `make format`
- `make lint`
- `go test ./... -skip TestBuildRecipeUsesGeneratedLSMetadataInLdflags` in `release-scripts`
- `bash -n release-scripts/trigger-endpoint-explorer-refresh.sh`
- YAML parse of `.circleci/config.yml`
- verified the live `snyk/endpoint-binary-explorer` workflow still accepts `workflow_dispatch` with `stable` and `preview` channel inputs

The unmodified `TestBuildRecipeUsesGeneratedLSMetadataInLdflags` currently fails because its temporary CLI fixture does not include the help Markdown now required by the build recipe; the remaining release-script suite passes.

## Deployment assumption

`HAMMERHEAD_GITHUB_PAT` must be able to dispatch Actions workflows in `snyk/endpoint-binary-explorer`. Missing or insufficient access is intentionally warning-only and does not fail the CLI release.

### DIFF
--- a/release-scripts/upload-artifacts.sh
+++ b/release-scripts/upload-artifacts.sh
@@ -170,6 +170,44 @@ trigger_repository_event() {
   fi
 }
 
+# Request a channel-specific Endpoint Explorer refresh without making the CLI
+# release depend on the dispatch or the resulting workflow.
+trigger_endpoint_explorer_refresh() {
+  channel=$1
+
+  if [ -z "${HAMMERHEAD_GITHUB_PAT:-}" ]; then
+    echo "WARNING: HAMMERHEAD_GITHUB_PAT is unavailable; Endpoint Explorer $channel refresh was not requested."
+    return 0
+  fi
+
+  echo "Triggering Endpoint Explorer $channel refresh..."
+  curl_status=0
+  response=$(curl \
+    --location \
+    --request POST \
+    --connect-timeout 2 \
+    --max-time 5 \
+    --header "Accept: application/vnd.github+json" \
+    --header "Authorization: Bearer $HAMMERHEAD_GITHUB_PAT" \
+    --header "X-GitHub-Api-Version: 2022-11-28" \
+    --data "{\"ref\":\"main\",\"inputs\":{\"channel\":\"$channel\"}}" \
+    --write-out "%{http_code}" \
+    --silent \
+    --show-error \
+    --output /dev/null \
+    "https://api.github.com/repos/snyk/endpoint-binary-explorer/actions/workflows/refresh-cli-data.yml/dispatches") || curl_status=$?
+
+  if [ "$curl_status" -ne 0 ]; then
+    echo "WARNING: Endpoint Explorer $channel refresh dispatch failed (curl exit $curl_status); continuing the CLI release."
+  elif [ "$response" != "204" ]; then
+    echo "WARNING: Endpoint Explorer $channel refresh was not requested (HTTP $response); continuing the CLI release."
+  else
+    echo "Endpoint Explorer $channel refresh requested."
+  fi
+
+  return 0
+}
+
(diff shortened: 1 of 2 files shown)


release-scripts/upload-artifacts.sh
```diff
@@ -170,6 +170,44 @@ trigger_repository_event() {
...
+    --connect-timeout 2 \
+    --max-time 5 \
+    --header "Accept: application/vnd.github+json" \
+    --header "Authorization: Bearer $HAMMERHEAD_GITHUB_PAT" \
+    --header "X-GitHub-Api-Version: 2022-11-28" \
+    --data "{\"ref\":\"main\",\"inputs\":{\"channel\":\"$channel\"}}" \
+    --write-out "%{http_code}" \
+    --silent \
+    --show-error \
+    --output /dev/null \
```
danskmt, line 197: Shouldn't we follow same pattern of **trigger_repository_event** and output to a file? Then we could have more information about a failure?

release-scripts/upload-artifacts.sh
```diff
@@ -170,6 +170,44 @@ trigger_repository_event() {
...
+    return 0
+  fi
+
+  echo "Triggering Endpoint Explorer $channel refresh..."
+  curl_status=0
+  response=$(curl \
+    --location \
+    --request POST \
+    --connect-timeout 2 \
+    --max-time 5 \
```
danskmt, line 189: should also include retries, for example`--retry 2 --retry-connrefused`?

release-scripts/upload-artifacts_test.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestDistributionTriggersRequestEndpointExplorerRefresh(t *testing.T) {
```
danskmt, line 12: I believe you have to plug these tests to run in the CI, e.g. in config.yml, under `jobs`, under `test-go`:
```yaml
- run:
      name: Running release script tests
      command: make test-release-scripts
```

release-scripts/upload-artifacts.sh
```diff
@@ -170,6 +170,44 @@ trigger_repository_event() {
...
+
+  if [ "$curl_status" -ne 0 ]; then
+    echo "WARNING: Endpoint Explorer $channel refresh dispatch failed (curl exit $curl_status); continuing the CLI release."
+  elif [ "$response" != "204" ]; then
+    echo "WARNING: Endpoint Explorer $channel refresh was not requested (HTTP $response); continuing the CLI release."
+  else
+    echo "Endpoint Explorer $channel refresh requested."
+  fi
+
+  return 0
```
danskmt, line 208: Should we have a different return value for the "WARNING" logs? Because right now it just logs (and might get lost / nobody sees it), but continues the flow normally in all cases.

PR #7133, opened 2026-08-14
feat: wire live Configuration into analytics instrumentation

nick-y-snyk:
## Summary
- `sendInstrumentation` now passes `analytics.WithConfiguration(eng.GetConfiguration())` alongside the existing `WithLogger(logger)` option to `GetV2InstrumentationObject`, so real CLI invocations opt into GAF's new shape-based secret scrubbing for analytics extension data.
- No other behavior of `sendInstrumentation` changes.

Implements ticket 02 of the `agent-feedback-secret-redaction` spec (see `cli-extension-axi` repo, `.scratch/agent-feedback-secret-redaction/`).

## Blocked — do not merge yet
This depends on `analytics.WithConfiguration`, added in [go-application-framework#704](https://github.com/snyk/go-application-framework/pull/704), which is still open. This branch does **not** build against the currently-pinned GAF version. Once #704 merges and a GAF version containing it is tagged, `cliv2/go.mod` needs a real dependency bump before this can merge — that bump is intentionally not included here.

Verified locally against a temporary `replace github.com/snyk/go-application-framework => ../../go-application-framework` pointed at #704's branch (not included in this diff): builds clean, `go vet` clean, tests pass.

## Test plan
- [x] `go test ./cliv2/pkg/core/...` passes (locally, with the temporary GAF replace)
- [x] New `Test_sendInstrumentation_passesEngineConfigurationToInstrumentationObject` asserts `eng.GetConfiguration()` is consulted for the new option via a gomock `MockEngine`
- [ ] CI — expected to fail until GAF#704 ships and `go.mod` is bumped for real

___

### **PR Type**
Enhancement, Dependencies

___

### **Description**
- Wire `go-application-framework` configuration into analytics.

- Ensure specific sensitive data is excluded from redaction.

- Update `go-application-framework` to v0.15.0.

___

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>instrumentation.go</strong><dd><code>Pass engine configuration to analytics instrumentation.</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/cli/pull/7133/changes#diff-bd3f842516ec1f463be60388de00b14040c2e4678d2eb8b112ac3e85508055bf">+8/-2</a>&nbsp; &nbsp; &nbsp; </td>

[...]

PR #7142, opened 2026-08-17
feat: wire live Configuration into analytics instrumentation

nick-y-snyk:
## Summary
- Passes the engine's live `Configuration` through to `analytics.WithConfiguration` at the report-analytics chokepoint (GAF#704), so `logging.REDACTION_TERMS`-based scrubbing covers the CLI's own analytics extension data, not just debug logs.
- `populateRedactionTerms` excludes the client machine id and any detected AI agent name from its unknown-value sweep — both echo a raw env var value verbatim into an extension (`studio::client_machine_id`, `persona.agent`), so without the exclusion the scrub chokepoint stripped them right back out as if they were unrecognized secrets.

Squashed recreation of #7133 into a single commit (that PR had accumulated several unsquashed merge commits flagged by Danger). Same diff, same review comments should still apply — #7133 left open, not closed automatically.

## Test plan
- [x] `go build ./...` (cliv2)
- [x] `go vet ./...`
- [x] `golangci-lint run ./pkg/core/...`
- [x] `go test ./pkg/core/...` — full suite green, including new `Test_populateRedactionTerms_excludesClientMachineId`, `Test_populateRedactionTerms_excludesDetectedAgent`, and extended `Test_sendInstrumentation_passesEngineConfigurationToInstrumentationObject`
- [x] Verified fresh `GOMODCACHE` resolves `go-application-framework v0.15.0` from scratch (addresses stale "unresolvable dependency" review comment on #7133)

### DIFF
--- a/cliv2/pkg/core/main_test.go
+++ b/cliv2/pkg/core/main_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/snyk/go-application-framework/pkg/local_workflows/content_type"
 	"github.com/snyk/go-application-framework/pkg/local_workflows/json_schemas"
 	"github.com/snyk/go-application-framework/pkg/local_workflows/local_models"
+	"github.com/snyk/go-application-framework/pkg/logging"
 	"github.com/snyk/go-application-framework/pkg/mocks"
 	"github.com/snyk/go-application-framework/pkg/utils/ufm"
 	"github.com/snyk/go-application-framework/pkg/workflow"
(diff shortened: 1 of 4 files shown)


cliv2/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/snyk/cli/cliv2
 
-go 1.26.6
```
robertolopezlopez, line 3: why the downgrade?

cliv2-private/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/snyk/cli/cliv2-private
 
-go 1.26.6
+go 1.26.5
```
robertolopezlopez: same here, why the downgrade

cliv2/pkg/core/main_test.go
```diff
@@ -68,6 +69,52 @@ func Test_mainWithErrorCode(t *testing.T) {
...
+	config := configuration.NewWithOpts(configuration.WithAutomaticEnv())
+	machineId := "studio-device-id-abc12345"
+	t.Setenv("INTERNAL_SNYK_CLIENT_MACHINE_ID", machineId)
+
+	terms := populateRedactionTerms(config, mockEngine)
+
+	assert.NotContains(t, terms, machineId, "client machine id must never be swept into REDACTION_TERMS, or the analytics scrub chokepoint strips it right back out of its own extension")
+}
+
+func Test_populateRedactionTerms_excludesDetectedAgent(t *testing.T) {
```
robertolopezlopez, line 103: Please, add more test cases here (table driven?). Specially AI_AGENT with spaces. Unlikely, but might happen.

robertolopezlopez (changes requested):

cliv2/go.mod
nick-y-snyk: image
explained in dms

nick-y-snyk: Looks like snyk fixed the issue today, returning it to 12.6.6

robertolopezlopez (approved):

PR #7153, opened 2026-08-19
chore: bump cli-extension-dep-graph to v2.10.0

gergo-papp:
Bumps `cli-extension-dep-graph` from v2.8.1 to v2.10.0 in **both** Go modules — `cliv2` and `cliv2-private`.

`cliv2-private` consumes `cliv2` through a `replace` directive, so it records the extension as an indirect dependency with its own resolved version. Left at v2.8.1 it would keep building the private binary against the old extension, so the two have to move together. That one was resolved with `go mod tidy` rather than by hand, since the requirement comes from the replaced module.

### What this picks up

- A placeholder .NET dependency resolver, registered behind the `internal-dotnet-resolver` feature flag and **disabled by default**. It discovers .NET target files and reports an empty dep graph, so it can be exercised end to end before any resolution logic exists.
- An ownership transfer and a CI context change, from v2.8.2.

### Behaviour

No change for users. The resolver sits behind two independent gates that are both off by default: the `--use-sbom-resolution` flag, and the org feature flag above. With either unset, dep-graph resolution is byte-identical to v2.8.1.

### Verification

- 3 lines per module across `go.mod` / `go.sum`, 6 in total — `go mod tidy` pulled in no transitive changes to either.
- `go build ./...` passes in both `cliv2` and `cliv2-private`.
- `go list -deps` confirms `pkg/ecosystems/dotnet/nuget` is linked into both binaries, so the flag has something to gate rather than this being a no-op bump.
- `release-scripts`, the repo's third Go module, has no dependency on the extension.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cliv2/go.mod
+++ b/cliv2/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/rs/zerolog v1.35.1
 	github.com/snyk/cli-extension-agent-scan v0.0.0-20260803112735-0ad462cba2af
 	github.com/snyk/cli-extension-ai-bom v0.0.0-20260817214817-d0a81ac40172
-	github.com/snyk/cli-extension-dep-graph/v2 v2.8.1
+	github.com/snyk/cli-extension-dep-graph/v2 v2.10.0
 	github.com/snyk/cli-extension-iac v0.0.0-20260515092252-505c498f1077
 	github.com/snyk/cli-extension-iac-rules v0.0.0-20260515141409-fb475901bb8e
 	github.com/snyk/cli-extension-os-flows v0.0.0-20260818092349-91c745a9ee7a
--- a/cliv2-private/go.mod
+++ b/cliv2-private/go.mod
@@ -212,7 +212,7 @@ require (
 	github.com/skeema/knownhosts v1.3.2 // indirect
 	github.com/snyk/cli-extension-agent-scan v0.0.0-20260803112735-0ad462cba2af // indirect
 	github.com/snyk/cli-extension-ai-bom v0.0.0-20260817214817-d0a81ac40172 // indirect
-	github.com/snyk/cli-extension-dep-graph/v2 v2.8.1 // indirect
+	github.com/snyk/cli-extension-dep-graph/v2 v2.10.0 // indirect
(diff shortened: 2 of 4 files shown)


octavian-snyk (approved):

cliv2/go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/rs/zerolog v1.35.1
 	github.com/snyk/cli-extension-agent-scan v0.0.0-20260803112735-0ad462cba2af
 	github.com/snyk/cli-extension-ai-bom v0.0.0-20260817214817-d0a81ac40172
-	github.com/snyk/cli-extension-dep-graph/v2 v2.8.1
+	github.com/snyk/cli-extension-dep-graph/v2 v2.10.0
```
CatalinSnyk, line 16: issue: Changes are not synchronised between private-public builds (which is why the builds fails in CI).

octavian-snyk (approved):

cliv2/go.mod
gergo-papp: Right diagnosis — the first commit bumped `cliv2` only. Fixed in 6eaae177, which bumps `cliv2-private` to v2.10.0 as well (indirect there, since it consumes `cliv2` via the `replace` directive).

Verified against the checks `make lint` actually runs, rather than by eye:

- `go mod tidy -diff` — clean in both `cliv2` and `cliv2-private`
- `make validate-gomod-sync` → `gomodsync --mode=validate` → **"go.mod files are in sync"**
- `make -C cliv2 tidy` produces no further diff, so the committed state is what the repo's own tooling generates
- `go build ./...` passes in both modules, and `go list -deps` confirms the new extension is linked into both binaries

Lesson taken: I bumped `cliv2` with `go get` and tidied `cliv2-private` by hand, when `make format` (which calls `make -C cliv2 tidy`, including the `gomodsync` step) is the documented path and would have caught the desync immediately. Using that from the start next time.

If `test_and_release` still fails on this head, it is something other than go.mod sync and I will dig into the job output.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #7162, opened 2026-08-20
fix: split Harness name and version in persona analytics (CLI-1771)

nick-y-snyk:
## Summary

- `persona.agent` fused Claude Code's name, version, and role into one string (`claude-code_2-1-233_agent`), fragmenting it into ~101 distinct values in analytics and making "how many Claude Code sessions" impossible to query. This splits name and version into two fields.
- Replaces the hand-ported `detect-agent` signature table (`cliv2/internal/persona/agent`) with the upstream `github.com/vercel/detect-agent` module — grows coverage 13 → 20 harnesses and fixes a live bug where plain-CLI OpenCode was undetected (our port matched the launcher's env var, not the one OpenCode itself sets).
- Adds a harness-agnostic regex that splits a version-shaped suffix off any detected name (`name[_/]version[_/role]`), and canonicalises the remaining fragment against `detect-agent`'s own `KnownAgents` vocabulary rather than a hand-written alias list.
- New `persona.agent_version` is written only when a version is found; `persona.agent` always resolves to `claude_code` for Claude Code regardless of whether it was detected via `AI_AGENT` or via environment signature.
- `cursor-cli`, `augment-cli`, and every other already-clean harness are unaffected — no version-shaped suffix, so the regex doesn't match and the value passes through unchanged.

**Signature renames, flagging ahead of release per the spec**: dropping the hand-ported table for upstream `detect-agent`'s own vocabulary changes the canonical string several harnesses report. None of these are remapped back to the old values — intentional, current analytics data for these harnesses is already messed up, so this is a good time to adopt vercel's naming instead of preserving a value we hand-picked ourselves:

| Detection case | Old `DetectAgent()` | New `Detect()` |
|---|---|---|
| Claude Code signature | `claude` | `claude_code` |
| Gemini signature | `gemini` | `gemini_cli` |
| Codex signature | `codex` | `codex_cli` |
| OpenCode signature | `opencode` | `open_code` |
| `AI_AGENT=github-copilot-cli` | `github-copilot` | `github-copilot-cli` |

This lands in the next release; will also be called out as a Jira comment on CLI-1771.

## Datadog: where the new field lands

Confirmed against live `analytics-service` logs (`service:analytics-service "analytics payload"`). Today:

```
custom.analytics-service.interaction.extension.persona.agent = "claude-code_2-1-220_harness"
```

After this change, `agent_version` is a new sibling key at the same nesting level, queryable as `@analytics-service.interaction.extension.persona.agent_version`:

```yaml
persona:
    agent: claude_code
    agent_version: "2.1.220"
    interactive: false
    interactive_mode: 0
```

`agent_version` is omitted from the payload entirely (not an empty string) whenever no version is parsed out.

## License bundling — fixed

`github.com/vercel/detect-agent` has no `LICENSE` file in its repo (confirmed via the GitHub API), despite tagged releases and external contributors. **This hard-failed CI and any local build.**

Every third-party Go dependency in `cliv2` ships its license text embedded in the CLI binary: `cliv2/scripts/prepare_licenses.go` runs `go-licenses save ./...` and writes one `LICENSE` (or `NOTICE`/`COPYING`) file per module path under `cliv2/internal/embedded/_data/licenses/`, which `cliv2/internal/embedded/file.go` pulls into the binary via `//go:embed _data` and surfaces to end users as the OSS notices. This step runs as part of `make configure` in `cliv2/Makefile`, ahead of `make build`.

Reproduced locally on this branch, matches CI exactly:

```
E... library.go:122] Failed to find license for github.com/vercel/detect-agent: cannot find a known open source license for ".../github.com/vercel/detect-agent@v1.2.0" ...
F... main.go:77] one or more libraries have an incompatible/unknown license: map["unknown":["github.com/vercel/detect-agent"]]
Error running go-licenses save: exit status 1
```

So `make configure` / `make build` fail outright with this dependency in place — this is not a merge that could accidentally ship unnoticed; the build gate stops it. `go test ./...` still passes (it doesn't go through `make configure`), which is why the test-plan checkboxes below are green despite this blocker.

The script already has a manual-override mechanism for packages `go-licenses` can't auto-detect (`github.com/davecgh/go-spew`, `github.com/alexbrainman/sspi`, `github.com/pmezard/go-difflib`, `go.dev` are each fetched from a hardcoded URL) — but `go-licenses save` fatal-exits *before* reaching that step for a package it finds zero license-like files for at all, so the override alone wasn't reachable. Fixed with two changes in `cliv2/scripts/prepare_licenses.go`:

1. `--ignore github.com/vercel/detect-agent` on the `go-licenses save` invocation, so it no longer requires a verdict on that package (dependencies of the ignored package would still be checked, but it has none).
2. A new manual-download entry supplying the Apache-2.0 license text directly (`https://www.apache.org/licenses/LICENSE-2.0.txt`), matching `github.com/vercel/detect-agent`'s own `package.json` declaration (`"license": "Apache-2.0"`) — its repo just never committed the file.

Verified both changes are needed: removing the `--ignore` and keeping only the manual-download entry still reproduces the exact same fatal error, since `go-licenses save` never reaches the manual-download loop. With both in place, `go run scripts/prepare_licenses.go` exits 0 and embeds a real Apache-2.0 `LICENSE` file for `github.com/vercel/detect-agent`.

Note this asserts Apache-2.0 based on `package.json`'s declaration, not on a license file the copyright holder actually published — still worth an explicit legal nod given that.

## Test plan

- [x] `go test ./...` passes in `cliv2` and `cliv2-private`
- [x] `go vet ./...` and `golangci-lint run` clean on changed packages
- [x] `go mod tidy -diff` clean in both `cliv2` and `cliv2-private`
- [x] New test coverage in `cliv2/internal/persona/persona_test.go` (`TestReport_Agent`) at the single seam where persona values are recorded onto analytics — extends the existing fake-analytics-recorder test rather than a new harness, using `t.Setenv` per case (the upstream module reads the process environment directly, no injection point). Covers every real-world shape observed in production: fused underscore (both `_agent` and `_harness` roles), legacy slash format normalising to the same bucket, bare signature detection, an already-clean harness, a fused identifier with no version-shaped segment, an identifier that fails to canonicalise, a redacted (`***`) identifier, and a version-shaped run with no trailing role.
- [x] Verified against 14 days of production `analytics-service` logs: no harness other than Claude Code currently reports a version-shaped `persona.agent` value at any real volume, confirming the ticket's Out-of-Scope call.
- [ ] Manual: after merge, re-run the distinct-`persona.agent`-value count and confirm Claude Code's ~101 buckets collapse to one, with `persona.agent_version` populated on those events.

Ref: CLI-1771


cliv2/internal/persona/agent/agent_test.go
robertolopezlopez, line 1: maybe adding some unit tests here wouldn't hurt - not full coverage, just some happy and unhappy paths

robertolopezlopez: although agent.go has shrinked quite much tbh

robertolopezlopez: ok nevermind I can see that this is already done indirectly in persona_test.go

cliv2/internal/persona/agent/agent.go
```diff
@@ -1,133 +1,72 @@
...
-			return err == nil
-		},
+// DetectAgent resolves the Harness identifier the current process declared
+// via AI_AGENT, or that detect-agent recognised from an environment
+// signature. The value is returned exactly as detect-agent produced it: an
+// explicit AI_AGENT declaration passes through verbatim (trimmed), a
+// signature match is already canonical. No version split or canonicalisation
+// is applied here — see SplitVersion.
+func DetectAgent() (string, bool) {
+	details, err := detectagent.Detect()
```
robertolopezlopez, line 20: apparently `Detect()` returns different signatures than the previous contract

|  Detection case    |  Old DetectAgent()  |  New Detect() |
|---|---|---|
|  Gemini signature    |  gemini | gemini_cli  |
|  Codex signature      | codex  | codex_cli  |
|   OpenCode signature  | opencode  |  open_code |
|  Claude Code signature | claude  | claude_code  |
|  AI_AGENT=github-copilot-cli | github-copilot  | github-copilot-cli  |

if I am not mistaken you are not remapping those differences to the previous values. As this (if I am not mistaken) won't be released in the closest release, I believe it will mess up the analytics.

Please fix this 🥺

cliv2/internal/persona/agent/agent.go
```diff
@@ -1,133 +1,72 @@
...
-	match func(l lookup) bool
-}
+// versionSuffix recognises a version-shaped run of digits at the end of a
+// Harness identifier, with an optional trailing role/surface segment that is
+// captured only so it isn't absorbed into the version, then discarded. Both
+// '_' and '/' are accepted as separators, so the legacy slash format
+// normalises identically to the current underscore format. Requiring at
+// least two numeric components avoids matching a name that merely ends in a
+// single digit.
+var versionSuffix = regexp.MustCompile(`^(?P<name>.+?)[_/](?P<version>\d+(?:[.\-]\d+){1,3})(?:[_/](?P<role>[A-Za-z0-9-]+))?$`)
```
robertolopezlopez: what about `name@version`? I do not know if I am misunderstanding the docs https://github.com/vercel/detect-agent/blob/v1.2.0/README.md#recommended-naming-convention

cliv2/internal/persona/persona_test.go
```diff
@@ -44,3 +44,101 @@ func TestReport(t *testing.T) {
...
+			env:         map[string]string{"AI_AGENT": "claude-code_2-1-233"},
+			wantAgent:   "claude_code",
+			wantVersion: "2.1.233",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			for k, v := range tc.env {
+				t.Setenv(k, v)
```
robertolopezlopez, line 179: you need to clean this up after 👀

cliv2/internal/persona/agent/agent.go
nick-y-snyk: It will be released in the next release. Current data is messed up anyways, it's a good time to use naming conventions from vercel package

robertolopezlopez (changes requested):
Last comment: probably, those posible corner cases in agent.go make agent_test.go really necessary after the fix

cliv2/internal/persona/agent/agent.go
nick-y-snyk: added support for it too

cliv2/internal/persona/persona_test.go
nick-y-snyk: done

cliv2/internal/persona/agent/vercel_contract_test.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// TestVercelDetectDoesNotNormalizeExplicitAIAgent proves the vercel package's
+// contract directly, independent of our own split/canonicalise code: an
+// explicit AI_AGENT declaration is returned byte-for-byte, version and role
+// still fused in, while the signature-fallback path (no AI_AGENT) returns
+// vercel's own clean vocabulary name. This is why SplitVersion exists.
+func TestVercelDetectDoesNotNormalizeExplicitAIAgent(t *testing.T) {
+	t.Run("explicit AI_AGENT is returned verbatim, not standardised", func(t *testing.T) {
+		t.Setenv("AI_AGENT", "claude-code_2-1-233_agent")
```
robertolopezlopez, line 59: you need to clean up, same as you did in persona_test.go

robertolopezlopez: same in other tests in this file

cliv2/internal/persona/agent/agent.go
robertolopezlopez: moving to a new thread

cliv2/internal/persona/agent/agent.go
```diff
@@ -1,133 +1,73 @@
...
-}
+// versionSuffix recognises a version-shaped run of digits at the end of a
+// Harness identifier, with an optional trailing role/surface segment that is
+// captured only so it isn't absorbed into the version, then discarded. '_',
+// '/', and '@' are all accepted as separators: '_' and '/' cover Snyk's own
+// fused and legacy formats, '@' covers detect-agent's documented
+// name@version convention for custom AI_AGENT declarations. Requiring at
+// least two numeric components avoids matching a name that merely ends in a
+// single digit.
+var versionSuffix = regexp.MustCompile(`^(?P<name>.+?)[_/@](?P<version>\d+(?:[.\-]\d+){1,3})(?:[_/@](?P<role>[A-Za-z0-9-]+))?$`)
```
robertolopezlopez: This regexp does not support, for example, `agent@1` format.

https://go.dev/play/p/hgsiEkr-oxO

cliv2/internal/persona/agent/vercel_contract_test.go
nick-y-snyk: done

cliv2/internal/persona/agent/agent.go
nick-y-snyk: nice catch, added support

cliv2/internal/persona/agent/agent.go
robertolopezlopez: OK, in such case please make this explicit in the PR description (it still shows `codex` instead of `codex_cli`, for example - review all updated signatures) and Jira (as comment)

cliv2/internal/persona/agent/vercel_contract_test.go
robertolopezlopez: The ambient-variable isolation is fixed, but cleanup still does not restore the caller environment. `t.Setenv` inside `t.Cleanup` registers another cleanup, which later unsets each restored variable; a sentinel set before the subtest becomes empty afterward. Please add that regression assertion and restore with `os.Setenv` plus error handling (with a focused `usetesting` suppression), or restructure cleanup so `t.Setenv` is registered before cleanup begins.

cliv2/internal/persona/agent/agent.go
```diff
@@ -1,133 +1,83 @@
...
 	}
 
-	for _, s := range signatures {
-		if s.match(l) {
-			return s.agent, true
-		}
+	fragment := m[versionSuffix.SubexpIndex("name")]
+	canonical, ok := knownNames[normalize(fragment)]
+	if !ok {
+		return raw, ""
```
robertolopezlopez: I found another corner case

`SplitVersion("custom-agent@2.0")` returns `("custom-agent@2.0", "")`, but it should return `("custom-agent", "2.0")`

robertolopezlopez: https://go.dev/play/p/ok2sVFpr48f

nick-y-snyk: added test case and handling

cliv2/internal/persona/agent/agent.go
robertolopezlopez: Done — I added the analytics rename note to CLI-1771 myself.

robertolopezlopez (approved):

PR #7163, opened 2026-08-20
fix: surface command timeouts correctly for unified test scanner [OSF-452]

jacek-rzrz:
When SNYK_TIMEOUT_SECS fires during a --all-projects run in the Go unified scanner path, the timeout was surfacing as a misleading "no testable projects found" (SNYK-CLI-0000) instead of the correct "Command timed out" (SNYK-CLI-0026), unlike the legacy TS CLI path.

Root cause spanned two extension repos: cli-extension-dep-graph was silently dropping plugin resolution errors instead of surfacing them (snyk/cli-extension-dep-graph#236), and cli-extension-os-flows was reformatting accumulated --all-projects failures into a plain string.


cliv2/pkg/core/errorhandling.go
```diff
@@ -27,13 +27,15 @@ func decorateError(err error) error {
 		return err
 	}
 
+	// decorate DeadlineExceeded with a more user-friendly error, even if something
```
CatalinSnyk: question/suggestion: Actually I think we could just scope it down to
```go
var catalogErr snyk_errors.Error
if errors.As(err, &catalogErr) {
	if catalogErr.ErrorCode == "SNYK-CLI-0011" &&
		errors.Is(err, context.DeadlineExceeded) {
		return cli.NewCommandTimeoutError(
			catalogErr.Detail,
			snyk_errors.WithCause(err),
		)
	}

	return err
}

if errors.Is(err, context.DeadlineExceeded) {
	return cli.NewCommandTimeoutError(err.Error(), snyk_errors.WithCause(err))
}
```
And then maybe revisit later to see what/where the timeout errors should be wrapped/preserved.

MoisBrigitte: We've pushed a fix  - does this work?

CatalinSnyk (approved):

PR #7176, opened 2026-08-26
fix: wait for npm publish before axi release

robertolopezlopez:

- [x] Highlights breaking API changes (none)
- [x] Updates relevant GitBook documentation (not applicable)
- [x] Includes product update to be announced in the next stable release notes (not applicable; release infrastructure only)

## What does this PR do?

Waits for the exact `snyk` CLI version to become visible on npm before dispatching the `cli_release` event to `snyk/axi`.

The release pipeline already runs distribution-channel triggers after `npm publish`, but npm can accept a publish and continue processing it asynchronously. During the `1.1307.0` release, `npm publish` completed successfully at 14:18:25 UTC, while the package became available at 14:43:37 UTC. The AXI workflow exhausted its own retries at 14:41:23 UTC and failed with `ETARGET` because the package was still unavailable.

This change polls npm every 30 seconds for up to 40 minutes. Other distribution channels are triggered immediately; only AXI waits for npm readiness. If the package never becomes available, the distribution job fails instead of dispatching a workflow that cannot succeed.

Failed AXI job: https://github.com/snyk/axi/actions/runs/32979700168/job/98213031126

`wait_for_npm_package` and its use immediately before the AXI dispatch in `release-scripts/upload-artifacts.sh`.

```shell
bash -n release-scripts/upload-artifacts.sh
```

Verify with mocked `npm` and `sleep` commands that:

1. Failed npm lookups are retried every 30 seconds.
2. AXI dispatch proceeds once the exact version becomes visible.
3. AXI is not dispatched after all 80 checks fail.

None. This only changes release orchestration.

Low. The change affects only AXI dispatch timing. Existing Snyk Images, Scoop, and Homebrew dispatches remain unchanged. Worst case, the distribution job waits up to 40 minutes and fails if npm never exposes the published version.

`npm publish` printed: `Your package is being processed and may take a few minutes to become available.` Its successful exit therefore confirms upload acceptance, not registry readiness.

None.

### DIFF
--- a/release-scripts/upload-artifacts.sh
+++ b/release-scripts/upload-artifacts.sh
@@ -123,6 +123,23 @@ upload_npm() {
   fi
 }
 
+wait_for_npm_package() {
+  local version="${VERSION_TAG#v}"
+
+  echo "Waiting for snyk@$version to become available on npm..."
+  for attempt in {1..80}; do
+    if npm view --prefer-online "snyk@$version" version >/dev/null 2>&1; then
+      echo "snyk@$version is available on npm."
+      return 0
+    fi
+    echo "npm availability check $attempt failed, retrying in 30s"
+    sleep 30
+  done
+
+  echo "snyk@$version did not become available on npm."
+  return 1
+}
+
 # Trigger event for a given repository
 # Failure mode is to log and continue as these steps
 # are non blocking to a release.
@@ -308,9 +325,8 @@ for arg in "${@}"; do
         DISTRIBUTION_FAILURE=1
     fi
 
-    # 4. Trigger axi
-    trigger_repository_event "axi" "cli_release"
-    if [ $? -ne 0 ]; then
+    # 4. Trigger axi once the published CLI version is available on npm
+    if ! wait_for_npm_package || ! trigger_repository_event "axi" "cli_release"; then
         DISTRIBUTION_FAILURE=1
     fi
 


release-scripts/upload-artifacts.sh
```diff
@@ -308,9 +325,8 @@ for arg in "${@}"; do
         DISTRIBUTION_FAILURE=1
     fi
 
-    # 4. Trigger axi
-    trigger_repository_event "axi" "cli_release"
-    if [ $? -ne 0 ]; then
+    # 4. Trigger axi once the published CLI version is available on npm
+    if ! wait_for_npm_package || ! trigger_repository_event "axi" "cli_release"; then
         DISTRIBUTION_FAILURE=1
     fi
```
danskmt, line 331: ~is this logic correct?~ (it is, see comment below)
1. The operator you are using **||** will not evaluate the second expression if the first one becomes **true**
2. If wait_for_npm_package returns 0, you're flipping it to 1, thus second expression will not take action, and you'll put DISTRIBUTION_FAILURE=1. But wait_for_npm_package returns 0 on success, no?

danskmt: Nevermind, bash works with inverted logic... 0 means no error. Suggestions for changing this part of the code:

Suggestion 1:
```bash
wait_for_npm_package && trigger_repository_event "axi" "cli_release" || DISTRIBUTION_FAILURE=1
```

Suggestion 2:
```bash
if ! { wait_for_npm_package && trigger_repository_event "axi" "cli_release"; }; then
  DISTRIBUTION_FAILURE=1
fi
```

danskmt (approved):

release-scripts/upload-artifacts.sh
```diff
@@ -123,6 +123,23 @@ upload_npm() {
   fi
 }
 
+wait_for_npm_package() {
```
PeterSchafer, line 126: Question: Is the CLI pipeline the right place to do a synchronization? shouldn't the axi package know about its dependencies and the cli pipeline naively just triggers axi?

robertolopezlopez: As you can see in [the linked job](https://github.com/snyk/axi/actions/runs/32979700168/job/98213031126), it keeps on failing with ENOENT error. So far we have always had to retrigger manually the axi job because of this precise issue.

I can see two possible solutions:

1. Increase timeout in the axi job. Advantage: the cli release does not get delayed. Disadvantage: this longer timeout does not guarantee any success. Actually Nick has increased it already but we have always had to re-run this job manually.
2. Waiting here until the npm package becomes available. I would consider that, if the npm package is not available yet, the release has not finished. Disadvantage, obvious: the cli release gets delayed.

Still, I may keep the warning message and remove the `DISTRIBUTION_FAILURE=1` in case of "timeout". What do you think?

release-scripts/upload-artifacts.sh
robertolopezlopez: thanks for your comments.

Leaving syntax aside, what's the change/advantage in your proposals?

danskmt: just readability. The options I suggested reads better IMHO.
But can leave as-is if no other changes are coming to this PR

robertolopezlopez: Thanks @danskmt . Let's wait for Peter answer and maybe I will change the PR a little bit with both aggregated suggestions. Your comment is reasonable.

PR #278, opened 2018-11-21
fix: Parse ARGs in Dockerfile FROM instructions 

robcresswell:
This patch raises the snyk-docker-plugin version, fixing an issue where
we failed to parse ARGs used in a Dockerfile FROM instruction

- [x] Ready for review
- [x] Follows [CONTRIBUTING](https://github.com/snyk/snyk/blob/master/.github/CONTRIBUTING.md) rules
- [x] Reviewed by Snyk internal team

PR #7180, opened 2026-08-27
fix: report SNYK-CLI-0008 for repos with no testable projects [OSF-481]

ariadna-roman:

- [x] Highlights breaking API changes (if applicable) — n/a, this restores previous behaviour
- [ ] Updates relevant GitBook documentation (PR link: n/a — no documented behaviour changes; exit code 3 and SNYK-CLI-0008 are already documented)

## What does this PR do?

Restores `SNYK-CLI-0008` ("No supported files found") and exit code **3** for `snyk test` in a directory with no supported manifest files, when the unified test API feature flag is enabled.

[Reported in Slack](https://snyk.slack.com/archives/C0BQRUH0UCR/p1787761635731579): with `FFUseUnifiedTestAPIForOSCliTest` enabled, routing no longer falls back to `useLegacy`, and the os-flows extension surfaced a generic `SNYK-CLI-0000` with `failed to get dependency graph: no testable projects found` and exit **2**. CI pipelines that scan repositories without testable projects ([snyk/homebrew-tap](https://github.com/snyk/homebrew-tap), [snyk/scoop-snyk](https://github.com/snyk/scoop-snyk)) treat `SNYK-CLI-0008` / exit 3 as "nothing to scan", so this broke them.

Three changes:

1. **Bumps `cli-extension-os-flows`** to pick up snyk/cli-extension-os-flows#278, which returns a typed Error Catalog error instead of a bare `fmt.Errorf` when no project is detected.
2. **Adds the missing exit-code mapping.** `SNYK-CLI-0008` was never in the central error-catalog → exit-code map, because the legacy TS CLI derived exit 3 itself by substring-matching its own message (`src/cli/main.ts`). With the flag on, that TypeScript code never runs, and `DeriveExitCode` only recognises an `exec.ExitError` or an `ErrorWithExitCode` — so a correctly typed error from Go still fell through to the generic exit 2. Follows the existing `aibom.NoSupportedFiles` precedent.
3. **Sends the `--json` error document to stdout** instead of the error stream. The error stream is deliberately routed to stderr whenever the default output is structured, so error text can't corrupt a JSON payload — but this document *is* the structured output, so `snyk test --json` left stdout empty and wrote the JSON to stderr. This affected every error on a Go workflow, not just this one, and diverged from both the legacy CLI and the rest of the CLI's own contract (the IaC acceptance tests already expect `"ok": false` on stdout).

`cliv2/cmd/cliv2/behavior/maperrortoexitcode.go` — the one-line map entry, and the reason it's needed.

Then `cliv2/pkg/core/main.go` (`displayError`) for the stdout/stderr change, and `test/jest/acceptance/snyk-test/equivalenceHelpers.ts` for why the equivalence suite couldn't catch this: `assertEquivalent` returned early for no-submission fixtures, with a comment stating that exit codes "legitimately differ (TS CLI exits 3, os-flows exits 2)". That escape hatch excused the exact divergence being fixed.

```bash
make build
mkdir /tmp/no-projects && echo hi > /tmp/no-projects/README.txt && cd /tmp/no-projects

# unified path
INTERNAL_SNYK_CLI_USE_UNIFIED_TEST_API_FOR_OS_CLI_TEST=true snyk test; echo "exit=$?"

# legacy path, for comparison
SNYK_FORCE_LEGACY_CLI=true snyk test; echo "exit=$?"
```

Both should print `No supported files found (SNYK-CLI-0008)` and exit **3**. Repeat with `--json > out.json` — `out.json` must be non-empty and identical on both paths (before this change it was empty on the unified path).

**Fixed:** `snyk test` on a repository with no supported manifest files again reports `SNYK-CLI-0008` and exits 3 instead of a generic `SNYK-CLI-0000` with exit 2, and `snyk test --json` again writes the error document to stdout instead of stderr.

**Low.** The exit-code mapping is additive — one entry keyed on an error code nothing else produces. The stdout change affects only the `--json` error path, moving output onto the stream the legacy CLI and the rest of the CLI already use; verified that `code test`, `iac test` and `sbom` in `--json` mode still emit a single payload on stdout with stderr empty.

### Automated tests

- `cliv2/pkg/core/exitcode_test.go` — `SNYK-CLI-0008` → exit 3, added to the existing `TestMapErrorToExitCode` table.
- `test/jest/acceptance/snyk-test/unified-test-api-equivalence.spec.ts` — the `no-supported-target-files` fixture is the regression guard. Now that `assertEquivalent` compares exit codes for no-submission fixtures, legacy 3 vs unified 2 fails the suite.
- Extension-side coverage lives in snyk/cli-extension-os-flows#278.

### Known cosmetic difference

The two paths are not byte-identical in **human-readable** output: the unified path additionally renders the catalog *description* line. This is not caused by this change — `error-catalog`'s JSON:API mapping never carries `Description` in either direction (`MarshalToJSONAPIError` doesn't write it, `MarshalFromJSONAPIError` doesn't read it), so every error the legacy CLI reports over the IPC loses its description while errors constructed in Go keep theirs. Worth fixing upstream via `meta.description`; out of scope here. `--json` output is identical.

- [OSF-481](https://snyksec.atlassian.net/browse/OSF-481)
- snyk/cli-extension-os-flows#278 (merged)

[OSF-481]: https://snyksec.atlassian.net/browse/OSF-481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cliv2-private/go.mod
+++ b/cliv2-private/go.mod
@@ -215,7 +215,7 @@ require (
 	github.com/snyk/cli-extension-dep-graph/v2 v2.11.0 // indirect
 	github.com/snyk/cli-extension-iac v0.0.0-20260515092252-505c498f1077 // indirect
 	github.com/snyk/cli-extension-iac-rules v0.0.0-20260818142329-df932cc1794f // indirect
-	github.com/snyk/cli-extension-os-flows v0.0.0-20260821101812-a52aa2dd8c3b // indirect
+	github.com/snyk/cli-extension-os-flows v0.0.0-20260831095357-9cc236689fc1 // indirect
 	github.com/snyk/cli-extension-sbom v0.0.0-20260818092356-16ee760085f9 // indirect
 	github.com/snyk/cli-extension-secrets v0.0.0-20260811115637-ee872c26b0cd // indirect
 	github.com/snyk/code-client-go v1.31.8 // indirect
--- a/cliv2-private/go.sum
+++ b/cliv2-private/go.sum
@@ -592,8 +592,8 @@ github.com/snyk/cli-extension-iac v0.0.0-20260515092252-505c498f1077 h1:lteivlSo
 github.com/snyk/cli-extension-iac v0.0.0-20260515092252-505c498f1077/go.mod h1:wRpQrWCjzWTPA5WK1xaHjWWI5zfY0qKe12PfKb+lcMk=
 github.com/snyk/cli-extension-iac-rules v0.0.0-20260818142329-df932cc1794f h1:EPaHfaWDJq/lrsrATMqAyKJRfstb9LZnPXYNwXdL32c=
 github.com/snyk/cli-extension-iac-rules v0.0.0-20260818142329-df932cc1794f/go.mod h1:JoubAvjOyuB3y0HxulaiAEiVthMOa9abElpL72C5zz4=
-github.com/snyk/cli-extension-os-flows v0.0.0-20260821101812-a52aa2dd8c3b h1:pAiAfZzN9JCu7/+PqO39sUUmvros+ckJ5n15qy+s0oA=
-github.com/snyk/cli-extension-os-flows v0.0.0-20260821101812-a52aa2dd8c3b/go.mod h1:ukSD97e6hc9w0UkrbI/OV2GeDpsEyUsIzE38vdzRbZ8=
+github.com/snyk/cli-extension-os-flows v0.0.0-20260831095357-9cc236689fc1 h1:r13WOBFc4Jtq7NM1Kj6whKts2B7DLci7CU24aNqNQx8=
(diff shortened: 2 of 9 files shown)


test/jest/unit/unified-test-api-equivalence-helpers.spec.ts
```diff
@@ -0,0 +1,122 @@
+import {
```
ariadna-roman: let's delete this entire file

ariadna-roman:
Deleted `test/jest/unit/unified-test-api-equivalence-helpers.spec.ts` per review. The `assertEquivalent` changes are still exercised by the equivalence suite itself, which now passes the `no-supported-target-files` fixture.

Also trimmed two comments I'd added down to the length their files' existing comments run to, and re-pinned `cli-extension-os-flows` to `dfbba750fc3a` (#278 was amended for its own review comment).

PeterSchafer:
Hey @ariadna-roman, the changes Look generally good, thank you! Since it seems that we were missing this case in the tests, how about adding another User Journey test for it?

PeterSchafer (approved):

PR #7183, opened 2026-08-27
fix: harden Scoop install in deployment monitor

robertolopezlopez:
## Summary

- Harden the `deployment: Scoop (windows)` install step in Deployment Tests: download the Scoop installer to disk, reject truncated downloads, propagate the installer exit code, and refresh PATH before calling `scoop`.
- Fixes intermittent `scoop is not recognized` failures where `Invoke-RestMethod | Invoke-Expression` silently no-ops (~2s install with no output) and the step proceeds without `scoop` on PATH.

Root cause analysis: [GHA run 33067468112](https://github.com/snyk/cli/actions/runs/33067468112/job/98509553287).

## Test plan

- [ ] Merge and trigger **Deployment Tests** via `workflow_dispatch`, or wait for the scheduled run
- [ ] Confirm `deployment: Scoop (windows)` completes with full installer output (~20–30s) and passes hash capture

### DIFF
--- a/.github/workflows/deployment-monitor.yml
+++ b/.github/workflows/deployment-monitor.yml
@@ -119,7 +119,10 @@ jobs:
         shell: pwsh
         run: |
           Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
-          Invoke-RestMethod -Uri https://get.scoop.sh | Invoke-Expression
+          Invoke-RestMethod -Uri https://get.scoop.sh -OutFile install-scoop.ps1
+          if ((Get-Item install-scoop.ps1).Length -lt 1000) { throw "Scoop installer download looks truncated" }
+          & ./install-scoop.ps1
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           scoop bucket add snyk https://github.com/snyk/scoop-snyk
           scoop install snyk
       - name: Capture version and shasum metadata


robertolopezlopez:
another failure https://github.com/snyk/cli/actions/runs/33089662327

.github/workflows/deployment-monitor.yml
```diff
@@ -119,7 +119,11 @@ jobs:
         shell: pwsh
         run: |
           Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
-          Invoke-RestMethod -Uri https://get.scoop.sh | Invoke-Expression
+          Invoke-RestMethod -Uri https://get.scoop.sh -OutFile install-scoop.ps1
+          if ((Get-Item install-scoop.ps1).Length -lt 1000) { throw "Scoop installer download looks truncated" }
```
PeterSchafer, line 123: Question: Where does 1000 come from? Looks a bit hacky.

.github/workflows/deployment-monitor.yml
```diff
@@ -119,7 +119,11 @@ jobs:
...
+          & ./install-scoop.ps1
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+          $env:PATH = "$env:USERPROFILE\scoop\shims;$env:PATH"
```
PeterSchafer: Question: how come this is now required?

.github/workflows/deployment-monitor.yml
robertolopezlopez: Sometimes get.scoop.sh returns error page instead of the installer

.github/workflows/deployment-monitor.yml
robertolopezlopez: This is just a safeguard. Possible to remove if you prefer. We do exactly the same in line 132 (`capture version and shasum metadata` step)

robertolopezlopez: anyway, I am now thinking that if installer fails (checked in line 125) we are already erroring out. And if the installer malfunctions with exit code 0 (unlikely), line 127 would fail (it is a failed test after all). I will remove this line 👍

